PR #191, opened 2014-06-28
Add column to chunkColumnLoad event

deathcap:
The 'chunkColumnLoad' event arguments currently only include the chunk position (`columnCorner`). This PR additionally adds the chunk data (`column`), for mineflayer-based bots to use for various purposes.

### DIFF
--- a/lib/plugins/blocks.js
+++ b/lib/plugins/blocks.js
@@ -94,7 +94,7 @@ function inject(bot) {
     }
 
     assert.strictEqual(offset, args.data.length);
-    bot.emit("chunkColumnLoad", columnCorner);
+    bot.emit("chunkColumnLoad", columnCorner, column);
   }
 
   function blockAt(absolutePoint) {


roblabla:
Right now, the Column class is a hidden implementation detail. It changed in 1.8, proving to be too much of a moving target to be exposed publicly. I think users should be required to use blockAt if they want to get aquire this information, or use their own map_chunk/map_chunk_bulk listener.

deathcap:
ok, I'll have to come up with a different solution. Was trying to use this column data in https://github.com/deathcap/voxel-clientmc/issues/5 but, of course, it incompatibly changed in 1.8…

Maybe a chunk class as part of the public API would be useful. Calling blockAt for every position in the chunk is a lot of overhead (it was the main bottleneck in voxel-clientmc, had to call it anyways to convert 16x16x256 chunks to 16x16x16, slow enough it required its own web worker), and writing a custom map_chunk packet listener locks in clients to a specific protocol version. No easy answer..

roblabla:
I understand the problem, but Chunk is a bit of a moving target, and exposing it would be the same as using map_chunk_bulk in terms of protocol lockup. Maybe adding a "forEachBlock" or something similar? We should open a new issue for this.

PR #194, opened 2015-01-04
Allow protocol version to be set in bot options

pcfreak30:
This will allow bots to be made for lower MC server versions.


roblabla:
Things aren't that simple. The underlying protocol encoder/decoder needs to be changed to be able to understand that version's packets.

roblabla:
This depends on andrewrk/node-minecraft-protocol#65

roblabla:
Closing this, as it doesn't actually fix anything. For discussion about multiprotocol support, #199 .

PR #195, opened 2015-03-09
Update for MC 1.8 protocol changes

deathcap:
For use with 1.8 version of https://github.com/PrismarineJS/node-minecraft-protocol

Not extensively tested but with these changes at least the examples/echo.js demo works

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ Note: This project is no longer maintained.
 
 ## Features
 
- * Supports Minecraft 1.6.2.
+ * Supports Minecraft 1.8.3.
  * Entity knowledge and tracking.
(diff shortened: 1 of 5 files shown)


roblabla:
Thank you for this, it's cool to see some people are still interested in mineflayer.

So as far as I can see, this PR mainly deals with chat and chunk data. That's a pretty good place to start, but I'll expect we'll need a lot more changes before we can really be 1.8-ready. Mainly, the enums will require some love, and I expect a lot of breakage in the different example.

Is this ready for merging, or do you plan on working on it more @deathcap ?

deathcap:
@roblabla I think it's good to pull, is at least a start to make 1.8 functional; enums can be updated separately

PR #197, opened 2015-03-11
Update biomes enum for MC 1.8

deathcap:
MC 1.8 update for lib/enums/biomes.json (probably the simplest of the enums). Based on burger.json from https://github.com/deathcap/Burger/tree/mc18 through `bin/generate_enums.java`, with the following caveats:

* Removed biome.height - does anyone know what this was for? (http://minecraft.gamepedia.com/Biome has no mention of a height property, the old pre-1.8 Burger had it but is not clear where in 1.8 the height comes from, if anywhere). In any case, nothing else seems to reference this value, I suspect it is fairly safe to remove.
* Some of the temperatures and rainfall values were not specified in the constructors for some reason (couldn't get Burger to extract), so I used http://minecraft.gamepedia.com/Biome and https://github.com/erich666/Mineways/blob/master/Win/biomes.cpp to fill in the gaps
* "Modified" biomes (128 + original ID) are not included, these appear to be handled differently (highly dynamic, may need to be analyzed manually, or dumped at runtime)
* Other enums not yet updated, will require more work to identify the structure for extraction (so far: https://gist.github.com/deathcap/6efd375e010a005040f8)

### DIFF
--- a/bin/generate_enums.js
+++ b/bin/generate_enums.js
@@ -95,7 +95,7 @@ function biomes(burger, cb) {
     biomeEnum[biome.id] = {
       id: biome.id,
       color: biome.color,
-      height: biome.height,
+      //height: biome.height,
       name: biome.name,
       rainfall: biome.rainfall,
       temperature: biome.temperature,
--- a/lib/enums/biomes.json
+++ b/lib/enums/biomes.json
@@ -2,10 +2,6 @@
   "0": {
     "id": 0,
     "color": 112,
-    "height": [
-      -1,
-      0.4
-    ],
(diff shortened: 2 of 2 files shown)


roblabla:
Is this ready to merge ? It looks good. Concerning "modified biomes", I think whenever we get a biome from the server, doing a simple if (biome > 128) biome -= 128 might work. From what I read on the gamepedia, only the name changes.

deathcap:
:+1: ok to merge (imho the modified biomes should be handled separately, since they inherit the same properties)

PR #200, opened 2015-03-12
Fix map handling

roblabla:
Merging #195 left a couple bugs in the way map is represented. This PR should fix it.

@rom1504 you'll want to use this for your work on making digger work properly.


roblabla:
Heh, being the only active member with commit access to the repo, I'm the only one who can merge my own PR :P

PR #205, opened 2015-03-13
Fix bot.players

rom1504:
This is only the first commit of the fixing of bot.players, maybe don't merge it yet.

What changed : 
* players now have a UUID
* packets often don't give the username anymore, only the UUID

bot.players was previously indexed by the player name, it should now probably be indexed by the UUID. A uuidToUsername function and a usernameToUuid function might be useful.
This means all mention of bot.players need to be changed throughout the code.

### DIFF
--- a/examples/chatterbox.js
+++ b/examples/chatterbox.js
@@ -12,7 +12,7 @@ bot.on('login', function() {
   console.log("settings", bot.settings);
 });
 bot.on('time', function() {
-  console.log("current time", bot.time);
+  //console.log("current time", bot.time);
 });
 bot.on('playerJoined', function(player) {
   bot.chat("hello, " + player.username + "! welcome to the server.");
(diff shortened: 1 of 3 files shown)


rom1504:
In the end I let bot.players be indexed by the username and added a bot.uuidToUsername, that's enough.

Chatterbox.js now works :)

Ready to merge

examples/chatterbox.js
```diff
@@ -12,7 +12,7 @@ bot.on('login', function() {
   console.log("settings", bot.settings);
 });
 bot.on('time', function() {
-  console.log("current time", bot.time);
+  //console.log("current time", bot.time);
```
roblabla: Could you uncomment this ? I have a feeling this isn't intentional :P

rom1504: Actually this is intentional, this display one line every ~200ms in the console, making everything else the chatterbox example display unreadable. You can see several other things are commented in that example for the same reason ;) (not by me)

roblabla: oh ok.

PR #217, opened 2015-03-16
Fix itemFromNotch : fix list and equip in inventory.js

rom1504:
The other commands (toss, craft, use, stop) in inventory.js are still broken, build is still broken in digger.js.


rom1504:
Ok, ready to merge.

PR #225, opened 2015-03-19
Fix dying (fix #224)

### LINKED ISSUE #224: Fixing dying
Dying is currently broken because we forget the player on entity_destroy (https://github.com/andrewrk/mineflayer/blob/master/lib/plugins/entities.js#L222), then when the bot spawn, player_info is not sent again but named_entity_spawn is sent (https://github.com/andrewrk/mineflayer/blob/master/lib/plugins/entities.js#L142).

Removing the line 222 and 223 of entities.js (https://github.com/andrewrk/mineflayer/blob/master/lib/plugins/entities.js#L222) fixes the issue but is it acceptable ? Is it ok to keep in memory the dead players ? Does that include the disconnected players ?

An other way to handle this issue would be to accept having unnamed player, but that break a lot of things and I'm not sure it can work ? Is https://github.com/andrewrk/mineflayer/blob/master/lib/plugins/entities.js#L348 sent at some point to give the username anyway ?

rom1504:
Keep the mapping between an uuid and a username when a player dies

### DIFF
--- a/lib/plugins/entities.js
+++ b/lib/plugins/entities.js
@@ -220,7 +220,6 @@ function inject(bot) {
       entity.isValid = false;
       if(entity.username && bot.players[entity.username]) {
         bot.players[entity.username].entity = null;
-        bot.uuidToUsername[bot.players[entity.username].entity.uuid] = null;
       }
       delete bot.entities[id];
     });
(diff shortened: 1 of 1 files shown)


rom1504:
I think it's ready to merge.

PR #227, opened 2015-03-22
Update items enum for 1.8

Kupferhirn:


rom1504:
Extracted (manually) from http://minecraft.gamepedia.com/Items , for example http://minecraft.gamepedia.com/Dark_Oak_Door for the Dark Oak Door.

A bit difficult to extract automatically hence to be sure there's everything, but it seems good to me.

rom1504:
Thanks :) merging.
I think it would be better to have something automatic to extract these but that will do for now.

PR #230, opened 2015-03-22
Update block enum for 1.8

Kupferhirn:


Kupferhirn:
not ready to merge

rom1504:
Ok I just tested digging without and with the new enum for prismarine.
It doesn't work without, and it works with the new enum.

What's the problem ?

Kupferhirn:
tested and ready to merge

rom1504:
Ok, merging.
(I think some automatic thing https://github.com/andrewrk/mineflayer/issues/229 would be more reliable but until it's done doing it manually works)

PR #235, opened 2015-03-24
Remove the obsolete Block 'add' field, minor cleanup

deathcap:
This was technically part of the API but could only be reached with custom block IDs (>255, unlikely without https://github.com/PrismarineJS/node-minecraft-protocol/issues/114), and it wasn't populated (either in `Block` or `Column`) since the MC 1.8 update. With the existing chunk parsing code, `block.type` is already up to 12-bits (`blockType >> 4`), incorporating the old 4-bit `add` field.


deathcap:
Rebased on top of master

PR #251, opened 2015-04-02
Add sound effect event

deathcap:
Adds a new API event for sound effects, `soundEffectHeard`, comparable to `noteHeard`. 

Included in a new plugin 'sound.js' (the sound effects are more general than block_actions, seemed fitting to keep separate). This PR doesn't include the world_particles (0x28 packet) sound effects, only 0x29 named_sound_effect, but they could be added later in this same plugin.

Known sound effect names: http://wiki.vg/Protocol#Sound_Effect (example: `mob.cow.say`, maybe could somehow include this list in https://github.com/PrismarineJS/minecraft-data but I'm not sure what data would be associated with it, if any - for now the name from the packet is passed directly through to the event).

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -201,6 +201,7 @@
       - ["blockUpdate:(x, y, z)" (oldBlock, newBlock)](#blockupdatex-y-z-oldblock-newblock)
       - ["chunkColumnLoad" (point)](#chunkcolumnload-point)
       - ["chunkColumnUnload" (point)](#chunkcolumnunload-point)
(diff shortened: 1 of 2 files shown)


wtfaremyinitials:
:+1:

roblabla:
I wonder why there are two packets for sound effect...

Anyhow, I'm in favor of adding a "sound API" to mineflayer. :+1:.

zuazo:
AFAIK all particles are graphical effects. Some of them include sounds, but are not sounds. It's a bit like the mobs, they make noise but are not sounds. I think it's okay not to include them :+1:

rom1504:
:+1:  merging

Issue #253, opened 2015-04-04
Plugin message custom payload for "MC|Brand" encoded incorrectly


deathcap:
lib/plugins/game.js:

```javascript
    bot.client.write('custom_payload', {channel: 'MC|Brand', data: new Buffer('vanilla')});
```

from http://wiki.vg/Plugin_channel#MC.7CBrand: this is "vanilla" (**encoded as a UTF-8 string**).

Mineflayer does not encode the buffer as a UTF-8 string (i.e., varint length prefix; `writeString()` in node-minecraft-protocol). This causes at least the [Glowstone](https://github.com/GlowstoneMC/Glowstone) server to choke:

```
20:18:26 [SEVERE] Error while handling PluginMessage(channel=MC|Brand, data=[118, 97, 110, 105, 108, 108, 97]) (handler: PluginMessageHandler)
java.lang.IndexOutOfBoundsException: readerIndex(1) + length(118) exceeds writerIndex(7): UnpooledHeapByteBuf(ridx: 1, widx: 7, cap: 7/7)
	at io.netty.buffer.AbstractByteBuf.checkReadableBytes(AbstractByteBuf.java:1171)
	at io.netty.buffer.AbstractByteBuf.readBytes(AbstractByteBuf.java:677)
	at io.netty.buffer.AbstractByteBuf.readBytes(AbstractByteBuf.java:685)
	at com.flowpowered.networking.util.ByteBufUtils.readUTF8(ByteBufUtils.java:46)
	at net.glowstone.net.handler.play.game.PluginMessageHandler.handleInternal(PluginMessageHandler.java:74)
	at net.glowstone.net.handler.play.game.PluginMessageHandler.handle(PluginMessageHandler.java:39)
	at net.glowstone.net.handler.play.game.PluginMessageHandler.handle(PluginMessageHandler.java:20)
	at com.flowpowered.networking.session.BasicSession.handleMessage(BasicSession.java:80)
	at com.flowpowered.networking.session.BasicSession.messageReceived(BasicSession.java:139)
	at net.glowstone.net.GlowSession.pulse(GlowSession.java:408)
	at net.glowstone.net.SessionRegistry.pulse(SessionRegistry.java:23)
	at net.glowstone.scheduler.GlowScheduler.pulse(GlowScheduler.java:174)
	at net.glowstone.scheduler.GlowScheduler.access$100(GlowScheduler.java:25)
[...]

PR #254, opened 2015-04-04
Add missing varint length prefix to MC|Brand packet (GH-253)

deathcap:
Fixes the "20:18:26 [SEVERE] Error while handling PluginMessage(channel=MC|Brand, data=[118, 97, 110, 105, 108, 108, 97]) (handler: PluginMessageHandler)
java.lang.IndexOutOfBoundsException: readerIndex(1) + length(118) exceeds writerIndex(7): UnpooledHeapByteBuf(ridx: 1, widx: 7, cap: 7/7)" error when connecting to Glowstone servers, as described in https://github.com/andrewrk/mineflayer/issues/253. 

I still think a better fix is possible (possibly, having node-minecraft-protocol know about plugin channel payload structures, or at least calling its writeString() encoding method), especially if later, mineflayer is enhanced to allow setting a custom "brand", but this PR at least fixes the immediate problem in the simplest way possible.

### DIFF
--- a/lib/plugins/game.js
+++ b/lib/plugins/game.js
@@ -21,7 +21,7 @@ function inject(bot) {
     bot.emit('login');
     bot.emit('game');
     bot.client.write('held_item_slot', {slotId: 0});
-    bot.client.write('custom_payload', {channel: 'MC|Brand', data: new Buffer('vanilla')});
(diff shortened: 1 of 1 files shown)


rom1504:
Okay merging, it probably should be handled directly in node-minecraft-protocol later though ;)

PR #255, opened 2015-04-05
added bot.recipesAll()

demipixel:
Also made a minor fix to api.md which was bugging me.

I needed to be able to access the recipe of a crafting table before I had made the planks (using Dutie). This seemed to be the best solution. It just doesn't check on wether the bot has enough materials or not.

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -1181,7 +1181,7 @@ with `metadata`.
  * `craftingTable` - a `Block` instance. If `null`, only recipes that can
    be performed in your inventory window will be included in the list.
 
-#### bot.recipesFor(itemType, metadata, craftingTable)
+#### bot.recipesAll(itemType, metadata, craftingTable)
 
 The same as bot.recipesFor except that it does not check wether the bot has enough materials for the recipe.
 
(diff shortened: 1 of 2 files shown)


doc/api.md
```diff
@@ -1180,6 +1181,10 @@ with `metadata`.
  * `craftingTable` - a `Block` instance. If `null`, only recipes that can
    be performed in your inventory window will be included in the list.
 
+#### bot.recipesFor(itemType, metadata, craftingTable)
```
roblabla: Woops, should be bot.recipesAll ^^

rom1504:
[Recipe.find](https://github.com/andrewrk/mineflayer/blob/master/doc/api.md#recipefinditemtype-metadata) is basically the same thing as your recipesAll.
I'm okay with the api.md change, it would be easier if it was in an other PR (you should use branches on your fork to do different PRs)

rom1504:
We could add a requirementsMetForRecipe boolean parameter to recipesFor if you want instead ?

demipixel:
I didn't want to add a parameters because I still wanted to be able to check for the crafting table (to better organize the possible recipes) as well as make it easier to understand when reading people's codes.

I don't want to use Recipe.find due to reasons of having to pass through a function rather than easily receiving an array. Otherwise it'd just be remaking a function that would always do the same thing.

rom1504:
Okay I guess it doesn't hurt to have one more function for this. Merging.

PR #282, opened 2015-04-29
Delete player from bot.players on leave

### LINKED ISSUE #280: Players list just keeps growing.
It seems like the player list just keeps getting populated but when a user leaves, it is not removed from the players list map. Is it supposed to be like that or is there a possible solution?

AlexKvazos:
Fixes #280

### DIFF
--- a/lib/plugins/entities.js
+++ b/lib/plugins/entities.js
@@ -79,7 +79,7 @@ var mobTypes = {
 };
 
 function inject(bot) {
-  
+
   bot.findPlayer = bot.findPlayers = function(filter) {
     var filterFn =  function(entity) {
       if(entity.type !== 'player') return false;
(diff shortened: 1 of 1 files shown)


lib/plugins/entities.js
```diff
@@ -79,7 +79,7 @@ var mobTypes = {
 };
 
 function inject(bot) {
-  
+
```
AlexKvazos: This is just whitespace that my text editor removes when I save any file.

rom1504:
Oh I didn't see this pull request, and I just fixed it in 39ce38be13677be8264ade2547501fc4d75471c8

AlexKvazos:
Nice, thanks. Quick question, what version should I be using on my package.json so that it pulls the latest mineflayer from this repo and master branch?

rom1504:
You can put "git://github.com/andrewrk/mineflayer.git#master" instead of the version number in your package.json.

rom1504:
@AlexKvazos and I just release 1.1.1 so you can use that ;)

AlexKvazos:
Thank you!

PR #285, opened 2015-05-02
Delete the nmp client when connection is ended

AlexKvazos:
Delete the instance of the client when `end` is called. This allows the GC to clear it from memory when it next runs.

### DIFF
--- a/index.js
+++ b/index.js
@@ -59,6 +59,7 @@ Bot.prototype.connect = function(options) {
   });
   self._client.on('end', function() {
     self.emit('end');
+    delete self._client;
   });
   for (var pluginName in plugins) {
     plugins[pluginName](self, options);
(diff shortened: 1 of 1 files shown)


rom1504:
Yeah but that doesn't delete the rest of the stuff mineflayer stores (chunk, entities,...)
Can't you just delete the whole bot instance when mineflayer end event is emmited ?

AlexKvazos:
Yes deleting the bot instance is also a thing, but it is important to delete any instances that each library makes.

rom1504:
If you don't have a reference to the bot instance anymore, then there will be no reference to _client and it will be GC'd.

AlexKvazos:
Well yeah. I mean, it was just a suggestion. You guys can decide if merging or not.

rom1504:
So, it doesn't seem like it's fixing much, it doesn't make much sense to use mineflayer without _client.

PR #290, opened 2015-05-20
swing arm immediately when digging

gipsy-king:
Nocheat kicks mineflayer if it does not send arm_animation immediately after block_dig. There is a setInterval but that will not get called if the block dig time is less than 350ms.


rom1504:
Yeah I tested it and it looks more natural indeed. I don't know why we didn't see that before.

Thanks !

Issue #284, opened 2015-04-30
Cannot read property 'id' of undefined at entities.js


AlexKvazos:
https://github.com/andrewrk/mineflayer/blob/c8a3c621517b4a6fe1fce3d818c51aa76254bd67/lib/plugins/entities.js#L315

```
server-0 (err): TypeError: Cannot read property 'id' of undefined
server-0 (err):     at Client.<anonymous> (/home/MinecraftChat/node_modules/mineflayer/lib/plugins/entities.js:315:33)
server-0 (err):     at Client.emit (events.js:107:17)
server-0 (err):     at afterParse (/home/MinecraftChat/node_modules/mineflayer/node_modules/minecraft-protocol/dist/client.js:97:10)
server-0 (err):     at parseNewStylePacket (/home/MinecraftChat/node_modules/mineflayer/node_modules/minecraft-protocol/dist/protocol.js:1080:5)
server-0 (err):     at prepareParse (/home/MinecraftChat/node_modules/mineflayer/node_modules/minecraft-protocol/dist/client.js:113:9)
server-0 (err):     at afterParse (/home/MinecraftChat/node_modules/mineflayer/node_modules/minecraft-protocol/dist/client.js:101:5)
server-0 (err):     at parseNewStylePacket (/home/MinecraftChat/node_modules/mineflayer/node_modules/minecraft-protocol/dist/protocol.js:1080:5)
server-0 (err):     at prepareParse (/home/MinecraftChat/node_modules/mineflayer/node_modules/minecraft-protocol/dist/client.js:113:9)
server-0 (err):     at Socket.<anonymous> (/home/MinecraftChat/node_modules/mineflayer/node_modules/minecraft-protocol/dist/client.js:124:5)
server-0 (err):     at Socket.emit (events.js:107:17)
```

rom1504:
When did this happen ?
Any way to reproduce it ?

AlexKvazos:
I believe this happens when any of the enchantment or experience events are fired.

AlexKvazos:
There was another stack trace before this one, let me go to my error logs and find it.

PR #291, opened 2015-05-23
Check effect exists before delete

Pietro210:
Fixes [#284](https://github.com/andrewrk/mineflayer/issues/284)

I was unable to reproduce the issue on a Vanilla server but it happened on a modded server right after the player spawned near a Beacon and the server removed the Speed effect.

### DIFF
--- a/lib/plugins/entities.js
+++ b/lib/plugins/entities.js
@@ -276,6 +276,14 @@ function inject(bot) {
     if(effect) {
       delete entity.effects[effect.id];
     }
+    else {
+      // unknown effect
+      effect = {
+        id: packet.effectId,
+        amplifier: -1,
+        duration: -1,
+      };
+    }
     bot.emit('entityEffectEnd', entity, effect);
   });
 
--- a/lib/plugins/entities.js
+++ b/lib/plugins/entities.js
@@ -273,7 +273,9 @@ function inject(bot) {
     // remove entity effect
     var entity = fetchEntity(packet.entityId);
     var effect = entity.effects[packet.effectId];
-    delete entity.effects[effect.id];
+    if(effect) {
+      delete entity.effects[effect.id];
+    }
     bot.emit('entityEffectEnd', entity, effect);
   });
 


rom1504:
The only way not checking would be a problem is if the remove entity effect comes before the entity effect packet.
That can come from out-of-order deserialization and that's a bug we just fixed in a (not released yet) node-minecraft-protocol change.

Let's see if that's the problem before adding that check (that is a bit a hack)

Pietro210:
That's exactly what happens in my case, the remove entity effect packet comes before the entity effect packet.

I tried again using the master branch of node-minecraft-protocol instead of the 0.13.4 release but the issue persists.

roblabla:
TBH, this should be merged anyway as we shouldn't crash if the server tries to remove an effect we didn't have. It's similar to scoreboard handling : notchian client used to crash if the server tried to remove a non-existing scoreboard entry. Instead of crashing, it should either ignore it, or gracefully error out (emit an error event), not crash in an uncatchable way :)

rom1504:
Ok, if that still happen with node-minecraft-protocol fix, it's possible the out-of-order problem actually from from spigot.

So ok I agree this should be checked in mineflayer. The problem with how it is fixed in this PR is this is going to emit the event entityEffectEnd with an undefined effect.
Should we still emit the event but with a { id:packet.effectId } as effect (and no amplifier and duration) or should we not emit that event at all ?

Pietro210:
I didn't want to ignore the packet completely because I thought that some server mods with custom effects might send it, either accidentally or on purpose, and the user should know about it.

Emitting the event with an undefined effect meant something like "the server removed an entity effect but we don't know that effect" but I see that it might be a bad idea.

I pushed another commit based on @rom1504 's suggestion, now if the server removes an unknown effect the bot emits the event with a special effect object with both amplifier and duration set to -1

It might not be the best solution but it avoids crashes and doesn't break existing code that tries to access the effect's properties.

rom1504:
Yeah that's a good enough solution.
Thanks for the PR !

Issue #262, opened 2015-04-06
Check the callback of the methods are called at the right time


rom1504:
Some callback are called too early or too late. I would be nice to check which one. Doing this automatically would be better (adding a test).
Known callback problem : 
- [ ] bot.equip callback return too soon
- [ ] placeBlock doesn't have a callback (also see #261 )

rom1504:
bot.lookAt (https://github.com/andrewrk/mineflayer/blob/master/doc/api.md#botlookatpoint-force) should have a callback, it's not instantaneous

rom1504:
the callback of bot.lookAt should be called when the move event id emmited = when the position/look has been sent.

Relevant for things depending on where you look : shooting an arrow, moving a vehicle (https://github.com/andrewrk/mineflayer/blob/master/doc/api.md#botmovevehicleleftforward)
it's especially important for arrows : if you shoot before aiming properly it will shoot at the wrong location.

rom1504:
An other way might be sending directly a http://wiki.vg/Protocol#Player_Look packet.

Check if this is not a problem to send that kind of packet too often.

rom1504:
Apparently NCP (NoCheat+) doesn't like instant updates, and mineflayer currently doesn't do any, maybe we should keep doing that.

rom1504:
bot.look and bot.lookAt now have a callback, so that part of this issue is fixed.

ViRb3:
Any update on `bot.equip callback return too soon`? It's causing some funky behavior here...
#326

ViRb3:
Every second time I log the bot on my server he has a pickaxe in his hand, but mineflayer truly believes it's cobblestone. I'm using the following code, modified from the scaffold plugin:
```
function equipBuildingBlock(cb) {
        console.log("Equipping building block")
        // return true if we're already good to go
        if (bot.heldItem && scaffoldBlockTypes[bot.heldItem.type]) { // says cobblestone
            console.log("Already equipped");
            cb();
            return true;
        }

        var scaffoldingItems = bot.inventory.items().filter(function (item) {
            return scaffoldBlockTypes[item.type];
        });

        var item = scaffoldingItems[0];
        if (!item) {
            changeState('stop', 'error');
            return false;
        }

        bot.equip(scaffoldingItems[0], 'hand', function (err) {
            if (err) {
                changeState('stop', 'error', err);
            }

            console.log("Callback in 500");
            setTimeout(function () {
                cb();
            }, 500);
        });
[...]

PR #292, opened 2015-05-24
Look immediately when digging or activating a block

Pietro210:
Somewhat related to #262 

I was having some trouble digging blocks on a modded server and it turned out that the server actually checks if the player is looking directly at the block when digging or interacting with blocks and it fails otherwise.

Calling look() or lookAt() before digging or interacting with a block currently doesn't work in this case, even if we set force to true, because the position update is actually sent up to 50 ms after the call and the server receives the block_dig or block_place packet before it knows we moved.

This also affects functions like activateBlock() and it causes problems when opening chests or interacting with some other blocks, it does not happen on a Vanilla server though.

A simple workaround would be to always wait for the 'move' event before proceeding but I think it's better to send the position update immediately because otherwise the documentation is misleading, it says:

> force - If present and true, skips the smooth server-side transition. Specify this to true if you need the server to know exactly where you are looking, such as for dropping items or shooting arrows.

but even when force is true you currently have to do something like this:

```javascript
bot.look(yaw, pitch, true)
bot.once('move', function() {
  // drop item, shoot arrow...
})
```

### DIFF
--- a/lib/plugins/digging.js
+++ b/lib/plugins/digging.js
@@ -13,7 +13,7 @@ function inject(bot) {
   function dig(block, cb) {
     if(bot.targetDigBlock) bot.stopDigging();
     cb = cb || noop;
-    bot.lookAt(block.position);
+    bot.lookAt(block.position, true);
     bot._client.write('block_dig', {
       status: 0, // start digging
       location: block.position,
--- a/lib/plugins/inventory.js
+++ b/lib/plugins/inventory.js
@@ -96,7 +96,7 @@ function inject(bot) {
 
   function activateBlock(block) {
     // TODO: tell the server that we are not sneaking while doing this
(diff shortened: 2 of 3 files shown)


rom1504:
That means we might send position updates faster than every 50ms. That's probably not a problem though.
Are you sure "force looking" is ok though ? look at that line https://github.com/andrewrk/mineflayer/blob/master/lib/plugins/physics.js#L238
Is the vanilla client properly displaying the bot look with this change ?

rom1504:
I did that in a7ba46e7473a73a7dc428476fde6554a8fd754c4 .
That's a better fix : no force move needed, no sending position faster than every 50ms needed.
That should fix your pb too on your modded server ;)

Pietro210:
Nice, thanks!

PR #294, opened 2015-05-28
Prevent crash when an unknown entity is spawned

Pietro210:
This is similar to what happened here: #291

There is no check that ensures we actually have the entity data before we use them. When an unknown entity is spawned, either because it was recently introduced in a new Minecraft update or because the entity type ID is used by a third-party mod, the bot crashes.

Also, is there an issue with PrismarineJS/minecraft-data or am I missing something? It seems that some entity IDs in [entities.json](https://github.com/PrismarineJS/minecraft-data/blob/master/enums/entities.json) are different from the ones reported [here](http://wiki.vg/Entities#Objects)

In my case the server sent a spawn_entity packet with type = 78, which should be an Armor Stand, but there are no entities with the same ID in that list, were some IDs changed in a protocol update?

Maybe this should be a separate issue, though.

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -359,6 +359,7 @@ Choices:
  * `object`
  * `global` - lightning
  * `orb` - experience orb.
+ * `other` - introduced with a recent Minecraft update and not yet recognized by Mineflayer or used by a third-party mod
 
 #### entity.username
 
--- a/lib/plugins/entities.js
+++ b/lib/plugins/entities.js
@@ -128,13 +128,19 @@ function inject(bot) {
   bot._client.on('spawn_entity', function(packet) {
     // spawn object/vehicle
     var entity = fetchEntity(packet.entityId);
-    entity.type = 'object';
     var entityData = entitiesData[packet.type];
-    entity.objectType = entityData.displayName;
-    entity.displayName = entityData.displayName;
-    entity.entityType = entityData.id;
-    entity.name = entityData.name;
-    entity.kind = entityData.type;
+    if(entityData) {
+      entity.type = 'object';
+      entity.objectType = entityData.displayName;
+      entity.displayName = entityData.displayName;
+      entity.entityType = entityData.id;
+      entity.name = entityData.name;
+      entity.kind = entityData.type;
+    } else {
+      // unknown entity
+      entity.type = 'other';
+      entity.entityType = packet.type;
+    }
     entity.position.set(packet.x / 32, packet.y / 32, packet.z / 32);
     entity.yaw = conv.fromNotchianYawByte(packet.yaw);
     entity.pitch = conv.fromNotchianPitchByte(packet.pitch);


rom1504:
[entities.json](https://github.com/PrismarineJS/minecraft-data/blob/master/enums/entities.json) is extracted from the [wiki page of entities](http://minecraft.gamepedia.com/Data_values/Entity_IDs) which seem more up to date than the [wiki.vg one](http://wiki.vg/Entities#Objects).
The wiki page says armor stand is now 30 and there are no id 78 entity. What version of the minecraft server are you connecting to ?
Anyway, the [wiki page](http://minecraft.gamepedia.com/Data_values/Entity_IDs) should be updated if some information on there are out dated.

There shouldn't be any entity for which we don't know the id so I'm not sure having a "other" entity type is a correct fix. It's different from #291 because entitiesData is static unlike entity.effects which is dynamic.

Pietro210:
I noticed the same issue on 2 different servers. The first runs Minecraft 1.8 (protocol version 47) and the other Minecraft 1.7.4 (protocol version 4)

I will probably test this on Minecraft 1.8.3/6 too just to be sure but I guess that if the ID changed Mineflayer will handle it just fine.

Both send 78 as the entity type when an Armor Stand is spawned nearby, but the same thing also happens with other entities listed on that page such as thrown potions, firework rockets, etc.

I guess it's reasonable that the most recent version of minecraft-data only supports the latest Minecraft protocol, but if the entity IDs can change at any time then Mineflayer is basically not backward compatible, it makes sense if it's by choice but if that's the case maybe it should be noted somewhere in the documentation so it's easier to debug this kind of issues :)

The only problem with this though is that some mods can potentially implement custom entities each with it's own unique ID and this can make the bot crash even on a server running the latest version, if the entity data don't exist we could ignore the event completely but I thought it made more sense to let the user handle the unknown type the way he prefers.

rom1504:
Yeah mineflayer, node-minecraft-protocol and minecraft-data only support the last version (1.8), in the long term we plan to support multiple versions but that's pretty complicated (look at the issues in these repo concerning that for more info)

So yeah we need to figure out if armor stand is 78 or 30 and update the minecraft wiki so minecraft data will be updated.

About the doc : there is a "Supports Minecraft 1.8.3." in mineflayer's readme, should we add something like "does not support multiple protocol version (yet)" ?

Oh custom entities can exist in mods ? If so, yeah your fix makes sense.

roblabla:
mods can customize everything, really, entity IDs, block IDs, effects, sounds, etc... We should assume all the data can be incomplete (in that we're missing some IDs), but the data we have should be right (mods rarely modify the behavior of existing stuff, and if they did the authors should be hit with a stick).

Pietro210:
@rom1504 I meant something like:
```markdown
# Support
[Release 1.2.0](#): Minecraft 1.8.3
[Release 1.1.1](#): Minecraft 1.7.2 to 1.7.4

Support to Minecraft 1.8.6 is coming soon.
```
of course this is just an example.

Since mods can always mess with IDs or at least add new ones I still think this should be merged just to be safe, there may be some other missing checks too but this is the only one I noticed so far.

roblabla:
We should support everything at a certain protocol leve, e.g. right now we (should have) support for mc version 1.8.1 up to 1.8.6, since they all share the same protocol and all that changed were a couple bugfixes left and right.

rom1504:
Ok so I checked : the data on the minecraft wiki is the correct one (so the data on minecraft data is correct).
But still I agree we should assume the data we have is incomplete to be able to handle mods, merging !

PR #296, opened 2015-05-31
Examples refactoring and cleanup

Pietro210:
I made a few changes in this one so here's a summary of what I did:

* Removed some code that was out of context, now each example should contain only what's relevant (there were some event listeners or chat commands that were probably added as a test)
* Added a description on top of each file so it's clear what each example does
* Changed the way chat commands are parsed and separated most reactions to commands in smaller  functions so it's easier to see what each command does without reading all the code
* Added comments explaining the syntax of the most complicated commands
* Reordered some code so functions are easier to find
* Separated a few other functions and statements so each example is easier to follow
* Fixed a few style inconsistencies

I tested every example manually to check that they were working and everything seems to be fine.

I think there are a few other optimizations that can be done, especially since there's some repetitive code that can be put in a reusable functions instead and some code that could be shared between examples but I didn't want to do too much at once. If everything looks good to you I might do that before merging.

### DIFF
--- a/examples/chatAddPattern.js
+++ b/examples/chatAddPattern.js
@@ -1,6 +1,19 @@
+/*
+ * How to add chat patterns to support your own server
+ *
+ * Adding more chat patterns is really simple.
+ *
+ * Use the Vanilla Launcher with visibility set to keep the Launcher open
+ * so you can see the raw messages.
+ *
+ * Copy the raw message into regexr.com or otherwise write a regex statement
+ * to match the username and the message.
+ *
+ * Add pattern, chat type, and an optional description with bot.chatAddPattern()
+ *
+ * Below you can see an example for the skyblock.net server
+ */
 var mineflayer = require('../');
-
-//Example for matching chat on skyblock.net.
 
 if(process.argv.length < 4 || process.argv.length > 6) {
   console.log("Usage : node chatAddPattern.js <host> <port> [<name>] [<password>]");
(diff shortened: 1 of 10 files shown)


rom1504:
Comments are nice, and creating new functions is nice too indeed !

Some problems :
* I see you uncommented some stuff in chatterbox.js : isn't it too spammy ?
* the `bot.on('experience` in chest.js was useful for the enchantment table (some enchantment can only be done at some level)
* I don't really like switch over if / else if, but ok I guess we can do that. What's the point of these switch true in chest.js though ?

Ok so my main complaint is about these switch case, imho if/else if are better. But we can improve on if/else if with something like this : 
```js
var commands = {
 /^list$/:listInventory,
 /^toss \d+ \w+$/:tossItem
}
for(var regex in Object.keys(commands))
{
  var results=message.exec(regex);
  if(results!=null)
  {
   results.shift();
   var f=commands[regex];
   f.apply(null,results);
   break;
  }
}
```
It makes the examples a little bit more complex but maybe that's worth it ?

Pietro210:
Just made a few more changes.

@rom1504:
* I tried chatterbox.js and it seems fine but I tested it on a server that was pretty much empty, it may cause some problems on a server with a lot of players but I think in that case it shouldn't be used anyway since it's just an example.
* You're right, I didn't notice that, I just put it back
* I used `switch(true)` when one or more strings are matched against a regex to keep things consistent between examples.

I personally like switches more because I found them easier to read, but I guess your idea might be even better. I don't think there's a way to use a regex as key but something like this should work:

```javascript
var commands = {
  '^list$': sayItems,
  '^toss (\\d+) (\\w+)$': tossItem,
};

bot.on('chat', function(username, message) {
  if (username === bot.username) return;
  parseCommand(message);

  function parseCommand(message) {
    var match;
    for (var key in commands) {
      match = new RegExp(key).exec(message);
      if (match) break;
    }
    commands[key].apply(null, match.slice(1));
    return match;
  }
});
```

rom1504:
I do intend to merge this but I need to look at some of the stuff, maybe add a commit or two, didn't have time to do that recently ;)

rom1504:
I finally took time to merge this !
Thanks for this PR @Pietro210 
In the end, the swich case look okay, we might put in place that solution with an array of commands later.

I added a commit to put findBlock in blocks.js, it's better to have the function in only one place in mineflayer and it's a good thing to have kept a similar API to mineflayer-blockfinder function.

PR #308, opened 2015-07-11
arm swing on placing block, and look at center of block on activate

gipsy-king:
nocheat is offended by the missing arm swing and often does not allow
for opening chest if looking at the corner of it.


rom1504:
Thanks ! I added a commit to use offset instead of clone().translate.

PR #311, opened 2015-07-15
Add clickWindow support to ContainerWindow

Gnomesley:
Small fix to allow clicking inside non-Vanilla windows.


rom1504:
Thanks ! I added a commit improving the code a little.

Issue #228, opened 2015-03-22
Handle new kinds of bounding boxes


rom1504:
For example a carpet is not really a block but we shouldn't fall through it.
See also : 
- fences
- doors
- bars

Currently we'll consider a carpet has an empy bounding box but that's not perfect.

deathcap:
also slabs, stairs, snow (etc.)? simplest non-cubic bounding box is of blocks of varying heights:

http://minecraft.gamepedia.com/Solid_block#Heights


Height	| Block types
--- | ---
2	| Door, Iron Door
1 1/2	| Fence, Cobblestone Wall
15/16	| Cactus
7/8	| Soul Sand, Chest, Ender Chest, Trapped Chest, Brewing Stand (stand part)
13/16	| End Portal Frame (without Eye of Ender)
3/4	| Enchantment Table, Cocoa Bean (plant), Head (on wall)
5/8	| Inside of Hopper
9/16	| Bed
1/2	| Bottom slab, Head (on ground), bottom part of Stair, Cake
3/8	| Daylight Sensor, Flower Pot
5/16	| Inside of Cauldron
3/16	| Bottom Trapdoor
1/8	| Brewing Stand (base), Redstone Repeater, Redstone Comparator, Snow
1/16	| Carpet
1/64	| Lily Pad

demipixel:
How do we want to handle all these? Different bounding boxes can be an issue, but things like doors or fences which can change depending on blocks or state, or even brewing stands which have multiple "bound **boxes**" will have an issue. One idea was to make bounding box an area of boxes, ex:
```
[ { x: 0, y: 0, z: 0, width: 1, length: 1, height: 0.5 } ]
```
Not ideal but better than nothing. Might cause it to get a bit slow.

As for actually supporting them, we have SOME support right now. We test for all blocks in the area (surrounding the character) but the [current method](https://github.com/andrewrk/mineflayer/blob/47517943f0d78d1fcb18712197ae85a0ac0e8078/lib/plugins/physics.js#L181) returns a bool. Instead we're going to need to return the bounding box(es). From there we can calculate the height each player would be on any given block (full block VS on top of brewing stand). Then limit the player's Y to the highest one (similar for X and Z).

[...]

PR #320, opened 2015-08-12
[WIP] start work on boundingBox (#228) : soulsand seems to work

rom1504:
Creating a pull request to talk about how to implement this.
I tried some stuff with these commits, but it doesn't seem very reliable, the server still sends some position correcting packets.
The second point is that this only handle soulsand, it should handle all bounding boxes types, but I think that should only been done when soulsand works.

The general idea of that implementation is finding the highest block under us and putting the bot at that block height when colliding. See the new collisionInRangeHighest function.

### DIFF
--- a/lib/plugins/physics.js
+++ b/lib/plugins/physics.js
@@ -68,6 +68,12 @@ function inject(bot) {
     doPhysicsTimer = null;
   }
 
+  function typeToBlockHeight(type)
+  {
+    if(type===88) return 7/8;
+    else return 1;
+  }
+
   function nextFrame(deltaSeconds) {
     if(deltaSeconds < EPSILON) return; // too fast
     var pos = bot.entity.position;
@@ -128,7 +134,7 @@ function inject(bot) {
 
     // limit speed
     var currentMaxGroundSpeed;
-    var underBlock=bot.blockAt(pos.offset(0,-1,0));
+    var underBlock=bot.blockAt(new Vec3(pos.x,Math.ceil(pos.y-1.002),pos.z));
     if(underBlock && underBlock.type===88)
       currentMaxGroundSpeed=physics.maxGroundSpeedSoulSand;
     else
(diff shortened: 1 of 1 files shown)


lib/plugins/physics.js
```diff
@@ -128,7 +134,7 @@ function inject(bot) {
 
     // limit speed
     var currentMaxGroundSpeed;
-    var underBlock=bot.blockAt(pos.offset(0,-1,0));
+    var underBlock=bot.blockAt(new Vec3(pos.x,Math.ceil(pos.y-1.002),pos.z));
```
Gjum: Yay, magic numbers! :)

rom1504: Haha I know :D 
That's because of https://github.com/andrewrk/mineflayer/blob/e3573541a5caf8286e07c50a1c61fa0f5d94b9c0/lib/plugins/physics.js#L155
So the bot position is a little bit too high, removing 0.002 gives the correct ceiling.

Not sure if there's a better way here, probably.

Gjum: Ah, I understand. See, if there was some global constant for this, it would be easier to understand where the 0.002 comes from and what it relates to.

rom1504: Yeah I might do that. I think it's possible 0.001 might be enough even.
The ceiling is needed because now it possible for pos.y to be 63+7/8, which correctly gives 64 with ceiling, but for normal blocks we sometimes have pos.y = 64.001 which should be ceiled to 64, not 65. 
Maybe a MARGIN_ABOVE_GROUND constant would make sense.

Gjum: Sounds good. Stuff like this should be documented (at least by using constants with talking names), or the code quickly gets difficult to understand.

rom1504:
We probably need something like 
`function correct(pos){return new Vec3(pos.x,Math.ceil(pos.y-0.002),pos.z)}` for users.

rom1504:
Yeah indeed the onGround flag is not set properly for soulsand. I'm not sure if that's the problem, but I'll try to fix that to see if that changes anything.

rom1504:
doesn't really work
Let's open a new pr when this gets actually implemented someday

PR #344, opened 2015-11-30
Changed jumpSpeed from 9.0 to 8.0.

jakibaki:
It works better with some anti cheat systems. So I assume it is nearer to the vanilla client.


rom1504:
Why did you close it ? I didn't have time to test this yet

jakibaki:
I noticed that the bot has sometimes problems jumping on blocks with the new value.

rom1504:
maybe see https://github.com/andrewrk/mineflayer/issues/327 , that should be fixed

PR #350, opened 2015-12-26
Added Scoreboard functionality

jakibaki:


rom1504:
Nice !
A few things that can be improved:
* adding some doc about bot.scoreboard content in doc/api.md
* emiting an event (bot.emit("someEvent")) when the scoreboard gets open
* I think it would be better to define the object manually instead of just copying the whole packet. So we can easily limit api changes when Minecraft packets change (instead of doing =packet , do = {someField:packet.someField,...}
* some minor style issues : `new Object()` is the same as {}

rom1504:
Oh and what would be nice too is adding a simple example with a /scoreboard command so people know how to use this and we can check it works.

rom1504:
Did you test this ?
An example would really be useful because I'm not sure how to test this, I never used scoreboards.

rom1504:
I made the example and changed a bit the code in c3f67f98c7002c8aad720fde9ad280bde85dd4ac

Thanks !

Issue #327, opened 2015-08-31
Bot cannot jump repeatedly


Darthfett:
When navigating using mineflayer-navigate, the bot is unable to jump as fast as necessary, due to physics.js preventing jumping from happening too quickly:  https://github.com/andrewrk/mineflayer/blob/master/lib/plugins/physics.js#L99

This bug causes this: https://github.com/Darthfett/helperbot/issues/41

rom1504:
Ah, good point !
That check was added to prevent the bot to jump to quickly was added to fix andrewrk/mineflayer#114 (which it did).
But apparently that's not exactly the correct fix.

We need to find the right way to allow both things to work correctly.

Darthfett:
I wonder if physics.js is setting the bot.entity.onGround too quickly?  It
seems like the reason constantly jumping wouldn't work is because the
server doesn't realize we landed.

It would probably be good to look at the positions sent by the Notchian
client and mineflayer client and compare.  Maybe we could send a second
position update with the on-ground position when we are supposed to have
landed.

Issue #114, opened 2013-02-03
jumping for a long time hurts the bot


vogonistic:
Can be reproduced by leaving `bot.setControlState('jump', true)` on until the bot dies with the text: `bot hit the ground too hard`

rom1504:
I just checked, this is still an issue, setting bot.setControlState('jump', true); : 
* make the bot not touch the ground : I don't know whether this is wrong or not
* after like 30s the bot die with "fell from a high place"

rom1504:
do we use that http://minecraft.gamepedia.com/The_Player#Movement 
"The maximum height a player can jump without the "jump boost" effect is about 1.252 blocks." 
?

rom1504:
Also see http://minecraft.gamepedia.com/Effect

roblabla:
I don't think we support status effects. It was added with 1.7, with the beacons, IIRC.

rom1504:
Yeah, but we should still jump only a maximum of 1.252 blocks. I'm pretty sure physics.js jump implementation doesn't do that.

PR #355, opened 2016-01-09
"fix" for #327

rom1504:
There's probably a better way to do it, but that's a quick fix (that will break #114 again) for people that need it


rom1504:
this is just stupid, let's do something else sometimes.

PR #356, opened 2016-01-12
Remove requireindex, for browserify support

deathcap:
Not the most ideal fix but I couldn't find any other feasible solution unfortunately. Background in https://github.com/deathcap/wsmc/issues/19#issuecomment-170791081 but long story short, this change allows using `require('mineflayer')` in a browserified environment.

Previously I was using `require('mineflayer/lib/block')` etc. and not accessing the mineflayer index.js at all, but this changed in https://github.com/PrismarineJS/mineflayer/commit/ca365c3c1bbdb39cf6632b71e7f3d6363d36c2ae so I tried to use `require('mineflayer').Block` instead but then it runs the unbrowserifiable requireindex module. With this PR, index.js can be browserified.

Another option is to include a separate browser.js listing the modules and keep requireindex in the main index.js, but this means it has to be maintained separately. I think making the change to index.js is the simplest solution, although it does mean that this file will have to be updated when new plugins are added. Open to other suggestions but this is the best I could come up with so far (see https://github.com/deathcap/wsmc/issues/19#issuecomment-170791081)

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,8 +2,34 @@ var mc = require('minecraft-protocol');
 var EventEmitter = require('events').EventEmitter;
 var util = require('util');
 var path = require('path');
-var requireIndex = require('requireindex');
-var plugins = requireIndex(path.join(__dirname, 'lib', 'plugins'));
+var plugins = {
+  bed: require('./lib/plugins/bed'),
+  block_actions: require('./lib/plugins/block_actions'),
+  blocks: require('./lib/plugins/blocks'),
+  chat: require('./lib/plugins/chat'),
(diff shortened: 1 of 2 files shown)


deathcap:
For what it's worth, I faced a similar problem with voxel-plugins (which also runs under browserify). Various tricks with `require()` are possible but it is more straightforward and easier to understand in my opinion to make each require explicit (https://github.com/voxel/voxel-plugins/issues/8).

rom1504:
Yeah, I think it's a good idea to get rid of dynamic requires in general anyway. It's not really a big effort to add a line in index.js for a new plugin.

PR #366, opened 2016-01-27
remove chat filter

rom1504:
So I can't see what was the point of this filter. Without it echo.js can correctly parse and send:
* '
* "
* \"
* test
* 我

Merging this soon if no one can figure out if this filter has still any use.


deathcap:
@rom1504 maybe the purpose of this filter was to block control characters? https://en.wikipedia.org/wiki/ANSI_escape_code#Sequence_elements

PR #382, opened 2016-02-19
https://github.com/roblabla/ProtoDef/pull/50 : update to new numerical names

rom1504:


rom1504:
omg I need to update all these things that want 0.x now...

I guess I'll finish later.

PR #442, opened 2016-07-22
fix tiny typo at "noteHeard"

Atmatm6:
"a Block instance, the block at emitted the noise"
became
"a Block instance, the block that emitted the noise"


rom1504:
Thanks !

Atmatm6:
You're welcome! :smile:

PR #447, opened 2016-08-17
Update example with more thorough explanation

Corgano:
Added more thorough explanations on what parameters are passed to the function and how to set up custom events via custom patterns


Corgano:
this is how i submit changes right? can't remember git

rom1504:
:+1:

PR #451, opened 2016-09-29
actionBar event

ArcticZeroo:
Adds an actionBar event emitted when the client receives an action bar message. Previously, this type of message would emit a 'message' event with no indication that it is any different.


rom1504:
thanks !

PR #453, opened 2016-10-05
villager trading support

plexigras:


rom1504:
looks like we need a new nmp release for this, will do

examples/trader.js
```diff
@@ -0,0 +1,331 @@
...
+      text += ' enchanted with ' + (ench || StoredEnchantments).value.value.map(function (e) {
+        var lvl = e.lvl.value;
+        var id = e.id.value;
+        return entchantment(id).displayName + ' ' + lvl;
+      }).join(' ');
+    }
+  }
+  return text;
+  function entchantment(id) {
+    var entchantments = [
```
rom1504: this should go in minecraft-data

rom1504:
Could you use camelCase instead of snake_case ? (it's already the case for most variables and fields, but for some reason some are in snake_case (for example `max_tradeuses`))

Issue #113, opened 2013-01-29
Interact with Entities


andrewrk:
#### Unable to open minecart chest
it's an entity not a block

---
Edit:

Interacting with all entities needs to be implemented, including interacting with villagers
- Changed by @nevercast

rom1504:
related to #179

Gjum:
Is this working in the latest version?

rom1504:
> interacting with villagers

This certainly isn't implemented.

Gjum:
There could be a clear checklist of things that still need work.

rom1504:
done !

PR #455, opened 2016-10-08
added support for Minecart Chest windows

plexigras:
a solution to https://github.com/PrismarineJS/mineflayer/issues/113


rom1504:
Nice !

PR #457, opened 2016-11-08
Add bot.canSeeBlock()

Nixes:
This pull request adds bot.canSeeBlock and some other methods it requires. Currently everything was added into blocks.js but there might be better places to add these functions.

bot.canSeeBlock() returns true when the bot is within visible range of the block parameter and false otherwise.

I've tested this on my own setup and it seems to work as described.

### DIFF
--- a/lib/plugins/blocks.js
+++ b/lib/plugins/blocks.js
@@ -164,7 +164,7 @@ function inject(bot) {
       na=a.plus(v);
       // check that blocks don't inhabit the same position
       if(!na.floored().equals(a.floored())) {
-        // check block block not transparent
+        // check block is not transparent
         if(blockIsNotEmpty(na)) return false;
       }
       a=na;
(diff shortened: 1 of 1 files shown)


lib/plugins/blocks.js
```diff
@@ -147,6 +147,41 @@ function inject(bot) {
...
+      a=na;
+    }
+    return true;
+  }
+
+  // if passed in block is within line of sight to the bot, returns true
+  // also works on anything with a position value
+  function canSeeBlock(block) {
+    // this emits a ray from the center of the bots body to the block
+    if(visiblePosition(bot.entity.position.offset(0,bot.entity.height*0.5,0),block.position) ) {
```
kenny-evitt: Isn't the height, if the bot player is standing or otherwise upright, relative to the Z (third) dimension? If so, shouldn't this be:

```
... bot.entity.position.offset(0,0,bot.entity.height*0.5) ...
```

Also, why half the bot entity's height? Why not the full height or 80% (roughly where the eyes would be)?

kenny-evitt:
This is a great idea.

lib/plugins/blocks.js
demipixel: Up in Minecraft is on the Y axis, not Z. Not sure about the half-bot-height thing.

Nixes: The half bot height is to calculate the center of the bot on the Y axis. It might be better to have this come from the camera height in a vanilla minecraft client but I can't find current info on it,

Nixes: Also if we do that we should probably have a central constant for it, and I'm not sure where that would go.

rom1504:
merged https://github.com/PrismarineJS/mineflayer/commits/master

thanks !

Issue #432, opened 2016-06-20
Using Minecraft 1.7.10?


dannysmc95:
Is there a way to get your bot to use Minecraft 1.7.10? and modded? I assume it should just connect as it doesn't need the mods? But I have no idea how to get it to work with 1.7.10...?

Thanks in advanced,

rom1504:
Mineflayer currently support only 1.8.
The underlying network library node-minecraft-protocol support multiple versions including 1.7. So we could consider bringing multiple versions support in mineflayer. But that would require some work, there are things that are different in 1.7.
You could try setting the "version" option to 1.7 in createBot, but that will only make node-minecraft-protocol use the 1.7 protocol, it probably will cause problem with mineflayer itself.
Feel free to try to implement the necessary changes in a new branch (in a fork)

dannysmc95:
Oh I see, that is a shame really! Well it would be amazing if you could get this to work, even spoof the mods list, so we can connect to modded servers, because if we are never displaying them, we should never really need the mods right?

Could you show me an example sorry?

roblabla:
The problem with mods is that some mods modify the protocol in ways that, if we are not aware of, break our ability to communicate properly with the server. It's possible to spoof the modlist with [node-minecraft-protocol-forge](https://github.com/PrismarineJS/node-minecraft-protocol-forge) but we never tried to use mineflayer with it AFAIK. But you can give it a try, it may work :).

rom1504:
mineflayer now has cross version supports from 1.8 to 1.11.
We could try adding 1.7 back.

extremeheat:
Seems still valid? https://github.com/PrismarineJS/node-minecraft-protocol supports it, only changes needed are to diff 1.8 and 1.7 and add handle the packet structure changes here

dgo2001:
What would need to be modified from the 1.8 files to make them compatible with 1.7.10? Is there any key differences between the versions? I would really like to help.

rom1504:
Open a 1.7 PR and follow the instructions above the fix the issues
You can also look at the dozen of merged PRs adding support for versions in mineflayer

PR #463, opened 2016-12-22
Clarify supported versions

kenny-evitt:
First off, and unrelated to this PR, thanks so much for this project – it's fantastic!

As for this PR, I had a hell of a time getting the simplest bot to visibly* connect/login/spawn to a game. I'm using [Gamocosm](https://gamocosm.com/) to create my Minecraft servers and the only 'version' of the Minecraft server software for which I finally got a bot to connect was "Forge (1.8.8-11.14.4.1581)". Version 1.10.2 and 1.7.10 of the "vanilla" Minecraft server didn't work; nor did Cuberite (whatever version is installed).

I definitely didn't interpret the feature "Supports Minecraft 1.8.9" to imply that that was the *only* version supported. [This comment](https://github.com/PrismarineJS/mineflayer/issues/432#issuecomment-227196657) pretty clearly states exactly that:

> Mineflayer currently support only 1.8.

That comment also outlines what would need to be done to add support for other (older) versions.

But until that happens, some clarification in the README would probably be helpful to other new users.

* I include my looking at the Minecraft server logs as part of this visibility.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ Create Minecraft bots with a powerful, stable, and high level JavaScript API.
 
 ## Features
 
- * Supports Minecraft 1.8.9.
+ * Supports Minecraft 1.8.x*.
  * Entity knowledge and tracking.
  * Block knowledge. You can query the world around you.
  * Basic physics and movement - currently blocks are either "solid" or "empty".
@@ -18,6 +18,8 @@ Create Minecraft bots with a powerful, stable, and high level JavaScript API.
  * Miscellaneous stuff such as knowing your health and whether it is raining.
  * Activating blocks and using items.
  * Chat.
+
+\* Supports *only* versions 1.8.*x*; definitely (seems to) work with version 1.8.8.
 
 ### Roadmap
 


roblabla:
Hey, thanks for the PR.

**TLDR**: I'd prefer replacing the line about versions to say "Supports minecraft from 1.8.0 to 1.8.9", without any asterix or w/e. Makes it nice and explicit which versions we support on a single line. 

Mineflayer is currently on life support really, given that the core developers sadly don't have enough time to dedicate to it. If you would like to help work on it to help bring it up to date, you should really stop by our gitter channel (https://gitter.im/PrismarineJS/mineflayer) and we'd be happy to help with anything you need :).

So about versions. Node-Minecraft-Protocol (NMP) (which is the low-level library that handles all the packet and networking stuff) actually supports multiple versions nowadays.

Whenever we update the Prismarine-Chunk to support 1.9-10-11, mineflayer should start to really be multi-versioned, since all the work to be make it support other versions have already been done for the other components (I think. We never actually had a chance to test it ^^'). I believe prismarine-chunk to be the last outdated piece, but I may be wrong.

So in the meantime, having a warning that we are only compatible with 1.8 seems reasonable. However, I think instead of the current wording, we should just have a single bullet point saying "Supports minecraft from 1.8.0 to 1.8.9", without any asterix. If you could do that change, I'll merge that.

That turned out to be a long post. Sorry :D

roblabla:
Oh and a small note. 1.8.0 to 1.8.9 are all the same protocol version with only extremely minor differences, which is why we support all of those :). It's pretty much 100% sure that it works across all of them, and if there's a broken version, it's probably a simple enough fix.

rom1504:
mineflayer now supports 1.8 -> 1.11

PR #466, opened 2017-01-08
update for 1.9

rom1504:
Need to cd node_modules/prismarine_chunk && npm install

Amazingly, it doesn't crash.

The bot doesn't fall because of https://github.com/PrismarineJS/prismarine-chunk/pull/23#issuecomment-271188512

But that's quite nice that there is no crash already.

The plan is cross version support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ Create Minecraft bots with a powerful, stable, and high level JavaScript API.
 
 ## Features
 
- * Supports Minecraft 1.8.9.
+ * Supports Minecraft 1.8, 1.9, 1.10 and 1.11.
  * Entity knowledge and tracking.
  * Block knowledge. You can query the world around you.
  * Basic physics and movement - currently blocks are either "solid" or "empty".
--- a/circle.yml
+++ b/circle.yml
@@ -1,12 +1,10 @@
 machine:
   environment:
-    MC_SERVER_JAR: /home/ubuntu/mineflayer/minecraft-server/minecraft_server.jar
+    MC_SERVER_JAR_DIR: /home/ubuntu/mineflayer/minecraft-server/
   node:
     version: 6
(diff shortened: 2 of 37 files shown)


rom1504:
I believe the problem is that since load is now slower, mineflayer physics engine doesn't react and doesn't make the bot fall.

rom1504:
Or.... that position packets probably changed between 1.8 and 1.9 which would explain why the bot doesn't move.
Trying to figure out what changed

rom1504:
looking at http://wiki.vg/index.php?title=Pre-release_protocol&oldid=7412

rom1504:
indeed position_look is sent but doesn't seem to change stuff (DEBUG=minecraft-protocol)

seems to be a 1.9 thing, wip

rom1504:
now working using the published prismarine-chunk.

I'll see about proper 1.9 support + cross version a bit later (to finish that PR)

rom1504:
build seems to fail for unrelated to pchunk reasons

rom1504:
Added a commit that makes cross version almost done.
Remaining here : 
* [ ] looking at http://wiki.vg/index.php?title=Pre-release_protocol&oldid=7412 for all protocol changes
* [ ] fixing the corresponding packets (mostly position datatype changes)
* [ ] trying all the example
* [ ] merge and publish as mineflayer 2.0.0

rom1504:
* http://wiki.vg/index.php?title=Pre-release_protocol&oldid=7972 : 1.10 : basically nothing changed that matters here
* http://wiki.vg/index.php?title=Pre-release_protocol&oldid=8249 : 1.11 : on thing changed that matter : cursor is now from 0 to 1 in place block

rom1504:
remaining : updating the doc + documenting the breaking changes in the history

Issue #56, opened 2012-06-30
Document how to use minecraft-protocol-forge with mineflayer


Kyso:
To some extent, you won't have to implement the mods directly, but at least send the packet that Forge wants to let it work? _shrug_

deathcap:
@rom1504 can you reopen this issue? Still would be a cool feature =) Requires lower-level protocol support first https://github.com/PrismarineJS/node-minecraft-protocol/issues/114

deathcap:
> MinecraftForge mod on server sends packet to mineflayer requesting for mods x, y and z, which are necessary to join.
> Mineflayer responds request by falsifying a succeeded packet, repeats for each requested mod.

Good idea, https://github.com/ORelio/Minecraft-Console-Client/pull/100 (a console client with Forge (1.8 and 1.7.10) support) used this same technique

rom1504:
Yeah reopening. Nothing can be done here until PrismarineJS/node-minecraft-protocol#114 is done though.

rom1504:
Now that nmp has a minecraft-forge plugin, something can be done here.
Not sure if it would be directly in mineflayer or as a mineflayer-forge (a plugin is probably better)

rom1504:
let's close this
It can be reopened if there's a real need

PR #472, opened 2017-01-30
Forge can be used

Cedware:
Solves Issue #56

To connect to a server, which uses forge the forge options can be passed to the createBot Method.

example:
```javascript
export const bot = mineflayer.createBot({
    host: 'localhost',
    port: 8080,
    username: '....',
    password: '...'
},{
    forgeMods:[
        {modid: 'treechopper', version: '1.1.1a'}
    ]
});
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,4 +1,5 @@
 var mc = require('minecraft-protocol');
+var mcForge = require('minecraft-protocol-forge');
 var EventEmitter = require('events').EventEmitter;
 var util = require('util');
 var path = require('path');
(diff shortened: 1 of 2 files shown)


roblabla:
So I'd rather forge support stays outside mineflayer. I believe it is already usable by doing this : 

```
export const bot = mineflayer.createBot({/* options */});
var forgeHandshake = require('minecraft-protocol-forge').forgeHandshake;
forgeHandshake(bot._client, { /* forgeOptions */ });
```

(It'd be much appreciated if you could confirm this works :D)

We might want to add some documentation to minecraft-protocol-forge to explain how to use it with mineflayer though.

The reason why I'm a bit iffy with adding explicit forge support in mineflayer is because none of the core devs actually use forge, so it's all pretty untested, and it looks like it should work rather well without actual support in the core.

Cedware:
Your suggestion works as well as my modifications.
So I think it's totally okay to keep the forge support outside the project

rom1504:
replaced by https://github.com/PrismarineJS/mineflayer/issues/522

PR #507, opened 2017-03-24
Fixed bot.sleep()

RyanRiffle:
Modified bot.sleep() to account for time being returned as ticks total and not ticks for the current day.

Documentation
```
bot.time.day

Time of the day, in ticks.

Time is based on ticks, where 20 ticks happen every second. There are 24000 
ticks in a day, making Minecraft days exactly 20 minutes long.

The time of day is based on the timestamp 
modulo 24000. 0 is sunrise, 6000 is noon, 
12000 is sunset, and 18000 is midnight.
```

### DIFF
--- a/lib/plugins/bed.js
+++ b/lib/plugins/bed.js
@@ -19,7 +19,8 @@ function inject(bot) {
   }
 
   function sleep(bedBlock, cb) {
-    if(!(bot.time.day >= 12541 && bot.time.day <= 23458))
+    var time = bot.time.day % 24000;
+    if(!(time >= 12541 && time <= 23458))
       cb(new Error("it's not night"));
     else if(bot.isSleeping)
       cb(new Error("already sleeping"));


rom1504:
"The time of day is based on the timestamp modulo 24000"
is it not already %24000 ?
Do you see value of bot.time.day that are > 24000 ?

roblabla:
Doesn't look like it : https://github.com/PrismarineJS/mineflayer/blob/cfbb1f5203092cd4d5e32c7ebc0fcb8e48ba0877/lib/plugins/time.js#L10

roblabla:
I think the server is meant to send a value between 0 and 24000 for the `bot.time.day`, but it's not actually enforced by the protocol. So We might want to `% 24000` ourselves in there.

RyanRiffle:
If you believe it would be better implemented in `bot.time.day` instead, just close the PR. I was just getting it to work for myself.

rom1504:
"enforced by the protocol" means the server send us that : it's the Mojang
server that defines the protocol.

Do you get values superior to 24000 ?

RyanRiffle:
Yes. I get much larger from Minecraft 1.11.2 running a single player game then opened up to the LAN. I get around double the max.

roblabla:
Fixed with 2f7abba

PR #515, opened 2017-05-02
Refactoring

gkal19:


rom1504:
why did you close the PR ?
These might be interesting changes

plexigras:
yeah i think we should updated the repo to use es6 but we should make it so its backwards compatible using babel.

while doing that we should also add a .eslintrc and an maybe a .editorconfig file.

rom1504:
We decided to just use es6 in node-minecraft-protocol and drop babel.
I think we can do that here too.
See https://github.com/PrismarineJS/node-minecraft-protocol/pull/319

plexigras:
oh ok then no babel then

PR #529, opened 2017-06-12
Chat formatting

plexigras:
ANSI and MOTD formatting support?

What do you think about it?


roblabla:
Does this work in windows cmd.exe ?

rom1504:
This is server specific. So I think at least a note should be added to say on what kind of server it works.

plexigras:
@rom1504 it should work on all servers.

@roblabla it should work with the newer version of the cmd (conhost.exe) which they added in windows 10 but not in the old cmd.exe i believe.

maybe someone could try it :)

plexigras:
maybe don't merge it for now there are still some things that don't seam right

plexigras:
ok fixed the stuff i wasn't so happy abut

rom1504:
Ok so : 
* who/what use version code such as §2 ? players ? server plugins ?
* anybody has an opinion about whether it's ok to put en_us.lang online ?

plexigras:
> who/what use version code such as §2 ? players ? server plugins ?

what do you mean by this?

> anybody has an opinion about whether it's ok to put en_us.lang online ?

i believe this is in compliance with [minecrafts brand and asset usage guidelines](https://account.mojang.com/documents/brand_guidelines)

plexigras:
i think i have a better solution dont merge for now

rom1504:
ANSI colors (http://wiki.vg/Chat#Colors) are part of the old system, so I guess the vanilla server doesn't send these kind of color code by itself ?
But I guess it's ok, we can support it.

Ok if it's possible legally, then I guess we can put it in minecraft-data (or https://github.com/rom1504/minecraft-assets idk).

Also possible to make a module to download it automatically when needed, not sure.

I guess putting it in minecraft-data would be the simpler solution. Maybe we can store it as json, it would help a little bit.

rom1504:
Apparently bungeecode does it (https://github.com/SpigotMC/BungeeCord/blob/master/chat/src/main/resources/mojang-translations/en_US.properties) so I guess let's do it.

Can you do a PR to minecraft-data of that file ? (maybe converted to json if it looks ok)

plexigras:
@rom1504 i just wrote a plugin so we can load minecaft assets like 

`bot.loadAsset('minecraft/lang/en_GB.lang')`

```
var assert = require('assert');
var path = require('path');
var fs = require('fs');
var mcPath = require('minecraft-folder-path');

module.exports = inject;

function inject(bot, options) {
  var indexPath = path.join(mcPath, 'assets/indexes', options.assetIndex || bot.majorVersion);
  try {
    bot.assetIndex = require(indexPath).objects;
    bot.assetFiles = Object.keys(bot.assetIndex);
  }
  catch (e){
    bot.assetIndex = null;
    bot.assetFiles = [];
  }

  function loadAsset(file) {
    if (!bot.assetIndex) return null;
    var hash = bot.assetIndex.objects[file].hash;
    var filePath = path.join(mcPath, 'assets/objects', hash.substring(0,2), hash);
    return fs.readFileSync(filePath, 'utf8');
  }

  bot.loadAsset = loadAsset;
}
```

plexigras:
> so I guess the vanilla server doesn't send these kind of color code by itself ?

the vanilla server dose send these codes for example for colored names.

```
/scoreboard teams add (Team name)
/scoreboard teams join (Team name)
/scoreboard teams option (Team name) color (The color you want)
```

roblabla:
I'd be very much for putting the lang files in minecraft-data. Relying on the minecraft files is a bit of a pain (what about headless servers ?)

plexigras:
@roblabla i believe we plan on putting a parsed version of the lang file up on minecraft-data

rom1504:
It would be nice to make a small script that automatically gets the data (possibly using https://github.com/rom1504/node-minecraft-wrap to download the jar) and output it as a .json.
That way we can easily get the language files for all the supported versions.

rom1504:
@plexigras can you update the pr using https://github.com/PrismarineJS/node-minecraft-data/commit/259345657a738dd390f7aa5e10a35b36423ca6f7 ?

plexigras:
@roblabla oh you put it into minecraft-data after all. i thought we would put it into minecraft-assets? wouldn't be minecraft-assets a much better place?

rom1504:
@plexigras I figured let's put transformed/not too heavy things in minecraft-data.
After all everything currently in minecraft-data could also be derived from the jar.

plexigras:
well ok no big deal.

what is kinda big of a deal is that  the chat message instance dose not know of bot, meaning it doesn't know the version and such.

also i would like to add translations so it automatically translates to the right language. what do you think @rom1504?

rom1504:
I think we should follow the loader pattern in the chat message file, so you can do require ('chat_message')(version)
Have a look at https://github.com/PrismarineJS/prismarine-block/blob/master/index.js#L3 for an example of how to implement this pattern.

About translations, why not, let's try to extract them and see how big they are.

plexigras:
oh yes that's better

rom1504:
ok let's merge now, we can add the translations afterwards

Issue #468, opened 2017-01-21
Consider adding auto version detection and making it the default


rom1504:
That's possible thanks to nmp auto version detection.
The main needed change would be to use the version in bot.client and to wait for version detection to be over before loading mineflayer plugins.

Making it the default would just mean putting false as defaultVersion in version.js.

Consequence of that would be to be able to connect a bot to any server of version 1.8 -> 1.11 without checking the version of the server first.

rom1504:
done by @plexigras

PR #530, opened 2017-06-14
Auto Version

plexigras:


rom1504:
Does that work ?
(https://github.com/PrismarineJS/mineflayer/issues/468 is the issue tracking this)
in particular : does options.version get updated to the correct version ? it's needed by plugins https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/entities.js#L27

plexigras:
> Does that work ?

i tried it on 1.9.4, 1.10.2, 1.11.2 and 1.12

it only didn't work on 1.10.2 but it didn't work on master either

rom1504:
what's the error on 1.10.2 ?

rom1504:
Ah I know, it doesn't exist.
1.11.2 exists.

rom1504:
Thanks !

PR #546, opened 2017-06-27
Asset index loading plugin.

plexigras:
#### bot.loadAssetIndex([version])

Loads asset indexes that can be found in `.minecraft/assets/indexes`.

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -1124,7 +1124,7 @@ Injects plugins see `bot.loadPlugin`.
 
 #### bot.loadAssetIndex([version])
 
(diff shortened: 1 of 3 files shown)


rom1504:
related https://github.com/PrismarineJS/minecraft-data/issues/184

We might not want to rely on the user having installed (and used) a minecraft vanilla client.

plexigras:
maybe better to have in a seperade pkg.

rom1504:
Yeah. This also wouldn't need to be a mineflayer plugin, that doesn't rely on anything of mineflayer.

PR #557, opened 2017-07-09
added plugins to bot options

plexigras:


rom1504:
is that easier to use ?
If we do that, the rest of the plugin api should be updated.

plexigras:
its just so you can control what internal plugins are loaded when you create a bot

PR #560, opened 2017-07-11
Fix code typo

SuperOP535:


examples/chatterbox.js
```diff
@@ -158,9 +158,8 @@ bot.on('playerJoined', (player) => {
   }
 })
 bot.on('playerLeft', (player) => {
-  if (player.username === bot.username) {
+  if (player.username === bot.username) return;
     bot.chat(`Bye ${player.username}`)
```
roblabla: Could you remove two spaces here, so it's properly indented ?

SuperOP535: Uhm could you tell me where exact?

roblabla: bot.chat, remove two spaces before it :)

SuperOP535:
I made a bonus and fixed playerJoined double space too ;)

roblabla:
No, the double space in playerJoined is OK.

When something is between braces, it should be intended. Basically, 

```
if (something) {
  ThatHappensIfSomething // So we make it visually "under" something
}
```

But

```
if (something) return;
ElseThisHappens // This doesn't happen if something, so we don't put it "under" it. We don't put spaces
```

SuperOP535:
Well, I have been coding JavaScript in a long timeand I did not notice that ._.

SuperOP535:
Cool now I have the Contributor tag

PR #569, opened 2017-07-28
fix mcdata import in trader example

plexigras:
autoversion is making trouble again.


rom1504:
if we're using it in the examples, I think inject_allowed should be in the doc

rom1504:
(ie I had to look at the code to know what it is)

plexigras:
maybe doing `bot._client.once('success', ... )` would be better?

rom1504:
would be nice to finish this, I don't like dead PRs....

PR #574, opened 2017-08-02
A very simple interactive chat

duralakun:
I couldn't find an example of an interactive chat, so I tried to make a very simple one.


duralakun:
Oh I guess this only works on ubuntu?

PR #575, opened 2017-08-02
A very simple interactive chat

duralakun:
I couldn't find an example of an interactive chat, so I tried to make a very simple one.

### DIFF
--- a/examples/chat.js
+++ b/examples/chat.js
@@ -26,14 +26,12 @@ const bot = mineflayer.createBot({
 // uses on 'message' instead of on 'chat'
 // because on 'chat' does not work on all servers 
 bot.on('message', (message) => {
-  var strMessage = `${message}` // removes json
-  console.log(strMessage)
+  console.log(`${message}`) // print chat
 })
 
 // console input
 var stdin = process.openStdin()
 stdin.addListener('data', function (d) {
-  // if something was typed in
-  var input = d.toString().trim() // get input
-  bot.chat(input) // tell everyone
+  var cInput = d.toString().trim() // get input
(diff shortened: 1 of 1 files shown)


duralakun:
If you don't like it, pls tell me why.

examples/chat.js
```diff
@@ -0,0 +1,37 @@
+/*
+ * This is a very simple chat client.
+ *
+ * That shows you how you can read and interact with any chat on a server
+ *
+ * Just type into the console and press enter to say something
+ *
+ * Be sure to stand somewhere save so you dont get killed by zombies :)
```
rom1504: safe

examples/chat.js
```diff
@@ -0,0 +1,37 @@
...
+  host: process.argv[2],
+  port: parseInt(process.argv[3]),
+  username: process.argv[4] ? process.argv[4] : 'echo',
+  password: process.argv[5],
+  verbose: true
+})
+
+// uses on 'message' instead of on 'chat'
+// because on 'chat' does not work on all servers 
+bot.on('message', (message) => {
```
rom1504: what about the fact it's json ?

plexigras: `console.log(message.toString())` might be better

duralakun:
It says "All checks have failed" but I think this is a bug? What am I doing wrong?

duralakun:
Thanks for the reviews I did some changes.
I was thinking about adding the on 'health' event and print out a message in the conole, to warn players before they die.
Also a quit option would be nice. So you can type for example ".quit" in the console and it will call the bot.quit() function.
But this would me make the program less simple. Maybe for a more advanced chat example?

PR #581, opened 2017-08-22
remove extraneous spaces in ChatMessage

Gjum:


plexigras:
aren't the spaces are required?

rom1504:
Yeah same question, what does it currently do ? What does it do with that
change ?

plexigras:
i think it converts
```
{ text: 'this', extra: ['is', 'an', 'example'] }
```
to `thisisanexample` instead of the current `this is an example`

Gjum:
The spaces are not only not required, adding them adds space where the vanilla client doesn't show any.

Take a look at this example, which is representative of many servers, vanilla and modded, and all messages on them.

chat message from server:
```js
{
    "text":"",
    "extra": [
        {
            "text":"Logging out safely in ", // <-- note the extra space here
            "color":"gray"
        },
        {
            "text":"5 seconds",
            "color":"aqua"
        },
        {
            "text":"...",
            "color":"gray"
        }
    ]
}
```

conversion:
```js
const chatMsg = new ChatMessage(JSON.parse(packet.message));
log(`[CHAT] ${chatMsg.toAnsi()}`);
```

original output: (extraneous spaces marked)
```
[CHAT]  Logging out safely in  5 seconds ...
       ^                      ^         ^
       1                      2         3
```

fixed output: (beginnings of `extra` entries marked)
```
[CHAT] Logging out safely in 5 seconds...
       ^                     ^        ^
```

The first additional space is because even the empty `text` at the outermost level gets separated by a space.
2 and 3 are where the colors change, and so they need to be in separate `extra` entries, which get separated as well.

For an extreme example, one could imagine a rainbow name, with a new color for each letter, which would get blown up to twice the length, a space after every letter.

plexigras:
if you remove the spaces this
```js
this.extra.forEach((entry) => {
  message += `${entry.toMotd(lang)} `
})
```
should turn into
```js
this.extra.forEach((entry) => {
  message += entry.toMotd(lang)
})
```
or even better
```js
message += this.extra.map((entry) => entry.toString(lang)).join('')
```

Gjum:
Good point, I cleaned them up.

PR #587, opened 2017-09-05
Add support for end dimension

iRath96:
In its current state, the code expects skylight data to be sent in all dimensions except for the nether. Skylight data is actually sent _only for the overworld_. The problem is that the code expects the end dimension to also send skylight data, which causes the bot to crash.

This PR updates the skylight logic so that all three dimensions work properly.


rom1504:
something is broken with the blockAt test, I tried to fix it but apparently it's not enough.
That needs fixing before this can be merged.

Anyway thanks for fixing this behavior ;)

iRath96:
@rom1504 seems like your fix actually worked, just the indentation broke the test :)

roblabla:
So, if I create a custom world name, I can't send skylight data ? Interesting.

iRath96:
@roblabla It's not the world name that matters, it's the dimension.

roblabla:
Right, I meant a custom dimension type ^^

iRath96:
@roblabla According to http://wiki.vg/Protocol, you cannot create custom dimensions. There's only -1, 0 and +1 in Vanilla :)

PR #588, opened 2017-09-08
Update length limit to 256

ypymc:
Seems like it's 256 and has been [since 1.11](https://minecraft.gamepedia.com/Talk:Multiplayer#1.11_Chat_Character_Increase).

screenshot

PR #589, opened 2017-09-10
Add dynamic chat lengths depending on options or version

ArcticZeroo:
Improvement on https://github.com/PrismarineJS/mineflayer/pull/588 (I was also looking to change it and saw the PR already made), this allows dynamic chat length depending on the version.

This PR does a few things:
* Adds `bot.protocolVersion`
* Adds a `chatLengthLimit` option to the `createBot` function
* Adds the dynamic chat limit

The chat limit is determined as follows:
1. If a chatLengthLimit is provided, use that
2. If the protocol version is 301 or higher (the first 1.11 snapshot or higher), the length is 256
3. Otherwise, it's 100

The chatLengthLimit option is provided in case the length changes in the future (to allow people to change it without diving into the module itself), and for servers with backwards compatibility that allow more characters even if the client thinks it's 1.8 (for instance, Mineplex allows this).

### DIFF
--- a/lib/plugins/chat.js
+++ b/lib/plugins/chat.js
@@ -1,4 +1,4 @@
-const PROTO_VER_1_10 = require('minecraft-data')('1.10.2').version.version;
+const PROTO_VER_1_10 = require('minecraft-data')('1.10.2').version.version
 
 module.exports = inject
 
--- a/lib/plugins/chat.js
+++ b/lib/plugins/chat.js
@@ -1,7 +1,9 @@
+const PROTO_VER_1_10 = require('minecraft-data')('1.10.2').version.version;
+
 module.exports = inject
 
 function inject (bot, options) {
-  const CHAT_LENGTH_LIMIT = (options.chatLengthLimit) ? options.chatLengthLimit : ((bot.protocolVersion >= 301) ? 256 : 100)
+  const CHAT_LENGTH_LIMIT = (options.chatLengthLimit) ? options.chatLengthLimit : ((bot.protocolVersion > PROTO_VER_1_10) ? 256 : 100)
 
   const ChatMessage = require('../chat_message')(bot.version)
   // add an array, containing objects such as {pattern:/regex pattern/, chatType:"string", description:"string"}
--- a/lib/plugins/chat.js
+++ b/lib/plugins/chat.js
@@ -1,8 +1,8 @@
-const CHAT_LENGTH_LIMIT = 100
-
 module.exports = inject
 
-function inject (bot) {
+function inject (bot, options) {
+  const CHAT_LENGTH_LIMIT = (options.chatLengthLimit) ? options.chatLengthLimit : ((bot.protocolVersion >= 301) ? 256 : 100)
+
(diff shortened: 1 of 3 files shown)


roblabla:
Is there a way to make the magical `301` taken from minecraft-data ? I'd rather have something like `const MC_11_PROTO = require("minecraft-data")("1.11").version;` and compare against that instead.

roblabla:
Furthermore, I wonder if this data shouldn't be in the `protocol.json` (or somewhere else in `minecraft-data`) instead. This promotes reusability of this kind of magic constants across our various projects.

How does the notchian client/server behave if we send a message that's longer than the limit ?

ArcticZeroo:
The only problem with using `minecraft-data` is that snapshots aren't accounted for in any of its data, though I suppose that's easily fixable by using the last 1.10 release.

Gjum:
The limit is only client -> server.

If the client sends a message over the limit, it gets kicked with a kick message saying so.

roblabla:
Right, so it's effectively a "soft" limit - so it makes sense to let it be overridable. Thanks for the info.

@ArcticZeroo I'm not sure what you mean ?

ArcticZeroo:
@roblabla in regards to what? I left a comment about protocol numbers and pretty much immediately realized it was dumb and deleted it, assuming you mean that?

non-notchian servers behave similarly to vanilla, they just kick the player for some packet-related message or a generic proxy disconnect.

lib/plugins/chat.js
```diff
@@ -1,8 +1,10 @@
-const CHAT_LENGTH_LIMIT = 100
+const PROTO_VER_1_10 = require('minecraft-data')('1.10.2').version.version
```
plexigras: why not hardcode this to `301`?

roblabla:
Right, I didn't realize you deleted it - I stayed on the page, so github showed the new comment as soon as you published it, but didn't remove it when you did xD

I'm forcing a rebuild, I think the CircleCI failure is just a blip.

lib/plugins/chat.js
```diff
@@ -1,8 +1,10 @@
...
 
 module.exports = inject
 
-function inject (bot) {
+function inject (bot, options) {
+  const CHAT_LENGTH_LIMIT = (options.chatLengthLimit) ? options.chatLengthLimit : ((bot.protocolVersion > PROTO_VER_1_10) ? 256 : 100)
```
plexigras: maybe like this
```js
const CHAT_LENGTH_LIMIT = options.chatLengthLimit || (bot.protocolVersion < 301 ? 100 : 256)
```

lib/plugins/chat.js
```diff
@@ -1,8 +1,10 @@
...
+
```
plexigras: ??

lib/plugins/chat.js
roblabla: Among other reasons, because it's ugly A.F., and confusing to newcomers and potential maintainers. The more explicit we are, the better.

PR #600, opened 2017-10-06
improved version checks

plexigras:


rom1504:
I feel like we're losing information here.
What about using these >= and <= you implemented ?

plexigras:
> I feel like we're losing information here.

any examples?

> What about using these >= and <= you implemented ?

its not necessary to use these functions and this is simpler.

roblabla:
I agree, I don't like this too much. 1.7, if ever implemented, it should follow the same code path as 1.8. Using >= and <= would be better I think.

rom1504:
Yes actually I've been thinking about this a bit and maybe we should
actually go the other way completely.
Have a version to feature map and only when we don't know about the version
try to guess (or maybe not even try to guess at all)

For example here we would have
```json
{
"1.8" : "representEntitiesPositionAsFixedPoint"
}
```

So we could then do if
(checkFeature(version,"representEntitiesPositionAsFixedPoint")) { ...}

It's unclear in which versions features live and we definitely don't have
much information about this, so I think we should try and keep as much
information we do have instead of guessing "this is always going to be like
that except for 1.8".

rom1504:
The benefit of doing that is that it would be clear what feature we're
acting on instead of vagues version ranges (that we cannot check without
looking at the code in the if)

rom1504:
let's not do it that way

PR #605, opened 2017-10-26
simplify chat plugin

plexigras:
its a quiet radical change so there might be side effects i don't know of

### DIFF
--- a/lib/plugins/chat.js
+++ b/lib/plugins/chat.js
@@ -33,105 +33,28 @@ function inject (bot) {
   }
 
   bot._client.on('chat', (packet) => {
-    // Function used to process "unidentified" messages, such as 1.6 chat and bukkit servers
-    function parseOldMessage (stringMsg) {
+    function checkForChatPatterns (msg) {
+      const stringMsg = msg.toString()
       // iterate through each object in chat.patterns array and test if .pattern matches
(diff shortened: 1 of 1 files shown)


rom1504:
Well well...
AFAIK chat is broken in 99% of real uses (ie people always come complaining it doesn't work)
I believe we should setup some automatic testing for chat for 
* spigot
* paper
* a selection of real servers out there

Testing would just be "check we can decode pseudo/message"

At least this PR doesn't break for 1.7 to 1.12 according to our tests.

Are you sure it's not needed for other servers ?

plexigras:
i don't know of any special packages that cant be parsed by the chat message plugin but if there are it would definitly cause problems

plexigras:
lets hope for the best

rom1504:
yeah, it's better if it's simpler anyway, easier to fix

Corgano:
Could we get an updated example now? How do we add custom patterns as we did before? I keep getting TypeError: bot.chatAddPattern is not a function when i run the included mineflayer example

plexigras:
this is not due to this pr but due to async plugin injection.

you can do

```js
bot.once('inject_allowed', () => bot.chatAddPattern(...))
```

or

```js
bot.loadPlugin(() => bot.chatAddPattern(...))
```

you should open an issue if you found an issue.

PR #609, opened 2017-11-04
Added sneaking option to controll states

Meldiron:
- added option 'sneak' to setControllState,
- changed api documentation, added this option there

### DIFF
--- a/lib/plugins/physics.js
+++ b/lib/plugins/physics.js
@@ -35,7 +35,7 @@ function inject (bot) {
     right: false,
     jump: false,
     sprint: false,
-    sneak: false,
+    sneak: false
   }
   let jumpQueued = false
   let lastSentYaw = null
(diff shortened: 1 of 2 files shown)


lib/plugins/physics.js
```diff
@@ -34,7 +34,8 @@ function inject (bot) {
     left: false,
     right: false,
     jump: false,
-    sprint: false
+    sprint: false,
+    sneak: false,
```
plexigras: eslint doesn't like the trailing commas, you will need to remove it to confirm to the code style rules

plexigras: eslint doesn't like the trailing commas, you will need to remove it to confirm to the code style rules.

plexigras: eslint doesn't like the trailing commas, you will need to remove it to confirm to the code style rules

rom1504:
Thanks !

Issue #639, opened 2018-01-28
Boss bar


ArtFect:
Am I blind or is there no way to get custom boss bar text? 
I tried to find wither or dragon in bot.entities but they aren't there

plexigras:
i dont think there is a mineflayer plugin that handles this

but you could write one or you could just do `bot._client.on('boss_bar", ... )`

http://prismarinejs.github.io/minecraft-data/?d=protocol#toClient_boss_bar

also then you might want to parse the `title` field using `new ChatMessage(title)`

rom1504:
We should probably handle that mineflayer. This is a new feature so it's not handled but it should be, if you have some code that implement it, please share it.
Otherwise it should still be implemented, reopening.

rom1504:
If someone is interested, please finish https://github.com/PrismarineJS/mineflayer/pull/659
I'm not merging a half-done feature like that, closing the pr since it's dead.

rom1504:
now done, thanks @wvffle

PR #659, opened 2018-04-17
Add bossBar events

czaarek99:
Add bossBar events that work for all minecraft versions including 1.8,
from minecraft 1.9 and onwards it's simple to listen to the boss bars as
you just have to listen to the boss_bar packet. This commit also adds
support for boss bars in 1.8 where it's simulated by using a renamed
dragon or wither

As requested in issue #639 

All the new events work in a similiar way. The first argument is always the entity type e.g either entityUUID or just entity. This argument indicates if the next argument will be an entity object or the uuid of the entity. The next arguments are then relevant to the event in question.

I decided against finding the entity in the bot.entities array from 1.9 and upward due to uncessarry overhead. These packets get sent quite a lot and it's pointless to find the entity each time if the user doesn't really need to know. Most of the time there will only be a single boss bar.

### DIFF
--- a/lib/plugins/entities.js
+++ b/lib/plugins/entities.js
@@ -419,6 +419,51 @@ function inject (bot, { version }) {
     }
   })
 
+  const bossBar = 'bossBar'
+  if (bot.majorVersion === '1.8') {
+    const enderDragon = 'Ender Dragon'
+    const wither = 'Wither'
+
+    bot._client.on('entity_metadata', (packet) => {
+      const entity = bot.entities[packet.entityId]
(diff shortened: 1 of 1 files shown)


rom1504:
Hi,
Sorry for the reaction time.
Thanks for the PR but could you : 

* use these new features in an example (a new or an existing one) so you can test it works
* add the events in the doc

rom1504:
could someone finish this ?

rom1504:
Closing. Feel free to reopen if you want to finish it.

PR #684, opened 2018-07-13
Rename LICENCE to LICENSE

DavidRazdolski:
Fixing a typo


rom1504:
Licence is correct in British English but apparently GitHub uses license, so ok let's merge.

PR #698, opened 2018-08-03
Multiple.js example improved

Vistril:
The provided multiple.js hardly works and no parent object for the bots, meaning they can't do anything.
My code allows for modular usernames, quantity and password field if needed.
Uses a loop instead of an interval so crashing probability is lower.

### DIFF
--- a/examples/multiple.js
+++ b/examples/multiple.js
@@ -11,7 +11,7 @@ newClient('Client', null, 5)
 
 function newClient(username, password, quantity) {
 	if (typeof quantity != 'number') throw 'Quantity must be a number!'
-	for (let i = 0 i < quantity i++) {
+	for (let i = 0; i < quantity; i++) {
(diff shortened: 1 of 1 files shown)


rom1504:
Why do you think a loop is better than an interval ?
The interval prevent bots to all start exactly at the same time, overloading the CPU and preventing the bots to start.

Didn't you notice that when starting many bots ?

Vistril:
Personally, I had no CPU load issues. However, if this is what should be done, then I can change it to use a timeout and interval. I just don't like how the current example doesn't allow for a usable multi-client object.

rom1504:
Try to connect 50 bots, you'll see it.

Vistril:
I've connected more than 50

rom1504:
https://github.com/PrismarineJS/mineflayer/issues/710 shows the interval is required

PR #705, opened 2018-08-17
Add files via upload

Malod219:
Added water slowdown to horizontal speed in physics.js. Unsure if value of 0.3 is accurate to actual water speed.


rom1504:
Can you move the file to the proper location ?

Malod219:
My bad, still rather new to Github Pull/Push requests. It's in proper file location now. It's essentially just the soulsand code repeated and tweaked. The multiplier of 0.3 for max water speed seems about right, but might be worth double checking in case

rom1504:
looks good, merged, thanks !

PR #764, opened 2019-01-14
Update physics.js

CheezBarger:
Updated maxGroundSpeed to value given in Minecraft wiki.
MaxGroundSpeed did not take into account the fact that the player could have been sprinting.


rom1504:
Thank you for the PR

PR #771, opened 2019-01-29
Prevent error from being thrown when chunk fails to load

ArcticZeroo:
Sometimes (usually on custom servers not running vanilla), a chunk fails to load, and the error is directly thrown instead of going to the error event emitter.

Also fixes a suggestion @plexigras made on my PR like two years ago


rom1504:
thanks

PR #786, opened 2019-02-25
Clarified when to use login and spawn

SuperVK:
It said spawn was fired when you load in for the first, but this is false. And at login it refered to spawn to use when you want to do something as soon as you  join the server, but as spawn isn't fired when you join the server this was false too.

(This is my first pull request, I'm not sure if I'm doing everything correctly)


SuperVK:
I released that I might have done something not the way it's supposed to....

rom1504:
I think it would be better to fix spawn instead

Issue #548, opened 2017-06-28
Read and use block entities nbt


rom1504:
http://wiki.vg/Protocol#Update_Block_Entity

and 

block entities of http://wiki.vg/Protocol#Chunk_Data

I believe this is where sign texts are now stored.

(will only be useful for >= 1.12)

rom1504:
tile_entity_data

wvffle:
I guess it should be closed by now

rom1504:
Indeed !

Issue #675, opened 2018-06-07
Sign text is undefined (1.12.2)


safazi:
```
Block {
  type: 68,
  metadata: 2,
  light: 0,
  skyLight: 0,
  biome:
   Biome {
     id: 0,
     color: 112,
     name: 'Ocean',
     height: undefined,
     rainfall: 0.5,
     temperature: 0.5 },
  position: Vec3 { x: -153, y: 64, z: 156 },
  name: 'wall_sign',
  hardness: 1,
  displayName: 'Wall Sign',
  boundingBox: 'empty',
  diggable: true,
  material: 'wood',
  harvestTools: undefined,
  drops: [ { drop: 323 } ],
  signText: undefined,
  painting: undefined }
```

Block found with bot.findBlock, and I am not changing the block in any way.

EDIT: this is on bukkit-Spigot.

safazi:
Also, the sign is not blank, before anyone asks :+1:

plexigras:
Good find, we might want to have a look at updating the signText property on update_sign events.

http://wiki.vg/Protocol#Update_Sign
http://minecraft-data.prismarine.js.org/?d=protocol#toServer_update_sign

plexigras:
it might be better to use something like [`updateBlock`](https://github.com/PrismarineJS/mineflayer/blob/7284dd436aac8c94d639af038a3a3f1efdf18473/lib/plugins/blocks.js#L146) over [`emitBlockUpdate`](https://github.com/PrismarineJS/mineflayer/blob/7284dd436aac8c94d639af038a3a3f1efdf18473/lib/plugins/blocks.js#L275) directly.

wvffle:
Any works on this?

rom1504:
No, you can work on it ;)

wvffle:
After some testing I can get sign text on demand on paper 1.12.2

```js
const getSignText = async location => new Promise(resolve => {
  const parse = text => {
    const ChatMessage = require('mineflayer/lib/chat_message')(bot.version)
    return new ChatMessage(JSON.parse(text))
  }

  bot._client.once('tile_entity_data', data => {
    const { Text1, Text2, Text3, Text4 } = data.nbtData.value
    const text = [ Text1, Text2, Text3, Text4 ].map(cm => parse(cm.value).toString()).join('\n')
    resolve(text)
  })

  bot._client.write('update_sign', { location, text1: '', text2: '', text3: '', text4: '' })
})
```

[...]

PR #796, opened 2019-04-23
Add support for block entities

wvffle:
Adds a `blockEntity` field to the result of `bot.blockAt(position)`. 

Kind of workaround for #675 as we can now read sign text as follows:
```js
const { Text1, Text2, Text3, Text4 } = sign.blockEntity
if (Text1 === bot.username) {
  bot.chat('That\'s my sign!')
}
```

This also allows to gain data about stuff defined [there](https://minecraft.gamepedia.com/Block_entity)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,6 @@
 # different than .npmignore
 node_modules
 package-lock.json
+yarn.lock
 /README.md
 versions/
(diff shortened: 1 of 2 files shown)


wvffle:
Closes #548

rom1504:
Cool stuff, code looks good, thanks for the contribution, merging !

Issue #752, opened 2018-11-27
Add displayName in user properties


A5005:
There is a well-known NickNames Bukkit plugin that displays a modified username for other players in the chat window (also in the TAB playerlist). I have not found a way to extract the nickname from the original username. Is there is a chance to realize that ?

The original username can be easily determined via bot.players.
Typical output (non-vanilla Server) shows always the the original username even if the player is nicked.

username: 'UserName',
  ping: 0,
  uuid: '*****-*****-****-*****-*****',
  entity:
   Entity {
     domain: null,
     _events: {},
     _eventsCount: 0,
     _maxListeners: undefined,
     id: 2932485,
     type: 'player',
     position: Vec3 { x: -4412.5, y: 59.001, z: 4114.5 },
     velocity: Vec3 { x: 0, y: 0, z: 0 },
     yaw: 1.2490457723982544,
     pitch: -1.0279847897285246,
     onGround: true,
     height: 1.62,
     effects: {},
     equipment: [ <5 empty items> ],
     heldItem: null,
     isValid: true,
     metadata:
      { '0': 0,
        '1': 300,
        '2': '',
        '3': 0,
        '4': 0,
        '6': 20,
        '7': 0,
        '8': 0,
        '9': 0,
        '10': -1,
        '16': 0,
        '17': 0,
        '18': 8 },
     username: 'UserName',
     timeSinceOnGround: 39.33799999999982 } }

rom1504:
Yes.
Can you add displayName: item.displayName at https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/entities.js#L372
And make a pr for it ?

It's undeed available in the packet
http://minecraft-data.prismarine.js.org/?d=protocol#toClient_player_info

rom1504:
now done, thanks @wvffle

Issue #747, opened 2018-11-24
Get contents of tab list


EvilSourcerer:
I am wondering if it's possible to get the contents of the list in Minecraft when you press tab in plain text

rom1504:
Yeah that packet
http://minecraft-data.prismarine.js.org/?d=protocol#toClient_tab_complete

bot._client.on("tab_complete", data => ...)

It's not mineflayer high level API though, maybe you can add it :)

EvilSourcerer:
wait but that's an event, not a variable... how do I use it

rom1504:
```
bot._client.on("tab_complete", data => console.log(data))
```

EvilSourcerer:
It has to first send the packet which 'tab_complete', the result is no output when I use that code. How do I send the packet.

plexigras:
http://minecraft-data.prismarine.js.org/?d=protocol#toServer_tab_complete

EvilSourcerer:
that seems to be the right packet... but how is it sent

EvilSourcerer:
wait thats not the player list... thats the command that autofills when you press tab. I was asking to make the messages on the player list readable as plaintext.

rom1504:
this.players

EvilSourcerer:
Fixed! The actual way to get the header and footer of the tab list is : 
bot._client.on("playerlist_header", data => console.log(data))

rom1504:
It would be great to add a mineflayer event (and doc) for this so other people can use it ;)

rom1504:
bot._client provides all the packets but it's low level and can change at any point (whenever mojang feels like changing stuff).
It's better to add stuff to mineflayer API so we can guarantee some stability.

rom1504:
now done, thanks @wvffle

Issue #739, opened 2018-11-02
let the bot sow and harvest plants?


ahibbe:
* what you want to do (the objective in english)
I want to let the bot sow and harvest plants?

* what you tried (the code)
nothing, I expected someting like:

 if (bot.canCanHarvest(block)) 
  bot.harvest(block, onHarvestCompleted)
....

I do not need a complete code just a few key words where to look, if that's possible. Unfortunately I did not find anything in the examples

Thank you

plexigras:
`bot.placeBlock`, `bot.dig` and maybe `bot.activateBlock`

https://github.com/PrismarineJS/mineflayer/blob/master/docs/api.md

rom1504:
I think an example showing how to do that would be nice, feel free to create open when you made it work @ahibbe

rom1504:
now done, thanks @wvffle

Issue #731, opened 2018-10-23
find the block the bot is looking at.


ahibbe:
* what you want to do (the objective in english):
I want to find the block the bot is looking at. (In the client, the block is indicated by a thin black frame)

* what you tried (the code):

```js
const yaw = player.entity.yaw;
const pitch = player.entity.pitch;
console.log("yaw=" + yaw + " pitch=" + pitch);
var vecLook = new vec3(
    -Math.cos(pitch) * Math.sin(yaw),
    Math.sin(pitch),
    -Math.cos(pitch) * Math.cos(yaw)
);
var block = mybot.blockAt(vecLook);
if (!block) {console.log(...
```

* what happened:
block is always null

* what you expected to happen:
the block the bot is looking at.

Thanks for this great API!

plexigras:
this is not the first time this has been asked :smile: https://github.com/PrismarineJS/mineflayer/issues/674

rom1504:
now done, thanks @wvffle

PR #797, opened 2019-04-23
Implement new features

wvffle:
# TODO
- [x] Add displayName (#752)
- [x] Add bossBars (#639)
- [x] Add blockInSight (#731)
- [x] Add scoreboard (#348)
- [x] Add chat completion (mentioned in #747)
- [x] Add tablist header/footer (#747)
- [x] Add consume function (#414)
- [x] Add fishing (#619)
- [x] Add tests
  - [x] Add blockEntity tests
  - [x] Add block.signText tests
  - [x] Add displayName tests
  - [x] Add blockInSight tests
  - [x] Add scoreboard tests
  - [x] Add tab complete tests
  - [x] Add consume tests
  - [x] Add fishing tests
- [x] Add examples
  - [x] Add blockEntity example
  - [x] Add displayName example
  - [x] Add blockInSight example
  - [x] Add scoreboard example
  - [x] Add tab complete example
  - [x] Add consume example
  - [x] Add fishing example
  - [x] Add example for sow and harvest (#739)
  - [x] Add example for flyTo (#665)
  - [x] Add repl example
- [x] Add docs
  - [x] Add blockEntity docs
  - [x] Add displayName docs
  - [x] Add blockInSight docs
  - [x] Add scoreboard docs
  - [x] Add tab complete docs
  - [x] Add consume docs
  - [x] Add fishing docs
  - [x] Add boss bar docs


wvffle:
I just saw that I broke something in my local repo ;/

wvffle:
bot.bossBars returns an array

```js
const bossBar = bot.bossBars[0]

bossBar.dividers // Returns number of dividers
bossBar.title // Returns ChatMessage
bossBar.isPink // Returns true if bar color is pink
```
All colors avilable in table [here](https://wiki.vg/Protocol#Boss_Bar)

wvffle:
bot.blockInSight() is a primitive ray tracer by now. Primitive because it doesn't support bounding boxes and checks only for position of a block.

wvffle:
This change breaks already broken API. Scoreboards were not handled correctly an right now they are
```js
bot.scoreboards // { my_name: scoreboard }
bot.scoreboard // { list: scoreboard, sidebar: scoreboard, belowName: scoreboard, 1: scoreboard, ... }

scoreboard.itemsMap // { dzikoysk: { name: 'dzikoysk', value: 1 }, wvffle: { name: 'wvffle', value: 3 } }
scoreboard.items // Sorted array: [ { name: 'wvffle', value: 3 }, { name: 'dzikoysk', value: 1 } ]
```

wvffle:
#### bot.consume()
`bot.consume()` waits for `entity_status` packet and times out after 10 seconds because it's not determined wether or not we canceled eating ;/
```js
bot.consume(err => {
  if (err) return console.log(err) // Time out after 10s
  // Consumed
}) 
```


#### bot.tabComplete()
tabComplete arguments are as follows
```js
bot.tabComplete(text, cb, assumeCommand = false, sendBlockInSight = true)
```
From my research `assumeCommand` is always false, but I might be wrong so there's an option to change that.

`sendBlockInSight` is a flag that determines if `packet.lookedAtBlock` should be true block to look at or block at `bot.position`. By default it's `true` but it may damage the performance since looking at a distant block is time consuming. (Of course in miliseconds)
```js
const cb = matching => console.log(matching)

// Exact mode
bot.tabComplete('/give wv', cb)

// Performance mode
bot.tabComplete('/give wv', cb, undefined, false) 
```

rom1504:
Very nice !
I'll take some time to review all that soon.

wvffle:
I'll also add tests for every feature that I add here ;) After I'm done with this PR I'll take my hands #725 and maybe #341

wvffle:
### bot.fish()
It automates fishing process. We're waiting for new hook/float entity. Then we wait for a special particle that is created after splash sound at a location nearly the same as our hook/float (if we look at `x` and `z` coordinates). If that particle is at it's place then we end the fishing and call the callback.

It's required to have the fishing rod equipped. Maybe later I'll add a check for that

```js
bot.fish(err => {
  if (err) return console.log(err)
  // We've fished something
  // Now gather the drops
})
```

wvffle:
Block entities are supported from 1.11, `update_sign` event is not supported in 1.10 and it seems like it's not fired at all in 1.9 but it's in the protocol. @rom1504, could it be node-minecraft-protocol bug?

Only 1.9 an 1.10 should fail right now

wvffle:
Weird, circleci says that only 1.9 fails due to double quotes, but on my machine 1.9 and 1.10 doesn't work at all

wvffle:
Rebase and merge would be nice :D

package.json
```diff
@@ -7,7 +7,7 @@
     "test": "mocha --reporter spec",
     "pretest": "npm run lint",
     "lint": "standard",
-    "prepare": "npm install require-self && require-self",
```
rom1504: Why ? Removing that means it's required to do npm install before doing npm publish

examples/farmer.js
```diff
@@ -0,0 +1,72 @@
...
+  port: parseInt(process.argv[3]),
+  username: process.argv[4] ? process.argv[4] : 'farmer',
+  password: process.argv[5],
+  verbose: true
+})
+
+// To fish we have to give bot the seeds
+// /give farmer wheat_seeds 64
+
+const ID_FARMLAND = 60
```
rom1504: Please retrieve that id from minecraft data, it changes every few versions

examples/block_entity.js
```diff
@@ -0,0 +1,39 @@
...
+
+bot.on('message', (cm) => {
+  if (cm.toString().includes('spawner')) {
+    spawner()
+  }
+})
+
+function spawner () {
+  const block = bot.findBlock({
+    matching: 52,
```
rom1504: Retrieve that id from minecraft data

docs/history.md
```diff
@@ -1,3 +1,6 @@
+## 2.8.0
```
rom1504: Let's not put that in this review, I'll add that in the release commit

examples/farmer.js
```diff
@@ -0,0 +1,72 @@
...
+    }
+  })
+}
+
+function loop () {
+  try {
+    const toHarvest = blockToHarvest()
+    if (toHarvest) {
+      return bot.dig(toHarvest, () => {
+        return setTimeout(loop, 1000)
```
rom1504: Why are these setTimeout needed ?

examples/fisherman.js
```diff
@@ -0,0 +1,88 @@
...
+function onCollect (player, entity) {
+  if (entity.kind === 'Drops' && player === bot.entity) {
+    bot.removeListener('playerCollect', onCollect)
+    startFishing()
+  }
+}
+
+function startFishing () {
+  bot.chat('Fishing')
+  bot.equip(346, 'hand', (err) => {
```
rom1504: Retrieve that id from minecraft data

examples/fisherman.js
```diff
@@ -0,0 +1,88 @@
...
+
+  if (nowFishing) {
+    bot.activateItem()
+  }
+}
+
+function eat () {
+  stopFishing()
+
+  bot.equip(349, 'hand', (err) => {
```
rom1504: Retrieve id from minecraft data

examples/repl.js
```diff
@@ -0,0 +1,24 @@
+const mineflayer = require('mineflayer')
+const repl = require('repl')
+
+if (process.argv.length < 4 || process.argv.length > 6) {
+  console.log('Usage : node scoreboard.js <host> <port> [<name>] [<password>]')
```
rom1504: Wrong file name

examples/tab_complete.js
```diff
@@ -0,0 +1,29 @@
+const mineflayer = require('mineflayer')
+
+if (process.argv.length < 4 || process.argv.length > 6) {
+  console.log('Usage : node scoreboard.js <host> <port> [<name>] [<password>]')
```
rom1504: Wrong file name

lib/plugins/chat.js
```diff
@@ -72,10 +72,34 @@ function inject (bot, options) {
...
+    if (sendBlockInSight) {
+      const block = bot.blockInSight()
+
+      if (block) {
+        position = block.position
+      }
+    }
+
+    bot._client.once('tab_complete', (packet) => {
+      cb(packet.matches)
```
rom1504: callback first parameter is error, so here you need to do cb(null, packet.matches)

package.json
```diff
@@ -1,13 +1,13 @@
 {
   "name": "mineflayer",
-  "version": "2.7.5",
+  "version": "2.8.0",
   "description": "create minecraft bots with a stable, high level API",
```
rom1504: Don't do that here, I'll do that in the release commit

test/externalTest.js
```diff
@@ -38,7 +38,7 @@ mineflayer.supportedVersions.forEach((supportedVersion, i) => {
   const PORT = Math.round(30000 + Math.random() * 20000)
   const mcData = require('minecraft-data')(supportedVersion)
   const version = mcData.version
-  const MC_SERVER_JAR_DIR = process.env.MC_SERVER_JAR_DIR
+  const MC_SERVER_JAR_DIR = process.env.MC_SERVER_JAR_DIR || `${process.cwd()}/server_jars`
```
rom1504: Why ?

examples/farmer.js
wvffle: Without the timeouts we could get `Maximum call stack size exceeded` error

package.json
wvffle: Then we can change `prepare` to `prepublish`. `prepare` is also called on `npm install` with no arguments so it's installed 2 times as it's already in `devDependencies`.

wvffle: Even in [usage](https://github.com/yortus/require-self#usage) it's like that

test/externalTest.js
wvffle: I think we should have some default option

examples/farmer.js
rom1504: Ok, do setTimeout(loop, 0)

Putting arbitrary timeout values (non 0) tend to make things work "most of the time" which is not great for the times when it doesn't work.

wvffle: setImmediate(loop) is just the same as setTimeout(loop, 0)

package.json
rom1504: I believe prepublish is deprecated. If you find a way to make it run only on publish it would be perfect though

wvffle: But is it required to install it every time? It's not specified in the usage of require-self

test/externalTest.js
rom1504: Ah, in case you run the tests locally, ok makes sense

package.json
rom1504: Every time what ?
It's required to install it if you want to run it.
If it's put in prepare it makes it possible to run npm publish on a clean copy without having to run npm install.

rom1504: Ideally, I would not want to install it not to run it in case of npm publish.
I only want to run it in case of a local install, not before publishing, but I believe that's not possible ?

docs/api.md
```diff
@@ -1043,25 +1138,55 @@ Fires when you wake up.
 
 Fires when `bot.experience.*` has updated.
 
-#### "scoreboardObjective" (scoreboardName, displayText)
+#### "scoreboardCreated" (scoreboard)
+
+Fires when a scorebard is added.
```
rom1504: Scorebard -> scoreboard (typo)

lib/scoreboard.js
```diff
@@ -1,8 +1,44 @@
...
-  this.displayText = null
-  this.items = {}
-  this.position = null
+const sortItems = (a, b) => {
+  if (a.value > b.value) return -1
+  if (a.value < b.value) return 1
+  return 1
+}
+
+function ScoreBoard (packet) {
```
rom1504: It would be more coherent with the rest of the code to use the "class" keyword for this here
Optional change though, what you did works too

wvffle: I'll change that if we refactor everything to classes as I proposed in #801

package.json
wvffle: I'll put it back then

safazi:
Good job wvffle!

rom1504:
@wvffle I added you as collaborator of mineflayer
please don't push directly to master though, create PR so code can be reviewed :)

wvffle:
Right!

Wqrld:
blockInSight seems to always be undefined for me?

wvffle:
@Wqrld Create new bug report

PR #811, opened 2019-05-07
Correct api.md

safazi:
api.md currently says that bot.equip takes an `Item` instance. This is incorrect.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1328,11 +1328,11 @@ Set the direction your head is facing.
 
 Changes the text on the sign.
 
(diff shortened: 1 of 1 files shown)


wvffle:
I'd rather make bot.equip and all  of similar functions support Item, id, block name, block display name and array of it all/filter function that returns matching items/blocks

safazi:
That's a good idea. Might give that a shot today.

wvffle:
I'd create a helper function that takes anything as input and spits out the id, import it in every file and add to every core function that uses item/block id

Issue #175, opened 2014-04-21
Chat modularization : make chat work for all non-vanilla


roblabla:
Currently, chat support is, IMO, broken. The current implementation kind-of works on vanilla minecraft, but doesn't work any kind of third-party chat systems, and there is no documentation about how to make it work, if possible at all.

As such, we need to modularize chat entirely. I propose removing chat.js entirely, and instead reworking it into a module system similar to #165. 

Essecially, we need to define three things : 
- Our own ChatMessage "class", which would work similarely to minecraft's, without the translation bit. It would have a toString method to return the raw string without any kind of color codes and a sender property to recover the sender.
- A ChatBase "interface", that defines convertMessage(string) and returns a ChatMessage. It also defines and implements a function, stringTojson (silly name, I know), that would transform the old format codes (those with the paragraph signs) into the equivalent proper json message as sent by newer servers. 
- A default chat implementation, that works on vanilla servers.

I think we don't need to actually provide support for any other chat systems, community will provide them as needed. It's not complicated to write one anyway. 

I'll start working on a PR to implement this tomorrow.

@nevercast

rom1504:
related https://github.com/andrewrk/mineflayer/tree/chat-plugins

zuazo:
Hey guys!

Thanks for all the work you are doing in these projects! I'm the one who did the old #165 PR and the [#chat-plugins](https://github.com/andrewrk/mineflayer/tree/chat-plugins) branch was also mine (it is the same code).

That code is old and I don't it will work as it is. Not to mention that it seems it was not the best solution. I think I was not very inspired at that moment. Also, I think that parsing the old (*non-JSON*) format no longer makes sense.

To avoid falling into the same error and seeing that it can be tricky or difficult to implement, I recommend starting by defining the exact API changes and how would be an ideal chat plugin example.

rom1504:
Last issue before 1.1 release, it would be nice to get this done :)

rom1504:
I'm not exactly sure what an "interface" would be in js ?
Also what of https://github.com/andrewrk/mineflayer/blob/master/lib/chat_message.js ?

I'm not sure fixing that issue needs to be blocking for 1.1 : vanilla server chat currently works, other formats probably not but they never really worked. Which mean no regression.
What do you think @roblabla @zuazo ?

roblabla:
Doesn't need to be blocking.

[...]

PR #812, opened 2019-05-08
Add universal chat pattern

wvffle:
To test the regex you can use https://www.regexpal.com/

**Chat regex**
```js
/^(?:\(.+\)|\[.+\]|.)*?(\w+)\s?[>:\-»\]\)~]+\s(.*)$/
```
**Chat tests**
```
Teleported flatbot to 0.5, 4.0, 0.5
[Server: Opped flatbot]
Given  (ID 17) * 255 to flatbot for 1 seconds
Set the display objective in slot 'list' to 'test2'
//-- above should drop --//
wvffle : meh
wvffle: meh
wvffle: meh
wvffle ~ meh
<wvffle> meh
<wvffle> meh (meh)
wvffle: (meh)
moderator wvffle: meh
[wvffle] meh
[a] [b]wvffle » meh
[a] [b] wvffle » meh
[a] ~wvffle: meh
[] wvffle - meh
```

**Whisper regex**
```js
/^(?:\(.+\)|\[.+\]|.)*?(\w+) whispers (?:to you)?:? (.*)$/
/^(?:\(.+\)|\[.+\]|.)*?(\w+) -> \w+\\s?\] (.*)$/ // - \w is there instead of 'me' to support translations
```
**Whisper tests**
```
[wvffle -> me] meh
wvffle whispers meh
wvffle whispers: meh
wvffle whispers to you meh
```
wvffle whispers to you: meh


wvffle:
All of those passes:
image

wvffle:
image

All above examples pass

wvffle:
What about #175?

wvffle:
The only error right now is that we match some server replies
```
username: Server
message: Opped flatbot]
```
and
```
username: Given
message: (ID 17) * 255 to flatbot for 1 seconds
```
image

package.json
```diff
@@ -19,7 +19,7 @@
   },
   "license": "MIT",
   "dependencies": {
-    "minecraft-data": "^2.31.0",
+    "minecraft-data": "^2.36.0",
```
rom1504: Why ?

package.json
```diff
@@ -29,7 +29,7 @@
     "prismarine-item": "^1.0.1",
     "prismarine-recipe": "^1.0.1",
     "prismarine-windows": "^1.1.0",
-    "protodef": "^1.6.7",
```
rom1504: Why ?

1. Unrelated with this review
2. Not needed because of the ^

wvffle: Just merged two of the update branches. I think that dependencies should be updated. Especially minecraft-data ;)

wvffle:
current regex to match `chat` is 
```js
/^(?:\(.+\)|\[.+\]|.)*?(\w+)\s?[>:\-»\]\)~]+\s(.*)$/
```
so basically it matches
image

rom1504:
How many chat plugins does it work on ? Can you add in the doc what chat plugins that add support of ?

wvffle:
I don't really know. I haven't tested it with plugins. I just typed 'minecraft chat' and tried to match all of different chat patterns in screenshots.

The only downside of this approach is that it can match some custom plugin messages.
Messages starting with `[Server:` are filtered out by default.

PR #823, opened 2019-06-05
BossBar update

didjee2:
**Changes:**
- Made BossBar class (located in libs/bossbar.js) to replace the previously used object
- Fixed an error / crash that happened when a bossbar update is sent without a bossbar being registered (or the bossbar already being deleted). New behaviour: now just ignores the bossbar state update when no bossbar is registered / when it already is unregistered (assuming this is normal mc behaviour).

_PS: Ignore my awful commit messages_

### DIFF
--- a/lib/plugins/boss_bar.js
+++ b/lib/plugins/boss_bar.js
@@ -8,7 +8,7 @@ function inject (bot, { version }) {
   const bars = {}
 
   function checkBarExistance (packet) {
-    if (!bars.has(packet.entityUUID)) {
+    if (!(packet.entityUUID in bars)) {
       console.log(' --- BAR DID NOT EXIST IN MAP --- ')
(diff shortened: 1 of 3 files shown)


lib/plugins/boss_bar.js
```diff
@@ -1,47 +1,43 @@
...
+        packet.color,
+        packet.flags
+      )
 
       bot.emit('bossBarCreated', bars[packet.entityUUID])
     } else if (packet.action === 1) {
       bot.emit('bossBarDeleted', bars[packet.entityUUID])
       delete bars[packet.entityUUID]
     } else {
+      if (!(packet.entityUUID in bars)) {
```
didjee2: This is the code that fixes the error / crash when a bar update is received without it being registered.

rom1504:
forgot about the pr sorry, thanks !

PR #830, opened 2019-07-05
accurate knockback handling

Vap0r1ze:


rom1504:
does this work ?

wvffle:
Any source or just pure estimation?

Vap0r1ze:
Yea just guessed and tested, it works really well. Although y axis might
need tweaking

rom1504:
Ok, I'll trust you on that, we can change it again if people complain

PR #853, opened 2019-08-22
fix(entities): make `parseMetadata()` reuse entity metadata

### LINKED ISSUE #828: Entity metadata is lost on every entity_metadata update
## Versions 
 - mineflayer: 2.9.0
 - server: vanilla
 - node: 8.9.4

 For example: Named mobs have name metadata on spawning, but the name metadata is lost soon.
This is because the packet does not contain all metadata. (updated metadata only?)
parseMetadata function in entities.js should use previous entity metadata.

wvffle:
## ABLE TO MERGE
- [x] Yes
- [x] No

## TODO
- [x] Make `parseMetadata()` reuse entity metdata
- [x] Create tests

Fixes #828

### DIFF
--- a/lib/plugins/entities.js
+++ b/lib/plugins/entities.js
@@ -226,7 +226,7 @@ function inject (bot, { version }) {
 
     const notchVel = new Vec3(packet.velocityX, packet.velocityY, packet.velocityZ)
     entity.velocity.update(conv.fromNotchVelocity(notchVel))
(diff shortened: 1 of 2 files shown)


rom1504:
why not merging this @wvffle ?

wvffle:
It'd be good to create tests for this. Unfortunately I have lots of work so basically no time to do this.

rom1504:
Looks like your new test is failing :)

PR #860, opened 2019-09-03
delete npmignore and use whitelist in package.json instead

EvilSourcerer:
Fixes a severe bundle size leak.


EvilSourcerer:
not worth it, closed

PR #864, opened 2019-10-08
Add teams packet to the scoreboard

IdanHo:
Stores data about scoreboard teams.

### DIFF
--- a/lib/plugins/scoreboard.js
+++ b/lib/plugins/scoreboard.js
@@ -4,6 +4,7 @@ module.exports = inject
 
 function inject (bot) {
   const scoreboards = {}
+  const teams = {}
 
   bot._client.on('scoreboard_objective', (packet) => {
     if (packet.action === 0) {
@@ -57,6 +58,47 @@ function inject (bot) {
     }
   })
 
+  bot.on('teams', (packet) => {
+    if (packet.mode === 0) {
+      let team = {};
+      team.displayName = packet.name
+      team.prefix = packet.prefix
+      team.suffix = packet.suffix
+      team.friendlyFire = packet.friendlyFire
+      team.nameTagVisibility = packet.nameTagVisibility
+      team.collisionRule = packet.collisionRule
+      team.color = packet.color
+      team.players = packet.players
+      teams[packet.team] = team
+    }
+
+    if (packet.mode === 1) {
+      delete teams[packet.team]
+    }
+
+    if (packet.mode === 2) {
+      if (teams[packet.team] === undefined) return
+      teams[packet.team].displayName = packet.name
+      teams[packet.team].prefix = packet.prefix
+      teams[packet.team].suffix = packet.suffix
+      teams[packet.team].friendlyFire = packet.friendlyFire
+      teams[packet.team].nameTagVisibility = packet.nameTagVisibility
+      teams[packet.team].collisionRule = packet.collisionRule
+      teams[packet.team].color = packet.color
+    }
+
+    if (packet.mode === 3) {
+            if (teams[packet.team] === undefined) return
+            teams[packet.team].players = teams[packet.team].players.concat(packet.players)
+    }
+
+    if (packet.mode === 4) {
+            if (teams[packet.team] === undefined) return
+            teams[packet.team].players = teams[packet.team].players.filter((el) => !packet.players.includes(el))
+    }
+  })
+
   bot.scoreboards = scoreboards
   bot.scoreboard = ScoreBoard.positions
+  bot.teams = teams
 }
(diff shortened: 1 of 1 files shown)


lib/plugins/scoreboard.js
```diff
@@ -57,6 +58,47 @@ function inject (bot) {
...
+      team.players = packet.players
+      teams[packet.team] = team
+    }
+
+    if (packet.mode === 1) {
+      delete teams[packet.team]
+    }
+
+    if (packet.mode === 2) {
+      if (teams[packet.team] === undefined) return
```
wvffle: Shouldn't we create a team then?

lib/plugins/scoreboard.js
```diff
@@ -57,6 +58,47 @@ function inject (bot) {
...
+      teams[packet.team].displayName = packet.name
+      teams[packet.team].prefix = packet.prefix
+      teams[packet.team].suffix = packet.suffix
+      teams[packet.team].friendlyFire = packet.friendlyFire
+      teams[packet.team].nameTagVisibility = packet.nameTagVisibility
+      teams[packet.team].collisionRule = packet.collisionRule
+      teams[packet.team].color = packet.color
```
wvffle: Are you sure that none of these values are `undefined`?

lib/plugins/scoreboard.js
```diff
@@ -57,6 +58,47 @@ function inject (bot) {
     }
   })
 
+  bot.on('teams', (packet) => {
```
wvffle: I'd rather keep it separate from the `scoreboard.js`

lib/plugins/scoreboard.js
```diff
@@ -57,6 +58,47 @@ function inject (bot) {
...
+    if (packet.mode === 0) {
+      const team = {}
+      team.displayName = packet.name
+      team.prefix = packet.prefix
+      team.suffix = packet.suffix
+      team.friendlyFire = packet.friendlyFire
+      team.nameTagVisibility = packet.nameTagVisibility
+      team.collisionRule = packet.collisionRule
+      team.color = packet.color
+      team.players = packet.players
```
wvffle: I'd map this array to the values of `bot.players` or add `bot.onlinePlayers` which does that

lib/plugins/scoreboard.js
IdanHo: The vanilla client stores team and scoreboard data in the same class, but we can just make a teams.js

lib/plugins/scoreboard.js
IdanHo: the vanilla client just assumes the team exist, its possible this line isnt even needed.

lib/plugins/scoreboard.js
IdanHo: none of these can be undefined, as they all have default values in the vanilla server/client

lib/plugins/scoreboard.js
IdanHo: im not sure what you mean by that, do you mean you want to instead of store a list of players in the team object, store the player's team in each of the team's players?

wvffle: I meant `team.onlinePlayers` which would map `team.players` to `bot.players` values. `player.team` is also a good idea

lib/plugins/scoreboard.js
Artiik373: Shouldn't this be `bot._client.on('scoreboard_team', (packet) => {`?

lib/plugins/scoreboard.js
wvffle: Have you actually tested this?

lib/plugins/scoreboard.js
wvffle: Have you actually tested this?

lib/plugins/scoreboard.js
IdanHo: No? That's just not the name of the packet:
https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/1.12.2/protocol.json#L2593

IdanHo: oh you it does need to be `bot._client` and not just `bot` though

lib/plugins/scoreboard.js
IdanHo: yes ive run it on my own bot on several non vanilla servers already, seems to work fine

lib/plugins/scoreboard.js
IdanHo: yes ive run it on my own bot on several non vanilla servers already, seems to work fine

IdanHo:
seems like i have previously accidentally deleted the fork this pull request represents, and as such im gonna have to close this pull request and reopen it in order to commit the requested changes.

PR #868, opened 2019-10-20
Fix #535

### LINKED ISSUE #535: Expose a bot.swingArm()
I am making a human-like PvP script and need the bot to swing its arm/sword and miss without actually attacking the entity. I can't seem to find this in the API, and was wondering if that was possible?

If not, I guess I could try to get the bot to place a block at the player or "open" the entity to swing its arm. But that may not work out.

Thanks in advance.

wvffle:


rom1504:
Can you use it everywhere we need it in mineflayer internals ?

wvffle:
What do you mean rom?

rom1504:
Everywhere we send arm_animation in mineflayer, instead use this new method

PR #881, opened 2019-11-10
unload chunks on mc version 1.9-1.12

IdanHo:
mc versions 1.9-1.12 dont send an empty map_chunk packet anymore, but rather an unload_chunk packet, implement functionality for it to prevent memory leaks


rom1504:
Thanks !

PR #882, opened 2019-11-10
Expose chunk columns & blockEntities

SiebeDW:


wvffle:
I don't think that this should be exposed in that way.
There should be a method `getColumn(x, y)` which returns deeply sealed object which is a copy of the original column. It'd be a vulnerablity to expose this as a modifiable variable. This should be read only.

rom1504:
honestly I think it doesn't matter too much
this is javascript you can always break things if you do it by purpose

everything under bot._something is anyway marked as "can change at any time, use at your own perils"

wvffle:
LGTM then

PR #891, opened 2019-12-18
Update inventory window slot count

IdanHo:
The current inventory size we use for initializing the slots array for the player inventory window is 44 for some arbitrary reason, the actual value is 46 (https://wiki.vg/index.php?title=Inventory&oldid=14093#Player_Inventory), and since js arrays have an unfixed length when the normal window_items packet is sent to update the player's inventory on login the slots array gets resized to 46 anyways.


wvffle:
LGTM

PR #900, opened 2020-01-20
add bot.tpa & bot.reply

1by:


rom1504:
I don't see the point of this review. We are not going to add one method for each command.
There is no benefit with these methods compared to the user doing bot.chat('/tpa')
Do tell if you see some advantage of this

PR #906, opened 2020-01-29
Add a callback to `bot.creative.setInventorySlot` and fix tests failing randomly

wvffle:
Well, this "randomly" was depending on the latency between client and server in the tests. We didn't have a callback in `bot.creative.setInventorySlot` and depended on `setTimeout`. This PR adds a callback and fixes this error.

### DIFF
--- a/lib/plugins/creative.js
+++ b/lib/plugins/creative.js
@@ -14,8 +14,27 @@ function inject (bot, { version }) {
     stopFlying
   }
 
-  function setInventorySlot (slot, item) {
+  const creativeSlotsUpdates = new Map()
+  bot._client.on('set_slot', ({ windowId, slot, item }) => {
+    if (windowId === 0 && creativeSlotsUpdates.has(slot)) {
+      const cb = creativeSlotsUpdates.get(slot)
+      creativeSlotsUpdates.delete(slot)
+
+      setImmediate(cb)
+    }
+  })
+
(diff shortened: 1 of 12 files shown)


lib/plugins/creative.js
```diff
@@ -14,8 +14,27 @@ function inject (bot, { version }) {
...
+
+  function setInventorySlot (slot, item, cb) {
     assert(slot >= 0 && slot <= 44)
+
+    if (cb != null) {
+      if (creativeSlotsUpdates.has(slot)) {
+        cb(new Error('Server has not set this slot yet'))
+      }
+
+      creativeSlotsUpdates.set(slot, cb)
```
wvffle: I'm storing callbacks in an update map to have only 1  listener on this event. If I'd create a listener per `setInventorySlot` request then quickly we'd run out of memory and exceed listeners limit.

rom1504: can you add in the doc (and comment of this method) that it should not be called on the same slot without waiting for the callback ? (that is a reasonable constraint, but you are indeed making that assumption in your code, so it needs to be clear to avoid surprises from users)

wvffle: Well, if user would set the same slot twice, we have to run the previous callback with an error. I've made a mistake and I've called current callback.

I'll specify in the the doc that you should not call the function on the same slot before the callback of the first run is called.

PR #907, opened 2020-01-29
Add automatically generated changelog

wvffle:


rom1504:
I don't think we should have that.
I write manually in history.md at each release by sumarrizing when needed. If people want to see the commit log, github provide that feature.

PR #921, opened 2020-03-12
update

marinaluisa2020:
extracting example from minecraft protocol

### DIFF
--- a/examples/proxy.socks5.js
+++ b/examples/proxy.socks5.js
@@ -1,0 +1,46 @@
+/*
+ *
+ * Update Proxy Minecraft Protocol Example for MineFlayer Proxy Example
+ *
+ */
+const mineflayer = require('mineflayer')
+const socks = require('socks').SocksClient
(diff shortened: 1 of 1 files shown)


examples/proxy.socks5.js
```diff
@@ -0,0 +1,46 @@
...
+    }, (err, info) => {
+      if (err) {
+        console.log(err)
+        return
+      }
+
+      client.setSocket(info.socket)
+      client.emit('connect')
+    })
+  },
```
wvffle: Is it required whilst using a proxy agent?

examples/proxy.socks5.js
```diff
@@ -0,0 +1,46 @@
...
+  agent: new ProxyAgent({ protocol: 'socks5', host: proxyHost, port: proxyPort }),
```
wvffle: There are no such variables as proxyHost and proxyPort

PR #922, opened 2020-03-22
Add water physics

Bayugoon:
Added some code to physics.js so the bot will be pushed by water and lava

### DIFF
--- a/lib/plugins/physics.js
+++ b/lib/plugins/physics.js
@@ -1,4 +1,3 @@
-//Mineflayer's Physics with water physics added
 const Vec3 = require('vec3').Vec3
 const assert = require('assert')
 const math = require('../math')
@@ -223,18 +222,18 @@ function inject (bot) {
     return vel.x * vel.x + vel.z * vel.z
   }
 
-  function liquidPush(pos){
-    let flow = liquidFlow(pos);
-    let push = flow.clone().scaled(physics.liquidPush);
+  function liquidPush (pos) {
+    const flow = liquidFlow(pos);
+    const push = flow.clone().scaled(physics.liquidPush);
     push.y = push.y/physics.gravity;
-    console.log(flow);
-    return push;
+    return push
   }
 
-  function liquidFlow(pos) {
-    let flow = new Vec3(0,0,0);
-    let base = bot.blockAt(pos).metadata;
-    let id = bot.blockAt(pos).type;
+  function liquidFlow (pos) {
+    const flow = new Vec3(0,0,0);
+    const block = bot.blockAt(pos)
+    const base = block.metadata;
+    const id = block.type;
     if(base > 7) {
       base = 0;
       flow.set(0,-1,0);
@@ -243,25 +242,24 @@ function inject (bot) {
       if(level > 7) return 0;
       return level; 
     }
-    let Blocks = {};
-    Blocks['x+'] = bot.blockAt(pos.offset(1,0,0));
-    Blocks['x-'] = bot.blockAt(pos.offset(-1,0,0));
-    Blocks['z+'] = bot.blockAt(pos.offset(0,0,1));
-    Blocks['z-'] = bot.blockAt(pos.offset(0,0,-1));
-    let waterBlocks = {}
-    for(let i = 0; i < Object.keys(Blocks).length; i++){
-      key = Object.keys(Blocks)[i];
+    const Blocks = {
+      'x+': bot.blockAt(pos.offset(1,0,0)),
+      'x-': bot.blockAt(pos.offset(-1,0,0)),
+      'z+': bot.blockAt(pos.offset(0,0,1)),
+      'z-': bot.blockAt(pos.offset(0,0,-1))
+    }
+    const waterBlocks = {}
+    for(block of Object.entries(Blocks)){
       if(Blocks[key].type == id) waterBlocks[key] = depth(Blocks[key].metadata)
     }
-    console.log(waterBlocks);
-    if(waterBlocks['x+'] !== null && waterBlocks['x+'] > base) flow = flow.offset(1,0,0);
-    if(waterBlocks['x-'] !== null && waterBlocks['x-'] > base) flow = flow.offset(-1,0,0);
-    if(waterBlocks['z+'] !== null && waterBlocks['z+'] > base) flow = flow.offset(0,0,1);
-    if(waterBlocks['z-'] !== null && waterBlocks['z-'] > base) flow = flow.offset(0,0,-1);
-    if(waterBlocks['x+'] !== null && waterBlocks['x+'] < base && flow.x == 0) flow = flow.offset(-1,0,0)
-    if(waterBlocks['x-'] !== null && waterBlocks['x-'] < base && flow.x == 0) flow = flow.offset(1,0,0);
-    if(waterBlocks['z+'] !== null && waterBlocks['z+'] < base && flow.z == 0) flow = flow.offset(0,0,-1)
-    if(waterBlocks['z-'] !== null && waterBlocks['z-'] < base && flow.z == 0) flow = flow.offset(0,0,1);
+    if(waterBlocks['x+'] !== null && waterBlocks['x+'] > base) flow = flow.add(1,0,0);
+    if(waterBlocks['x-'] !== null && waterBlocks['x-'] > base) flow = flow.add(-1,0,0);
+    if(waterBlocks['z+'] !== null && waterBlocks['z+'] > base) flow = flow.add(0,0,1);
+    if(waterBlocks['z-'] !== null && waterBlocks['z-'] > base) flow = flow.add(0,0,-1);
+    if(waterBlocks['x+'] !== null && waterBlocks['x+'] < base && flow.x == 0) flow = flow.add(-1,0,0)
+    if(waterBlocks['x-'] !== null && waterBlocks['x-'] < base && flow.x == 0) flow = flow.add(1,0,0);
+    if(waterBlocks['z+'] !== null && waterBlocks['z+'] < base && flow.z == 0) flow = flow.add(0,0,-1)
+    if(waterBlocks['z-'] !== null && waterBlocks['z-'] < base && flow.z == 0) flow = flow.add(0,0,1);
     return flow;
   }
 
--- a/lib/plugins/physics.js
+++ b/lib/plugins/physics.js
@@ -223,44 +223,44 @@ function inject (bot) {
   }
 
   function liquidPush (pos) {
-    const flow = liquidFlow(pos);
-    const push = flow.clone().scaled(physics.liquidPush);
-    push.y = push.y/physics.gravity;
+    const flow = liquidFlow(pos)
+    const push = flow.clone().scaled(physics.liquidPush)
+    push.y = push.y / physics.gravity
     return push
   }
 
   function liquidFlow (pos) {
-    const flow = new Vec3(0,0,0);
+    const flow = new Vec3(0, 0, 0)
     const block = bot.blockAt(pos)
-    const base = block.metadata;
-    const id = block.type;
-    if(base > 7) {
-      base = 0;
-      flow.set(0,-1,0);
+    const base = block.metadata
+    const id = block.type
+    if (base > 7) {
+      base = 0
+      flow.set(0, -1, 0)
     }
     const depth = level => {
-      if(level > 7) return 0;
-      return level; 
+      if (level > 7) return 0
+      return level
     }
     const Blocks = {
-      'x+': bot.blockAt(pos.offset(1,0,0)),
-      'x-': bot.blockAt(pos.offset(-1,0,0)),
-      'z+': bot.blockAt(pos.offset(0,0,1)),
-      'z-': bot.blockAt(pos.offset(0,0,-1))
+      'x+': bot.blockAt(pos.offset(1, 0, 0)),
+      'x-': bot.blockAt(pos.offset(-1, 0, 0)),
+      'z+': bot.blockAt(pos.offset(0, 0, 1)),
+      'z-': bot.blockAt(pos.offset(0, 0, -1))
     }
     const waterBlocks = {}
-    for(block of Object.entries(Blocks)){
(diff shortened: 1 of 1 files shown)


lib/plugins/physics.js
```diff
@@ -1,3 +1,4 @@
+//Mineflayer's Physics with water physics added
```
wvffle: unneeded

lib/plugins/physics.js
```diff
@@ -134,7 +136,8 @@ function inject (bot) {
     const inBlock = bot.blockAt(pos.offset(0, 0, 0))
     if (underBlock && underBlock.type === 88) {
       currentMaxGroundSpeed = physics.maxGroundSpeedSoulSand
-    } else if (inBlock && inBlock.type === 9) {
+    } else if (inBlock && (inBlock.type === 9 || inBlock.type === 11)) {
```
wvffle: Does it work the same for water and lava? isn't lava slower?

lib/plugins/physics.js
```diff
@@ -220,6 +223,48 @@ function inject (bot) {
...
+    }
+    const depth = level => {
+      if(level > 7) return 0;
+      return level; 
+    }
+    let Blocks = {};
+    Blocks['x+'] = bot.blockAt(pos.offset(1,0,0));
+    Blocks['x-'] = bot.blockAt(pos.offset(-1,0,0));
+    Blocks['z+'] = bot.blockAt(pos.offset(0,0,1));
+    Blocks['z-'] = bot.blockAt(pos.offset(0,0,-1));
```
wvffle: Assigning new keys is slower than creating new object with all of the fields.

lib/plugins/physics.js
```diff
@@ -220,6 +223,48 @@ function inject (bot) {
...
+    let id = bot.blockAt(pos).type;
+    if(base > 7) {
+      base = 0;
+      flow.set(0,-1,0);
+    }
+    const depth = level => {
+      if(level > 7) return 0;
+      return level; 
+    }
+    let Blocks = {};
```
wvffle: use const

lib/plugins/physics.js
```diff
@@ -220,6 +223,48 @@ function inject (bot) {
...
+    }
+    console.log(waterBlocks);
+    if(waterBlocks['x+'] !== null && waterBlocks['x+'] > base) flow = flow.offset(1,0,0);
+    if(waterBlocks['x-'] !== null && waterBlocks['x-'] > base) flow = flow.offset(-1,0,0);
+    if(waterBlocks['z+'] !== null && waterBlocks['z+'] > base) flow = flow.offset(0,0,1);
+    if(waterBlocks['z-'] !== null && waterBlocks['z-'] > base) flow = flow.offset(0,0,-1);
+    if(waterBlocks['x+'] !== null && waterBlocks['x+'] < base && flow.x == 0) flow = flow.offset(-1,0,0)
+    if(waterBlocks['x-'] !== null && waterBlocks['x-'] < base && flow.x == 0) flow = flow.offset(1,0,0);
+    if(waterBlocks['z+'] !== null && waterBlocks['z+'] < base && flow.z == 0) flow = flow.offset(0,0,-1)
+    if(waterBlocks['z-'] !== null && waterBlocks['z-'] < base && flow.z == 0) flow = flow.offset(0,0,1);
```
wvffle: Isn't there `flow.add()`? If so make flow a const and use it.

lib/plugins/physics.js
```diff
@@ -220,6 +223,48 @@ function inject (bot) {
...
+    Blocks['x+'] = bot.blockAt(pos.offset(1,0,0));
+    Blocks['x-'] = bot.blockAt(pos.offset(-1,0,0));
+    Blocks['z+'] = bot.blockAt(pos.offset(0,0,1));
+    Blocks['z-'] = bot.blockAt(pos.offset(0,0,-1));
+    let waterBlocks = {}
+    for(let i = 0; i < Object.keys(Blocks).length; i++){
+      key = Object.keys(Blocks)[i];
+      if(Blocks[key].type == id) waterBlocks[key] = depth(Blocks[key].metadata)
+    }
+    console.log(waterBlocks);
```
wvffle: Remove the log

lib/plugins/physics.js
```diff
@@ -220,6 +223,48 @@ function inject (bot) {
...
+    let push = flow.clone().scaled(physics.liquidPush);
+    push.y = push.y/physics.gravity;
+    console.log(flow);
+    return push;
+  }
+
+  function liquidFlow(pos) {
+    let flow = new Vec3(0,0,0);
+    let base = bot.blockAt(pos).metadata;
+    let id = bot.blockAt(pos).type;
```
wvffle: use const

wvffle: add `const block = bot.blockAt(pos)` and use it instead of calling blockAt multiple times

lib/plugins/physics.js
```diff
@@ -220,6 +223,48 @@ function inject (bot) {
     return vel.x * vel.x + vel.z * vel.z
   }
 
+  function liquidPush(pos){
+    let flow = liquidFlow(pos);
+    let push = flow.clone().scaled(physics.liquidPush);
+    push.y = push.y/physics.gravity;
+    console.log(flow);
```
wvffle: Remove

lib/plugins/physics.js
```diff
@@ -220,6 +223,48 @@ function inject (bot) {
     return vel.x * vel.x + vel.z * vel.z
   }
 
+  function liquidPush(pos){
+    let flow = liquidFlow(pos);
+    let push = flow.clone().scaled(physics.liquidPush);
```
wvffle: use const

lib/plugins/physics.js
```diff
@@ -220,6 +223,48 @@ function inject (bot) {
...
+    const depth = level => {
+      if(level > 7) return 0;
+      return level; 
+    }
+    let Blocks = {};
+    Blocks['x+'] = bot.blockAt(pos.offset(1,0,0));
+    Blocks['x-'] = bot.blockAt(pos.offset(-1,0,0));
+    Blocks['z+'] = bot.blockAt(pos.offset(0,0,1));
+    Blocks['z-'] = bot.blockAt(pos.offset(0,0,-1));
+    let waterBlocks = {}
```
wvffle: use const

lib/plugins/physics.js
```diff
@@ -220,6 +223,48 @@ function inject (bot) {
...
+    for(let i = 0; i < Object.keys(Blocks).length; i++){
+      key = Object.keys(Blocks)[i];
```
wvffle: use for with `Object.entries()`

lib/plugins/physics.js
wvffle: Also try to add it to the above for. It can be simpler.

lib/plugins/physics.js
Bayugoon: lava isn't slower

lib/plugins/physics.js
Bayugoon: sorry I forgot to remove that

lib/plugins/physics.js
Bayugoon: I will fix it and run standard --fix

Bayugoon: but don't I need the key so I can add entries to the waterBlocks object?

lib/plugins/physics.js
```diff
@@ -220,6 +222,47 @@ function inject (bot) {
...
+      if(Blocks[key].type == id) waterBlocks[key] = depth(Blocks[key].metadata)
+    }
+    if(waterBlocks['x+'] !== null && waterBlocks['x+'] > base) flow = flow.add(1,0,0);
+    if(waterBlocks['x-'] !== null && waterBlocks['x-'] > base) flow = flow.add(-1,0,0);
+    if(waterBlocks['z+'] !== null && waterBlocks['z+'] > base) flow = flow.add(0,0,1);
+    if(waterBlocks['z-'] !== null && waterBlocks['z-'] > base) flow = flow.add(0,0,-1);
+    if(waterBlocks['x+'] !== null && waterBlocks['x+'] < base && flow.x == 0) flow = flow.add(-1,0,0)
+    if(waterBlocks['x-'] !== null && waterBlocks['x-'] < base && flow.x == 0) flow = flow.add(1,0,0);
+    if(waterBlocks['z+'] !== null && waterBlocks['z+'] < base && flow.z == 0) flow = flow.add(0,0,-1)
+    if(waterBlocks['z-'] !== null && waterBlocks['z-'] < base && flow.z == 0) flow = flow.add(0,0,1);
```
IdanHo: flow is a constant, add does inplace editing of the vector, removing the `flow = ` in all of these lines
also, it looks like you didnt run standard --fix after doing wvvfle's requested changes

Bayugoon: oops

Bayugoon: I also need to change something

lib/plugins/physics.js
```diff
@@ -223,44 +223,44 @@ function inject (bot) {
...
-    if(waterBlocks['z-'] !== null && waterBlocks['z-'] < base && flow.z == 0) flow = flow.add(0,0,1);
-    return flow;
+    if (waterBlocks['x+'] !== null && waterBlocks['x+'] > base) flow = flow.add(1, 0, 0)
+    if (waterBlocks['x-'] !== null && waterBlocks['x-'] > base) flow = flow.add(-1, 0, 0)
+    if (waterBlocks['z+'] !== null && waterBlocks['z+'] > base) flow = flow.add(0, 0, 1)
+    if (waterBlocks['z-'] !== null && waterBlocks['z-'] > base) flow = flow.add(0, 0, -1)
+    if (waterBlocks['x+'] !== null && waterBlocks['x+'] < base && flow.x == 0) flow = flow.add(-1, 0, 0)
+    if (waterBlocks['x-'] !== null && waterBlocks['x-'] < base && flow.x == 0) flow = flow.add(1, 0, 0)
+    if (waterBlocks['z+'] !== null && waterBlocks['z+'] < base && flow.z == 0) flow = flow.add(0, 0, -1)
+    if (waterBlocks['z-'] !== null && waterBlocks['z-'] < base && flow.z == 0) flow = flow.add(0, 0, 1)
```
IdanHo: still trying to update a constant...
Also, flow.x/flow.z should be compared with strict type comparsion (===)

wvffle: .add updates current value. Don't reassign it.

lib/plugins/physics.js
```diff
@@ -220,6 +222,47 @@ function inject (bot) {
...
+    }
+    const Blocks = {
+      'x+': bot.blockAt(pos.offset(1, 0, 0)),
+      'x-': bot.blockAt(pos.offset(-1, 0, 0)),
+      'z+': bot.blockAt(pos.offset(0, 0, 1)),
+      'z-': bot.blockAt(pos.offset(0, 0, -1))
+    }
+    const waterBlocks = {}
+    for (block of Object.entries(Blocks)) {
+      if (Blocks[key].type == id) waterBlocks[key] = depth(Blocks[key].metadata)
```
wvffle: You haven't even tested this...

```js
for (const [key, block] of blocks) {
  if (block.type === id) {
    waterBlocks[key] = depth(block.metadata)
  }
}
```

lib/plugins/physics.js
wvffle: Still, make a map how flow should change at each neighbor and apply it dynamically in the `for` above

lib/plugins/physics.js
```diff
@@ -220,6 +222,47 @@ function inject (bot) {
...
+    const id = block.type
+    if (base > 7) {
+      base = 0
+      flow.set(0, -1, 0)
+    }
+    const depth = level => {
+      if (level > 7) return 0
+      return level
+    }
+    const Blocks = {
```
wvffle: pascal case?

lib/plugins/physics.js
```diff
@@ -220,6 +222,47 @@ function inject (bot) {
...
+    const base = block.metadata
+    const id = block.type
+    if (base > 7) {
+      base = 0
+      flow.set(0, -1, 0)
+    }
+    const depth = level => {
+      if (level > 7) return 0
+      return level
+    }
```
wvffle: Not a descriptive name

lib/plugins/physics.js
```diff
@@ -220,6 +222,47 @@ function inject (bot) {
...
+
+  function liquidFlow (pos) {
+    const flow = new Vec3(0, 0, 0)
+    const block = bot.blockAt(pos)
+    const base = block.metadata
+    const id = block.type
+    if (base > 7) {
+      base = 0
+      flow.set(0, -1, 0)
+    }
```
wvffle: No newlines, unreadable code

lib/plugins/physics.js
```diff
@@ -220,6 +222,47 @@ function inject (bot) {
     return vel.x * vel.x + vel.z * vel.z
   }
 
+  function liquidPush (pos) {
+    const flow = liquidFlow(pos)
+    const push = flow.clone().scaled(physics.liquidPush)
```
wvffle: why do you clone the flow?

wvffle:
Recalculating flow every physics tick is slow. You can calculate water flow once for example when entity enters water or when the chunk is loaded and recalculate it only when the water block is updated. Then simply apply the force of the current block that the entity is at.

lib/plugins/physics.js
rom1504: Could you retrieve the block IDs from minecraft data instead of hardcoding them ?

lib/plugins/physics.js
Cj1m: flow.add(0, 0, 1) won't work, add takes a Vec3 as a parameter, needs to be flow.add(new Vec3(0, 0, 1))

rom1504:
Am I right that this could now be closed @Karang ?

Karang:
> Am I right that this could now be closed @Karang ?

yes this can be closed

PR #934, opened 2020-04-02
Updated entities.js

Snoekoog:
Fix for a bug I encountered. Happened to me on a bukkit/spigot server, not sure if Vanilla has the same issue. Not sure if this is the right way to fix this but worked for me so far.


wvffle:
Which version have you been using when the bug happened? Let me know minecraft, mineflayer and node versions.

Snoekoog:
Node: v13.12.0
Minecraft: 1.8.9
Mineflayer: 2.11.0

PR #937, opened 2020-04-05
修复1.11.2钓鱼不收杆问题

hutu13879513663:


rom1504:
Why 1.23 ?

hutu13879513663:
调试发现数值在1.113左右，往大一点的数我比较喜欢1.23，因为 1 2 3 ,哈哈
Debugging found that the value is around 1.113, I prefer 1.23 to the larger number, because 1 2 3  --Translated by Google

PR #938, opened 2020-04-05
Bump license year

1by:


rom1504:
actually it's better to keep the oldest date, since copyright is valid since then

Issue #661, opened 2018-05-02
fix crafting toss items


rom1504:
that bug came back (found with rbot)...

Kshitij-Jande:
Hey, so how about using an alternative method like *bot.moveSlotItem()* and crafting?

Kshitij-Jande:
And, how to do it?

PR #940, opened 2020-04-10
Fix issue: Item tossed when crafted

### LINKED ISSUE #905: Crafting Problem (Item tossed when crafted)
**Versions**
 - mineflayer: 1.10.0
 - server: Spigot 1.12.2
 - node: 10.16.3

Ok, so the question I have it that, I want a bot to craft 36x Stacks of Iron Ingots from 4 Stacks of Iron Blocks. I want to achieve this only using the 2x2 grid in the inventory. Also, I want to make it Fast, and Efficient (I mean, that since some servers lag, the crafting becomes a bit slow). I've tried this:
```javascript
function craftItem (name, amount) {
  amount = parseInt(amount, 10)
  const item = require('minecraft-data')(bot.version).findItemOrBlockByName(name)
  const craftingTable = bot.findBlock({
    matching: 58
  })
  blockClick(2832, 95, 2101)
  if (item) {
    const recipe = bot.recipesFor(item.id, null, 1, null)[0]
    if (recipe) {
      console.log(`Crafting Iron Ingots...`)
      bot.craft(recipe, amount, null, (err) => {
        if (err) {
          console.log(`An Error Occured!`)
        } else {
          console.log(`Crafted all Iron Ingots successfully.`)
        }
      })
    } else {
      console.log(`I cannot make ${name}`)
    }
  } else {
    console.log(`unknown item: ${name}`)
  }
}
```
This is from the Examples, but as soon as the bot crafts it, it drops the Iron Ingots.
Also, how do I use the *bot.moveSlotItem()* to maybe craft the items in the inventory.
I've tried this:
```javascript
bot.on('chat', (username, message) => {
    if (message == 'craft') {
        bot.clickWindow(36)
        setTimeout(() => {
            bot.putAway(1)
            setTimeout(() => {
                bot.clickWindow(0)
                setTimeout(() => {
                    bot.putAway(9)
                }, 1000);
            }, 1000);
        }, 1000);
    }
})
```
It'd be great if you can help me! Thanks in advance! :)

Karang:
When picking the crafting result, the server update the slot before acknowledging the click:

```
client->server: play window_click : {"windowId":0,"slot":0,"mouseButton":0,"action":5,"mode":0,"item":{"blockId":58,"itemCount":1,"itemDamage":0}}
client<-server: play.unlock_recipes :{"action":1,"craftingBookOpen":false,"filteringCraftable":false,"recipes1":[]}
client<-server: play.set_slot :{"windowId":0,"slot":0,"item":{"blockId":-1}}
client<-server: play.set_slot :{"windowId":0,"slot":0,"item":{"blockId":72,"itemCount":1,"itemDamage":0}}
client<-server: play.set_slot :{"windowId":0,"slot":0,"item":{"blockId":143,"itemCount":1,"itemDamage":0}}
client<-server: play.set_slot :{"windowId":0,"slot":0,"item":{"blockId":-1}}
client<-server: play.transaction :{"windowId":0,"action":5,"accepted":true}
```

This result in slots[0] being set to null before `window.acceptSwapAreaLeftClick(click)` is called, so it cannot set `window.selectedItem` to the picked item (it is set to null).
Because of that, `putSelectedItemRange` does nothing and the item is toss later because the item was never stored in an inventory slot.

The solution I have is to save the item that was in the slot at the moment of the click (in the click object itself) so we can access it when the click is accepted, even if the slot has been changed (possibly multiple time) by the server.

I also noticed another bug in https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/craft.js#L154

When forcing the crafting result to be the result of the recipe, instead of what the server sent, it can be that the metadata of the object is changed (for instance, when crafting planks from different kind of logs), so I just removed the lines 153-154.

Goes with https://github.com/PrismarineJS/prismarine-windows/pull/6

Close #905 #661

### DIFF
--- a/lib/plugins/craft.js
+++ b/lib/plugins/craft.js
@@ -149,10 +149,7 @@ function inject (bot, { version }) {
 
       function grabResult () {
         assert.strictEqual(window.selectedItem, null)
-        // put the recipe result in the output
-        const item = new Item(recipe.result.id, recipe.result.count, recipe.result.metadata)
-        window.updateSlot(0, item)
-        // shift click result
+        // move the result to inventory
         bot.putAway(0, (err) => {
           if (err) {
             cb(err)
--- a/lib/plugins/inventory.js
+++ b/lib/plugins/inventory.js
@@ -508,7 +508,8 @@ function inject (bot, { version }) {
       mouseButton,
       mode,
       id: actionId,
-      windowId: window.id
+      windowId: window.id,
+      item: slot === -999 ? null : window.slots[slot]
     }
     windowClickQueue.push(click)
     bot._client.write('window_click', {


wvffle:
Seems like the tests fail, are they working after merging PrismarineJS/prismarine-windows#6?

Karang:
> Seems like the tests fail, are they working after merging [PrismarineJS/prismarine-windows#6](https://github.com/PrismarineJS/prismarine-windows/pull/6)?

Yes, they are expected to fail without the changes in prismaring-windows. Is there a way to tell circle-ci to use another branch of prismarine-window?

rom1504:
test are failing since merging this https://circleci.com/gh/PrismarineJS/mineflayer/1064
https://circleci.com/gh/PrismarineJS/mineflayer/tree/master

if you know how to fix this Karang then please fix, otherwise I'll revert later

PR #941, opened 2020-04-10
Improve bot.sleep function

### LINKED ISSUE #568: bot.sleep() does not give error for when monsters are nearby.
bot.sleep() should give an error when it fails, and does so when it's not night, but doesn't give an error if it fails due to monsters nearby.
Can easily be circumvented externally, just thought i'd let you know since it seems like it'd be an internal thing.

ImHarvol:
Fixes #568 by throwing an error if you try to sleep when there's a monster nearby


lib/plugins/bed.js
```diff
@@ -24,8 +24,18 @@ function inject (bot) {
     } else if (bedBlock.type !== 26) {
       cb(new Error('wrong block : not a bed block'))
     } else {
-      bot.activateBlock(bedBlock)
-      cb()
+      bot.activateBlock(bedBlock, () => {
+        const sleepTimeout = setTimeout(() => {
+          const sleepCb = bot.listeners('sleep').find(f => f.name === 'sleepCb')
```
wvffle, line 29: sleepCb should be available in this scope

lib/plugins/bed.js
```diff
@@ -24,8 +24,18 @@ function inject (bot) {
...
+          bot.removeListener('sleep', sleepCb)
+          cb(new Error('there are monsters nearby'))
+        }, 500)
```
wvffle, line 32: I don't think that a timeout would do it. What if bot has high ping and exceed the timeout?

ImHarvol: Would 1.5s be okay or what value would you suggest?

ImHarvol: So what about [this](https://github.com/PrismarineJS/mineflayer/issues/568#issuecomment-578296191) solution?

ImHarvol:
I've looking for how many blocks away should a monster be from a bed to sleep in the bed. In https://minecraft.gamepedia.com/Bed it says its 8 blocks, but I've been testing it and it seems to depend on the orientation of the bed. If anyone knows about it, would be helpful
I'll modify the PR

wvffle:
> I've looking for how many blocks away should a monster be from a bed to sleep in the bed. In [minecraft.gamepedia.com/Bed](https://minecraft.gamepedia.com/Bed) it says its 8 blocks, but I've been testing it and it seems to depend on the orientation of the bed. If anyone knows about it, would be helpful
> I'll modify the PR

It's probably 8 blocks away from one of the 2 blocks

lib/plugins/bed.js
```diff
@@ -24,17 +54,33 @@ function inject (bot) {
...
+          clearTimeout(sleepTimeout)
+          cb()
+        }
+
         const sleepTimeout = setTimeout(() => {
-          const sleepCb = bot.listeners('sleep').find(f => f.name === 'sleepCb')
           bot.removeListener('sleep', sleepCb)
           cb(new Error('there are monsters nearby'))
-        }, 500)
+        }, 1000)
```
wvffle, line 81: The problem will still exist, adding timeout won't help. On the other side 1s ping is kind of unplayable...

lib/plugins/bed.js
```diff
@@ -1,8 +1,38 @@
...
 
 function inject (bot) {
   bot.isSleeping = false
 
+  function parseBedMetadata (bedBlock) {
+    const metadata = {}
+    const bMetadata = bedBlock.metadata.toString(2).padStart(4, '0')
+
+    metadata.part = Number(bMetadata[0])
+    metadata.occupied = Number(bMetadata[1])
```
wvffle, line 13: This will be a bit faster in V8:
```js
const metadata = {
  part: // ...
  occupied: // ...
}
```

Also add the default outcome (judging by L10 it's `00`) to the object at creation time and then simply modify the string and update the vector without creating a new instance

And my personal preference is to use `+"8"` rather than `Number` or `parseInt`. Less writing and kind of understandable if you know how javascript works

lib/plugins/bed.js
```diff
@@ -1,8 +1,38 @@
+const Vec3 = require('vec3').Vec3
+
 module.exports = inject
 
 function inject (bot) {
   bot.isSleeping = false
 
+  function parseBedMetadata (bedBlock) {
+    const metadata = {}
+    const bMetadata = bedBlock.metadata.toString(2).padStart(4, '0')
```
wvffle, line 10: Name it something like `bitMetadata`

lib/plugins/bed.js
```diff
@@ -24,17 +54,33 @@ function inject (bot) {
     } else if (bedBlock.type !== 26) {
       cb(new Error('wrong block : not a bed block'))
     } else {
+      const metadata = parseBedMetadata(bedBlock)
+      let headBlock = bedBlock
+
+      if (metadata.part === 0) {
+        headBlock = bot.blockAt(bedBlock.position.add(metadata.headOffset))
+        if (headBlock.type !== 26 || parseBedMetadata(headBlock).part === 0) {
```
wvffle, line 62: Do we need to parse the metadata again?

lib/plugins/bed.js
wvffle: @ImHarvol I opt for that solution you've proposed before. The thing is that with timeouts it's not that easy to do. Remember that you may lay in bed for entire night just because other players won't go to sleep. In that case any timeout fails.

But unfortunately some non-vanilla servers may disable this message or send a custom one

lib/plugins/bed.js
ImHarvol: I don't see what the `+"8"` would do
Thanks for the feedback btw!

wvffle: Oh, it's just a quick way to convert to a number

lib/plugins/bed.js
```diff
@@ -2,32 +2,33 @@ const Vec3 = require('vec3').Vec3
...
+      occupied: parseInt(bitMetadata[1], 2),
+      facing: 0,
+      headOffset: new Vec3(0, 0, 1)
+    }
+
+    switch (bitMetadata.slice(2, 4)) {
       case '01':
-        metadata.facing = 'west'
+        metadata.facing = 1
         metadata.headOffset = new Vec3(-1, 0, 0)
```
wvffle, line 23: Use `metadata.headOffset.set(-1, 0, 0)`

lib/plugins/bed.js
```diff
@@ -55,20 +56,43 @@ function inject (bot) {
       cb(new Error('wrong block : not a bed block'))
     } else {
       const metadata = parseBedMetadata(bedBlock)
-      let headBlock = bedBlock
+      let headPoint = bedBlock.position
+
+      if (metadata.occupied === 1) return cb(new Error('the bed is occupied'))
```
wvffle, line 61: In my opinion single line if statements are kind of unreadable but it's up to you if standard isn't throwing an error there

lib/plugins/bed.js
```diff
@@ -2,32 +2,33 @@ const Vec3 = require('vec3').Vec3
...
-    switch (bMetadata.slice(2, 4)) {
-      case '00':
-        metadata.facing = 'south'
-        metadata.headOffset = new Vec3(0, 0, 1)
-        break
+    const bitMetadata = bedBlock.metadata.toString(2).padStart(4, '0')
+
+    const metadata = {
+      part: parseInt(bitMetadata[0], 2),
+      occupied: parseInt(bitMetadata[1], 2),
```
wvffle, line 15: Why are you using binary there?

wvffle:
Could you add an internal test to check if it works as expected?

ImHarvol:
> Could you add an internal test to check if it works as expected?

Sure, I'll add tests after I add the monster handling

lib/plugins/bed.js
ImHarvol: Are you suggesting to use Booleans?

wvffle: no, just thinking why binary and not decimal?

ImHarvol: `bedBlock.metadata` returns a int and if I want to know what that number means I have to parse it to binary and interpret 
 each bit (https://minecraft.gamepedia.com/Bed). I don't know if there's any workaraund

ImHarvol: `bedBlock.metadata` returns a int and if I want to know what that number means I have to parse it to binary and interpret 
 each bit (https://minecraft.gamepedia.com/Bed). I don't know if there's any workaraund

ImHarvol:
The bot should be able to sleep in a thunderstorm even if it's not night. It's easy to check if its raining but, is there some way to know if it's a thunderstorm?

lib/plugins/bed.js
```diff
@@ -78,34 +81,39 @@ function inject (bot) {
...
-      if (botPos.x > seCorner.x || botPos.x < nwCorner.x || botPos.y > seCorner.y || botPos.y < nwCorner.y || botPos.z > seCorner.z || botPos.z < nwCorner.z) return cb(new Error('the bed is too far'))
-
-      bot.activateBlock(bedBlock, () => {
-        function sleepCb () {
-          clearTimeout(sleepTimeout)
-          cb()
-        }
+      const nwClickCorner = headPoint.offset(clickRange[1], -2, clickRange[2]) // North-West lower corner
+      const seClickCorner = headPoint.offset(clickRange[3], 2, clickRange[0]) // South-East upper corner
+      if (botPos.x > seClickCorner.x || botPos.x < nwClickCorner.x || botPos.y > seClickCorner.y || botPos.y < nwClickCorner.y || botPos.z > seClickCorner.z || botPos.z < nwClickCorner.z) {
```
wvffle, line 100: Are you sure it's a square around and not radius?

ImHarvol: I've been doing some testing, so I'm quite sure

test/internalTest.js
```diff
@@ -343,6 +343,84 @@ mineflayer.supportedVersions.forEach((supportedVersion, i) => {
...
+        })
+
+        done()
+      })
+
+      server.once('login', (client) => {
+        bot.time.day = 18000
+        client.write('login', {
+          entityId: 0,
+          levelType: 'fogetaboutit',
```
wvffle, line 373: foget?

ImHarvol: I did just copy the code from the 'blockAt' test. Didn't notice!

ImHarvol:
I'm almost done making it work with 1.13 too (beds changed a bit in that version), but I need some help to commit:
Now that 1.13 support has been released, my fork branch is many commits behind mineflayer's master, so how should I update my branch? Fetch from upstream, merge and push? Or should I rebase?

lib/plugins/bed.js
```diff
@@ -10,35 +10,64 @@ function noop (err) {
...
+          metadata.headOffset.set(-1, 0, 0)
+          break
+        case '11':
+          metadata.facing = 3
+          metadata.headOffset.set(1, 0, 0)
+      }
+    } else {
+      const bitMetadata = bedBlock.metadata.toString(2).padStart(4, '0') // PART (1 bit) | OCCUPIED (1 bit) | FACING (2 bits)
+      metadata.part = Boolean(parseInt(bitMetadata[0], 2))
+      metadata.occupied = Boolean(parseInt(bitMetadata[1], 2))
```
wvffle, line 56: Wooow, use `!!`, V8 should optimize that

lib/plugins/bed.js
```diff
@@ -10,35 +10,64 @@ function noop (err) {
...
-      case '10':
-        metadata.facing = 2
-        metadata.headOffset.set(0, 0, -1)
-        break
-      case '11':
-        metadata.facing = 3
-        metadata.headOffset.set(1, 0, 0)
+    if (bot.majorVersion === '1.13') {
+      const state = bedBlock.stateId - mcData.blocksByStateId[bedBlock.stateId].minStateId
+      const bitMetadata = state.toString(2).padStart(4, '0') // FACING (2 bits) | PART (1 bit) | OCCUPIED (1 bit)
```
wvffle, line 36: The comment here is a bit misleading, it suggests as everything is ORed on two bits.
Either use `frist 2 bits`, `3rd bit` `4th bit` or change separator to `,`

lib/plugins/bed.js
```diff
@@ -10,35 +10,64 @@ function noop (err) {
...
+      metadata.part = !parseInt(bitMetadata[3], 2)
+      metadata.occupied = !parseInt(bitMetadata[2], 2)
```
wvffle, line 38: As those are single characters, couldn't you just compare like `bitMetadata[3] === '0'`?

lib/plugins/bed.js
```diff
@@ -10,35 +10,64 @@ function noop (err) {
...
+
+      switch (bitMetadata.slice(0, 2)) {
+        case '00':
```
wvffle, line 41: shouldn't this be `01`?
You're doing `.padStart(4, '0')` on bitMetadata, so by default you handle `00`

lib/plugins/bed.js
```diff
@@ -10,35 +10,64 @@ function noop (err) {
...
+          metadata.headOffset.set(-1, 0, 0)
+          break
+        case '10':
+          metadata.facing = 2
+          metadata.headOffset.set(0, 0, -1)
+          break
+        case '11':
+          metadata.facing = 3
+          metadata.headOffset.set(1, 0, 0)
+      }
```
wvffle, line 70: Repeated part of code, you can move it down. Out of the if statement

lib/plugins/bed.js
```diff
@@ -112,14 +141,17 @@ function inject (bot) {
...
 
-      const nwMonsterCorner = headPoint.offset(monsterRange[1], -6, monsterRange[2]) // North-West lower corner
-      const seMonsterCorner = headPoint.offset(monsterRange[3], 4, monsterRange[0]) // South-East upper corner
-      for (const key of Object.keys(bot.entities)) {
-        const entity = bot.entities[key]
-        if (entity.kind === 'Hostile mobs') {
-          const entityPos = entity.position.floored()
-          if (entityPos.x <= seMonsterCorner.x && entityPos.x >= nwMonsterCorner.x && entityPos.y <= seMonsterCorner.y && entityPos.y >= nwMonsterCorner.y && entityPos.z <= seMonsterCorner.z && entityPos.z >= nwMonsterCorner.z) {
-            return cb(new Error('there are monsters nearby'))
+      if (!(bot.majorVersion === '1.13' && bot.gameMode === 'creative')) { // If in creative mode the bot should be able to sleep even if there are monster nearby (starting in 1.13)
```
wvffle, line 144: This might be somehow confusing. Simplify this statement.

lib/plugins/bed.js
```diff
@@ -10,35 +10,64 @@ function noop (err) {
 
 function inject (bot) {
   bot.isSleeping = false
+  const mcData = require('minecraft-data')(bot.version)
+
   const beds = new Set(['white_bed', 'orange_bed', 'magenta_bed', 'light_blue_bed', 'yellow_bed', 'lime_bed', 'pink_bed', 'gray_bed',
     'light_gray_bed', 'cyan_bed', 'purple_bed', 'blue_bed', 'brown_bed', 'green_bed', 'red_bed', 'black_bed', 'bed'])
 
+  const hostileMobs = new Set(['blaze', 'creeper', 'drowned', 'endermite', 'enderman', 'evoker', 'giant', 'guardian', 'elder_guardian',
```
wvffle, line 18: Isn't there any way to get the list from the data? I can't imagine updating it with every new minecraft version

test/internalTest.js
```diff
@@ -351,24 +351,30 @@ mineflayer.testedVersions.forEach((supportedVersion, i) => {
...
       bot.once('chunkColumnLoad', (columnPoint) => {
         assert.doesNotThrow(() => { // Should sleep
           bot.sleep(bot.blockAt(bed1Pos))
         })
 
-        assert.throws(() => { // The bed is too far
+        assert.throws(() => {
           bot.sleep(bot.blockAt(bed2Pos))
-        })
+        }, Error('the bed is too far'))
```
wvffle, line 373: You can use just string there

lib/plugins/bed.js
ImHarvol: At first I ussed `entity.kind === 'Hostile mobs`, but it has been removed in 1.13 from minecraft-data, would it be possible to add it again?

wvffle: @rom1504

lib/plugins/bed.js
```diff
@@ -10,35 +10,64 @@ function noop (err) {
 
 function inject (bot) {
   bot.isSleeping = false
+  const mcData = require('minecraft-data')(bot.version)
+
   const beds = new Set(['white_bed', 'orange_bed', 'magenta_bed', 'light_blue_bed', 'yellow_bed', 'lime_bed', 'pink_bed', 'gray_bed',
```
wvffle, line 15: Isn't it the same id just with another damage value?

lib/plugins/bed.js
ImHarvol: Only the `case '11':` is repeated
Below 1.13: 00 is south, 01 west, 10 north and 11 east
In 1.13: 00 is north, 01 south, 10 west and 11 east, that's why I made 2 switches

lib/plugins/bed.js
ImHarvol: This is similar to the [other comment](https://github.com/PrismarineJS/mineflayer/pull/941/files/de5897cef5e558821805fc7db78371b55c8939a6#r410135932) metadata handles by default south (00 below 1.13 and 01 in 1.13), that's why here I added the `case '00':` and removed the `case '01':`

lib/plugins/bed.js
rom1504: No it's not

lib/plugins/bed.js
rom1504: I guess but the only source for that is the minecraft wiki. you can either add manually there or run the extractor that use both wiki.vg and minecraft wiki. (Or add the data to wiki.vg)

lib/plugins/bed.js
wvffle: Okay, then it should be tested if it works correctly in the tests.

ImHarvol:
This PR needs a minecraft-data to be updated with this PrismarineJS/minecraft-data#271 to work
With that change, test passes

rom1504:
Looks like tests are not passing. Can you fix ?

ImHarvol:
Test failed because I was waiting for a new version of minecraft-data with [this changes](https://github.com/PrismarineJS/minecraft-data/commit/b5d6eaf88653eded18308e5d8ea9e3922a1067a4). So now test should pass

rom1504:
so do it ? fix mcdata version to 2.47.0
tests should pass before merging not after

rom1504:
We should definitely have a more generic way to parse metadata but thanks for the feature, merging

PR #942, opened 2020-04-11
Fix error when trying to place a block

Karang:
When trying to place a block after equipping the item without any delay, the check for the item being held gets triggered, but the block placing doesn't get cancelled as it should, resulting in the callback being called twice (first with an error because the item held has not been updated, secondly when the block placement succeed).

I also moved the check in the callback of the lookat action, because the extra delay seems to be giving the time for the item to update.

While this fix is solving the problem in my tests, I think the `bot.equip` function should watch the destination slot and only return successfully when the server acknowledged that the item is in it.

### DIFF
--- a/lib/plugins/inventory.js
+++ b/lib/plugins/inventory.js
@@ -378,8 +378,8 @@ function inject (bot, { version }) {
   }
 
   function placeBlock (referenceBlock, faceVector, cb = noop) {
-    if (!bot.heldItem) cb(new Error('must be holding an item to place a block'))
     bot.lookAt(referenceBlock.position.offset(0.5, 0.5, 0.5), false, () => {
+      if (!bot.heldItem) return cb(new Error('must be holding an item to place a block'))
       // TODO: tell the server that we are sneaking while doing this
       bot.swingArm()
       const pos = referenceBlock.position
--- a/lib/plugins/inventory.js
+++ b/lib/plugins/inventory.js
@@ -378,8 +378,8 @@ function inject (bot, { version }) {
   }
 
   function placeBlock (referenceBlock, faceVector, cb = noop) {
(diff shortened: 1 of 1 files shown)


lib/plugins/inventory.js
```diff
@@ -378,8 +378,8 @@ function inject (bot, { version }) {
   }
 
   function placeBlock (referenceBlock, faceVector, cb = noop) {
-    if (!bot.heldItem) cb(new Error('must be holding an item to place a block'))
```
rom1504, line 381: The return should be in this line

rom1504:
Moving inside look callback is not a good solution to solve the issue of "too soon".
A correct solution would be adding a callback in bot.equip

Karang:
> Moving inside look callback is not a good solution to solve the issue of "too soon".
> A correct solution would be adding a callback in bot.equip

`bot.equip` already have a callback, but it is called too soon. I think the problem come from the way the transactions are confirmed in clickWindow. Do you remember why there was a need for the client to confirm the transaction without waiting for the server ? 

On 1.12.2, dropping the client-side confirmation https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/inventory.js#L530 seems to be working.

Karang:
> On 1.12.2, dropping the client-side confirmation https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/inventory.js#L530 seems to be working.

But it doesn't solve the issue... If I revert the changes of this PR (test before lookat, I get the `Error: must be holding an item to place a block` error.

Karang:
> But it doesn't solve the issue... If I revert the changes of this PR (test before lookat, I get the `Error: must be holding an item to place a block` error.

I found a fix, running the tests and pushing :)

lib/plugins/inventory.js
Karang: done

PR #948, opened 2020-04-13
Fix block_entity example

SiebeDW:
mob_spawner got replaced by spawner

### DIFF
--- a/examples/block_entity.js
+++ b/examples/block_entity.js
@@ -28,13 +28,14 @@ bot.on('message', (cm) => {
 })
 
 function spawner () {
+  let block;
   if (bot.majorVersion === '1.13') {
-    const block = bot.findBlock({
+    block = bot.findBlock({
       matching: mcData.blocksByName.spawner.id,
       point: bot.entity.position
     })
   } else {
-    const block = bot.findBlock({
+    block = bot.findBlock({
       matching: mcData.blocksByName.mob_spawner.id,
       point: bot.entity.position
     })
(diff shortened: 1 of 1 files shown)


examples/block_entity.js
```diff
@@ -28,10 +28,18 @@ bot.on('message', (cm) => {
 })
 
 function spawner () {
-  const block = bot.findBlock({
-    matching: mcData.blocksByName.mob_spawner.id,
-    point: bot.entity.position
-  })
+  if (bot.majorVersion === '1.8' || bot.majorVersion === '1.9' || bot.majorVersion === '1.10' || bot.majorVersion === '1.11' || bot.majorVersion === '1.12') {
```
wvffle, line 31: It's better to check if it's 1.13, less writing for now

examples/block_entity.js
```diff
@@ -28,10 +28,18 @@ bot.on('message', (cm) => {
...
+    const block = bot.findBlock({
+      matching: mcData.blocksByName.mob_spawner.id,
+      point: bot.entity.position
+    })
```
wvffle, line 35: `block` variable is defined in other scope than it's used. Tests will fail due to that

examples/block_entity.js
```diff
@@ -28,10 +28,18 @@ bot.on('message', (cm) => {
...
+  } else {
+    const block = bot.findBlock({
+      matching: mcData.blocksByName.spawner.id,
```
wvffle, line 38: Do something like
```js
const blockName = (bot... === '1.13')
  ? 'spawner'
  : 'mob_spawner'

const block = bot...({
  matching: ...[blockName].id
})
```

PR #950, opened 2020-04-14
use pool for faster chunk processing

rom1504:


rom1504:
Yeah not sure I want to merge this
The consequence of this are : 
* will use other cores
* will use json for communicating with other cores : json serialization/parsing is fast but json is big so it takes a bunch of time to communicates with cores

Tests are failing due to slowness so it's not a great fix.

https://github.com/PrismarineJS/prismarine-chunk/pull/74 might be a better solution

wvffle:
>json is big so it takes a bunch of time

We have 1.13 implementation right? Cannot we just serialize the chunks using the same method they're serialized whilst sent to the client? Provided, there won't be larger impact on performance than before.

For pre 1.13: I didn't really look at the format but maybe it's possible to adjust the implementation slightly?

rom1504:
No that kills the whole point.
The point of this pr was to offload the loading of the chunk to another
core. If we use the normal serialisation to communicate with that core it
means you will parse the chunk in the current core (to retrieve the chunk)

But anyway karang managed to optimize 1.13 chunks so this will most likely
not be necessary.

rom1504:
now not needed thanks to https://github.com/PrismarineJS/prismarine-chunk/pull/74

PR #952, opened 2020-04-15
Support node < 10.0.0

FroggyYT:
Fix error on line 17
Unexpected Token {

### DIFF
--- a/lib/scoreboard.js
+++ b/lib/scoreboard.js
@@ -14,7 +14,7 @@ class ScoreBoard {
   setTitle (title) {
     try {
       this.title = JSON.parse(title).text // version>1.13
-    } catch {
+    } catch(err) {
       this.title = title
     }
   }


wvffle:
This is not an error, this is a syntax that is available from node v10.0.0

rom1504:
Lts is node 12.
Is there something inside npm or some module to check the user node and
tell him to upgrade if < 10 ?

(Basically let's support only > 10)

wvffle:
>Is there something inside npm or some module to check the user node and
tell him to upgrade if < 10 ?

`"engines"` in package.json

rom1504:
Does it do anything ?

wvffle:
https://stackoverflow.com/a/29349773/5565538

So basically when we create .npmrc with
```
engine-strict=true
```
it should throw an error when installing. Should also work with yarn as stated in the comments

PR #954, opened 2020-04-18
fix transfer item count

khang-hoang:
bug reproduce step:
1. put 66 `gold_block` in chest
2. then withdraw 32 `gold_block`
3. then withdraw 34 `gold_block`
4. the chest remain 2 `gold_block`


wvffle:
Can you add tests for that?

rom1504:
fixed in #992 
thanks !

PR #956, opened 2020-04-20
Fix villager assert + failing test

ImHarvol:
In 1.8 villager has a capital letter, so in 1.8 the assert throws when it shouldn't

### DIFF
--- a/test/externalTests/fishing.js
+++ b/test/externalTests/fishing.js
@@ -23,7 +23,7 @@ module.exports = () => (bot, done) => {
     })
   })
 
-  bot.on('playerCollect', (collector, collected) => {
+  bot.once('playerCollect', (collector, collected) => {
     if (collected.type === 'object') {
       bot.test.sayEverywhere('I caught: ' + collected.displayName)
       done()
--- a/lib/plugins/villager.js
+++ b/lib/plugins/villager.js
@@ -8,7 +8,6 @@ function noop (err) {
 }
 
 function inject (bot, { version }) {
-  const mcData = require('minecraft-data')(version)
   const Item = require('prismarine-item')(version)
   const windows = require('prismarine-windows')(version).windows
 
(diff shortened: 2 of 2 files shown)


rom1504:
Tests are failing, can you fix ?

ImHarvol:
> Tests are failing, can you fix ?

I don't think it has anything to do with my PR, It passes all tests on my computer. 
It's the same error that I got in #941 (which also passes all test on my computer)

Anyways I think it just need to change a 'on' to a 'once' for the test not to fail

test/externalTests/fishing.js
```diff
@@ -23,7 +23,7 @@ module.exports = () => (bot, done) => {
     })
   })
 
-  bot.on('playerCollect', (collector, collected) => {
+  bot.once('playerCollect', (collector, collected) => {
```
wvffle, line 26: This is not the correct way of checking this in my opinion. If `collected.type` is not 'object' then the test would fail due to timeout. 

My guess is that you catch two items, hence `playerCollect` called twice. I don't know if it's possible so there may be an option that it needs to be inspected more closely. In such case it deserves it's own issue in my opinion.

ImHarvol: Right, It could be the Ink sac tho the probability is 1.2%, and It happened in both my tests which is suprising
> **1.11**: Ink sac chances have been changed from 10 chances for 1 ink sac to 1 chance for 10 ink sacs.
https://minecraft.gamepedia.com/Fishing

wvffle: I haven't looked closer, but the check of `collected.type` is there for some reason. If you catch it simply once it may still fail and the `collected.type` is not as expected then a timeout will occur and the test fails once again.

rom1504:
this was fixed in 1.14 PR

PR #959, opened 2020-04-22
Improve block placement

Karang:
Makes the bot look at the face it's clicking instead of the center of the block. This looks better especially when the bot is making a bridge (placing block while standing on the reference block).


rom1504:
Good idea, it makes sense

PR #960, opened 2020-04-27
Incorrect window id in apologize packet

JungleDome:
To fix successive window click being rejected due to wrong window id in the apologize packet.
(refer to https://wiki.vg/Protocol#Window_Confirmation_.28serverbound.29)

### DIFF
--- a/lib/plugins/inventory.js
+++ b/lib/plugins/inventory.js
@@ -480,7 +480,7 @@ function inject (bot, { version }) {
 
     function onRejected () {
       bot._client.write('transaction', {
-        windowId: 0,
+        windowId: window.Id,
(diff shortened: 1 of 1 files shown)


lib/plugins/inventory.js
```diff
@@ -480,7 +480,7 @@ function inject (bot, { version }) {
 
     function onRejected () {
       bot._client.write('transaction', {
-        windowId: 0,
+        windowId: window.Id,
```
wvffle, line 483: `window.id`

rom1504:
this PR is incorrect and not sure what bug it fixes, please reopen if you still want it

PR #968, opened 2020-05-04
add 1.14 to supported versions

rom1504:
nothing ready yet, just to see what circle ci says


rom1504:
https://wiki.vg/Protocol_History#1.14.4 will be pretty useful for this PR

rom1504:
plan is :
1. fix tests
2. update stuff from history page
3. check untested examples

rom1504:
changes : 
* [ ] Added Update Light (0x57)
* [ ] Added Entity Sound Effect (0x4E)
* [ ] Replaced the minecraft:open_book plugin channel with a new packet, Open Book (0x2C). The layout is the same as with the channel: it only contains a hand parameter.
* [ ] Replaced the minecraft:trader_list plugin message with a dedicated Trade List (0x5A) packet.
* [ ] Removed the ability to specify number of slots, which cleans up some parts of the code but does make it impossible to create 1/2/4/5 row sized chest windows (MC-142082)
* [ ] Menu IDs are no longer sent as strings, but instead numeric IDs specified by the minecraft:menu registry.
* [ ] Split Open Window (formerly 0x14) into 2 packets: Open Horse Window (0x14) and Open Window (0x59)
* [x] Restructured the Player Block Placement packet, including the addition of a new field indicating whether the player's head is inside of a block (used by scaffolding when holding shift).
* [ ] Removed object IDs; Spawn Object (0x00) now uses normal entity type IDs.
* [ ] Removed the Use Bed packet.
* [ ] Added new fields to Trade List: XP, Special Price, Price Multiplier (in the trades array), villager level, experience, is regular villager (only one for the packet)
* [ ] Removed Difficulty field from Join Game and Respawn

rom1504:
tests currently blocked by 
>  Removed object IDs; Spawn Object (0x00) now uses normal entity type IDs.

Karang:
The creative.js test is failing because the server is sending 2 send_slot packets. I think the logic for setInventorySlot is a bit messy: there is a protection that cancelled the first callback when called twice on the same slot. It seems to work in older versions because the server only sent the set_slot packet for the last `set_creative_slot` request, but in 1.14 each `set_creative_slot` gets an answer (which seems better). The problem with that is the test is expecting the 2nd item to trigger the callback, while in 1.14 the 1st item is triggering the 2nd callback (that replaced the 1st callback). In order to fix that (and make the function a bit more robust), I suggest to change the logic: instead of cancelling the first callback, just prevent the 2nd:
```javascript
  // WARN: This method should not be called twice on the same slot before first callback exceeds
  function setInventorySlot (slot, item, cb) {
    assert(slot >= 0 && slot <= 44)

    if (cb != null) {
      if (creativeSlotsUpdates.has(slot)) {
        /* const oldCb = creativeSlotsUpdates.get(slot)
        oldCb(new Error(`Setting slot ${slot} cancelled due to calling bot.creative.setInventorySlot(${slot}, ...) again`)) */
        cb(new Error(`Setting slot ${slot} cancelled due to calling bot.creative.setInventorySlot(${slot}, ...) again`))
        return
      }

      creativeSlotsUpdates.set(slot, cb)
    }

    bot._client.write('set_creative_slot', {
      slot,
      item: Item.toNotch(item)
    })
  }
```
That would require to change the test but will work in every version. What do you think ?

rom1504:
Sounds reasonable. Feel free to commit on this branch

rom1504:
the fix you made for fishing is just half of a fix, we need to properly handle the new ids of object (which are now the entity ids and not object ids) in entities.js

rom1504:
I update to a mcdata with more data

rom1504:
next we need to go through that list, do all these stuff and it should be ok

rom1504:
We have more tests now, it's great.
But this is a lot of changes, so let's still check the example : 
Most important examples to check : 
* [ ] digger (has auto tests)
* [ ] jumper
* [ ] inventory  (has auto tests)
* [ ] chest : broken

Issue #969, opened 2020-05-04
Move to features pattern in mineflayer, like flying squid


rom1504:
See https://github.com/PrismarineJS/flying-squid/blob/master/src/lib/features.json and https://github.com/PrismarineJS/flying-squid/blob/master/src/lib/supportFeature.js

this is a good pattern

the current if (version) is a bad pattern

rom1504:
I'm working on this
This will make migration to 1.14 much cleaner

rom1504:
done

PR #970, opened 2020-05-04
Uses feature pattern instead of if version

rom1504:
fixes issue #969


rom1504:
some of these feature flags could be merged together / some could be removed, but it's already much better than the hardcoded version check solution
about using intervals instead of hard major version, I guess we could, but I think it adds in complexity without that much of a clear gain

PR #972, opened 2020-05-07
just fix entity_teleport for 1.14

kamesa7:
Please close freely


kamesa7:
Thanks for the checking well and fix.
Next time I'll check the old protocol.

PR #976, opened 2020-05-09
Added translated README.md to Russian language

shketov:
I decided to translate some documentation so that it would be easier for Russian-speaking users to understand Mineflayer


rom1504:
Is that not a duplicate from your other pull request ?
You can make several changes in a single pull request

shketov:
No. This is a translated document README.md into Russian. Speaking about the document that you mean - a switch was added there, with which it will be convenient to switch between languages

PR #977, opened 2020-05-09
Added translated api.md to Russian language

shketov:
I decided to translate some documentation so that it would be easier for Russian-speaking users to understand Mineflayer


rom1504:
Cool idea.
I won't be able to maintain that documentation and for sure cannot ask contributors to update the Russian documentation though.
So it means that people (you for example) may have to regularly update the documentation from the English one so it's correct.

Could you add a warning at the top of this Russian doc to explain that it might not be fully up to date and to refer to the English one for perfectly accurate doc ?

Anyway, thanks for this work !

PR #978, opened 2020-05-09
Updated README.MD

shketov:
Added language switch


rom1504:
Looks like you committed the wrong filename (it should be README.md)

rom1504:
thanks !

rom1504:
previous commit was correct, now it's wrong, you remove the main readme

rom1504:
I added a commit directly to master to add that toggle

PR #980, opened 2020-05-09
initial 1.15 commit

rom1504:


rom1504:
to be done : 
1. all changes up from https://wiki.vg/Protocol_History#1.14.4
* [ ] Removed Metadata field from Spawn Mob and Spawn Player, bringing these packets in line with Spawn Object, Spawn Experience Orb, and Spawn Painting
* [ ] Added array of 1024 biomes as ints to Chunk Data (only sent when full chunk is true), representing biomes in 4x4x4 clumps. Uncertain how the client derives the rest of the biome information.
* [ ] Added enable respawn screen boolean to end of Join Game, set to the opposite of the new doImmediateRespawn gamerule
* [ ] Added hashed seed long field to Join Game (between Dimension and Max Players) and Respawn (between Dimension and Gamemode)
* [ ] Particle X/Y/Z positions are now doubles instead of floats
2. test these examples manually : 
* [ ] chest
* [ ] jumper
* [ ] a few others to make sure

PR #983, opened 2020-05-10
Handle disabled physics

ImHarvol:
If you try to open a chest (for example) with the physics plugin disabled the bot tries to call `bot.lookAt(...` and as the plugin that handles that function is disabled, the bot throws error. I don't think this shold happen

Edit: If you encountered the same problem, an easy _fix_ is to add a bot.look function yourself that calls the callback. Something like this:
```js
  // ...
})
bot.lookAt = (a, b, cb) => {
  cb()
}
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -92,6 +92,7 @@ class Bot extends EventEmitter {
     const self = this
     self._client = mc.createClient(options)
     self.username = self._client.username
+    self.plugins = options.plugins
     self._client.on('session', () => {
       self.username = self._client.username
     })
--- a/lib/plugins/digging.js
+++ b/lib/plugins/digging.js
@@ -10,7 +10,8 @@ function inject (bot) {
   function dig (block, cb) {
     if (bot.targetDigBlock) bot.stopDigging()
     cb = cb || noop
-    bot.lookAt(block.position.offset(0.5, 0.5, 0.5), false, () => {
+
+    function lookCb () {
       bot._client.write('block_dig', {
         status: 0, // start digging
         location: block.position,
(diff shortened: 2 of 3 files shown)


rom1504:
I don't think it makes a lot of sense to do this.
Currently mineflayer is mostly a set of interconnect plugins, it doesn't really work to disable a plugin.
We could make the plugins independent but it would require some redesign work.

Why do you disable physics ?

wvffle:
> I don't think it makes a lot of sense to do this.

At first I thought that those functions should throw but then after some rethinking, disabling physics should not impact other functionalities like `activateBlock()` that do not REALLY need physicss to work

ImHarvol:
> Why do you disable physics ?

I'm working on a bot that trades with one villager, and the bot only has to move to arrive to the trading location. I tried to use mineflayer-navigate, but the bot stucks on the carpets, so I disabled physics and sent the position packets _by hand_ (that worked). Then at the trading location I try to open a chest and surprise! the bot throws due to the bot.lookAt.
Sure there is some way to get past the carpets without disabling the plugin, but I don't really understand how it works, so I decided to just disable it as I also don't need it once the bot arrives at the position.

I agree with what @wvffle said, so if that's preferred I'll modify the PR

rom1504:
So basically you reimplemented your own "one use case only" physics plugin.

Would be so much better to improve the current physics plugin.

I do think that modularising the plugins would be interesting, but that
means a redesign : it means defining properly what eachs plugin depends on,
and recoding the plugins to be as uncoupled as possible.

I would be ok with that.

I'm not very ok to add hacks so that it kind of work by chance and putting
if not disabled everywhere

rom1504:
I looked again, and I don't think this is a good fix, it makes the code more complicated and is just a workaround...
Would be good to solve this properly.

PR #984, opened 2020-05-10
Promisify

### LINKED ISSUE #725: promisify
[Issuehunt badges](https://issuehunt.io/r/PrismarineJS/mineflayer/issues/725)

we can do something like https://nodejs.org/api/fs.html#fs_fs_promises_api to provide promises

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $6.00)

- [rom1504 rom1504](https://issuehunt.io/u/rom1504) ($2.00)
- [imharvol imharvol](https://issuehunt.io/u/imharvol) ($2.00)
- [louis030195 louis030195](https://issuehunt.io/u/louis030195) ($2.00)

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/PrismarineJS/mineflayer/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

wvffle:
I'll start by promisifying entire lib and then adding support for callbacks back.

The tests will fail until the support for callbacks is back, that way we can check if everything works correctly and then rewrite tests to use async/await

Aiming to close #725

### DIFF
--- a/lib/plugins/bed.js
+++ b/lib/plugins/bed.js
@@ -4,10 +4,6 @@ module.exports = inject
 
 const CARDINAL_DIRECTIONS = ['south', 'west', 'north', 'east']
 
-function noop (err) {
-  if (err) throw err
-}
-
 function inject (bot) {
   bot.isSleeping = false
   const mcData = require('minecraft-data')(bot.version)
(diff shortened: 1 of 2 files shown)


wvffle:
As I'm going through the entire code I've decided to make it a little bit more readable and optimize it in some places

rom1504:
Yeah that's one way to go at it.
Another way would be to expose callback+promise for all the external stuff, then convert example and tests to use that.
And only then converting the internal to use promises. An advantage of this approach is it'll be easier to provide the first version that expose promise+callback.

Both approach are ok.

(The end result will be the same)

rom1504:
Maybe it would be wiser to expose callback and promise for every function you convert even with your approach though, that way you can keep the test running with every commit.
Fixing all the tests at the end will be a pain.

wvffle:
Just exposed the callbacks and fixed some failing tests

wvffle:
Key question right now, should we use async functions for everything or just things that have callbacks right now? 

If one would use `await` for every mineflayer function, then it's no problem to have non-async functions.

However, let's say that we have `bot.updateSign(block, text)` now. What if we changed the logic (some time after merging this PR) in a way that it resolves when the server returns a packet? User who doesn't use `await` for everything could face unhandled rejection error if the function failed in its logic.

Above case can happen even if we make every function an async one, of course. It's up to the user whether they use `await` or not but in my opinion if we say that EVERY function is async, then they're more likely to use the `await` keyword.

---

Second thing that comes to my mind right now are functions that return just value or an error with many values in the callback. I can't tell if we have the latter, but speaking of the first we have `bot.tabComplete(str, cb)` which returns just `matches`. This is a problem with current approach as `util.callbackify` expects the last argument to be a `cb(err, value)` style callback. 

Well, **the last argument**. This is how the full function looks like: `bot.tabComplete(str, cb, [assumeCommand], [sendBlockInSight])`. This is also a problem

Of course I could add some ifs or switches here: [index.js#L92](https://github.com/PrismarineJS/mineflayer/blob/wvffle/promisify/index.js#L92) but I'd rather consider changing this when we get on a road to mineflayer@3.0.0

rom1504:
For the first point, could you list all the function that currently do not accept a callback and you think they could in the future. That will make taking the decision easier.

For the second point this is a bug imho. We should be respecting the node convention which is one error and one return value in callback. We should just fix everything that is not like this. I don't believe it is a breaking change in this case as it's not working as intended.
Do you see many such cases ?

About 3.0.0, it will be best if we can avoid it. Incrementing the major version means breaking compatibility, so it means all dependent code need to be updated (that's hundred of repos).
We could consider if only if it's truly necessary.

rom1504:
any blocker in this PR or just didn't have time ?
I think it would be cool to finish this

wvffle:
lack of time, I'll probably continue next week.

If anyone wants to contribute, feel free

wvffle:
Hmmm, there's some inconsistency between bot.consume and bot.fish. Both methods seem to work in the same manner but yet:

https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/inventory.js#L73

rom1504:
Yes it should

rom1504:
I think this would be great to progress on

rom1504:
@wvffle do you think we could do this incrementally?
ie add promise support for some method, and merge them now, and progressively add everything else
I think it'd have better chance to work then this "do everything now" approach

wvffle:
To be honest, it could be a good approach in a way that, we'd probably get it working faster than all in one PR. The main con is that if many contributors will convert it in different PRs some problems (like the ones we've discussed before with bot.dig() for example) may be implemented in different ways

wvffle:
> I think it would be better to split it up as you risk diverging from the master branch in long-lived, unmerged branches. Plus, it would be easier for the community to contribute across multiple PRs anyway instead of a single PR.

Then in #725 we should discuss the common ways to solve some of the problems

wvffle:
to be honest I totally forgot about this PR
answering your question: there will be a lot of conflicts

lib/plugins/bed.js
```diff
@@ -69,91 +69,95 @@ function inject (bot) {
     return metadata
   }
 
-  function wake (cb) {
+  async function wake () {
```
NotSugden, line 72: why does this function need to be async? it doesn't utilize await

lib/plugins/blocks.js
```diff
@@ -19,13 +19,13 @@ function inject (bot, { version }) {
...
   const Chunk = require('prismarine-chunk')(version)
   const ChatMessage = require('../chat_message')(version)
-  let columns = {}
   const signs = {}
   const paintingsByPos = {}
   const paintingsById = {}
-
   const blockEntities = new Map()
 
+  let columns = {}
```
NotSugden, line 27:
```suggestion
  const columns = {}
```
i haven't seen this be re-assigned anywhere, so it can be a const

lib/plugins/blocks.js
```diff
@@ -413,24 +440,26 @@ function inject (bot, { version }) {
   })
 
   bot.updateSign = (block, text) => {
-    const lines = text.split('\n')
+    const lines = [...text.split('\n'), '', '', '']
+      .slice(0, 4)
+      .map(JSON.stringify)
+
     if (lines.length > 4) {
```
NotSugden, line 447: the above change `.slice(0, 4)` prevents the array from having more than 4 elements, this check no longer does anything

lib/plugins/blocks.js
```diff
@@ -413,24 +440,26 @@ function inject (bot, { version }) {
   })
 
   bot.updateSign = (block, text) => {
-    const lines = text.split('\n')
+    const lines = [...text.split('\n'), '', '', '']
+      .slice(0, 4)
+      .map(JSON.stringify)
```
NotSugden, line 445: this would pass extra arguments to JSON.stringify, im not sure what effect it would have so would it be better to `json => JSON.stringify(json)`?

Karang:
you should always use rebase instead of merge.
You can fix it with something like:
```
git reset --hard HEAD~2
git rebase origin/master
git push --force
```

Karang:
Do a reset --hard and rewrite history. Do not revert commits.

Karang:
Yes you can do a `git status` and `git log` to confirm your head is on the right commit and your unwanted changes have been removed before you push.

lib/plugins/blocks.js
wvffle: L479

lib/plugins/blocks.js
wvffle: yeah

wvffle:
Closing in favour of #1475

PR #992, opened 2020-05-11
Fix 2 transfer bugs

### LINKED ISSUE #988: Optimize chest.deposit
It could be, chest.deposit is throwing `(node:16890) UnhandledPromiseRejectionWarning: Error: missing source item 86:null in (54,90)` after a few stacks are deposited
Edit: This error has nothing to do with mineflayer, I was a problem of my code

Right now the bot deposits the item one by one, and if you need to deposit an entire inventory of stacks, that's going to be a hell lol of sent packets, which is not ideal: maybe a server could start rejecting transactions after so many packets?

Instead of deposit items one by one, deposit the entire slot in the chest

I haven't look yet at how the bot currently handles chest.deposit, but I'll take a look

Karang:
This PR fix 2 bugs in the transfer function:
* Closes #988 thanks @ImHarvol for finding the bug
* I also confirm the bug in https://github.com/PrismarineJS/mineflayer/pull/954/files but the fix doesn't require to add a dependency on mcData, so I cleaned-up and included the fix of @khang-hoang

To give more insights:
* first bug is due to the use of a variable declared outside of the scope (`sourceItemCount`)
* second bug is that `destSlot` is an integer and never contains a count, `destItem` is the variable we want to use.

Lastly, in his PR @khang-hoang fixed a potential issue for items that don't have a `stackSize` of 64 (like enderpearls with a `stackSize` of 16), so it is included as well.

### DIFF
--- a/lib/plugins/inventory.js
+++ b/lib/plugins/inventory.js
@@ -261,7 +261,6 @@ function inject (bot, { version }) {
         if (!sourceItem) return cb(new Error(`missing source item ${itemType}:${metadata} in (${sourceStart},${sourceEnd})`))
         if (firstSourceSlot === null) firstSourceSlot = sourceItem.slot
         // number of item that can be moved from that slot
-        var sourceItemCount = sourceItem.count
         clickWindow(sourceItem.slot, 0, 0, (err) => {
           if (err) {
             cb(err)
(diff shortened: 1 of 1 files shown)


rom1504:
Looks like none of our tests catch this.
Do you think it would be doable to add a regression test ?

Karang:
> Looks like none of our tests catch this.
> Do you think it would be doable to add a regression test ?

For the second bug, the wrong amount is transferred, so it should be easy to catch by testing the number of items in chest (modifying the chest.js test to do the procedure described in https://github.com/PrismarineJS/mineflayer/pull/954)
For the first bug it's more tricky because the result is correct, it just does too many actions. Unless we add hooks to count the number of clicks I don't see how to test it.

rom1504:
ok let's just merge it, it looks fine

PR #998, opened 2020-05-15
Mineflayer TypeScript definitions

IdanHo:
I documented all functions listed in the docs. Unfortunately quite a lot of the prismarine-* packages are missing typescript definitions, and as such, i used `object`s as placeholders, any missing type is annotated like this:
`object /*prismarine-package Type*/`
This should not be merged until the missing types are added to the appropriate repositories and the placeholders and replaced.

### Currently missing types:
- [x] prismarine-entity Entity
- [x] prismarine-block Block
- [x] prismarine-biome Biome
- [x] prismarine-windows Window
- [x] prismarine-item Item
- [x] prismarine-recipe Recipe
- [x] vec3 Vec3

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,671 +1,681 @@
 /// <reference types="minecraft-protocol" />
 /// <reference types="node" />
+/// <reference types="vec3" />
+/// <reference types="prismarine-item" />
+/// <reference types="prismarine-windows" />
+/// <reference types="prismarine-recipe" />
+/// <reference types="prismarine-block" />
+/// <reference types="prismarine-entity" />
 
 import {EventEmitter} from 'events';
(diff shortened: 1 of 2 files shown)


rom1504:
all dependencies released, please update

rom1504:
do you know if it's possible to automatically check whether these type definitions match the js code ?

IdanHo:
Pretty sure the only way is to convert mineflayer's tests to ts, but instead of checking for intended behaviour, checking it doesn't crash with a type mismatch error

rom1504:
ok let's merge it
since it's not automatically checked by anything, they may be discrepancies that creep up with actual code in the future, but I guess it's not that of a big deal since it's just IDE help, we can fix it when we notice

PR #1004, opened 2020-05-20
explicitely use last version of protodef and nmp to get compiler speed up

rom1504:
cc @Karang


rom1504:
Seems that is doesn't work. We'll need to fix as this is released

rom1504:
I'll probably release a mineflayer fixing to previous version of node-minecraft-protocol in the mean time

Karang:
> Seems that is doesn't work. We'll need to fix as this is released

Ok I'll have a look

rom1504:
it's fixed, thanks

PR #1009, opened 2020-05-25
fix type definitions

dada513:
Before:

image
image

After:

image

What I changed:

Removed declare module mineflayer which caused the issue
Now everything is exported normally and not in the declare module block
 
image


rom1504:
Thanks

PR #1013, opened 2020-05-30
The Physic Update

### LINKED ISSUE #228: Handle new kinds of bounding boxes

### LINKED ISSUE #258: Random teleportation
Experience with my bot (some with rbot as well) has shown that on connection (most reliable) and at other random times, the bot will teleport downwards a certain amount (Around 15-20 blocks although I don't think this is very reliable) into a cave, lava, the void, etc. Any clue why this might be happening and how we can fix it?

### LINKED ISSUE #288: add previous position in move events
The doc currently says `if you want the previous position, use bot.entity.position.minus(bot.entity.velocity)`
That's not true in case of forced moves. 

In order to add the previous position it's needed to encapsulate the position setting to be able to store the previous position.

### LINKED ISSUE #621: Implement knockback
Hi,

is it normal that my bot doesn't move when he get damage? I tested zombies, a sword with knockback II and a simple punch with the bare hand, but he doesn't move by "himself".

### LINKED ISSUE #671: bot get kicked for flying
hi i am testing my bot on an BungeeCord MC server, which's version is 1.12
and all the other works, the bot receive message, and type login info and got respawn.
but then after login->respawn, the bot got kicked for the server thoughts it was flying

i was wondering if some attribute in physics.js were too optimistic?

### LINKED ISSUE #788: Account not being pushed by water, and taking no knockback
Version tested: 1.8 Spigot, 1,12 Spigot, 1.12 vanilla
Mineflayer version: 2.7.5 
Code:

const mineflayer = require('mineflayer');

var options = {
    host: "",
    port: "",
    username: "",
    password: "",
    version: "",
    verbose: true
};
let fbot = mineflayer.createBot(options);

What happens:
the account will log in but when i try to use water to move it around it wont move, and when i hit the account it wont move, but will fall.

Extra: I am lost and i don't know if its a issue but i think by default this sort of stuff should be enabled if i need to enable anything. i just seem lost and looked in the api, examples, everywhere (even looked at other bots).

This seems like some sort of issue

### LINKED ISSUE #814: Underwater walking bug
## Versions
 - mineflayer: 2.8.1
 - server: vanilla 1.12
 - node: 10.15.3

A bot, when told to jump in water, instead of swimming starts jumping as if it was on ground. This causes the bot to drown in a few seconds.

## Expected behavior
The bot, when in water and the function bot.setControlState('jump', true) is called, should start swimming and float to the top (or at least, walk on water as if it was swimming).

## Additional context
This problem has been there since the beginning and is pretty limiting, since pathfinding with mineflayer-navigate becomes completely useless near the sea and in places with many lakes/rivers, making the bot effectively useless.

### LINKED ISSUE #835: Avoid being detected by cheat detection plugins
Hi :D

I tried to run the bot with an anti-cheat (I was allowed to do that), the problem is that the acted my bot and has banned. The administrators then told me that it was because of the fact that it was not an instance of minecraft

How to make the bot an instance of Minecraft ?

### LINKED ISSUE #896: Jumping sprint
Bot disallow jump controlstate, when navigate.walk(). How to enable sprinting?

### LINKED ISSUE #993: handle the server resending the rotation of the bot
there https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/physics.js#L373

probably needed to check the flags for this, see https://wiki.vg/Protocol#Player_Position_And_Look_.28clientbound.29

Karang:
### The Physic Update

Complete rework of mineflayer's physic plugin to be closer to minecraft. This should fix some physic related issues. It makes the following modifications:
* Vanilla clients send a packet_position, packet_look or packet_position_and_look, depending on what was updated.
* The packet_position is sent every second if no update occured (not the packet_flying)
* Before 1.12, at least one position update must be sent every tick (packet_flying if no move). After 1.12 no need to spam the server.
* Vanilla yaw is not constrained to -PI, +PI. While it is easier to work with constrained yaw in code, anticheat could detect that. We keep track of the last sent yaw and use it to send an unconstrained value decoupled from the api.
* Unify physic updates and send position intervals, so they are executed in a deterministic order (and keep the code simpler to read).
* No need to redefine Math.sign
* Using min and max is a more efficient implementation of math.clamp
* Use flags to determine if position is absolute or relative Closes #993
* Add previous position in move events Closes #288
* entity_velocity set twice for the bot
* Fix the timestep! https://gafferongames.com/post/fix_your_timestep/ also because minecraft physic is computed in ticks and not in seconds
* Heading should not rotate the bot (moving left/right should strafe, not rotate the view)
* Minecraft time unit for physic is tick not seconds: simplify computations as deltaTime is now always 1 tick
* entity_velocity unit is 1/8000 block per server tick Closes #621
* Add a physic tick event emitted just after physic update (useful for pathfinder or other plugins that need to monitor movement)
* Correct slipperiness for slime, ice and blue ice https://www.mcpk.wiki/w/index.php?title=Slipperiness
* Implement collision with block shapes Closes #228 #778
* Blocks id shouldnt be hardcoded but instead taken from mcData
* Implement sprint + jump Closes #896
* Implement Water and lava movement Closes #814
* Multiple soulsand blocks can affect the bot (it cumulates)
* Implement webs slowdown
* add bot.physics.adjustPositionHeight function to apply gravity to a position
* Implement stepHeight
* Implement flowing water interactions Closes #788
* Implement slime block bounce
* Implement ladders and vines

To go with it, the pathfinder plugins need to be updated (I'll update mineflayer-pathfinder, but we need to decide what to do with mineflayer-navigate).
In particular:
* Use physicTick event to monitor movement
* Be aware of block height when setting targets
* Add swim moves
* Add ladder moves

Issues that should be closed, but need checking (close and reopen if needed):
* Random teleportation Closes #258
* Detected by anticheats Closes #835
* Get kicked for flying Closes #671

Sources:
* https://www.mcpk.wiki/wiki/Main_Page
* https://www.mcpk.wiki/wiki/SourceCode
* https://github.com/ddevault/TrueCraft/wiki/Entity-Movement-And-Physics

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -81,6 +81,7 @@ higher level API on top of Mineflayer.
 
  * [navigate](https://github.com/andrewrk/mineflayer-navigate/) - get around
    easily using A* pathfinding. [YouTube Demo](https://www.youtube.com/watch?v=O6lQdmRz8eE)
+ * [pathfinder](https://github.com/Karang/mineflayer-pathfinder) - advanced A* pathfinding with a lot of configurable features
  * [radar](https://github.com/andrewrk/mineflayer-radar/) - web based radar
    interface using canvas and socket.io. [YouTube Demo](https://www.youtube.com/watch?v=FjDmAfcVulQ)
  * [blockfinder](https://github.com/Darthfett/mineflayer-blockFinder) - find blocks in the 3D world
--- a/lib/aabb.js
+++ b/lib/aabb.js
@@ -1,0 +1,107 @@
+class AABB {
+  constructor (x0, y0, z0, x1, y1, z1) {
+    this.minX = x0
+    this.minY = y0
+    this.minZ = z0
+    this.maxX = x1
+    this.maxY = y1
+    this.maxZ = z1
+  }
+
+  clone () {
+    return new AABB(this.minX, this.minY, this.minZ, this.maxX, this.maxY, this.maxZ)
+  }
+
+  floor () {
+    this.minX = Math.floor(this.minX)
+    this.minY = Math.floor(this.minY)
+    this.minZ = Math.floor(this.minZ)
+    this.maxX = Math.floor(this.maxX)
+    this.maxY = Math.floor(this.maxY)
+    this.maxZ = Math.floor(this.maxZ)
+  }
+
+  extend (dx, dy, dz) {
+    if (dx < 0) this.minX += dx
+    else this.maxX += dx
+
+    if (dy < 0) this.minY += dy
+    else this.maxY += dy
+
+    if (dz < 0) this.minZ += dz
+    else this.maxZ += dz
+
+    return this
+  }
+
+  contract (x, y, z) {
+    this.minX += x
+    this.minY += y
+    this.minZ += z
+    this.maxX -= x
+    this.maxY -= y
+    this.maxZ -= z
+    return this
+  }
+
+  expand (x, y, z) {
+    this.minX -= x
+    this.minY -= y
+    this.minZ -= z
+    this.maxX += x
+    this.maxY += y
+    this.maxZ += z
+    return this
+  }
+
+  offset (x, y, z) {
+    this.minX += x
+    this.minY += y
+    this.minZ += z
+    this.maxX += x
+    this.maxY += y
+    this.maxZ += z
+    return this
+  }
+
+  computeOffsetX (other, offsetX) {
+    if (other.maxY > this.minY && other.minY < this.maxY && other.maxZ > this.minZ && other.minZ < this.maxZ) {
+      if (offsetX > 0.0 && other.maxX <= this.minX) {
+        offsetX = Math.min(this.minX - other.maxX, offsetX)
+      } else if (offsetX < 0.0 && other.minX >= this.maxX) {
+        offsetX = Math.max(this.maxX - other.minX, offsetX)
+      }
+    }
+    return offsetX
+  }
+
+  computeOffsetY (other, offsetY) {
+    if (other.maxX > this.minX && other.minX < this.maxX && other.maxZ > this.minZ && other.minZ < this.maxZ) {
+      if (offsetY > 0.0 && other.maxY <= this.minY) {
+        offsetY = Math.min(this.minY - other.maxY, offsetY)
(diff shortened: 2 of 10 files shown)


rom1504:
Looks promising !

What do you think about testing it (I mean manually) on a few servers with anticheat to see if it's better than before ?

Karang:
> Looks promising !
> 
> What do you think about testing it (I mean manually) on a few servers with anticheat to see if it's better than before ?

I'm testing the code by simultaneously modifying mineflayer-pathfinder so I can set the bot in follow mode and test its capabilities through some challenges. I'm testing on a vanilla 1.13 server, but I also have a paper server with nocheatplus installed (I havent tested yet).

So far, it seems to have improved the similarity to vanilla (I compare with another account running the vanilla client). I also looked at the move packets with a proxy and it also behave more like a vanilla client (it stop spamming the server with move packet, which I think could have triggered more than one anticheat).

I can release an experimental branch of mineflayer-pathfinder if you want to test the new physics. It now handle any block shape properly, but the navigation algorithm needs more work. Basically, the minecraft physic makes every block a bit slippery (which was not simulated properly before), so the pathfinder should anticipate a bit because it cannot stop in 1 tick like before (it's all work in progress).

Simulating the mc physic properly is a necessity to implement more movements possibility (ladders, swiming, sprint jump, etc..) but it comes with its own difficulties :)

IdanHo:
just wanted to pipe in and say this looks very similar to my try at recreating mc physics more accurately in mineflayer (just better made), so i went through your constants and confirmed all of them indeed match mine (which i extracted directly from the source of 1.12).

rom1504:
I also think it would be good to have some new tests for the new stateless functions (for example stuff in aabb but it could also apply to lib/physics.js)
this can be done later though

Karang:
> I also think it would be good to have some new tests for the new stateless functions (for example stuff in aabb but it could also apply to lib/physics.js)
> this can be done later though

I think this should be done when we move lib/physics.js to prismarine-physics and have the new tests there

Vap0r1ze:
🙌

PR #1014, opened 2020-05-30
Sleeper didn't work for me without mineflayer-navigate

uAliFurkanY:
I don't know if this is related to navigatePlugin but it works with this code.

also this is most likely a small bug

### DIFF
--- a/examples/sleeper.js
+++ b/examples/sleeper.js
@@ -7,6 +7,7 @@
  * You can ask the bot to sleep or wake up by sending a chat message.
  */
 const mineflayer = require('mineflayer')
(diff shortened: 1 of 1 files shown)


rom1504:
This doesn't make sense.
Can you explain what happens when you tried to use the normal sleeper example on a vanilla server ?

uAliFurkanY:
I did not use a vanilla server but a paper server, 0b0t.
When I tried the code from the example, it didn't sleep but when I added `mineflayer-navigate` for an unrelated purpose, it started working.

uAliFurkanY:
I don't know, I can't test it right now, I'll just close this.

PR #1019, opened 2020-06-01
Create node.js.yml

rom1504:


rom1504:
impressive... faster than circle ci at first try
and worked at first try.

PR #1021, opened 2020-06-02
Add auto totem example

AlexProgrammerDE:
Something simple, but usefull.

### DIFF
--- a/examples/auto_totem.js
+++ b/examples/auto_totem.js
@@ -1,0 +1,24 @@
+/*
+ * This script will automaticly set if a totem is in the inventory the totem in the off-hand.
+ * It checks for a totem every secound.
+ */
+const mineflayer = require('mineflayer')
+
+if (process.argv.length < 4 || process.argv.length > 6) {
+  console.log('Usage : node echo.js <host> <port> [<name>] [<password>]')
+  process.exit(1)
+}
+
+const bot = mineflayer.createBot({
+  host: process.argv[2],
+  port: parseInt(process.argv[3]),
+  username: process.argv[4] ? process.argv[4] : 'totem',
+  password: process.argv[5]
+})
+
+let helpbc = setInterval(() => {  
+  var totem = bot.inventory.findInventoryItem(449, null);
+  if (totem && totem.type == 449) {
+  	bot.equip(totem, "off-hand");
+  }
+}, 1000)


examples/auto_totem.js
```diff
@@ -0,0 +1,24 @@
...
+
+const bot = mineflayer.createBot({
+  host: process.argv[2],
+  port: parseInt(process.argv[3]),
+  username: process.argv[4] ? process.argv[4] : 'totem',
+  password: process.argv[5]
+})
+
+setInterval(() => {
+  var totem = bot.inventory.findInventoryItem(449, null)
```
Karang, line 20: You shouldn't use the numeric id, because it changes from one version to an other. Instead, you can get the id from mcData like this:
```javascript
bot.on('spawn', () => {
  const mcData = require('minecraft-data')(bot.version) // You know the version when the bot has spawned

  if (mcData.itemsByName.totem_of_undying) { // Some versions don't have the totem
    setInterval(() => {
      const totemId = mcData.itemsByName.totem_of_undying.id // Get the correct id
      const totem = bot.inventory.findInventoryItem(totemId, null)
      if (totem) {
        bot.equip(totem, 'off-hand')
      }
    }, 1000)
  }
})
```

examples/auto_totem.js
```diff
@@ -0,0 +1,24 @@
...
+  if (totem && totem.type === 449) {
+    bot.equip(totem, 'off-hand')
+  }
+}, 1000)
```
Karang, line 24: why not checking every tick ? (50 ms)

examples/auto_totem.js
```diff
@@ -0,0 +1,24 @@
...
+const bot = mineflayer.createBot({
+  host: process.argv[2],
+  port: parseInt(process.argv[3]),
+  username: process.argv[4] ? process.argv[4] : 'totem',
+  password: process.argv[5]
+})
+
+setInterval(() => {
+  var totem = bot.inventory.findInventoryItem(449, null)
+  if (totem && totem.type === 449) {
```
Karang, line 21: no need to check again the type of the object

AlexProgrammerDE: @Karang The player might have no totem in his inventory. That would return null. If i try to read the property type of a undefined object if would throw an error.

examples/auto_totem.js
AlexProgrammerDE: @Karang When does the Bot get the inventory? I think it wouldnt be a good idea to search for a totem if the bot still has no inventory recieved.

AlexProgrammerDE: I think i will use spawn

rom1504:
Can you explain what's the objective of this example ? What is a totem ?

Karang:
@AlexProgrammerDE use `npm run fix` to automatically make the linter happy :)

AlexProgrammerDE:
@rom1504 you know not what a totem of undying is? It basicaly safes your from dying when its in your main or off hand. Bots often die. That can prevent it. https://minecraft.gamepedia.com/Totem_of_Undying

AlexProgrammerDE:
@Karang wow im dumb.

Karang:
> Can you explain what's the objective of this example ? What is a totem ?

A totem of undying is an item that was introduced in 1.9 https://minecraft.gamepedia.com/Totem_of_Undying

When you hold it in your off-hand it will save you from dying (consuming the item). Auto-totem is a mod you'll find in most cheat clients that will replace a totem in your hand each time one is consumed (as long as you have some in your inventory).

PR #1028, opened 2020-06-03
fix CI badge

IdanHo:
it currently points to a random python deep learning project instead of mineflayer


rom1504:
ah yeah my bad, wrong copy pasting

PR #1035, opened 2020-06-06
Fast findBlocks

### LINKED ISSUE #813: Merge mineflayer-blockfinder into mineflayer
`bot.findBlocks` is one of the most commonly used functions. Current implementation is slow and shouldn't be used.

I'd rather make minefalyer-blockfinder a part of mineflayer. It's one of the most used plugins so I'd consider merging it. 

Alternatively we can remove `bot.findBlocks` from mineflayer if it's too high level function

## Additional context
None ;)

Karang:
Finding blocks may seems a trivial task, but finding them in a reasonable amount of time may be challenging. Mineflayer expose a bot.findBlock function that return the first matched block in a cube arround the player: this will be our baseline. We also need some benchmark to compare the baseline to new solutions, with the player located on the surface we search for:

* find diamond_ore (maxDistance = 128, minCount = 1) : diamond ores are usually near bedrock (y<13)
* find crafting_table (maxDistance = 128, minCount = 1) : we placed a crafting table just in front of the player
* find diamond_block (maxDistance = 128, minCount = 1) : there is no diamond block arround the player (worst case scenario)

So, let's see how the baseline behave:

* find diamond_ore: 35.59 ms (less than a tick, its very good. this is because the algorithm start searching at y=0 and go up, so it is faster at finding blocks near bedrock)
* find crafting_table: 2572.27 ms (2.5 seconds! that's a lot :( the crafting table was just in front of the player, but we checked all blocks at the bottom of the world first)
* find diamond_block: 5290.04 ms (very bad. we just tested every block in a 128x128x128 cube and found nothing)

A third party plugin was created to improve the search algorithm: https://github.com/Darthfett/mineflayer-blockfinder it has gone through a lot of iterations and came up with a clever algorithm to return in priority the blocks closer to the player. The algorithm is based on an Octahedron iterator (3D diamond shape) that goes layer by layer arround the player. Let's see how it behaves on the benchmark:

* find diamond_ore: 196.79 ms (slower than the baseline, but 4 ticks is acceptable for a block that is far from the player)
* find crafting_table: 0.25 ms (very good, it finds blocks near the player almost instantly)
* find diamond_block: 1036.72 ms (worst case of 1s, that's still a bit slow, but 5x faster than the baseline)

The octahedron approach is very good at finding blocks near the player. In fact, it guarantees that the returned block will always be the closest one. But when the block is far or not present, we sill go through a lot of blocks.

Can we do better ?

To do better, we need to test less blocks. In minecraft, the world is divided in columns (16x256x16 blocks) and these columns are themselves divided in sections of (16x16x16 = 4096 blocks). If a section is `null`, we know that all the blocks in it are blocks of air, so we don't need to check them (unless you are searching for air, but why would you ?). In this case we can skip the whole section. Let's modify the octahedron iterator to iterate on sections instead of blocks and search in the section using a simple cube loop but only if the section is not `null`:

* find diamond_ore: 153.52 ms (a bit faster)
* find crafting_table: 8.8 ms (slightly slower, but still very fast)
* find diamond_block: 666.8 ms (x1.5 faster)

This allowed to improve the worst case, but empty sections are not that common (unless you are in the sky).

The last modification takes advantage of the palette system which was introduced in 1.13. Each section contains a palette that map the block id local to the section to the global stateId. This palette is usually quite small (<20 ids) and allow to test if a certain block is present in the section. We modify our previous version to first test the palette:

* find diamond_ore: 5.80 ms (x6 faster than baseline)
* find crafting_table: 1.82 ms (x1413 faster than baseline)
* find diamond_block: 7.27 ms (x727 faster than baseline)

This allows to speed the search significantly.

This MR adds the faster version. Closes #813

I also added 2 other iterators (Manathan and Raycast that I think might be useful for mineflayer)

### DIFF
--- a/lib/iterators.js
+++ b/lib/iterators.js
@@ -1,0 +1,162 @@
+const Vec3 = require('vec3').Vec3
+
+// 2D spiral iterator, useful to iterate on
+// columns that are centered on bot position
+class ManathanIterator {
+  constructor (x, y, maxDistance) {
+    this.maxDistance = maxDistance
+    this.startx = x
+    this.starty = y
+    this.x = 2
+    this.y = -1
+    this.layer = 1
+    this.leg = -1
+  }
+
+  next () {
+    if (this.leg === -1) {
+      // use -1 as the center
+      this.leg = 0
+      return { x: this.startx, y: this.starty }
+    } else if (this.leg === 0) {
+      if (this.maxDistance === 1) return null
+      this.x--
+      this.y++
+      if (this.x === 0) this.leg = 1
+    } else if (this.leg === 1) {
+      this.x--
+      this.y--
+      if (this.y === 0) this.leg = 2
+    } else if (this.leg === 2) {
+      this.x++
+      this.y--
+      if (this.x === 0) this.leg = 3
+    } else if (this.leg === 3) {
+      this.x++
+      this.y++
+      if (this.y === 0) {
+        this.x++
+        this.leg = 0
+        this.layer++
+        if (this.layer === this.maxDistance) {
+          return null
+        }
+      }
+    }
+    return new Vec3(this.startx + this.x, 0, this.starty + this.y)
+  }
+}
+
+class OctahedronIterator {
+  constructor (start, maxDistance) {
+    this.start = start.floored()
+    this.maxDistance = maxDistance
+    this.apothem = 1
+    this.x = -1
+    this.y = -1
+    this.z = -1
+    this.L = this.apothem
+    this.R = this.L + 1
+  }
+
+  next () {
+    if (this.apothem > this.maxDistance) return null
+    this.R -= 1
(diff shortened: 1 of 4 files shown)


rom1504:
would be good to add some tests (for example by making an external test for the example with some /setblock of diamonds before), but let's do that later, merging

Thanks for adding this!

lib/iterators.js
```diff
@@ -0,0 +1,162 @@
+const Vec3 = require('vec3').Vec3
+
+// 2D spiral iterator, useful to iterate on
+// columns that are centered on bot position
+class ManathanIterator {
```
Darthfett, line 5: Spellcheck - should be "Manhattan"?

Karang: oups, yes it should

PR #1036, opened 2020-06-07
Add minetelegram

hexatester:
Minetelegram can be used as standalone client from cli, or imported as plugin.


Karang:
Is it a project or a plugin ? Why add it twice ?

hexatester:
It's a project, but also can be used as plugin.

PR #1051, opened 2020-06-14
Restrict found blocks to be within maxDistance

### LINKED ISSUE #1048: findBlocks is not working properly
## Versions
 - mineflayer: 2.19.0
 - server: vanilla 1.15.2
 - node: 13.9.0

findBlocks method does not return the nearest blocks

## Current code
```js
const blocks = bot.findBlocks({
	point: bot.entity.position,
	matching: 9, // dirt
	maxDistance: 2,
	minCount: 2
})
				
console.log(bot.entity.position)
// Vec3 { x: 50.12733864306501, y: 4, z: -35.27180865196587 }

console.log(blocks.length)
// 512
// Why so much?

console.log(bot.blockAt(blocks[0]))
// Block {
//    type: 9,
//    position: Vec3 { x: 48, y: 1, z: -33 }, // why so far?
//    name: 'dirt',
//    // another block data...
// }
```

## Expected behavior
I want to find the nearest blocks in radius 2. But this method returns me 512 blocks that are far from the coordinates I need

Karang:
Closes #1048

### DIFF
--- a/lib/plugins/blocks.js
+++ b/lib/plugins/blocks.js
@@ -135,13 +135,12 @@ function inject (bot, { version }) {
 
   bot.findBlocks = (options) => {
     const matcher = getMatchingFunction(options.matching)
-    const point = options.point || bot.entity.position
-    let start = point.floored()
-    const maxDistance = Math.ceil((options.maxDistance || 16) / 16)
+    const point = (options.point || bot.entity.position).floored()
+    const maxDistance = options.maxDistance || 16
     const minCount = options.minCount || 1
     const blocks = []
-    start = new Vec3(Math.floor(start.x / 16), Math.floor(start.y / 16), Math.floor(start.z / 16))
-    const it = new OctahedronIterator(start, maxDistance)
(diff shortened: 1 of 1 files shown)


lib/plugins/blocks.js
```diff
@@ -135,13 +135,12 @@ function inject (bot, { version }) {
 
   bot.findBlocks = (options) => {
     const matcher = getMatchingFunction(options.matching)
-    const point = options.point || bot.entity.position
-    let start = point.floored()
-    const maxDistance = Math.ceil((options.maxDistance || 16) / 16)
+    const point = (options.point || bot.entity.position).floored()
```
rom1504, line 138: Are you sure about this part ? It seems to be undoing the previous PR

Karang: yeah I'm sure, it's to correct a small bias in the distance. The previous pr was to use the point instead of the section in which the point was. This fix floor the point to avoid biasing toward one side of the block.

PR #1068, opened 2020-06-17
Emit chestLidMove only once

### LINKED ISSUE #1064: "chestLidMove" is triggered twice on a double chest
When catching the "chestLidMove" event, the handler trigger twice if the chest is double.

## Current code
```js
bot.on("chestLidMove", () => {
  console.log("Chest updated");
})
```

## Expected behavior
The handler is triggered once, regardless of the type of chest.

ImHarvol:
Even thought the vanilla server emits the event twice when it's a double chest, I think it's kind of confusing for a mineflayer user, so I think this change makes sense

Not sure if this is the best approach, but it does the job!

Closes #1064

### DIFF
--- a/lib/plugins/block_actions.js
+++ b/lib/plugins/block_actions.js
@@ -2,19 +2,77 @@ const Vec3 = require('vec3').Vec3
 
 module.exports = inject
 
+const CARDINALS = {
+  north: new Vec3(0, 0, -1),
+  south: new Vec3(0, 0, 1),
+  west: new Vec3(-1, 0, 0),
+  east: new Vec3(1, 0, 0)
+}
+
 function inject (bot, { version }) {
   const { instruments, blocks } = require('minecraft-data')(version)
+
+  function parseChestMetadata (chestBlock) {
+    const chestTypes = ['single', 'right', 'left']
+    let data
+
+    if (bot.supportFeature('doesntHaveChestType')) {
+      data = {
+        facing: Object.keys(CARDINALS)[chestBlock.metadata - 2]
+      }
+    } else {
+      data = {
+        waterlogged: !(chestBlock.metadata & 1),
+        type: chestTypes[(chestBlock.metadata >> 1) % 3],
+        facing: Object.keys(CARDINALS)[Math.floor(chestBlock.metadata / 6)]
+      }
+    }
+    return data
+  }
+
+  function getChestType (chestBlock) { // Returns 'single', 'right' or 'left'
+    if (bot.supportFeature('doesntHaveChestType')) {
+      const facing = parseChestMetadata(chestBlock).facing
+
+      // We have to check if the adjacent blocks in the perpendicular cardinals are the same type
+      if (facing === 'north' || facing === 'south') {
+        if (bot.blockAt(chestBlock.position.plus(CARDINALS.west)).name === chestBlock.name) {
+          return facing === 'north' ? 'left' : 'right'
+        } else if (bot.blockAt(chestBlock.position.plus(CARDINALS.east)).name === chestBlock.name) {
+          return facing === 'north' ? 'right' : 'left'
+        }
+      } else if (facing === 'west' || facing === 'east') {
+        if (bot.blockAt(chestBlock.position.plus(CARDINALS.north)).name === chestBlock.name) {
+          return facing === 'west' ? 'right' : 'left'
+        } else if (bot.blockAt(chestBlock.position.plus(CARDINALS.south)).name === chestBlock.name) {
+          return facing === 'west' ? 'left' : 'right'
+        }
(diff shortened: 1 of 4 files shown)


lib/plugins/block_actions.js
```diff
@@ -2,19 +2,77 @@ const Vec3 = require('vec3').Vec3
...
+      data = {
+        facing: Object.keys(CARDINALS)[chestBlock.metadata - 2]
+      }
+    } else {
+      data = {
+        waterlogged: !(chestBlock.metadata & 1),
+        type: chestTypes[(chestBlock.metadata >> 1) % 3],
+        facing: Object.keys(CARDINALS)[Math.floor(chestBlock.metadata / 6)]
+      }
+    }
```
wvffle, line 29:
```js
return bot.supportFeature('doesntHaveChestType')
  ? { ... }
  : { ... }
```

lib/plugins/block_actions.js
```diff
@@ -2,19 +2,77 @@ const Vec3 = require('vec3').Vec3
...
+    return data
+  }
+
+  function getChestType (chestBlock) { // Returns 'single', 'right' or 'left'
+    if (bot.supportFeature('doesntHaveChestType')) {
+      const facing = parseChestMetadata(chestBlock).facing
+
+      // We have to check if the adjacent blocks in the perpendicular cardinals are the same type
+      if (facing === 'north' || facing === 'south') {
+        if (bot.blockAt(chestBlock.position.plus(CARDINALS.west)).name === chestBlock.name) {
```
Karang, line 39: Can this test be type/metadata based instead of string based ?

lib/plugins/block_actions.js
```diff
@@ -2,19 +2,77 @@ const Vec3 = require('vec3').Vec3
...
+          return facing === 'north' ? 'right' : 'left'
+        }
+      } else if (facing === 'west' || facing === 'east') {
+        if (bot.blockAt(chestBlock.position.plus(CARDINALS.north)).name === chestBlock.name) {
+          return facing === 'west' ? 'right' : 'left'
+        } else if (bot.blockAt(chestBlock.position.plus(CARDINALS.south)).name === chestBlock.name) {
+          return facing === 'west' ? 'left' : 'right'
+        }
+      }
+
```
wvffle: Too over complicated. A double chest cannot be next to another chest.

Provided the above assumption, the facing won't be on the same axis as the chest so we can make a `facingMap` like so:
```js
const facingMap = {
  north: { east: 'left', west: 'right' },
  south: { east: 'right', west: 'left' },
  ...
}

for (const [cardinal, vec] of Object.entries(CARDINALS)) {
  if (bot.blockAt(chestBlock.position.plus(vec)).name === chestBlock.name) {
    return facingMap[cardinal][facing]
  }
}
```

lib/plugins/block_actions.js
ImHarvol: Right, that would be better :+1:

lib/plugins/block_actions.js
```diff
@@ -2,19 +2,72 @@ const Vec3 = require('vec3').Vec3
...
+        west: { north: 'left', south: 'right' },
+        east: { north: 'right', south: 'left' }
+      }
+
+      const facing = parseChestMetadata(chestBlock).facing
+
+      // We have to check if the adjacent blocks in the perpendicular cardinals are the same type
+      const perpendicularCardinals = Object.keys(facingMap[facing])
+      for (const cardinal of perpendicularCardinals) {
+        const cardinalOffset = CARDINALS[cardinal]
```
wvffle, line 41: This can be compressed to 
```js
for (const [cardinal, cardinalOffset] of Object.entries(CARDINALS)) {
```

ImHarvol: I changed it a bit so it only checks left and right and not the front and back too

wvffle: Hmmm.. You're right

test/externalTests/chestEvents.js
```diff
@@ -0,0 +1,34 @@
...
+    bot.test.placeBlock(chestSlot, chestLocation1, (err) => {
+      assert.ifError(err)
+      bot.test.placeBlock(chestSlot, chestLocation2, (err) => {
+        assert.ifError(err)
+
+        bot.openChest(bot.blockAt(chestLocation2))
+
+        bot.on('chestLidMove', (block, isOpen) => {
+          assert.strictEqual(isOpen, 1)
+          done() // If 'chestLidMove' is emitted twice, done() would be called twice and thus failing tests
```
wvffle, line 29:
```js
// NOTE: If ...
done()
```

rom1504:
this is failing

PR #1072, opened 2020-06-18
Added utilty getEquppedItem() function to simple_inventory.js

TheDudeFromCI:
It's often really useful to know what the bot is currently wearing.

Also Added QUICK_BAR_START constraint from the simple_inventory.js to the TypeScript header.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -154,6 +154,10 @@ export class Bot extends EventEmitter {
     callback?: () => void
   ): void;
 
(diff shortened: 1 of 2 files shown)


rom1504:
There's already
https://github.com/PrismarineJS/prismarine-entity#entityhelditem for this.
I think we need a way to include subpackages docs into the main doc so it's
easier to find this kind of thing

PR #1073, opened 2020-06-18
Remove double installation of require-self

wvffle:
As stated in [usage of require-self](https://www.npmjs.com/package/require-self#usage) it's not required to install it in `prepare` script anymore. Dev dependencies are enough.

Current approach breaks test env if modules are installed using yarn and generates both `package-lock.json` and `yarn.lock` which is highly not recommended.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,10 +6,10 @@
   "types": "index.d.ts",
   "scripts": {
     "test": "mocha --reporter spec --exit",
-    "pretest": "require-self",
+    "pretest": "npm run lint && require-self",
     "lint": "standard",
     "fix": "standard --fix",
-    "prepare": "npm install require-self && require-self",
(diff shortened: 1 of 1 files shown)


wvffle:
Good point, I removed linting instead of npm install by accident

wvffle:
Does it now? Or did it break in the past? It's already in devDependencies so it shouldn't matter unless you want to edit a single line without installing dependencies in a fresh clone.

npm probably installs dependencies again after yarn. Everything works fine unless you want to run tests. In such case it says that vec3 is missing. Which was weird for me the frist time I have seen it.

Well, using multiple package managers is a bad practice but in my setup I had only node and yarn installed, without npm. I had to install npm just to install dependencies of mineflayer which is not good in my opinion.

PR #1074, opened 2020-06-18
Converted block finder to use square distance.

TheDudeFromCI:
A standard distance function uses a square root operation which can be slow when called in extremely high quantities. By swapping this out by a distance squared comparison, several orders of performance can be gained. As there is no distance squared function within the vec3 library, (at the time of writing) I inlined this. The syntax can be cleaned up back to a single line if this function is added to the library.

This patch should yield the exact same behavior as the original implementation but without the need for a costly square root operation for each block.

This PR requires https://github.com/andrewrk/node-vec3/pull/7 to work and should be merged after vec3 is updated to the newest version.

### DIFF
--- a/lib/plugins/blocks.js
+++ b/lib/plugins/blocks.js
@@ -137,6 +137,7 @@ function inject (bot, { version }) {
     const matcher = getMatchingFunction(options.matching)
     const point = (options.point || bot.entity.position).floored()
     const maxDistance = options.maxDistance || 16
+    const maxDistanceSq = maxDistance * maxDistance
     const minCount = options.minCount || 1
     const useExtraInfo = options.useExtraInfo || false
     const blocks = []
@@ -157,7 +158,15 @@ function inject (bot, { version }) {
             for (cursor.y = next.y * 16; cursor.y < next.y * 16 + 16; cursor.y++) {
               for (cursor.z = next.z * 16; cursor.z < next.z * 16 + 16; cursor.z++) {
                 const found = bot.blockAt(cursor, useExtraInfo)
-                if (matcher(found) && cursor.distanceTo(point) <= maxDistance) blocks.push(cursor.clone())
+                if (matcher(found)) {
+                  // Calculate distance squared
+                  let dx = cursor.x - point.x
+                  let dy = cursor.y - point.y
+                  let dz = cursor.z - point.z
+                  let distSq = dx * dx + dy * dy + dz * dz
+                  if (distSq <= maxDistanceSq)
+                    blocks.push(cursor.clone())
+                }
               }
             }
           }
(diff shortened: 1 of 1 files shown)


lib/plugins/blocks.js
```diff
@@ -137,6 +137,7 @@ function inject (bot, { version }) {
     const matcher = getMatchingFunction(options.matching)
     const point = (options.point || bot.entity.position).floored()
     const maxDistance = options.maxDistance || 16
+    const maxDistanceSq = maxDistance * maxDistance
```
wvffle, line 140:
```js
     const maxDistanceSq = maxDistance ** 2
```

rom1504:
Can you run the block finder example and check whether it's actually faster
?
(I agree it makes sense to used squaered distance but wondering how much of
a difference it makes, sqrt is quite optimized)

lib/plugins/blocks.js
TheDudeFromCI: Done

Karang:
Can you do the comparison on the test cases described here: https://github.com/PrismarineJS/mineflayer/pull/1035

The findBlocks performance highly depends on the use case, and was optimized to be good accross different scenariis.

While I don't think using squared distance would be harmfull, I am more concerned about caching which can introduce memory overheads (when searching for large amounts of blocks).

A side note on sqrt: even in non interpreted languages, sqrt is not as slow as it used to be, there is even dedicated instructions for it on modern cpus: https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=sqrt_ss&expand=5389

When optimizing, you should always start by measuring the bottlenecks to decide whether it is worse it to do the optimization. Also put your measurements in relation with your application: a minecraft tick is 50ms, a difference of 0.02ms or even 0.5ms will not change anything compared to other functions that takes 100ms or 5000ms to process.

wvffle:
If it comes to benchmarking it's not quite that simple. You're measuring now an average of not only the function but also `i < operations`, `i++`, initial `let i = 0` and GC that happens at the end of every single iteration.

You cannot assume that it'll take the same time every time. What if some other program borrows a lot of cpu in the meantime? Either GC or other measured stuff can take more time than in another test run.

My point is that your tests aren't wrong, squared distance is surely faster to calculate than a normal one. They just show false values.


On the other hand, let's do quick math. 
```
t - measured time
i - number of iterations
t/i - time per operation

0.91/10000 = 0.0091ms/op
0.89/10000 = 0.0089ms/op
```

How much is that? If I flash you a light for 13ms (less than 1 frame of 60fps) and 10ms you probably couldn't tell the difference. So what's the point of optimizing this even further?

wvffle:
Don't get me wrong, optimizations are sincerely welcomed, it's just my point of view about optimizing such small units of time by such small factors. (I'm not counting caching as Karang already said about memory problem)

PR #1083, opened 2020-06-23
Improve digTime computation

### LINKED ISSUE #1067: Fast digging
## Versions
 - mineflayer: 2.20.0
 - server: vanilla 1.15.2
 - node: 13.9.0

## Detailed description of a question
Is there any way to make the bot dig blocks as quickly as possible?

I wrote this code:
(I also use a [pathfinder](https://github.com/Karang/mineflayer-pathfinder) plugin)
```js
const currentBlock = mcData.blocksByName.gold_block

const dig = () => {
	const block = bot.findBlock({
       		matching: currentBlock.id,
		maxDistance: 1
	})

	if (!block) {
		return bot.chat('Не могу найти этот блок поблизости')
	}

	bot.pathfinder.setMovements(defaultMove)
	bot.pathfinder.setGoal(new GoalNear(block.position.x, block.position.y, block.position.z, 1))

	bot.once('goal_reached', () => {
		bot.chat('Копаю')
      		bot.dig(block, () => {
			bot.chat('Выкопал')
			dig()
		})
	})
}
```

The problem is that he slowly digs block by block. I want to achieve the effect of digging as when holding the left mouse button. I tried other ways to find the blocks that need to be dug, but this is not the problem and it still digs slowly

Karang:
Needs https://github.com/PrismarineJS/prismarine-block/pull/11
Closes #1067


lib/plugins/digging.js
```diff
@@ -10,7 +13,7 @@ function inject (bot) {
   function dig (block, cb) {
     if (bot.targetDigBlock) bot.stopDigging()
     cb = cb || noop
-    bot.lookAt(block.position.offset(0.5, 0.5, 0.5), false, () => {
+    bot.lookAt(block.position.offset(0.5, 0.5, 0.5), true, () => {
```
rom1504, line 16: Doing instant looking might get people kicked in some servers, could you add an option to dig (default to false) to do "slow" looking ?
(You can have default parameters by doing some tricks with cb, there are other examples of that in the code)

lib/plugins/digging.js
```diff
@@ -7,10 +10,9 @@ function inject (bot) {
   bot.targetDigBlock = null
   bot.lastDigTime = null
 
-  function dig (block, cb) {
+  function dig (block, cb = noop, forceLook = false) {
```
rom1504, line 13: cb should still be the last parameters, this is a node convention

rom1504: I think making force look true by default is ok

rom1504:
pblock 1.6.0 is available

rom1504:
bit hacky but first solution of https://stackoverflow.com/a/33805148/1658314 works for having a default value

rom1504:
(this will be better when we switch to promises)

PR #1099, opened 2020-06-29
Fix 1.14 raining detection

Leftier:


Leftier:
1.14 is the only version from 1.7 to 1.16 that has raining state as 1 begin, 2 end

test/externalTests/rain.js
```diff
@@ -0,0 +1,13 @@
+const assert = require('assert')
+
+module.exports = () => (bot, done) => {
+  bot.test.sayEverywhere('/weather rain')
```
rom1504, line 4: You'll probably need to wait for an event

lib/plugins/rain.js
```diff
@@ -2,11 +2,20 @@ module.exports = inject
...
 function inject (bot) {
   bot.isRaining = false
+  let rainingBeginState, rainingEndState
+  if (bot.supportFeature('rainingStateBeginsWithOne')) {
+    rainingBeginState = 1
+    rainingEndState = 2
+  } else {
+    rainingBeginState = 2
+    rainingEndState = 1
+  }
```
wvffle, line 12:
```js
const rainingState = bot.supportFeature('rainingStateBeginsWithOne')
  ? [1, 2]
  : [2, 1]
```

And use array or move output of supportFeature to a variable and use it as 
```js
const rainingBeginState = beginsWithOne
  ? 1
  : 2
```

Leftier:
I'm a bit stuck i've been debugging for a bit and it looks like the protocol doesn't change in 1.14.4.
Every version emit one when is raining and two when it stops

rom1504:
Maybe try to change back the implementation to that and see if the tests pass ?

PR #1100, opened 2020-06-29
Fix off by 1 error in findBlocks

Karang:
Thanks to LucKYsHoTSzZz for finding this bug.


rom1504:
Could you explain what was the bug and how that fixes it ?

Karang:
> Could you explain what was the bug and how that fixes it ?

Sure,
If the bot is at x=15 in a chunk (lets say xChunk=0), maxDistance is set to 16, it should be able to check blocks in xChunk = 2, but with the old code (Math.ceil(maxDistance/16)) it would only check xChunk = 1
The new code, round the chunk distance properly so that if the bot position is > 8 in local chunk coordinates, the chunk distance is expanded by 1.

rom1504:
Ok makes sense thanks

Issue #1096, opened 2020-06-28
bot.dig freezes after 100th or so call


thesourceoferror:
## Versions
 - mineflayer: 2.21.0
 - server: vanilla/spigot/paper 1.15.2
 - node: 12.18.1

## Detailed description of a problem
I am developing a famer bot, and the bot walks over a layer of soul sand to both harvest and sow nether warts. However, after a couple minutes or so of harvesting, the bot freezes.

## Expected behavior
Simply expect the bot to break a block, without freezing.

## Additional context
I need to call the function bot.dig(block, cb) quite frequently, which is where I encounter a problem. Most of the time, bot.dig works just perfectly, however, once every 100th or so function call, it freezes, and with it the whole bot.

I looked into the source code and where it stalls seems to be in a function called "checkYaw" in plugins/physics.js, in which there is an if statement that requires a math equation to be less than 0.001, however, when the bug is encountered, this statement is at a consistent 2PI instead of dropping towards 0.001.

I havent investigated the issue further.

rom1504:
Interesting. 2Pi is indeed the same as 0
Probably something somewhere doesn't reset it to 0.

thesourceoferror:
Fixed with pull request #1101

PR #1101, opened 2020-06-29
Update physics.js

thesourceoferror:
Fix to issue #1096: The statement
`((lastSentYaw % PI_2) - bot.entity.yaw)`
can yield a result which approaches PI_2, which messes with the second modulo operation. Thus, the statement will not be less than 0.001, even though it should be.

### DIFF
--- a/lib/plugins/physics.js
+++ b/lib/plugins/physics.js
@@ -104,8 +104,8 @@ function inject (bot) {
     bot._client.write('position_look', lastSent)
     bot.emit('move', oldPos)
   }
-  
-  let deltaYaw;
+
+  let deltaYaw
   function updatePosition (dt) {
     // If you're dead, you're probably on the ground though ...
     if (!bot.isAlive) bot.entity.onGround = true
@@ -195,8 +195,7 @@ function inject (bot) {
 
   let lookAtCb = noop
   function checkYaw () {
-        
-    if (Math.abs(deltaYaw)<0.001) {
+    if (Math.abs(deltaYaw) < 0.001) {
       lookAtCb()
       lookAtCb = noop
     }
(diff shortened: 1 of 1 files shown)


Karang:
Indeed, nice catch. The computation at https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/physics.js#L113 should be correct and is a bit simpler. Could you use it so we use only one way of computing the delta through all the code ? Ideally it could be made into a function and called in both place.

thesourceoferror:
There we go! That should do the trick.

Karang:
> There we go! That should do the trick.

Looks good to me!

Can you run `npm run fix` to make the linter happy ?

thesourceoferror:
I'm not at all familiar with `npm run fix`, but I attempted to run `npm run fix physics.js`, however it did not seem to make any difference?

Karang:
> I'm not at all familiar with `npm run fix`, but I attempted to run `npm run fix physics.js`, however it did not seem to make any difference?

just run `npm run fix` in the mineflayer directory

thesourceoferror:
Accidentally closed the pull request - I'm not completely familiar with github yet. But, the changed code should fit with the standard now.

lib/plugins/physics.js
```diff
@@ -194,7 +195,7 @@ function inject (bot) {
 
   let lookAtCb = noop
   function checkYaw () {
-    if (Math.abs(((lastSentYaw % PI_2) - bot.entity.yaw) % PI_2) < 0.001) {
+    if (Math.abs(deltaYaw) < 0.001) {
```
rom1504, line 198: can you use a function to compute this ?
Defining a variable at the top like you did is confusing and potentially wrong.

Issue #1003, opened 2020-05-18
noteHeard event not passing correct data


msf567:
## Versions
 - mineflayer: 2.15.0
 - server: paperspigot 1.15.2
 - node: 12.16.3

## Detailed description of a problem
noteHeard event is not passing accurate data. instrument name is always harp and pitch is always 0

## Current code
```js
bot.on('noteHeard',(block,instrument,pitch)=>
{

  bot.chat(`Music for my ears! I just heard a ${instrument.name}`)
  bot.chat('i hear a note at pitch ' + pitch)
  if(pitch == '10')
  {
    bot.setControlState('jump', true)
    bot.setControlState('jump', false)
  }
})
```

## Expected behavior
instrument pitch and type should be accurate to the block

Karang:
Indeed, the way to retreive the note changed in 1.13 https://wiki.vg/Block_Actions#Note_Block
This needs to be fixed here: https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/block_actions.js#L13

rom1504:
fixed

PR #1115, opened 2020-07-06
Fix noteHeard event not working in 1.13 and above

Naomi-alt:
This will fix the 'noteHeard' event not working. #1003


rom1504:
Seems tests are failing

Naomi-alt:
Whoops, I forgot to check the formatting, should be good now.

PR #1122, opened 2020-07-08
Add ability to change the skin parts of a bot

Naomi-alt:


Naomi-alt:
Yeah, I guess that makes more sense.

I've also added an example, as suggested by @rom1504

PR #1125, opened 2020-07-10
Add looker example

AlexProgrammerDE:
A small example which looks at the closest entity.

### DIFF
--- a/examples/looker.js
+++ b/examples/looker.js
@@ -31,8 +31,8 @@ bot.once('spawn', function () {
 
 function nearestEntity (type) {
   let id, entity, dist
-  const best = null
-  const bestDistance = null
+  let best = null
+  let bestDistance = null
 
   for (id in bot.entities) {
     entity = bot.entities[id]
--- a/examples/looker.js
+++ b/examples/looker.js
@@ -12,7 +12,7 @@ if (process.argv.length < 4 || process.argv.length > 6) {
 const bot = mineflayer.createBot({
   host: process.argv[2],
   port: parseInt(process.argv[3]),
-  username: process.argv[4] ? process.argv[4] : 'totem',
+  username: process.argv[4] ? process.argv[4] : 'looker',
   password: process.argv[5]
 })
 
@@ -21,7 +21,7 @@ bot.once('spawn', function () {
     var entity = nearestEntity()
     if (entity) {
       if (entity.type === 'player') {
-        bot.lookAt(entity.position.offset(0, 1, 0))
+        bot.lookAt(entity.position.offset(0, 1.6, 0))
       } else if (entity.type === 'mob') {
         bot.lookAt(entity.position)
       }
@@ -30,18 +30,24 @@ bot.once('spawn', function () {
 })
 
 function nearestEntity (type) {
-  var id, entity, dist
-  var best = null
-  var bestDistance = null
+  const id, entity, dist
+  const best = null
+  const bestDistance = null
+  
   for (id in bot.entities) {
     entity = bot.entities[id]
+    
     if (type && entity.type !== type) continue
+    
     if (entity === bot.entity) continue
+    
     dist = bot.entity.position.distanceTo(entity.position)
+    
     if (!best || dist < bestDistance) {
       best = entity
       bestDistance = dist
     }
   }
+  
   return best
 }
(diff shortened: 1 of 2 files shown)


examples/looker.js
```diff
@@ -0,0 +1,47 @@
...
+
+if (process.argv.length < 4 || process.argv.length > 6) {
+  console.log('Usage : node echo.js <host> <port> [<name>] [<password>]')
+  process.exit(1)
+}
+
+const bot = mineflayer.createBot({
+  host: process.argv[2],
+  port: parseInt(process.argv[3]),
+  username: process.argv[4] ? process.argv[4] : 'totem',
```
Karang, line 15: name it 'looker' ?

examples/looker.js
```diff
@@ -0,0 +1,47 @@
...
+  password: process.argv[5]
+})
+
+bot.once('spawn', function () {
+  setInterval(() => {
+    var entity = nearestEntity()
+    if (entity) {
+      if (entity.type === 'player') {
+        bot.lookAt(entity.position.offset(0, 1, 0))
```
Karang, line 24: If you want to look in the player eye, you can add 1.6 (the player eye height)

examples/looker.js
```diff
@@ -0,0 +1,47 @@
...
+        bot.lookAt(entity.position)
+      }
+    }
+  }, 50)
+})
+
+function nearestEntity (type) {
+  var id, entity, dist
+  var best = null
+  var bestDistance = null
```
wvffle, line 35: Use const/let

examples/looker.js
```diff
@@ -0,0 +1,47 @@
...
+    entity = bot.entities[id]
+    if (type && entity.type !== type) continue
+    if (entity === bot.entity) continue
+    dist = bot.entity.position.distanceTo(entity.position)
+    if (!best || dist < bestDistance) {
+      best = entity
+      bestDistance = dist
+    }
+  }
+  return best
```
wvffle, line 46: separate with newlines for readability

AlexProgrammerDE:
oops forgot that i have opened this. I will adress your requests now.

examples/looker.js
```diff
@@ -0,0 +1,53 @@
...
+  let bestDistance = null
+
+  for (id in bot.entities) {
+    entity = bot.entities[id]
+
+    if (type && entity.type !== type) continue
+
+    if (entity === bot.entity) continue
+
+    dist = bot.entity.position.distanceTo(entity.position)
```
wvffle, line 44: this could be const

examples/looker.js
```diff
@@ -0,0 +1,53 @@
...
+    }
+  }, 50)
+})
+
+function nearestEntity (type) {
+  let id, entity, dist
+  let best = null
+  let bestDistance = null
+
+  for (id in bot.entities) {
```
wvffle, line 37: id can be const

examples/looker.js
```diff
@@ -0,0 +1,53 @@
...
+    entity = bot.entities[id]
```
wvffle, line 38: this can be const

AlexProgrammerDE:
@wvffle i don't understand why declaring the type of the variable is so important. It works fine without it.

rom1504:
it's fine

just FYI javascript is a very unsafe language. To avoid one kind of error, it's good to use const to avoid a variable that is supposed not to change to be reassigned.

wvffle:
It's an **example**. Many new javascript devs  will look at it and use it as a source of their knowledge. We should teach good patterns, not bad ones. Having a global variable or a variable that is in the upper scope is a most likely a memleak. 

Also, changing the value type of a variable is really slow. (idk if it's slow with `undefined` but I doubt, dunno what with `null`)

examples/looker.js
```diff
@@ -0,0 +1,53 @@
...
+const bot = mineflayer.createBot({
+  host: process.argv[2],
+  port: parseInt(process.argv[3]),
+  username: process.argv[4] ? process.argv[4] : 'looker',
+  password: process.argv[5]
+})
+
+bot.once('spawn', function () {
+  setInterval(() => {
+    var entity = nearestEntity()
```
wvffle, line 21: There is also a `var` which is pretty much unsafe. You should use `const` there.

wvffle:
I'm for reverting this and fixing this issues.

AlexProgrammerDE:
ok do it if you want.

rom1504:
@wvffle fixed there https://github.com/PrismarineJS/mineflayer/pull/1147

Issue #1089, opened 2020-06-25
Clearly address the status for each kind of user target : make a guide


rom1504:
I think we have different kind of users
We can choose not to address some categories, or focus on some
The big categories :
1. People that want to play with advanced things like ai or Network stuff : we have at most 10 of them, for us the most interesting is well designed stuff and interesting problems
2. People that want to code, do something for an use case for minecraft, for them it's important to have a good doc and some advanced tutorials, and some help on discord
3. People that might be ok with coding a little bit with a lot of hand holding, for them a YouTube tutorial might be good
4. People that just want to use a bot or a server, and don't want to code, the only thing that work for them is an easy ui to use, click button

I think we address 1 and 2 pretty well
3 a little via discord but it's inefficient
4 produces noise and is just painful today, we don't handle it

Let's say in a period of one year :
Category 1 : 6 people
Category 2 : 100 people
Category 3 : 1000 people
Category 4 : actually hard to estimate since they just bounce currently, come to GitHub don't find install link, quit, but I'd say like 10000 people

But this is a free time project so it's completely fine to focus only on 1 and 2

However I think it would be useful to create a guide to redirect each category of user to the right thing.

rom1504:
Just to clarify before this goes into any direction I didn't meant it to go :
This is a classification of use cases, of how to handle better questions and support better users.
Not everybody is at the same knowledge level and not everybody care to invest the same about of time into this project.

The objective is definitely not to judge people, but only to try to guide people in the right way : this is no point advising a beginner to start working on the hardest task, and it also inefficient to advise someone who knows a lot about programming in js to read some js tutorial.

So please don't start using this as a "category system", it is not a ranking and the objective is not to say some people are better than others. It is perfectly fine not to know anything about programming.

I'll make this very clear In a future guide

rom1504:
https://github.com/Chocobozzz/PeerTube readme + https://github.com/Chocobozzz/PeerTube/blob/develop/.github/CONTRIBUTING.md is great

PR #1126, opened 2020-07-10
Improve tutorial

Naomi-alt:
I've improved the tutorial so that the people of category 3 from #1089 will be able to get started.

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -3,200 +3,457 @@
 **Table of Contents**
 
 - [Basics](#basics)
+  - [Javascript basics](#javascript-basics)
+    - [Installing Node](#installing-node)
+    - [Javascript variables](#javascript-variables)
+    - [Javascript functions](#javascript-functions)
+    - [Show output](#show-output)
+    - [Javascript types](#javascript-types)
+    - [If-statements](#if-statements)
+    - [Loops](#loops)
+    - [General tips](#general-tips)
+    - [Node Package manager](#node-package-manager)
   - [Creating a bot](#creating-a-bot)
+    - [Javascript objects](#javascript-objects)
+    - [Logging in](#logging-in)
+  - [Passing along functions](#passing-along-functions)
   - [Listening for an event](#listening-for-an-event)
-    - [Echo bot](#echo-bot)
-    - [Welcome Bot](#welcome-bot)
   - [Callbacks](#callbacks)
-    - [Craft Stick Bot](#craft-stick-bot)
-    - [Consuming Bot](#consuming-bot)
-    - [Fisherman Bot](#fisherman-bot)
 - [Advanced](#advanced)
+  - [Asynchronousy](#asynchronousy)
   - [Creating an event from chat](#creating-an-event-from-chat)
     - [Answer Hello Bot](#answer-hello-bot)
-    - [Custom chat](#custom-chat)
+    - [Custom Chat](#custom-chat)
 - [FAQ](#faq)
-  - [How to create discord bot with mineflayer](#how-to-create-discord-bot-with-mineflayer)
-  - [How to run bot on android](#how-to-run-bot-on-android)
+  - [How to run a bot on android](#how-to-run-a-bot-on-android)
+    - [Install Termux](#install-termux)
+    - [Setup](#setup)
+    - [Start your bot](#start-your-bot)
+
+## Introduction
+
+This tutorial will help you get started with Mineflayer, even if you know nothing about coding.  
+If you already know some things about Node and NPM, you can go to the [Create a bot](#creating-a-bot) section, otherwise start here.
 
 ## Basics
 
-For the following, we assume Mineflayer is installed and you have a text editor.
+The following sections are about basics concepts you need to know to get started using Mineflayer.
 
-### Creating a Bot
+### Javascript basics
 
-To get a feeling for the Mineflayer and how to use it, please open a text editor and follow the next few steps.
+#### Installing Node
 
-First, create an instance of the Bot.
+In this section you will learn the basics about Javascript, Node and NPM.
+
+Javascript, often abbreviated to JS, is a programming language designed for the web. It is what makes most interactivity on the web possible.  
+Node.js, often just Node, makes it possible to use Javascript outside of web browsers.
+
+So the first thing you have to do to get started is to install Node. You can get it [here](https://nodejs.org/en/download/).  
+After it is installed, open a command prompt (also known as a terminal) and then type `node -v`  
+If you have installed Node correctly, it should return a version number. If it says it can't find the command, try installing it again.
+
+Now you have Node, you could start writing code, but we need to do 1 more thing.  
+Javascript can be written in any basic text editor, but it is much easier if you use what is called an [Integrated development environment](https://en.wikipedia.org/wiki/Integrated_development_environment)(IDE)  
+An IDE will help you write code because it can give you suggestions, or tell you if your code has potential problems. A good IDE to start with is [Visual Studio Code](https://code.visualstudio.com/)(VSCode)  
+Once you have installed and set-up VSCode, create a new file and then save it somewhere with a name ending with `.js`, e.g. `bot.js`  
+This will let VSCode know we are working with Javascript, and give you the correct suggestions.
+
+#### Javascript variables
+
+Start by typing the following:
 
 ```js
-const mineflayer = require('mineflayer')
-const options = {
-  host: 'localhost', // optional / minecraft server address
-  port: 25565, // optional
-  username: 'email@example.com', // email and password are required only for
-  password: '12345678', // online-mode=true servers
-  version: false // false corresponds to auto version detection (that's the default), put for example '1.8.8' if you need a specific version
+var test = 5
+```
+
+This will create a new variable named `test` and assign it the value `5`  
+Variable are used to save data and use it later in the code.
+
+Now save the file so we can run the code. Open a terminal again (or a new terminal in VSCode) and navigate to the same folder the file is saved in. This can be done using the `cd` command, for example: `cd Documents\javascript`  
+Once your terminal is in the same folder as your Javascript file, you can run `node filename.js`  
+If you have done everything correctly, you should see nothing.
+
+#### Javascript functions
+
+Next you will learn about functions. Functions are a piece of code that can be used multiple times throughout your code.  
+These can be useful because you don't have to type something multiple times.
+
+```js
+function addition(a, b) {
+  var result = a + b
+  return result
 }
+
+var test1 = addition(5, 10)
+var test2 = addition(1, 0)
+```
+
+The `function` keyword is used to define a function.  
+After that you have the name of the function, in this case `addition`  
+After that comes the parameter list, everything between the round brackets `()` are parameters, separated by a comma.  
+Parameters are variables you can give to your function so that your function can work with them.  
+Then at last comes the function body, this is everything between the curly brackets `{}`  
+This is where you put the code of the function.
+
+As you can see, this code takes the parameters `a` and `b` and adds them together in the `result` variable.  
+Then the function will return the result.  
+When a function is defined, the code in the function body is not yet executed. To run a function you have to call it.  
+You can call a function by using the name of a function followed by round brackets. In this case `addition()`  
+However, the `addition` function requires 2 parameters. These can be passed along by putting them inside the round brackets, comma separated: `addition(1, 2)`  
+When the function is done, you can imagine that the function call is replaced by whatever the function has returned. So in this case `var test1 = addition(5, 10)` will become `var test1 = result` (You will not actually see this, but this can help you understand the concept)
+
+#### Show output
+
+Now when you save and run the code, you will still see nothing (If you did everything correctly)  
+To see something, for example if you want to see whether your addition function actually returns the correct value, we you to 'print' something to the terminal.  
+In Javascript, we can do this using the `console.log()` function.  
+
+```js
+function addition(a, b) {
+  var result = a + b
+  return result
+}
+
+var test1 = addition(5, 10)
+var test2 = addition(1, 0)
+
+console.log(test1)
+console.log(test2)
+```
+
+Now when you save and run this code, you should finally see something:
+
+```
+15
+1
+```
+
+#### Javascript types
+
+So far we have only worked with numbers, but Javascript can work with more variable types:
+
+- A string is a piece of text that can contain multiple characters. Strings are defined by using the quotes `''`
+- An array is a type that can hold multiple variables inside itself. Arrays are defined by using the square brackets `[]`
+- Object are basically advanced arrays, you will learn more about it later in this tutorial. Their defined by curly brackets `{}` 
+- Functions are also their own type.
+- A boolean is a type that can only be `true` or `false`
+- When something is not (yet) defined, its type is `undefined`
+
+```js
+var number = 1                  // Number type
+var string = 'This is a string' // String type
+var array = [1, 2, 3]           // Array type
+var object = {}                 // Object type
+var function = Function() {}    // Function type
+var boolean = true              // Boolean type
+var nothing                     // Undefined type
+```
+
+#### If-statements
+
+Sometimes you want to do different things based on a certain condition.  
+This can be achieved using if-statements.
+
+```js
+var name = 'Bob'
+
+if (name == 'Bob') {
+  console.log('Your name is Bob')
+}
+else if (name == 'Alice') {
+  console.log('Your name is Alice')
+}
+else {
+  console.log('Your name is not Bob or Alice')
+}
+```
+
+An if-statement is created using the `if` keyword. After that you have a condition between the round brackets `()` followed by the body between the curly brackets `{}`
+A condition has to be something that computes to a boolean.  
+In this case it uses an equal operator `==` which will be `true` if the value in front is the same as the value after. Otherwise it will be `false`
+If the condition is `true` the code in the body will be executed.  
+You can chain an if-statement with an if-else-statement or an else-statement.  
+You can have multiple if-else-statements, but there can't be any code between the statements.  
+If you have an else-statement, it will be called only if all the chained statements before it are `false`
+
+```js
+if (true) {
+  ...
+}
+console.log(5); // This is invalid.
+else {
+  ...
+}
+```
+
+#### Loops
+
+Loops are used to repeat certain code until a certain conditional is met.
+
+```js
+var countDown = 5
+
+while (countDown > 0) {
+  console.log(countDown);
+  countDown = countDown - 1
+}
+
+console.log('Finished!')
+```
+
+The above code will print the following
+
+```txt
+5
+4
+3
+2
+1
+Finished!
+```
+
+The `while` loop has a condition `()` and a body `{}`  
+It will execute the code in the body, as long as the condition is still `true`  
+Each loop, the code prints the current `countDown` number, and then decrements it by 1.  
+After the 5th loop, the condition `0 > 0` will be `false`, and thus the code will move on.
+
+#### General tips
+
+If you want to help yourself and other people understand your code better, you can use comments.  
+Comments can be created using `//` and everything after that is completely ignored by Javascript.
+
+In general, it is good practice to use the `const` keyword instead of the `var` keyword when defining a variable. A variable defined with `const` can't be modified later and is thus constant.  
+Javascript is then able to make your code run more efficiently because it knows it doesn't have to account for variable changes for that variable.  
+If you want a modifiable variable, you will still have to use `var` of course.
+
+```js
+const test = 5
+test = 10 // This line is not valid, as test is constant.
+```
+
+#### Node Package manager
+
+The last thing you need to know is how to use the [Node Package Manager](https://www.npmjs.com/).  
+NPM is automatically installed when you install Node.  
+NPM is used to get useful packages that other people created that can do useful things for you.  
+You can search for packages on [their website](https://www.npmjs.com/), and then install them using the `npm install` command in your terminal.  
+To install Mineflayer for example, run `npm install mineflayer`  
+
+Then, Node can access installed modules by using the `require()` function.
+
+```js
+const mineflayer = require('mineflayer');
+```
+
+After this, the `mineflayer` variable can be used to access all the features of Mineflayer.
+
+### Creating a bot
+
+Now that you know the basis of Javascript, Node and NPM, you're ready to start creating your first bot!  
+If you don't know any of the terms above, you should go back to the [previous section](#javascript-basics)
+
+Below is the absolute minimum necessary to create a Mineflayer bot.
+
+```js
+const mineflayer = require('mineflayer');
+
+const bot = mineflayer.createBot()
+```
+
+However, this bot isn't quite useful, as by default this will connect to a Minecraft server running on your machine with the port 25565.  
+If you want to choose which server you want your bot to connect to, you have to pass along a few options.
+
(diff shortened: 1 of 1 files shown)


docs/tutorial.md
```diff
@@ -3,200 +3,457 @@
...
+if (true) {
+  ...
+}
+console.log(5); // This is invalid.
+else {
+  ...
+}
+```
+
+#### Loops
```
Karang, line 189: Maybe, add something about `for` loops ?

docs/tutorial.md
```diff
@@ -3,200 +3,457 @@
...
 
 ```js
-const mineflayer = require('mineflayer')
-const options = {
-  host: 'localhost', // optional / minecraft server address
-  port: 25565, // optional
-  username: 'email@example.com', // email and password are required only for
-  password: '12345678', // online-mode=true servers
-  version: false // false corresponds to auto version detection (that's the default), put for example '1.8.8' if you need a specific version
+var test = 5
```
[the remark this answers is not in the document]
Naomi-alt: I guess I didn't want to overwhelm people with the difference between `var` and `let`, but I guess for people who are new, they wouldn't even know about the existence of `var`

docs/tutorial.md
```diff
@@ -3,200 +3,457 @@
...
+
+```js
+function addition(a, b) {
+  var result = a + b
+  return result
 }
-const bot = mineflayer.createBot(options)
+
+var test1 = addition(5, 10)
+var test2 = addition(1, 0)
```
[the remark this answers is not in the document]
Naomi-alt: Yeah that makes sense, done.

docs/tutorial.md
Naomi-alt: Done!

docs/tutorial.md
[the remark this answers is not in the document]
Karang: You could have kept the other version, to show the 2 syntaxes next to each others and explain that it means the same thing.

docs/tutorial.md
Karang: You could also add the other variants of for loops:
```js
for (const index in myarray) { // iterate over the keys (indices) of myarray
  const item = myarray[index]
  // for each item in myarray
}
for (const item of myarray) { // iterate directly over the elements of myarray
  // for each item in myarray
}
```

docs/tutorial.md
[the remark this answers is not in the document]
Naomi-alt: Wouldn't that just cause confusion? Isn't it better to only show the best syntax available?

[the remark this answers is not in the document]
Karang: I'm ok with that. It's just that some of our code still uses the old syntax (and there is some special case where we cannot do otherwise) so users may encounter both syntaxes.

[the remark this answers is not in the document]
Naomi-alt: I've added a line about the `function` syntax, without going to deep into it.

docs/tutorial.md
Naomi-alt: Done, although I feel it's a bit vague and hard to explain easily.

docs/tutorial.md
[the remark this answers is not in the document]
Karang: Looks good :)

wvffle:
Quite a lot of text, right now I have no time to read it, but in a couple of hours I'll review it.

docs/tutorial.md
```diff
@@ -3,200 +3,490 @@
...
 
-Here we're crafting oak logs into sticks.
-Craft oak logs into oak wood planks, wait for crafting finish. Then craft the oak wood planks into sticks.
+The last thing you need to know is how to use the [Node Package Manager](https://www.npmjs.com/).  
+NPM is automatically installed when you install Node.  
+NPM is used to get useful packages that other people created that can do useful things for you.  
+You can search for packages on [their website](https://www.npmjs.com/), and then install them using the `npm install` command in your terminal.  
+To install Mineflayer for example, run `npm install mineflayer`  
 
-Incorect aproach ❌ :
```
rom1504, line 119: any reason to remove this part ? I thought it was useful

Naomi-alt: Hmmm, I agree with you. I guess when I reorganized/restructured the tutorial, I removed a bit too much.
I'll add something like it back.

docs/tutorial.md
```diff
@@ -3,200 +3,490 @@
...
+  setTimeout( () => {
+    resolve("Success!")  // Yay! Everything went well!
+  }, 1000) 
+}) 
+
+myPromise.then((successMessage) => {
+  console.log(successMessage) 
+})
+
+myPromise.catch((failureMessage) => {
```
wvffle, line 458: it should be `error` instead of `failureMessage`

docs/tutorial.md
```diff
@@ -3,200 +3,490 @@
...
+The `.then` and `.catch` function can also be chained together with the promise to simplify the code.
+
+```js
+const myPromise = new Promise((resolve, reject) => {
+  setTimeout( () => {
+    resolve("Success!")  // Yay! Everything went well!
+  }, 1000) 
+}).then((successMessage) => {
+  console.log(successMessage) 
+}).catch((failureMessage) => {
```
wvffle, line 478: it should be `error` instead of `failureMessage`

docs/tutorial.md
```diff
@@ -3,200 +3,490 @@
...
-  version: false // false corresponds to auto version detection (that's the default), put for example '1.8.8' if you need a specific version
+let test = 5
+```
+
+This will create a new variable named `test` and assign it the value `5`  
+Variable are used to save data and use it later in the code.
+
+Now save the file so we can run the code. Open a terminal again (or a new terminal in VSCode) and navigate to the same folder the file is saved in. This can be done using the `cd` command, for example: `cd Documents\javascript`  
+Once your terminal is in the same folder as your Javascript file, you can run `node filename.js`  
+If you have done everything correctly, you should see nothing.
```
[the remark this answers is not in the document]
Naomi-alt: Yeah, that makes more sense, done!

docs/tutorial.md
```diff
@@ -3,200 +3,490 @@
...
+
+#### Javascript functions
+
+Next you will learn about functions. Functions are a piece of code that can be used multiple times throughout your code.  
+These can be useful because you don't have to type something multiple times.
+
+```js
+let addition = (a, b) => {
```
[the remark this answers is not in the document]
Naomi-alt: I've moved the const section to the `javascript variables` chapter

docs/tutorial.md
```diff
@@ -3,200 +3,490 @@
...
-- `bot.once(eventName, listener)`
-  Execute the `listener` function, the first time the event named `eventName` triggered.
-- `bot.removeListener(eventName, listener)`
-  Removes the specified `listener` for the event named `eventName`.
+So far we have only worked with numbers, but Javascript can work with more variable types:
 
-Not only bot object, [`Chest`](http://mineflayer.prismarine.js.org/#/api?id=mineflayerchest), [`Furnace`](http://mineflayer.prismarine.js.org/#/api?id=mineflayerfurnace), [`Dispenser`](http://mineflayer.prismarine.js.org/#/api?id=mineflayerdispenser), [`EnchantmentTable`](http://mineflayer.prismarine.js.org/#/api?id=mineflayerenchantmenttable), [`Villager`](http://mineflayer.prismarine.js.org/#/api?id=mineflayervillager) object also have their own events!
+- A string is a piece of text that can contain multiple characters. Strings are defined by using the quotes `''`
+- An array is a type that can hold multiple variables inside itself. Arrays are defined by using the square brackets `[]`
+- Object are basically advanced arrays, you will learn more about it later in this tutorial. Their defined by curly brackets `{}`
```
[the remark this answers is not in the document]
Naomi-alt: I've replaced the empty function with the `addition` function from the previous chapter.

docs/tutorial.md
```diff
@@ -3,200 +3,490 @@
...
-Examples :
+```js
+let number = 1                  // Number type
+let string = 'This is a string' // String type
+let array = [1, 2, 3]           // Array type
+let object = {}                 // Object type
+let function = () => {}         // Function type
+let boolean = true              // Boolean type
+let nothing                     // Undefined type
+```
```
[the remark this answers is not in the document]
Naomi-alt: Done

docs/tutorial.md
```diff
@@ -3,200 +3,490 @@
...
-function echo (username, message) {
-  if (username !== bot.username) {
-    bot.chat(message)
-  }
+let name = 'Bob'
+
+if (name == 'Bob') {
+  console.log('Your name is Bob')
+}
+else if (name == 'Alice') {
```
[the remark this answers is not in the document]
Naomi-alt: Do you mean `} else if` because I think it wouldn't be too confusing if we do this

docs/tutorial.md
```diff
@@ -3,200 +3,490 @@
...
+else {
+  console.log('Your name is not Bob or Alice')
+}
+```
+
+An if-statement is created using the `if` keyword. After that you have a condition between the round brackets `()` followed by the body between the curly brackets `{}`
+A condition has to be something that computes to a boolean.  
+In this case it uses an equal operator `==` which will be `true` if the value in front is the same as the value after. Otherwise it will be `false`
+If the condition is `true` the code in the body will be executed.  
+You can chain an if-statement with an if-else-statement or an else-statement.
```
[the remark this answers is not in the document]
Naomi-alt: Yup, changed

docs/tutorial.md
```diff
@@ -3,200 +3,490 @@
...
+
+```js
+if (true) {
+  ...
+}
+console.log(5) // This is invalid.
+else {
+  ...
 }
+```
```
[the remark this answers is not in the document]
Naomi-alt: Wow, never thought about it like that, I've changed it to be on the same line.

docs/tutorial.md
```diff
@@ -3,200 +3,490 @@
...
+4
+3
+2
+1
+Finished!
 ```
 
-`bot.on('chat', echo)` - will listen for [`chat`](http://mineflayer.prismarine.js.org/#/api?id=quotchatquot-username-message-translate-jsonmsg-matches) event, and execute `echo` function with arguments `username`, `message`, `translate`, `jsonMsg`, and `matches` when the event emitted.
+The `while` loop has a condition `()` and a body `{}`  
+It will execute the code in the body, as long as the condition is still `true`
```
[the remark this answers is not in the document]
Naomi-alt: I've made it more clear what happens after each loop.

docs/tutorial.md
```diff
@@ -3,200 +3,490 @@
...
-#### Welcome Bot
+Instead of only a condition, the for loops has 3 different parts  
+These parts are separated by a semi-column.  
+The first parts `let countDown = 5` is only executed once, at the start of the loop.  
+The second part `countDown > 0` is the condition, this is the same as the while loop.  
+The third part `countDown = countDown - 1` is executed after each loop.:
 
-Here we're listening for [`playerJoined`](http://mineflayer.prismarine.js.org/#/api?id=quotplayerjoinedquot-player) event.
-And sending welcome message to chat.
+If you want to do something for every item in an array, a `for in` or `for of` loop can be useful.
```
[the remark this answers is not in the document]
Naomi-alt: I already wasn't to certain about this section, but hearing your explanation, I've removed the `for..in`

docs/tutorial.md
```diff
@@ -3,200 +3,490 @@
...
 }
 
-bot.on('playerJoined', joined)
+for (const item of array) {
+  console.log(item)
+}
 ```
 
-### Callbacks
+A `for in/of` loop needs to have a variable before the in/of, this is the variable that can be used to access the current item.
```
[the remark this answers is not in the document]
Naomi-alt: What do you mean with this?

docs/tutorial.md
```diff
@@ -3,200 +3,490 @@
...
+test = 10 // This line is not valid, as test is constant.
+```
 
-#### Craft Stick Bot
+#### Node Package manager
 
-Here we're crafting oak logs into sticks.
-Craft oak logs into oak wood planks, wait for crafting finish. Then craft the oak wood planks into sticks.
+The last thing you need to know is how to use the [Node Package Manager](https://www.npmjs.com/).  
+NPM is automatically installed when you install Node.
```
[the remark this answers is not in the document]
Naomi-alt: I also use Arch (btw) and when I did my research for this and found out it was automatically installed on windows, it felt weird :P

docs/tutorial.md
```diff
@@ -3,200 +3,490 @@
...
+```
+
+However, this bot isn't quite useful, as by default this will connect to a Minecraft server running on your machine with the port 25565.  
+If you want to choose which server you want your bot to connect to, you have to pass along a few options.
+
+
+```js
+const mineflayer = require('mineflayer')
+
+const options = {host: 'localhost', port: 25565} // Change this to the ip and port you want
```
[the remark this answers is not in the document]
Naomi-alt: I've changed it to variant b.

docs/tutorial.md
```diff
@@ -3,200 +3,490 @@
...
+#### Javascript objects
+
+The curly brackets `{}` are used to create an object.  
+Objects contain what is called a key-value pair.  
+A key-value pair consist of a colon `:` and a key before the colon, and the value of that key after the colon.  
+The keys can then be used to retrieve their value.  
+You can have multiple key-value pairs by separating them by commas.
+
+```js
+const object = {number: 10, another: 5}
```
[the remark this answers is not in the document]
Naomi-alt: Wouldn't it be nice to have variant b like you said above everywhere possible?

docs/tutorial.md
```diff
@@ -3,200 +3,490 @@
...
+console.log(object.number) // This will print the value 10
+```
+
+This concept is often used to create what is named 'named parameters'  
+The advantage of this is that you don't have to use all the options available, and their position does not matter.  
+The value can be anything, even other object. If the value is a function, that function is often called a method for that object.  
+You can also create the object in-line.
+
+```js
+const bot = mineflayer.createBot({host: 'localhost', port: 25565})
```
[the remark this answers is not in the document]
Naomi-alt: Same as above.

docs/tutorial.md
```diff
@@ -3,200 +3,490 @@
...
-    bot.consume(onFinishEating)
-    // `bot.consume()` method takes `onFinishEating` as callback, and execute `onFinishEating` when its finish or stumble an error.
-  }
-  function onFinishEating (err) {
-    if (err !== null) {
-      bot.chat(err.message)
-      return
-    }
-    bot.chat(`I'm finish eating ${heldItemName}`)
+function welcome() {
```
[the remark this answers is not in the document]
Naomi-alt: I guess I did, didn't I. Changed it to arrow functions.

docs/tutorial.md
Naomi-alt: Done

docs/tutorial.md
Naomi-alt: Done

docs/tutorial.md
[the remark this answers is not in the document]
wvffle: yeah, I mean that

docs/tutorial.md
```diff
@@ -71,20 +71,50 @@ Variable are used to save data and use it later in the code.
 
 Now save the file so we can run the code. Open a terminal again (or a new terminal in VSCode) and navigate to the same folder the file is saved in. This can be done using the `cd` command, for example: `cd Documents\javascript`  
 Once your terminal is in the same folder as your Javascript file, you can run `node filename.js`  
-If you have done everything correctly, you should see nothing.
+If you have done everything correctly, you should see nothing.  
+In the next chapter we will show you how you can 'print' things to the terminal.
+
+In general, it is good practice to use the `const` keyword instead of the `let` keyword when defining a variable. A variable defined with `const` can't be modified later and is thus constant.
```
wvffle, line 77: > and is thus constant

i think that `and thus is a constant` would be more correct

docs/tutorial.md
```diff
@@ -71,20 +71,50 @@ Variable are used to save data and use it later in the code.
 
 Now save the file so we can run the code. Open a terminal again (or a new terminal in VSCode) and navigate to the same folder the file is saved in. This can be done using the `cd` command, for example: `cd Documents\javascript`  
 Once your terminal is in the same folder as your Javascript file, you can run `node filename.js`  
-If you have done everything correctly, you should see nothing.
+If you have done everything correctly, you should see nothing.  
+In the next chapter we will show you how you can 'print' things to the terminal.
+
+In general, it is good practice to use the `const` keyword instead of the `let` keyword when defining a variable. A variable defined with `const` can't be modified later and is thus constant.  
+Javascript is then able to make your code run more efficiently because it knows it doesn't have to account for variable changes for that variable.
```
wvffle, line 78: > variable changes

value changes

docs/tutorial.md
```diff
@@ -71,20 +71,50 @@ Variable are used to save data and use it later in the code.
...
+In general, it is good practice to use the `const` keyword instead of the `let` keyword when defining a variable. A variable defined with `const` can't be modified later and is thus constant.  
+Javascript is then able to make your code run more efficiently because it knows it doesn't have to account for variable changes for that variable.  
+If you want a modifiable variable, you will still have to use `let` of course.
+
+```js
+const test = 5
+test = 10
+```
+
+The second line is invallid because you can't reasign the `test` variable.
```
wvffle, line 86: > reasign

rea**s**sign

docs/tutorial.md
```diff
@@ -103,25 +133,7 @@ When the function is done, you can imagine that the function call is replaced by
...
-
-let test1 = addition(5, 10)
-let test2 = addition(1, 0)
-
-console.log(test1)
-console.log(test2)
-```
-
-Now when you save and run this code, you should finally see something:
+The above should output the following:
```
wvffle, line 136: Above snippet is missing console.logs

docs/tutorial.md
```diff
@@ -133,20 +145,38 @@ Now when you save and run this code, you should finally see something:
...
+- Object are basically advanced arrays, you will learn more about it later in this tutorial. Their defined by curly brackets `{}`
+
+```js
+const object = {}
+```
+
 - Functions are also their own type.
+
+```js
+const function = (a, b) => {return a+b}
```
wvffle, line 167: Again padding: `{ return ... }`

docs/tutorial.md
```diff
@@ -133,20 +145,38 @@ Now when you save and run this code, you should finally see something:
...
 
 ```js
-let number = 1                  // Number type
-let string = 'This is a string' // String type
-let array = [1, 2, 3]           // Array type
-let object = {}                 // Object type
-let function = () => {}         // Function type
-let boolean = true              // Boolean type
-let nothing                     // Undefined type
+const nothing
```
wvffle, line 179:
```js
// Both variables will be undefined
const nothing
const notDefined = undefined
```

docs/tutorial.md
```diff
@@ -213,14 +231,16 @@ Finished!
...
+After the end of the body is reached, the condition is checked again, and if `true`, the body executed again.  
+This will happen for as long as the condition check is still `true`  
+Each loop, this code prints the current `countDown` number, and then decrements it by 1.  
 After the 5th loop, the condition `0 > 0` will be `false`, and thus the code will move on.
 
 A `for` loop is also often used, and differs slightly from a `while` loop.  
 
 ```js
-for (let countDown = 5; countDown > 0; countDown = countDown - 1) {
+for (let	 countDown = 5; countDown > 0; countDown = countDown - 1) {
```
wvffle, line 243: Remove \t

docs/tutorial.md
```diff
@@ -432,6 +444,51 @@ When the consuming ends, the function that is passed along is called.
 We can then do other things that we want to do after.  
 The function could also be called when an error occurs.
 
+#### Correct and incorrect aproach
```
wvffle, line 447: missing from table of contents

docs/tutorial.md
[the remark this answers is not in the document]
wvffle: I see no example how to iterate over object. That could be in the advanced stuff

docs/tutorial.md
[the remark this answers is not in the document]
wvffle: You fixed it well

docs/tutorial.md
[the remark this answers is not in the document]
Naomi-alt: I've added a chapter about in the advanced section.

docs/tutorial.md
Naomi-alt: That does sound nicer.

docs/tutorial.md
Naomi-alt: Changed

docs/tutorial.md
Naomi-alt: Done

docs/tutorial.md
Naomi-alt: Added

docs/tutorial.md
Naomi-alt: Added

docs/tutorial.md
Naomi-alt: Added

docs/tutorial.md
Naomi-alt: Removed

docs/tutorial.md
Naomi-alt: Added

docs/tutorial.md
wvffle: Seems to be lacking the comment. Someone may miss the point covered by it

docs/tutorial.md
```diff
@@ -537,6 +543,66 @@ const myPromise = new Promise((resolve, reject) => {
 })
 ```
 
+### Loop over an object
```
wvffle, line 546: I'd add some warning to not use for..in.
Key arguments against for..in:
a) confusion with for..of
b) inherited properties

docs/tutorial.md
Naomi-alt: I didn't want to use a comment because the comment was explained in a later chapter, but I guess I could just move the comment section higher up.

docs/tutorial.md
Naomi-alt: I've added some info about the `for in` loop.

docs/tutorial.md
```diff
@@ -290,7 +287,9 @@ Then, Node can access installed modules by using the `require()` function.
 const mineflayer = require('mineflayer')
 ```
 
-After this, the `mineflayer` variable can be used to access all the features of Mineflayer.
+After this, the `mineflayer` variable can be used to access all the features of Minef#### General tips
+
+layer.
```
wvffle, line 292: well, here you messed up

docs/tutorial.md
```diff
@@ -603,6 +602,11 @@ These loops are possible because `Object.values(obj)` and `Object.keys(obj)` bot
 `Object.entries(obj)` returns an array where each item is an array with 2 items: a key and its corresponding value.  
 It's important to know that, unlike the `Object.values()` and `Object.keys()` functions, the `Object.entries()` function does not guarantee that the order is the same as the order when the object was defined.
 
+There is also a `for in` loop. However, you will most often want to use `for of` instead of `for in` because there are key differences.  
+The `for in` loop loops over the keys of an object instead of the values. (The index incase it is an array)
```
wvffle, line 606: in case

`incase` is a verb (technically misspelling of `encase`) and means `to enclosing something`

wvffle:
As this got merged, please fix it in new PR @Naomi-alt

Issue #1123, opened 2020-07-09
Chunks don't unload, memory leak


algiuxass:
## Versions
 - mineflayer: 2.23.0
 - server: spigot 1.12.2
 - node: 12.13.1

## Detailed description of a problem
Memory leak when traveling, chunks don't unload.
Node.js will crash if you load too many chunks.

## Current code
```js
let mineflayer = require('mineflayer');

//heap grows when traveling
setInterval(()=>{
    let v = process.memoryUsage();
    console.log(v.rss/1000000+"MB", v.heapTotal/1000000+"MB");
},1000);

var bot = mineflayer.createBot({
    host: "127.0.0.1",
    port: 25565,
    username: 'test_bot',
    version: "1.12.2"
});

bot.on('message', function(message) {
  bot.chat(message);
});
bot.on('login', i=>console.log('Joined.'))
bot.on('end', i=>{
    console.log('Left.')
    process.exit();
})
bot.on('error', err => console.log(err))
```

## Expected behavior
Unload chunks when bot is out of rendering distance.

## Additional context
I manually teleported bot around the world just to make sure that it wasn't some bug in my code. I noticed that heap memory doesn't go lower.

algiuxass:
```js
bot.on('chunkColumnUnload',(point)=>{
    console.log(point);
})
// output:
// Vec3 { x: NaN, y: 0, z: NaN }
```
This is weird.

Karang:
In 1.12, the x/z values are names chunkX, chunkZ: https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/1.12.2/protocol.json#L1478
It needs to be fixed here https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/blocks.js#L262
But first, check if it was always missnamed, or if it change from one version to an other.

algiuxass:
Made a PR. I wonder how this problem was unnoticed for so long?

rom1504:
I guess people don't move their bot around that much. If you don't look for
it it'll get noticable only if you load like 10x the view size.
You'd have to do something a bit special (auto exploration I guess) to get
to this point

Thanks for finding and fixing this bug !

IdanHo:
must have missed it when i wrote that, my bad :disappointed_relieved:

rom1504:
thanks

PR #1127, opened 2020-07-10
delColumn returning NaN (chunk memory leak) fix

algiuxass:
columnCorner in delColumn fix.
Expected: { x: Number, x: Number }
Got: { chunkX: Number, chunkZ: Number }
Weird bug, but fixed.
Added the same fix to addColumn just in case.
Issue #1123

### DIFF
--- a/lib/plugins/blocks.js
+++ b/lib/plugins/blocks.js
@@ -74,14 +74,14 @@ function inject (bot, { version }) {
 
   function delColumn (chunkX, chunkY) {
     const columnCorner = new Vec3(chunkX * 16, 0, chunkY * 16)
-    const key = columnKeyXZ(columnCorner.x, columnCorner.z)
+    const key = columnKeyXZ(columnCorner.chunkX || columnCorner.x, columnCorner.chunkZ || columnCorner.z)
     delete columns[key]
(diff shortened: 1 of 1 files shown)


lib/plugins/blocks.js
```diff
@@ -74,14 +74,14 @@ function inject (bot, { version }) {
 
   function delColumn (chunkX, chunkY) {
     const columnCorner = new Vec3(chunkX * 16, 0, chunkY * 16)
-    const key = columnKeyXZ(columnCorner.x, columnCorner.z)
+    const key = columnKeyXZ(columnCorner.chunkX || columnCorner.x, columnCorner.chunkZ || columnCorner.z)
```
Karang, line 77: All versions that defines unload_chunk packet uses chunkX/chunkZ you do not need the ||

Also you need to fix the issue at the source: https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/blocks.js#L262 (columnCorner is a Vec3, not the packet)

algiuxass: Sorry, stupid mistake. My first serious PR.

Karang: You did not need to close it, you can push more commit to the PR to fix it. The bug you found needs fixing :)

Issue #334, opened 2015-10-01
Use prismarine-world


rom1504:
This will make blockAt async and break everything depending on it.
But it's for the good cause of being able to implement a world representation with any kind of backend (especially async backend like databases)

https://github.com/PrismarineJS/prismarine-world

https://github.com/andrewrk/mineflayer/blob/master/lib/plugins/blocks.js

rom1504:
One way to make the transition smoother is to just deprecate blockAt, implement it using prismarine-world-sync and use bot.world for everything else.

rom1504:
Something interesting about this is a prismarine-world-blockfinder could be made and used in mineflayer-blockfinder that would then be just a few lines.

rom1504:
requires https://github.com/PrismarineJS/prismarine-chunk/issues/16#issuecomment-169614381

rom1504:
(using prismarine-world-sync would break mineflayer in the browser though, maybe it should be an optional dependency and we'd have no blockAt function for the no-prismarine-world-sync case, or maybe we could just remove blockAt altogether, but that would mean doing PRs on all projects using mineflayer)

rom1504:
Related idea : have a prismarine-world-snapshot that would take a snapshot of a given region and allow a sync api on that region. That might be useful to speed up blockfinder.
(or maybe doing it asyncly would be fast enough, better benchmark before doing that)

rom1504:
Alright, let's start with using prismarine-chunk in mineflayer. That will give us cross platform chunk (https://github.com/PrismarineJS/prismarine-chunk/issues/19) + we don't have to deal with the async problem for now.

rom1504:
p-chunk is now used.
using prismarine-world could be considered but would involve a lot of async everywhere. Not not obvious to do

rom1504:
https://github.com/PrismarineJS/mineflayer/pull/1145 made this happen

Follow up : 
* [ ] move find block to somewhere outside of mineflayer, so it can be used with any prismarine-world (either its own package, either directly in prismarine-world)
* [ ] make the viewer use prismarine-world only
* [ ] create an example showing how to save a mineflayer world to anvil (rendered trivial by using prismarine-world)
* [ ] A way to store block entities https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/blocks.js#L27 it would make sense to have it directly in prismarine-chunks similar to what mojang do (would simplify things for storing chunks), but that could go first in prismarine-world
[...]

PR #1145, opened 2020-07-13
use prismarine-world

rom1504:


rom1504:
next steps of this plan : 
* move find block to somewhere outside of mineflayer, so it can be used with any prismarine-world (either its own package, either directly in prismarine-world)
* make the viewer use prismarine-world only
* create an example showing how to save a mineflayer world to anvil (rendered trivial by using prismarine-world)

rom1504:
Then (from @Karang )
* A way to store block entities https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/blocks.js#L27 it would make sense to have it directly in prismarine-chunks similar to what mojang do (would simplify things for storing chunks), but that could go first in prismarine-world
* Some events, world is already an EventEmitter, but no events are emitted (blockUpdate, chunkColumnLoad/Unload, etc..) eventually those events could be removed or proxyed in mineflayer
* entities ? If we want to be able to use only pworld in pviewer, we also need the world to emit entity related events: https://github.com/PrismarineJS/prismarine-viewer/blob/master/lib/mineflayer.js#L82

(this can be done in other PRs, I may copy this content to an issue later)

rom1504:
https://github.com/PrismarineJS/prismarine-world/issues/42 is required so tests won't fail because we don't have anvil support for all versions

lib/plugins/blocks.js
```diff
@@ -194,9 +196,8 @@ function inject (bot, { version }) {
 
   function blockAt (absolutePoint, extraInfos = true) {
     const loc = new Location(absolutePoint)
-    const key = columnKeyXZ(loc.chunkCorner.x, loc.chunkCorner.z)
 
-    const column = columns[key]
+    const column = bot.world.sync.getColumnAt(loc.chunkCorner.x, loc.chunkCorner.z)
```
Karang, line 200: we should be able to use directly:
```js
const block = bot.world.sync.getBlock(absolutePoint)
```
also, I think that pworld should do the line:
```js
block.position = loc.floored
```

lib/plugins/blocks.js
```diff
@@ -74,21 +75,22 @@ function inject (bot, { version }) {
...
     const columnCorner = new Vec3(args.x * 16, 0, args.z * 16)
-    const key = columnKeyXZ(columnCorner.x, columnCorner.z)
     if (!args.bitMap) {
       // stop storing the chunk column
       delColumn(args.x, args.z)
       return
     }
-    let column = columns[key]
-    if (!column) columns[key] = column = new Chunk()
+    let column = bot.world.sync.getColumn(args.x, args.z)
```
wvffle, line 89: I'd add a newline to separate the former if statement from next code section

Karang: I think we should make it work first :) but I agree we need more newline in general

wvffle: > I think we should make it work first :) 

Yup but alongside all of those (small) changes we can make code more readable!

rom1504:
Yeah didn't have time to continue this. I think it's quite doable, must be just a small detail that isn't right

lib/plugins/blocks.js
rom1504: yeah totally, I was mainly trying to make as little change as possible while making it works
I'll clean this up

rom1504:
needed to merge this:
* [ ] fix tests
* [ ] update viewer (_columns is gone)
* [ ] reviews

rom1504:
After the followup on this, bot.blockAt will be deprecated, bot.world.getBlock should be used. (that'll be when everything of the world is ported to prismarine-world)

rom1504:
Ok fixed the tests
Could you review @Karang ?
The idea here is to introduce prismarine-world.
To fully take advantage of it, I put the follow up at https://github.com/PrismarineJS/mineflayer/issues/334#issuecomment-670987395

lib/plugins/blocks.js
```diff
@@ -74,21 +75,22 @@ function inject (bot, { version }) {
 
   function delColumn (chunkX, chunkY) {
     const columnCorner = new Vec3(chunkX * 16, 0, chunkY * 16)
-    const key = columnKeyXZ(columnCorner.x, columnCorner.z)
-    delete columns[key]
+    bot.world.unloadColumn(chunkX, chunkY)
```
Karang, line 78: chunkY -> chunkZ

lib/plugins/blocks.js
```diff
@@ -103,6 +105,21 @@ function inject (bot, { version }) {
     bot.emit('chunkColumnLoad', columnCorner)
   }
 
+  function waitForChunksToLoad (cb) {
+    // check 5x5 chunks around us
+    for (let x = -2; x <= 2; x++) {
+      for (let y = -2; y <= 2; y++) {
+        if (bot.blockAt(bot.entity.position.plus(new Vec3(x, 0, y).scaled(16))) == null) {
```
Karang, line 112: Instead of checking the block, it would be best to check the column !== null directly.
Something like:
```js
const chunkX = Math.floor(bot.entity.position.x / 16) + x
const chunkZ = Math.floor(bot.entity.position.z / 16) + z
if (bot.world.getColumn(chunkX, chunkZ) !== null) {
```

rom1504: done

PR #1147, opened 2020-07-14
Make looker example better

rom1504:


rom1504:
imho the setInterval is also wrong but the alternative is listening to like 10 events, maybe we should have a "anEntityPositionChanged" event or something

PR #1148, opened 2020-07-14
Fix typo and small improvements based on feedback for the tutorial

Naomi-alt:
I had my little brother(17) who is interested in learning programming follow the tutorial and changed a few little things that I missed.

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -425,7 +425,7 @@ To remove specific listener you can use `bot.removeListener()` method.
 - `bot.once(eventName, listener)`
   Execute the `listener` function, only once, the first time the event named `eventName` triggered.
 - `bot.removeListener(eventName, listener)`
-  Removes the specified `listener` for the event named `eventName`. This will only work with named functions `function myNamedFunc() {}`
+  Removes the specified `listener` for the event named `eventName`. In order to use this you either need to define your function with `function myNamedFunc() {}` or put your function in a variable with `const myNamedFunc = () => {}`. You can then use `myNamedFunc` in the listener argument.
 
 Not only bot object, [`Chest`](http://mineflayer.prismarine.js.org/#/api?id=mineflayerchest), [`Furnace`](http://mineflayer.prismarine.js.org/#/api?id=mineflayerfurnace), [`Dispenser`](http://mineflayer.prismarine.js.org/#/api?id=mineflayerdispenser), [`EnchantmentTable`](http://mineflayer.prismarine.js.org/#/api?id=mineflayerenchantmenttable), [`Villager`](http://mineflayer.prismarine.js.org/#/api?id=mineflayervillager) object also have their own events!
 
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -217,7 +217,7 @@ If you have an else-statement, it will be called only if all the chained stateme
 Loops are used to repeat certain code until a certain conditional is met.
 
 ```js
-const countDown = 5
+let countDown = 5
 
 while (countDown > 0) {
   console.log(countDown)
@@ -287,9 +287,7 @@ Then, Node can access installed modules by using the `require()` function.
 const mineflayer = require('mineflayer')
 ```
 
-After this, the `mineflayer` variable can be used to access all the features of Minef#### General tips
-
-layer.
+After this, the `mineflayer` variable can be used to access all the features of Mineflayer.
 
 ### Creating a bot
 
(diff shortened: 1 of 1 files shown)


docs/tutorial.md
```diff
@@ -426,7 +425,7 @@ To remove specific listener you can use `bot.removeListener()` method.
 - `bot.once(eventName, listener)`
   Execute the `listener` function, only once, the first time the event named `eventName` triggered.
 - `bot.removeListener(eventName, listener)`
-  Removes the specified `listener` for the event named `eventName`.
+  Removes the specified `listener` for the event named `eventName`. This will not work with annonymous functions, as they don't have a name.
```
wvffle, line 428: It's a bit confusing, `const a = function () {}` is still an anonymous function.

docs/tutorial.md
```diff
@@ -436,7 +435,7 @@ In Mineflayer, callbacks are often used to handle errors.
 
 ```js
 bot.consume((error) => {
-  if (error) {
+  if (error != undefined) { // If no error occored, 'error' will be undefined. If it is, an error occured.
```
wvffle, line 438: most likely it'll be `null`. Previous check was better as it covers both undefined and null and does not involve an equality operator (You should avoid those and use identity operators.)

docs/tutorial.md
```diff
@@ -468,21 +467,20 @@ Correct approach with callbacks ✔️:
 ```js
 const plankRecipe = bot.recipesFor(5)[0]
 
-bot.craft(plankRecipe, 1, null, function (err) {
+bot.craft(plankRecipe, 1, null, (error) => {
   // After bot.craft(plankRecipe, ...) is finished, this callback is called and we continue. ✔️
-  if (err !== null) { // Check if an error happened.
-    bot.chat(err.message)
-    return
+  if (error != undefined) { // Check if an error happened.
```
wvffle, line 472: Same as above

docs/tutorial.md
```diff
@@ -468,21 +467,20 @@ Correct approach with callbacks ✔️:
...
   } else {
     const stickRecipe = bot.recipesFor(280)[0]
 
-    bot.craft(stickRecipe, 1, null, function (err) {
+    bot.craft(stickRecipe, 1, null, (error) => {
       // After bot.craft(stickRecipe, ...) is finished, this callback is called and we continue. ✔️
-      if (err !== null) { // Check if an error happened.
-        bot.chat(err.message)
-        return
+      if (error != undefined) { // Check if an error happened.
```
wvffle, line 479: Same as above

docs/tutorial.md
Naomi-alt: Ow wait it is? So passing it `a` won't remove the listener?

Karang: `const a = function() {}` and `const a = () => {}` are both function expressions (the variable a gets assigned an anonymous function. In both case it will not remove the listener.

`function a() {}` is a function declaration, this is what you need to be able to remove the listener.

https://www.w3schools.com/js/js_function_definition.asp

Naomi-alt: Well, TIL

I'll mention it in the tutorial.

wvffle: > Ow wait it is? So passing it `a` won't remove the listener?

passing `a` will remove `a` listner. I meant that it's been described in a bit confusing way. What you said there was that even if we assign an anonymous function to a variable and pass that variable it won't work. But it will.

https://nodejs.org/api/events.html#events_emitter_removelistener_eventname_listener
Even example in node api shows so. 

Idk what you meant @Karang, that it won't work.

docs/tutorial.md
```diff
@@ -425,7 +425,7 @@ To remove specific listener you can use `bot.removeListener()` method.
 - `bot.once(eventName, listener)`
   Execute the `listener` function, only once, the first time the event named `eventName` triggered.
 - `bot.removeListener(eventName, listener)`
-  Removes the specified `listener` for the event named `eventName`. This will not work with annonymous functions, as they don't have a name.
+  Removes the specified `listener` for the event named `eventName`. This will only work with named functions `function myNamedFunc() {}`
```
wvffle, line 428: That is not true, discussion [here](https://github.com/PrismarineJS/mineflayer/pull/1148#discussion_r454716261)

rom1504: suggestion for @Naomi-alt , put this : 
> Removes the specified `listener` for the event named `eventName`. In order to use this you either need to define your function with `function myNamedFunc() {}` or put your function in a variable with `const myNamedFunc = () => {}`. You can then use myNamedFunc in the `listener` argument

suggestion for @wvffle : I know it's clear for you but it doesn't seem clear for @Naomi-alt , sometimes suggesting a fix is a good way to help understand and unlock the situation. Reviews are meant to help improve the quality of pull requests, not to block them and we should do our best to reach that objective.

Naomi-alt: Changed it.
I guess my confusion came from the fact that anonymous functions assigned to a variable are still anonymous, while I thought they weren't.
Thank for clearing it up.

PR #1161, opened 2020-07-18
Emit chestLidMove only once

### LINKED ISSUE #1064: "chestLidMove" is triggered twice on a double chest

ImHarvol:
I'm back!
This is a duplicate of #1068, unfortunately I deleted the other fork

> Even thought the vanilla server emits the event twice when it's a double chest, I think it's kind of confusing for a mineflayer user, so I think this change makes sense
> Not sure if this is the best approach, but it does the job!

I also ran `standard --fix` on the example of the README.md. Those semicolon have been bothering me for pretty long.

Closes #1064

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -59,18 +59,21 @@ For example `version:"1.8"`.
 
 ### Echo Example
 ```js
-var mineflayer = require('mineflayer');
+var mineflayer = require('mineflayer')
+
 var bot = mineflayer.createBot({
-  host: "localhost", // optional
+  host: 'localhost', // optional
   port: 25565,       // optional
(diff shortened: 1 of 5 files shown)


wvffle:
what about https://github.com/PrismarineJS/mineflayer/pull/1068#pullrequestreview-433192292?

rom1504:
I think this implementation is ok for now. I agree it could be improved, maybe you can open an issue to track this @wvffle ?

PR #1179, opened 2020-07-24
Adding options for main hand selection.

Colten-Covington:
Was wondering why my bots always held items in the "wrong" hand. Wasn't wrong hand. Was just left hand main is default.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -28,12 +28,14 @@ export interface BotOptions extends ClientOptions {
   chat?: ChatLevel;
   colorsEnabled?: boolean;
   viewDistance?: ViewDistance;
+  mainHand?: MainHands;
   difficulty?: number;
   chatLengthLimit?: number;
 }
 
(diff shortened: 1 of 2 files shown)


IdanHo:
From what i can tell he is using mainHand (options.mainHand in the initialization of the settings property of bot in the inject method), and mineflayer uses the left hand by default because it is not specificed in the 'settings' packet it sends out. (which means it uses the default, 0, with 0 being left and 1 being right)

Colten-Covington:
RIP my discord access because I said "ACK" after a bot spam pinged us all....

Colten-Covington:
Colten45#4375

Colten-Covington:
@u9g 

Ty for checking into this! Was quite sad when it happened. Not sure if it was an accident or whoever banned didn't know what "ACK" meant. But it was just a syn/ack joke because of the mass spam pings. I sent an email to rom as well. Again. Ty for the quick reply!

Issue #1171, opened 2020-07-21
Knowing which entity a mined block has dropped


Garfield100:
Since this isn't a bug I've left out versions etc.

Is there a way to pick up the item a block drops it it mined? It would be nice if the dig function returned the entity ID's of any items dropped as a result of mining the block. I've tried simply waiting for entitySpawn in bot.dig's callback. This works to a degree but isn't reliable. A single dropper dropping items on a clock nearby was enough to cause the event to fire for the dropper's entity, not the mined one.

Is there a more reliable way to do this?

TheDudeFromCI:
While you can't directly find what item drop fall from a block, (there could be multiple or none at all) the most simple approach is to listen for entity spawns of the object type "Item" which occur within one block of the block's center position. Usually, this will occur within a few ticks, so you can cancel the event if not entities are spawned during that time. You can see a sort of example of this, [here.](https://github.com/PrismarineJS/prismarine-gameplay/blob/c29688d5e6e92751ed7b3faae1c085042d0f1f88/src/strats/waitForItemDrop.js#L51)

ImHarvol:
Something that could help is checking that the spawned entity (dropped item) is one of the items that the block drops when it's mined.

[PrismarineJS/minecraft-data](https://github.com/PrismarineJS/minecraft-data) might help you here. Each block has a property `drops` with an array of the items that block can drop.

Garfield100:
> 
> 
> Something that could help is checking that the spawned entity (dropped item) is one of the items that the block drops when it's mined.
> 
> [PrismarineJS/minecraft-data](https://github.com/PrismarineJS/minecraft-data) might help you here. Each block has a property `drops` with an array of the items that block can drop.

This is what I would have liked to do, but unless I missed something there is no way to tell what item id the entity has, which I found confusing. The entity returned by the entitySpawn event has no metadata.

ImHarvol:
The server first fires ["entitySpawn"](https://github.com/PrismarineJS/mineflayer/blob/master/docs/api.md#entityspawn-entity) and then fires a ["Entity Metadata"](https://wiki.vg/Protocol#Entity_Metadata) which seems to contain the item's id.

If this is right must wait a bit for the client to receive the metadata packet, then you can get the item's id by looking for the dropped item's entity in `bot.entities`

Garfield100:
> 
> 
[...]

PR #1181, opened 2020-07-25
Add 'itemDrop' event

ImHarvol:
Instead of the user listening to 'entitySpawn' and then waiting an arbitrary amount of time for the item's metadata to be set, it would be good to have an 'itemDrop' event. Related to #1171

### DIFF
--- a/lib/plugins/entities.js
+++ b/lib/plugins/entities.js
@@ -324,8 +324,10 @@ function inject (bot, { version }) {
   bot._client.on('entity_metadata', (packet) => {
     // entity metadata
     const entity = fetchEntity(packet.entityId)
-    entity.metadata = parseMetadata(packet.metadata, entity.metadata)
+    const metadata = parseMetadata(packet.metadata, entity.metadata)
+    entity.metadata = metadata
     bot.emit('entityUpdate', entity)
+    if (entity.type === 'object' && metadata[6]) bot.emit('itemDrop', entity)
   })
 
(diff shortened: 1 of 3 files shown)


ImHarvol:
There are some problems. It looks like starting 1.13, minecraft-data's entities.json is pretty confusing: in https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/1.13/entities.json there are several objects that share properties, which doesn't seem normal.

rom1504:
What do you mean exactly by "share properties" ?

ImHarvol:
See lines 23 and 940, both have ID 2, which is wrong, isn't it?

Karang:
Neither are wrong. Since 1.13 objects and mob entities are 2 separate sets of ids. I dont remember the details but we have some code in mineflayer to handle that

lib/plugins/entities.js
```diff
@@ -324,8 +324,10 @@ function inject (bot, { version }) {
   bot._client.on('entity_metadata', (packet) => {
     // entity metadata
     const entity = fetchEntity(packet.entityId)
-    entity.metadata = parseMetadata(packet.metadata, entity.metadata)
+    const metadata = parseMetadata(packet.metadata, entity.metadata)
+    entity.metadata = metadata
     bot.emit('entityUpdate', entity)
+    if (entity.name === 'item' && metadata[6]) bot.emit('itemDrop', entity)
```
wvffle, line 330: Add `{` and `}`

PR #1191, opened 2020-07-27
Added FAQ checkbox to question template format.

TheDudeFromCI:


wvffle:
As seen in #1194, prepending `-` before checkbox will work.

wvffle:
If it comes to question template, I've a suggestion to add one more checkbox: 

- [ ] I have carefully read the [API documentation](https://github.com/PrismarineJS/mineflayer/blob/master/docs/api.md) and it does not contain an answer to my question.

Where `carefully` means with clicking into linked pages. Or something similar. That way we could eliminate questions like #1195

wvffle:
In my opinion 2 checkboxes (especially in question template) is better. As a person who wants to ask a question I have to tick 2 of them so it's more likely that I will even read them (both questions in checkboxes and the docs)

wvffle:
Also I'd add a comment under the checkbox:

```<!-- To mark the checkbox, insert `x` into it: [x] -->```
or anything liek that

wvffle:
To sum it up:
- Yeah, `- [ ]` will render empty checkbox and `- [x]` will render marked checkbox, so dash has to be added
- I've added a suggestion to add another checkbox regarding API
- I've added a suggestion to add a comment that will guide how to mark the checkbox (new GH user that registered just to add an issue there won't know that)

rom1504:
I agree about 2 checkbox, indeed it's more likely they will actually click
both links

Might be worth mentioning the examples too ? (A link to the table of
examples in the main readme maybe ?)

PR #1192, opened 2020-07-27
Added topic of console input to FAQ.

TheDudeFromCI:
This is a question I've seen come up several times. Might as well write an answer for it now.

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -40,3 +40,5 @@ One way is to increase the [checkTimeoutInterval](https://github.com/PrismarineJ
 ### How to get the lore / text of an item ?
 You can use the .nbt property of an item. prismarine-nbt nbt.simplify method may be useful
 
+### How can I send message from the console to the server?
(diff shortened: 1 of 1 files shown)


ARandomAssGuy:
> This is a question I've seen come up several times. Might as well write an answer for it now.

you can also use process.stdin, like such:
```js
// Get process.stdin as the standard input object.
var standard_input = process.stdin;

// Set input character encoding.
standard_input.setEncoding('utf-8')
// When user input data and click enter key.
standard_input.on('data', function (data) {
//example data: 'test'
})
```

rom1504:
we have an example for this https://github.com/PrismarineJS/mineflayer/blob/master/examples/repl.js ; we can you add a link to it ?

ARandomAssGuy:
> @ARandomAssGuy
> You can, but using a library is usually safer, imo.
> 
> @rom1504
> Oh, is that what repl is? I was wondering what that example was. Haven't heard of repl, yet.
> Sure, I'll add that now.

yea, theres many examples. I just put it out there for the people that try to use minimal libs

Issue #1098, opened 2020-06-29
Raining state is wrong


mdashlw:
It appears that reason codes for begin/end raining were changed from version to version:
- [Current protocol](https://wiki.vg/Protocol#Change_Game_State) (+ most versions) has 1 end, 2 begin
- [1.14.4 protocol](https://wiki.vg/index.php?title=Protocol&oldid=15346#Change_Game_State) has 1 begin, 2 end

Not sure if there are other versions with 1 begin, 2 end except 1.14.
The issue is in [rain plugin](https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/rain.js) that hardcodes reasons to 1 begin and 2 end.
I'm not PRing it because idk how to do version check properly

rom1504:
We have a feature.json file which specify behavior per version. You can
then use bot.supportFeature('yourFlag') Boolean in the rain plugin.
Would be great if you can PR a fix

Leftier:
#1099

Karang:
As shown in https://github.com/PrismarineJS/mineflayer/pull/1202 1.14.4 behaves exactly like other versions and for all versions:
1 = begin
2 = end
This must be an error in the wiki.

PR #1202, opened 2020-07-29
Add rain test

Karang:
Closes https://github.com/PrismarineJS/mineflayer/issues/1098


Karang:
Tests are passing locally for every versions. I'm not sure why they are hanging here. Will investigate later.

PR #1203, opened 2020-08-02
Fix bot.time.day and add more values

Naomi-alt:
I'm not sure whether bot.time.time is the best name.
And also someone else needs to update the russian api doc.

### DIFF
--- a/lib/plugins/bed.js
+++ b/lib/plugins/bed.js
@@ -84,7 +84,7 @@ function inject (bot) {
 
   function sleep (bedBlock, cb) {
     cb = cb || noop
-    if (!(bot.time.day >= 12541 && bot.time.day <= 23458)) {
+    if (!(bot.time.timeOfDay >= 12541 && bot.time.timeOfDay <= 23458)) {
       cb(new Error("it's not night"))
     } else if (bot.isSleeping) {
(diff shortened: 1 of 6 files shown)


lib/plugins/bed.js
```diff
@@ -84,7 +84,7 @@ function inject (bot) {
 
   function sleep (bedBlock, cb) {
     cb = cb || noop
-    if (!(bot.time.day >= 12541 && bot.time.day <= 23458)) {
+    if (!(bot.time.timeOfDay >= 12541 && bot.time.timeOfDay <= 23458)) {
```
Karang, line 87: What about defining isDay to timeOfDay out of [12542, 23458] so it can be used here ?

Naomi-alt: I guess that just depends on when you consider it to be night.
That definition would be the time at which you can sleep.
I used the moment mobs are able to spawn.
https://minecraft.gamepedia.com/Daylight_cycle

Karang: ok, fair enough.

rom1504:
Kind of breaking but let's count it as a bug fix, the previous thing was weird, I doubt many people used it

PR #1205, opened 2020-08-03
Parse entity_metadata packet for crouching

### LINKED ISSUE #904: crouch and uncrouch are not correctly implemented
The event listeners at:
https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/entities.js#L122
and
https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/entities.js#L126

listen for events that are not emitted anywhere.

IdanHo:
fixes #904, this isnt implementing full parsing like karang suggested, but its good enough.

### DIFF
--- a/lib/plugins/entities.js
+++ b/lib/plugins/entities.js
@@ -326,6 +326,14 @@ function inject (bot, { version }) {
     const entity = fetchEntity(packet.entityId)
     entity.metadata = parseMetadata(packet.metadata, entity.metadata)
     bot.emit('entityUpdate', entity)
+    
+    const bitField = packet.metadata.find(p => p.key === 0)
+    if (bitField === undefined) return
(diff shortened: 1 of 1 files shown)


lib/plugins/entities.js
```diff
@@ -326,6 +326,16 @@ function inject (bot, { version }) {
     const entity = fetchEntity(packet.entityId)
     entity.metadata = parseMetadata(packet.metadata, entity.metadata)
     bot.emit('entityUpdate', entity)
+
+    const bitField = packet.metadata.find(p => p.key === 0)
+    if (bitField === undefined) return
```
wvffle, line 331: Add `{` and `}` with newlines

IdanHo: done

PR #1213, opened 2020-08-05
Task Queue Utility

TheDudeFromCI:
This is a helpful utility object that allows players to define a list of async tasks to perform without needing to working about making a nested pyramid of callbacks. This adds a procedural style, much closer to what new programmers are used to coding with, so it'll help make the transition to callbacks a bit easier. Plus, it looks much nicer.

### DIFF
--- a/examples/taskqueuer.js
+++ b/examples/taskqueuer.js
@@ -29,7 +29,7 @@ bot.on('spawn', () => {
   queue.addSync(() => bot.chat('Going to mine down 10 blocks')) // Support for sync functions, too.
 
   for (let i = 1; i <= 10; i++) {
-    queue.add(cb => setTimeout(cb, 1000)) // Works on any async function
+    queue.add(cb => setTimeout(cb, 1000))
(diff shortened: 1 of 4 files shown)


Karang:
I think this pr is a good idea. But I don't see what you mean about pworld : it has both a sync and async api. The sync pworld was added especially for mineflayer

rom1504:
Hi,
I think adding such an example is useful to help people with async.
However I don't think it makes sense to add this utility.
I think it would be better in the example to use some existing lib such as https://caolan.github.io/async/v3/ or to use promises. A third way can be to put this queue inside the example.

It does not make a lot of sense to maintain this kind of task queue inside mineflayer as it doesn't contain a good subset of what people may want to do with async and I bet we would have to add many things to it, eventually leading to having some kind of clone of async.js

rom1504:
Putting the queue in the example might be a reasonable choice, to show that implementing such a queue is one way to handle async but not the only way (and it could be good to talk about the other ways in comments in the example)

rom1504:
Why close it ? I do mean adding the example is useful

wvffle:
Indeed, there are libraries that do even more than this util, I opt for adding an example both in examples/ and tutorial

Issue #1217, opened 2020-08-08
Calling bot.fish() directly inside bot.fish() callback throws an error


GroobleDierne:
- [X] I have carefully read the [FAQ](https://github.com/PrismarineJS/mineflayer/blob/master/docs/FAQ.md) and it does not contain an answer to my question.
- [X] I have carefully read the [API documentation](https://github.com/PrismarineJS/mineflayer/blob/master/docs/api.md) and it does not contain an answer to my question.
- [X] I have carefully read the [examples](https://github.com/PrismarineJS/mineflayer/tree/master/examples) and they do not contain an answer to my question.

## Versions
 - mineflayer: 2.25.0
 - server: paper 1.15.2
 - node: 12.11.1

## Detailed description of a question

Hello, I made a little bot using mineflayer.
The goal of the bot was to make it fish automatically. To do that I used the bot.fish(cb) method but when I called bot.fish() inside the callback passed to bot.fish() I got the error "Fishing cancelled due to calling bot.fish() again".
Then I used setTimeout() to work around this problem.
I wanted to know if this is a normal behavior or not, if it's not the case I can do a PR to fix it.

```javascript
bot.once('spawn', () => {
    bot.fish(fishingEnd);
});

function fishingEnd(err) {
    if (err) throw err;

    bot.fish(fishingEnd); //throws "Fishing cancelled due to calling bot.fish() again"
}
```

```javascript
bot.once('spawn', () => {
    bot.fish(fishingEnd);
});

function fishingEnd(err) {
    if (err) throw err;

    setTimeout(() => {
        bot.fish(fishingEnd)
    }, 30); //no problem
}
```

Karang:
Indeed thats an error because the callback is reset to undefined after being called https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/inventory.js#L56

Feel free to make a pr to fix it, but be careful there might be multiple place where it needs to be change (and maybe bot.consume() have a similar issue)

GroobleDierne:
Yes, I looked at that before I posted that issue, I'll also look for bot.consume()

PR #1218, opened 2020-08-08
Fix 'Fishing cancelled due to calling bot.fish() again' when calling bot.fish() inside the callback

GroobleDierne:
I've tested, there's no such problem for bot.consume()

### DIFF
--- a/lib/plugins/inventory.js
+++ b/lib/plugins/inventory.js
@@ -53,20 +53,18 @@ function inject (bot, { version }) {
     const pos = lastFloat.position
     if (packet.particleId === 4 && packet.particles === 6 && pos.distanceTo(new Vec3(packet.x, pos.y, packet.z)) <= 1.23) {
       activateItem()
-      let cb = lastFishing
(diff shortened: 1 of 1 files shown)


GroobleDierne:
I forgot the npm run fix :/

IdanHo:
Just do it and add another commit called formatting :) also suggest adding the words "closes #1217" to make it auto close your issue once this pull request gets merged

Karang:
> I've tested, there's no such problem for bot.consume()

Indeed, consume use setImmediate so the cb is effectively called when the function returns (and thus after lastEating is reset to undefined)
```js
lastEating = (err) => setImmediate(cb, err)
```

Can you reset lastFloat also before calling the callback ? (for consistency)

GroobleDierne:
I'm currently waiting for the end of npm install ^^

rom1504:
Why did you close it ?!

GroobleDierne:
It was an error sorry

PR #1222, opened 2020-08-09
Add anvil saver example

rom1504:
#334

### DIFF
--- a/lib/plugins/blocks.js
+++ b/lib/plugins/blocks.js
@@ -14,13 +14,18 @@ const paintingFaceToVec = [
   new Vec3(1, 0, 0)
 ]
 
-function inject (bot, { version }) {
+const dimensionNames = {
+  '-1': 'minecraft:nether',
+  0: 'minecraft:overworld',
+  1: 'minecraft:end'
+}
+
+function inject (bot, { version, storageBuilder }) {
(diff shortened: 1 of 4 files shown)


lib/plugins/blocks.js
```diff
@@ -14,13 +14,13 @@ const paintingFaceToVec = [
...
 
-function inject (bot, { version }) {
+function inject (bot, { version, storageBuilder }) {
   const nbt = require('prismarine-nbt')
   const Block = require('prismarine-block')(version)
   const Chunk = require('prismarine-chunk')(version)
   const ChatMessage = require('prismarine-chat')(version)
   const World = require('prismarine-world')(version)
-  bot.world = new World().sync
+  bot.world = new World(null, storageBuilder ? storageBuilder({ version: bot.version, worldName: 'overworld' }) : null).sync
```
Karang, line 23: What happen if we move to nether or the end ? Should a single World instance deal with all dimensions ?

rom1504: That's in the respawn line in blocks.js, a new world instance is created. worldName will be nether in this case
That's why storageBuilder is a function, so you can call it several time.
One world should handle only one dimension imho

(but indeed I actually hardcoded overworld in blocks.js, will add at least nether)

PR #1243, opened 2020-08-14
Add nearestEntity function to API

### LINKED ISSUE #1231: Bring nearestEntity() function from examples/looker.js into mineflayer

In order to get the nearest entity to a bot, (to my knowledge) one must implement a `nearestEntity()` function something like in `examples/looker.js`, seen [here](https://github.com/PrismarineJS/mineflayer/blob/8540d0838d0c0108bb07df55b01e42c50519509e/examples/looker.js#L32). I think this task is fairly common, and we could prevent code re-use by adding it to the library. 

It would be great if something like `bot.nearestEntity()` existed that simply did what is done in the `looker.js` example. I'd be happy to make a PR copying that function over if someone can advise on what the appropriate place would be in the codebase for it to go.

The main alternative I considered is to simply re-implement that function on each project every time myself or someone else needs it, or stuff it in some sort of 'misc' plugin.

### LINKED ISSUE #1242: How to get Username of nearest player
- [x] I have carefully read the [FAQ](https://github.com/PrismarineJS/mineflayer/blob/master/docs/FAQ.md) and it does not contain an answer to my question.
- [x] I have carefully read the [API documentation](https://github.com/PrismarineJS/mineflayer/blob/master/docs/api.md) and it does not contain an answer to my question.
- [x] I have carefully read the [examples](https://github.com/PrismarineJS/mineflayer/tree/master/examples) and they do not contain an answer to my question.

## Versions
 - mineflayer: #.#.#
 - server: vanilla/spigot/paper #.#.#
 - node: #.#.#

## Detailed description of a question
How can I get the Username of the player which is next to the bot?

Karang:
Closes #1231 

Closes #1242
```js
const nearestPlayerUsername = bot.nearestEntity((entity) => { return entity.type === 'player' }).username
```

### DIFF
--- a/lib/plugins/entities.js
+++ b/lib/plugins/entities.js
@@ -68,14 +68,19 @@ function inject (bot, { version }) {
   bot.nearestEntity = (match = (entity) => { return true }) => {
     let best = null
     let bestDistance = Number.MAX_VALUE
+
     for (const entity of Object.values(bot.entities)) {
-      if (entity === bot.entity || !match(entity)) continue
+      if (entity === bot.entity || !match(entity)) {
+        continue
+      }
+
       const dist = bot.entity.position.distanceSquared(entity.position)
       if (dist < bestDistance) {
         best = entity
         bestDistance = dist
       }
     }
+
     return best
   }
 
--- a/lib/plugins/entities.js
+++ b/lib/plugins/entities.js
@@ -65,6 +65,20 @@ function inject (bot, { version }) {
   bot.uuidToUsername = {}
   bot.entities = {}
 
+  bot.nearestEntity = (match = (entity) => { return true }) => {
+    let best = null
+    let bestDistance = Number.MAX_VALUE
+    for (const entity of Object.values(bot.entities)) {
+      if (entity === bot.entity || !match(entity)) continue
+      const dist = bot.entity.position.distanceSquared(entity.position)
+      if (dist < bestDistance) {
+        best = entity
+        bestDistance = dist
+      }
+    }
+    return best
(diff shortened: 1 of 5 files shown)


lib/plugins/entities.js
```diff
@@ -65,6 +65,20 @@ function inject (bot, { version }) {
...
+    let bestDistance = Number.MAX_VALUE
+    for (const entity of Object.values(bot.entities)) {
+      if (entity === bot.entity || !match(entity)) continue
+      const dist = bot.entity.position.distanceSquared(entity.position)
+      if (dist < bestDistance) {
+        best = entity
+        bestDistance = dist
+      }
+    }
+    return best
```
wvffle, line 79: I'd only add a couple of newlines to make it more readable and add brackets to L72. Other than that, LGTM

Karang: done

rom1504: It's completely fine both ways but I don't find adding these empty lines more readable, more the opposite actually since it takes more room and requires more scrolling.

rom1504:
@Karang failing test is saying prismarine-physics is broken in some way

Karang:
> @Karang failing test is saying prismarine-physics is broken in some way

Looks like a missing null check, I'm fixing it.

lib/plugins/entities.js
Karang: I tend to write more compact code for the reasons rom cited. But mineflayer is also aimed at being read by beginners, so maybe it is more readable for them ? Idk, its up to you

PR #1250, opened 2020-08-15
Fix typo breaking findBlocks

Karang:
Thanks @Naomi-alt for finding the bug


Joeri-G:
Thank you for fixing this, I just spent an hour or two researching the workings of `bot.waitForChunksToLoad` because `findBlocks` was not returning anything.

PR #1256, opened 2020-08-17
Fix 'itemDrop'

ImHarvol:
The changes implemented in #1181 only work in 1.9, 1.11 and 1.12 I think, maybe 1.13.
In this PR I add mostly test for this event.

The reason that is not working in 1.14 and so on is because `entity.name` is 'arrow', which is why [I was confused](https://github.com/PrismarineJS/mineflayer/pull/1181#issuecomment-663882914)

Edit: What a mess I did with the commits! Sorry

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -7,7 +7,7 @@ assignees: ''
 
 ---
 
-[ ] The [FAQ](https://github.com/PrismarineJS/mineflayer/blob/master/docs/FAQ.md) doesn't contain a resolution to my issue 
+- [ ] The [FAQ](https://github.com/PrismarineJS/mineflayer/blob/master/docs/FAQ.md) doesn't contain a resolution to my issue 
 
 ## Versions
  - mineflayer: #.#.#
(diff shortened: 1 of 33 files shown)


rom1504:
looks like tests are not passing for some versions ;)

ImHarvol:
The implementation is simple, but tests are pretty messy, sorry

I decided to hardcode some versions https://github.com/PrismarineJS/mineflayer/blob/f1d8f41ced51e4fc4cec4904b2cceee5e00f72df/test/internalTest.js#L455
as the mcData is pretty messy and I wasn't getting the type correctly

Also notice this error (which isn't really causing problems): https://github.com/PrismarineJS/mineflayer/pull/1256/checks?check_run_id=1002948588#step:6:322
It seems to happen due to the change of the type of the properties `x`, `y`, `z` from [i32 in 1.8](https://minecraft-data.prismarine.js.org/?v=1.8&d=protocol#toClient_spawn_entity) to [f64 in 1.9](https://minecraft-data.prismarine.js.org/?v=1.9&d=protocol#toClient_spawn_entity)

wvffle (approved):

PR #1272, opened 2020-08-23
Update saver.js

Clickydude:
Changing of the semicolon to underscore in worldName to avoid forbidden directory name in Windows.

Avoids the following error:
`Error: ENOENT: no such file or directory, mkdir 'minecraft:overworld'`


rom1504 (changes requested):
Can you run `npm run fix` to fix the lint ?

rom1504 (approved):

PR #1320, opened 2020-09-17
More accurate blockInSight

Karang:
Use RayTracingIterator to find the block in sight. It checks all blocks on the ray exactly once. 

Replaced maxStep and vectorLength by maxDistance that is easier to reason about.

### DIFF
--- a/lib/iterators.js
+++ b/lib/iterators.js
@@ -1,4 +1,4 @@
-const Vec3 = require('vec3').Vec3
+const { Vec3 } = require('vec3')
 
 // 2D spiral iterator, useful to iterate on
 // columns that are centered on bot position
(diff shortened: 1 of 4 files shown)


wvffle (changes requested):
the code looks good though I'm not so sure about the change in the API. Rather than making a breaking change, I'd choose a new name for it like `blockAtCursor` or something like that

Karang:
> the code looks good though I'm not so sure about the change in the API. Rather than making a breaking change, I'd choose a new name for it like `blockAtCursor` or something like that

You're right, I'll add a new function and a deprecation warning on the old one.

lib/plugins/ray_trace.js
```diff
@@ -1,28 +1,35 @@
...
+    const snYaw = Math.sin(yaw)
+    return new Vec3(-snYaw * csPitch, snPitch, -csYaw * csPitch)
+  }
 
-    const x = -Math.sin(yaw) * Math.cos(pitch)
-    const y = Math.sin(pitch)
-    const z = -Math.cos(yaw) * Math.cos(pitch)
+  bot.blockInSight = (maxSteps = 256, vectorLength = 5 / 16) => {
+    console.warn('[deprecated] use bot.blockAtCursor instead')
+    bot.blockAtCursor(maxSteps * vectorLength)
```
wvffle, line 15: Missing return.

Also I'd change `null` to `undefined` to stay 1:1 with previous API.

lib/plugins/ray_trace.js
```diff
@@ -1,28 +1,35 @@
...
-    const step = new Vec3(x, y, z).scaled(vectorLength)
+  bot.blockAtCursor = (maxDistance = 256) => {
+    const { position, height, pitch, yaw } = bot.entity
 
-    for (let i = 0; i < maxSteps; ++i) {
-      cursor.add(step)
+    const eyePosition = position.offset(0, height, 0)
+    const viewDirection = getViewDirection(pitch, yaw)
 
-      // TODO: Check boundingBox of block
```
wvffle, line 19: I'd preserve this comment. Maybe in future we'll add it

lib/plugins/ray_trace.js
```diff
@@ -1,28 +1,35 @@
...
+    const csYaw = Math.cos(yaw)
+    const snYaw = Math.sin(yaw)
+    return new Vec3(-snYaw * csPitch, snPitch, -csYaw * csPitch)
+  }
 
-    const x = -Math.sin(yaw) * Math.cos(pitch)
-    const y = Math.sin(pitch)
-    const z = -Math.cos(yaw) * Math.cos(pitch)
+  bot.blockInSight = (maxSteps = 256, vectorLength = 5 / 16) => {
+    console.warn('[deprecated] use bot.blockAtCursor instead')
```
wvffle, line 14: Hmmm... I'm not a fan of `console.*` methods. It would break `mineflayer-dashboard` if user does not reassign it to the dashboard's log function ;/

Also, in my opinion we should call it only once.

test/externalTests/rayTrace.js
```diff
@@ -3,11 +3,11 @@ const assert = require('assert')
...
   const { position } = bot.entity
   bot.lookAt(position.offset(0, 3, 0), true, () => {
-    const block = bot.blockInSight()
-    assert.strictEqual(block, undefined)
+    const block = bot.blockAtCursor()
+    assert.strictEqual(block, null)
 
     bot.lookAt(position.offset(0, -3, 0), true, () => {
-      const block = bot.blockInSight()
+      const block = bot.blockAtCursor()
```
wvffle, line 10: Can we have the old test as it was and add a new one below? Just to be sure, that new `blockInSight()` logic stays 1:1 (as the `undefined` changes to `null`)

wvffle (changes requested):

lib/plugins/ray_trace.js
Karang: What do you suggest then? The goal is to really drop the old api as it makes no sense to keep it, other that to ensure a backward compatibility period. If it breaks some code, at least it will put a warning to tell the user what to do to fix it.

wvffle: I understand that. Leave it, I'll just push a patch that overwrites this function and logs into the dashboard log.

lib/plugins/ray_trace.js
Karang: done

test/externalTests/rayTrace.js
Karang: done

lib/plugins/ray_trace.js
Karang: done

wvffle (approved):

rom1504 (approved):

rom1504:
Lgtm

wvffle:
On a second thought, I'm not sure we should deprecate `blockInSight()`. Looking at the implementation of the iterator, it's late and I'm quite sleepy so correct me if I'm wrong, you're iterating by whole blocks.

Logic of `blockInSight()` was more flexible with checking bounding boxes (someday). Maybe it's good to leave the choice for the user to either use `blockAtCursor()` for faster search and `blockInSight()` for better result in case checking bounding boxes is implemented?

We could also add a second parameter to `blockAtCursor()` that tells whether user wants fast or precise search.

wvffle:
LGTM

lib/iterators.js
```diff
@@ -99,29 +99,29 @@ const BlockFace = {
   EAST: 5
 }
 
-// This iterate along a ray starting at (x,y,z) in (dx,dy,dz) direction
+// This iterate along a ray starting at `pos` in `dir` direction
 // It steps exactly 1 block at a time, returning the block coordinates
 // and the face by which the ray entered the block.
 class RaycastIterator {
-  constructor (x, y, z, dx, dy, dz, maxDistance) {
+  constructor (pos, dir, maxDistance) {
```
TheDudeFromCI, line 106: Could this be a breaking change for existing projects that use the old constructor? This class is a public class. Might check the parameter types for backward compatibility.

TheDudeFromCI (changes requested):

lib/iterators.js
Karang: it was undocumented and added recently, I don't think its used anywhere

TheDudeFromCI: Oh, I thought these iterators were all publicly exported for plugin use. Apparently, they aren't okay then, nevermind. xD

TheDudeFromCI (approved):

Issue #1325, opened 2020-09-20
Fix bot.setCommandBlock(...)


ImHarvol:
- [x] The [FAQ](https://github.com/PrismarineJS/mineflayer/blob/master/docs/FAQ.md) doesn't contain a resolution to my issue 

## Versions
 - mineflayer: 1.9-1.X
 - server: vanilla
 - node: v12.18.0

## Detailed description of a problem
I was working on #973 I was using the function [bot.setCommandBlock(pos, command, track_output)](https://github.com/PrismarineJS/mineflayer/blob/master/docs/api.md#botsetcommandblockpos-command-track_output) which is defined here: https://github.com/PrismarineJS/mineflayer/blob/3da7fbb6cd53b758fd418748bb32c310c74b9176/lib/plugins/command_block.js#L6
And with the code below, it seems to work in 1.8 but not in the following versions (I've tested it in 1.9 and 1.12)
It would be understandable if it didn't work, as it hasn't been updated since its creation.

## Your current code
```js
    bot.setCommandBlock(vec3(1, 5, 1), '/summon Villager ~ ~1 ~ {Offers:{Recipes:[0:{maxUses:5,buy:{id:"minecraft:emerald",Count:1},sell:{id:"minecraft:bread",Count:1},uses:0}]}}', false) // bot.setCommandBlock(pos, command, track_output)
```

## Expected behavior
I see the command block before and after the bot calls that function, and I expect the comand to be set in the command block after the function call

## Additional context
I noticed at least 2 possible breaking changes:
- From 1.8 to 1.9 some options were added to the command block
- In 1.13-pre4 (see https://wiki.vg/Protocol_History) the way you update the command block changed from a plugin channel to a standard packet

ImHarvol:
Okay I just noticed this:
> In 1.8 and earlier the name of this channel was incorrectly spelled as MC|AdvCdm. This was fixed in snapshot 15w34a.
https://wiki.vg/index.php?title=Plugin_channels&diff=13401&oldid=13314#MC.7CAdvCmd

So if we take that in count, it'll probably work from 1.8 to 1.12

PR #1327, opened 2020-09-20
Fix bot.setCommandBlock() for versions 1.13-1.16 and add tests

ImHarvol:
Related to #1325

I dont know why `standard --fix` did such a mess, sorry

### DIFF
--- a/lib/plugins/command_block.js
+++ b/lib/plugins/command_block.js
@@ -5,9 +5,10 @@ const ProtoDef = require('protodef').ProtoDef
 module.exports = inject
 
 function inject (bot) {
-  function setCommandBlock (pos, command, trackOutput, flags) {
+  function setCommandBlock (pos, command, trackOutput, mode) {
     assert.notStrictEqual(pos, null)
     assert.notStrictEqual(command, null)
+    trackOutput = trackOutput || 2
     if (bot.supportFeature('usesAdvCmd') || bot.supportFeature('usesAdvCdm')) {
       const pluginChannelName = bot.supportFeature('usesAdvCdm') ? 'MC|AdvCdm' : 'MC|AdvCmd'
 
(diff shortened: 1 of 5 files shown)


rom1504:
not passing the tests

wvffle (approved):

lib/plugins/command_block.js
```diff
@@ -1,101 +1,114 @@
...
 const ProtoDef = require('protodef').ProtoDef
 
 module.exports = inject
 
 function inject (bot) {
-  function setCommandBlock (pos, command, trackOutput) {
-    const pluginChannelName = bot.supportFeature('usesAdvCdm') ? 'MC|AdvCdm' : 'MC|AdvCmd'
-
-    const proto = new ProtoDef()
+  function setCommandBlock (pos, command, trackOutput, flags) {
```
rom1504, line 8: Can you update the doc with this flags parameter please ?

rom1504 (changes requested):

lib/plugins/command_block.js
ImHarvol: The thing is that I'm not sure which parameters it accepts and what they do
https://wiki.vg/Protocol#Update_Command_Block

rom1504: ah yes sorry I forgot this pr
so basically if you don't know what it does, probably you should expose it in the API
mineflayer should abstract away as much as possible the protocol details, so either remove that flags from the exposed API, either clarify in the doc what values it take and make sure that's stable with versions

rom1504:
left a comment above about mode what is it ?

rom1504 (approved):

rom1504:
ok let's merge for now, but no guaranty of stability on that mode param, might go at any time

Issue #1143, opened 2020-07-12
Add language specific directories in docs/


wvffle:
We should create `docs/en/README.md`, `docs/ru/README.md` and so on. 

Could be implemented in #1142

Leftier:
We could try using https://gitlocalize.com/
It allows to select a specific directory to only translate certain files and allows you to track the progress of each one of them
Also automatically generates a new folder with the iso language code

TheDudeFromCI:
This sounds like a good way to organize the translations and would act as a good indicator for knowing if translations are outdated or not when making PRs.

rom1504:
only problem with that is AFAIK github only check readme in docs/README or README
is there a solution for this ?

Leftier:
I created a repository to test gitlocalize I added rom1504 as moderator but couldn't add anyone else as their usernames don't appear in the search bar

After translating a file I noticed that the folders containing the translated files are actually ignored so there is no need to move the readme just leave it as it is 

Repo: https://github.com/Leftier/gitlocalize-test
GitLocalize: https://gitlocalize.com/repo/5006

PR #1342, opened 2020-10-03
Update README_ES.md

ImHarvol:


rom1504 (approved):
thanks, I think we should setup a more automated process to update those
maybe something mentioned in https://github.com/PrismarineJS/mineflayer/issues/1143

PR #1343, opened 2020-10-03
Simplify setCommandBlock arguments

ImHarvol:
- Simplify setCommandBlock arguments by replacing trackOutput and mode arguments to a options argument
- options argument is more friendly as now the user doesn't have to guess what number does it need to replicate the behavior
- Also adds checks: the bot shouldn't be able to open/edit the command block when it's in survival mode, and the block at `pos` should be a command block

`flags` values and the command block properties output:
 - 0:  Unconditional - Needs Redstone - Tracks Output
 - 1:  Unconditional - Needs Redstone - Doesn't track output
 - 2:  Conditional   - Needs Redstone - Tracks Output
 - 3:  Conditional   - Needs Redstone - Doesn't track output
 - 4:  Unconditional - Always Active  - Tracks Output
 - 5:  Unconditional - Always Active  - Doesn't track output
 - 6:  Conditional   - Always Active  - Tracks Output
 - 7:  Conditional   - Always Active  - Doesn't track output

Related to #1327 where I wasn't sure what `flags` was.

There are still things missing but it's pretty complete for the latests minecraft versions

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1621,9 +1621,20 @@ Returns an `Villager` instance which represents the trading window you are openi
 
 Uses the open `villagerInstance` to trade.
 
-#### bot.setCommandBlock(pos, command, track_output, [mode])
+#### bot.setCommandBlock(pos, command, [options])
 
-Set a command block `command` at `pos`.
+Set a command block's properties at `pos`.
+Example `options` argument:
+```js
+{
+  mode: 2,
+  trackOutput: true,
+  conditional: false,
+  alwaysActive: true
+}
+```
+options.mode can have 3 values: 0 (SEQUENCE), 1 (AUTO), 2 (REDSTONE)
+All options attributes are false by default, except mode which is 2 (as to replicate the default command block in Minecraft).
 
 ### Lower level inventory methods
 
--- a/lib/plugins/command_block.js
+++ b/lib/plugins/command_block.js
@@ -17,9 +17,9 @@ function inject (bot) {
     options.mode = options.mode || 2 // Possible values: 0: SEQUENCE, 1: AUTO and 2: REDSTONE
 
     let flags = 0
-    if (!options.trackOutput) flags += 1
-    if (options.conditional) flags += 2
-    if (options.alwaysActive) flags += 4
+    flags |= +options.trackOutput << 0 // 0x01
+    flags |= +options.conditional << 1 // 0x02
+    flags |= +options.alwaysActive << 2 // 0x04
 
     if (bot.supportFeature('usesAdvCmd') || bot.supportFeature('usesAdvCdm')) {
       const pluginChannelName = bot.supportFeature('usesAdvCdm') ? 'MC|AdvCdm' : 'MC|AdvCmd'
--- a/lib/plugins/command_block.js
+++ b/lib/plugins/command_block.js
@@ -1,14 +1,26 @@
 const assert = require('assert')
-
 const ProtoDef = require('protodef').ProtoDef
 
 module.exports = inject
 
 function inject (bot) {
-  function setCommandBlock (pos, command, trackOutput, mode) {
+  function setCommandBlock (pos, command, options = {}) {
+    assert.strictEqual(bot.player.gamemode, 1, new Error('The bot has to be in creative mode to open the command block window'))
     assert.notStrictEqual(pos, null)
     assert.notStrictEqual(command, null)
-    trackOutput = trackOutput || 2
+    assert.strictEqual(bot.blockAt(pos).name, 'command_block', new Error("The block isn't a command block"))
+
+    // Default values when a command block is placed in vanilla minecraft
+    options.trackOutput = options.trackOutput || false
+    options.conditional = options.conditional || false
+    options.alwaysActive = options.alwaysActive || false
+    options.mode = options.mode || 2 // Possible values: 0: SEQUENCE, 1: AUTO and 2: REDSTONE
+
+    let flags = 0
+    if (!options.trackOutput) flags += 1
+    if (options.conditional) flags += 2
+    if (options.alwaysActive) flags += 4
+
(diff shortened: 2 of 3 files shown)


lib/plugins/command_block.js
```diff
@@ -1,14 +1,26 @@
...
+    // Default values when a command block is placed in vanilla minecraft
+    options.trackOutput = options.trackOutput || false
+    options.conditional = options.conditional || false
+    options.alwaysActive = options.alwaysActive || false
+    options.mode = options.mode || 2 // Possible values: 0: SEQUENCE, 1: AUTO and 2: REDSTONE
+
+    let flags = 0
+    if (!options.trackOutput) flags += 1
+    if (options.conditional) flags += 2
+    if (options.alwaysActive) flags += 4
```
ImHarvol, line 22: Comments on this would be appreciated

docs/api.md
```diff
@@ -1621,9 +1621,20 @@ Returns an `Villager` instance which represents the trading window you are openi
 
 Uses the open `villagerInstance` to trade.
 
-#### bot.setCommandBlock(pos, command, track_output, [mode])
+#### bot.setCommandBlock(pos, command, [options])
```
wvffle, line 1624: Changing the API results with a breaking change. You can leave the previous arguments as another docs entry. And mark is as deprecated. Then just check if track_output is a boolean or an object and handle it correctly.

lib/plugins/command_block.js
wvffle: Add flags:

```js
const FLAG_TRACK_OUTPUT = 1 << 0
const FLAG_CONDITIONAL = 1 << 1
const FLAG_ALWAYS_ACTIVE = 1 << 2
```
Just a suggestion how you could use them:
```js
// ...
const { trackOutput = false, conditional = false, alwaysActive = false } = options
const flags = (+trackOutput * FLAG_TRACK_OUTPUT) 
  | (+conditional * FLAG_CONDITIONAL) 
  | (+alwaysActive * FLAG_ALWAYS_ACTIVE)

```
or you can:
```js
let flags = 0
if (trackOutput) flags |= FLAG_TRACK_OUTPUT
```

lib/plugins/command_block.js
```diff
@@ -1,14 +1,26 @@
...
     assert.notStrictEqual(pos, null)
     assert.notStrictEqual(command, null)
-    trackOutput = trackOutput || 2
+    assert.strictEqual(bot.blockAt(pos).name, 'command_block', new Error("The block isn't a command block"))
+
+    // Default values when a command block is placed in vanilla minecraft
+    options.trackOutput = options.trackOutput || false
+    options.conditional = options.conditional || false
+    options.alwaysActive = options.alwaysActive || false
+    options.mode = options.mode || 2 // Possible values: 0: SEQUENCE, 1: AUTO and 2: REDSTONE
```
wvffle, line 17:
```js
const MODE_SEQUENCE = 0
const MODE_AUTO = 1 << 1
const MODE_REDSTONE = 1 << 2
```
And export the modes

docs/api.md
wvffle: Though I'd just add and export the flags I described in some comments below.
And let users do:
```js
const { TRACK_OUTPUT, ALWAYS_ACTIVE } = bot.commandBlockFlags
const { MODE_REDSTONE } = bot.commandBlockModes
bot.setCommandBlock(pos, command, TRACK_OUTPUT | ALWAYS_ACTIVE, MODE_REDSTONE)
```

wvffle (changes requested):
All of the comments are suggestions, though I request changes due to the breaking change possibility

docs/api.md
rom1504: I said while merging the previous pr that this mode parameter is basically not part of the API and it will get changed, so it's ok to "break" it here

lib/plugins/command_block.js
ImHarvol: Thanks! That's what I was thinking of but didn't know how to do

docs/api.md
ImHarvol: I don't understand that the mode parameter is not part of the API, I don't get it.

Another reason that supports the breaking changes is that bot.setCommandBlock() has been only working in 1.8 since its implementation untill a week ago or so, so certainly not many (if any) people are using it.

rom1504: Not sure what you mean. I said in the previous PR that it was not part of the API so it's not, that's all. That's what makes it ok to remove/change it now.

lib/plugins/command_block.js
```diff
@@ -1,14 +1,26 @@
...
+
+    let flags = 0
+    flags |= +options.trackOutput << 0 // 0x01
+    flags |= +options.conditional << 1 // 0x02
+    flags |= +options.alwaysActive << 2 // 0x03
```
wvffle, line 22: 1<<2 = 4
0x03 = 3

wvffle: If you want to display it as bit values then use `0b0100`

test/externalTests/trade.js
```diff
@@ -51,13 +57,14 @@ module.exports = () => (bot, done) => {
...
 
-  bot.test.setInventorySlot(36, new Item(mcData.itemsByName.emerald.id, 64, 0), (err) => {
+  bot.test.setInventorySlot(36, new Item(mcData.itemsByName.emerald.id, 64, 0), async (err) => {
     assert.ifError(err)
     bot.on('entitySpawn', onEntitySpawn)
 
     // A command block is needed to spawn the villager due to the chat's character limit in some versions
     bot.test.sayEverywhere(`/setblock ${commandBlockPos.toArray().join(' ')} command_block`)
-    bot.setCommandBlock(commandBlockPos, summonCommand, 1, 2)
+    await sleep(500)
```
rom1504, line 66: don't mix callback and async/await please

rom1504 (changes requested):

rom1504 (approved):

PR #1345, opened 2020-10-05
Exposed getDestSlot function

TheDudeFromCI:
This utility is useful for converting named inventory slots into raw slot IDs. I believe it would be useful to add this as a future-proof way to safely access raw inventory slots.

```js
const helmet = bot.inventory.slots[bot.getEquipmentDestSlot('helmet')]
const hand = bot.inventory.slots[bot.getEquipmentDestSlot('hand')]
const offHand = bot.inventory.slots[bot.getEquipmentDestSlot('off-hand')]
// etc
```
I believe it would be better to also keep the return type an integer value as it works smoother with the other low-level inventory functions like moving items to specific slots and such.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -278,6 +278,7 @@
       - [bot.openEntity(entity, Class)](#botopenentityentity-class)
       - [bot.moveSlotItem(sourceSlot, destSlot, cb)](#botmoveslotitemsourceslot-destslot-cb)
       - [bot.updateHeldItem()](#botupdatehelditem)
+      - [bot.getEquipmentDestSlot(destination)](#botgetequipmentdestslot-destination)
     - [bot.creative](#botcreative)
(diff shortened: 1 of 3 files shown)


docs/api.md
```diff
@@ -1677,6 +1677,10 @@ Move an item from `sourceSlot` to `destSlot` in the current window.
 
 Update `bot.heldItem`.
 
+#### bot.getEquipmentDestSlot(destination)
+
+Gets the inventory equipment slot id for the given equipment destination name.
```
wvffle, line 1682: I'd add destination values there

wvffle (approved):

docs/api.md
TheDudeFromCI: Done

rom1504 (approved):

PR #1351, opened 2020-10-07
Equations for do a perfect shot with a bow

sefirosweb:
Hello!, I made a package for calculate the yaw and pitch for a bow attack, it also detects movement to anticipate :D

I tried pack to simply files for easy read and use, you can see all https://github.com/sefirosweb/minecraftHawkEye

### DIFF
--- a/examples/perfectShotBow.js
+++ b/examples/perfectShotBow.js
@@ -1,0 +1,30 @@
+const mineflayer = require('mineflayer');
+const { hawkEye, getPlayer } = require('minecrafthawkeye');
+
+const bot = mineflayer.createBot({
+    host: process.argv[2],
+    port: parseInt(process.argv[3]),
+    username: process.argv[4] ? process.argv[4] : 'Archer',
+    password: process.argv[5]
+})
+
+hawkEye.load(bot);
+
+bot.on('spawn', function() {
+    bot.chat('/give Archer bow{Enchantments:[{id:unbreaking,lvl:100}]} 1');
+    bot.chat('/give Archer minecraft:arrow 300');
+    bot.chat('/time set day');
+    bot.chat('/kill @e[type=minecraft:arrow]');
+
+    bot.chat('Ready!');
+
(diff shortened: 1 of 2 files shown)


rom1504:
can you run npm run fix to fix the linting for your example ?

rom1504:
and would be great to do this https://github.com/sefirosweb/minecraftHawkEye/issues/1
(you can check any of the plugins listed in the readme as other examples)

examples/perfectShotBow.js
```diff
@@ -0,0 +1,30 @@
...
+    bot.chat('/give Archer bow{Enchantments:[{id:unbreaking,lvl:100}]} 1');
+    bot.chat('/give Archer minecraft:arrow 300');
+    bot.chat('/time set day');
+    bot.chat('/kill @e[type=minecraft:arrow]');
+
+    bot.chat('Ready!');
+
+    let target = getPlayer(bot);
+    if (!target)
+        return false;
```
wvffle, line 23: Add `{` and `}`

wvffle (changes requested):

examples/perfectShotBow.js
sefirosweb: OK! i try to fix this =)

sefirosweb:
Hi guys I done a update to 1.0.5

Now works has plugin,

Also I fixed some bugs, and now the precision to the target is better :D

sefirosweb:
> Looks good to me.
> Just run `npx standard --fix` inside the example code to fix the linting errors so the build will pass.
> After that, it should be good to go. 👍

Done! you can do double check if fine?

rom1504:
you need to push after running it ;)

sefirosweb:
> you need to push after running it ;)

sorry im stupid =P

Pushed,

Also today had fixed a minor bugs (the height targets, the spiders is too small =P )

TheDudeFromCI (approved):

rom1504 (approved):

PR #1354, opened 2020-10-08
Update action versions

imharvol:
This removes some warnings that were appearing in CI
More info: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/


ImHarvol:
**Before:**
Before

---

**After:**
After

wvffle (approved):

rom1504 (approved):

rom1504:
ok but then we also have to update all other prismarine repo...

PR #1357, opened 2020-10-08
Add chat position to message event

### LINKED ISSUE #1312: message event doesn't contain the position field

## Versions
 - mineflayer: 2.28.1
 - node: v13.3.0

The default “Minecraft ChatMessagePacket“ has three position values.
```
CHAT = 0 # A player-initiated chat message.
SYSTEM = 1 # The result of running a command.
GAME_INFO = 2 # Displayed above the hotbar in vanilla clients.
```
At the moment the “message” event will trigger on all three positions.
And the “jsonMsg” Object has no type or position identifier to identify a specific message type.

## What did you try yet?
I have tried to find an identifier in the “jsonMsg” Object to identify the massage position

## Your current code
```js

bot.on('message', function (jsonMsg) {
    console.log(jsonMsg)
    }
})


```

## Expected behavior
A possibility to identify the position of a message to exclude “GAME_INFO” messages from this event.

## Additional context
All “jsonMsg” Objects have the same format it is not possible to show differences only with the row json body.

larspapen:
This will add a parameter to check from which position the chat message event came from.
Updated english and russian documentation.

Closes #1312


TheDudeFromCI (approved):

wvffle (approved):

docs/api.md
```diff
@@ -998,12 +998,14 @@ Emitted for every server message which appears on the Action Bar.
...
  * `jsonMsg` - unmodified JSON message from the server
 
-#### "message" (jsonMsg)
+#### "message" (jsonMsg, position)
 
 Emitted for every server message, including chats.
 
  * `jsonMsg` - unmodified JSON message from the server
 
+ * `positon` - position of Chat message
```
rom1504, line 1007: can you add here what is the meaning of this value?
Plus a warning that it isn't available in versions before 1.16?

rom1504 (changes requested):

docs/api.md
larspapen: image

https://wiki.vg/Chat

Do you have any suggestions for the meaning of this value?

rom1504: no, that's why I think if it is exposed it should be well documented
Most likely we don't want to expose it directly but add some abstraction on top so it doesn't break in the next minecraft release.

docs/api.md
```diff
@@ -998,12 +998,17 @@ Emitted for every server message which appears on the Action Bar.
...
  * `jsonMsg` - unmodified JSON message from the server
 
-#### "message" (jsonMsg)
+#### "message" (jsonMsg, position)
 
 Emitted for every server message, including chats.
 
  * `jsonMsg` - unmodified JSON message from the server
 
+ * `positon` - (>= 1.8.1): position of Chat message can be
```
mdashlw, line 1007:
```suggestion
 * `position` - (>= 1.8.1): position of Chat message can be
```

rom1504 (approved):

PR #1358, opened 2020-10-09
Added hasAttackCooldown Feature

TheDudeFromCI:
Since versions 1.9+ use the attack cooldown system, but not 1.8, it's useful to check for this feature at runtime to know how to correctly handle attack speeds.


rom1504 (changes requested):
yes but where are you using it ?

rom1504 (approved):
ok, we should add support feature function to API.md probably

PR #1363, opened 2020-10-09
Added supportFeature to API.md

TheDudeFromCI:
Requested in https://github.com/PrismarineJS/mineflayer/pull/1358

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -268,6 +268,7 @@
       - [bot.openVillager(villagerEntity)](#botopenvillagervillagerentity)
       - [bot.trade(villagerInstance, tradeIndex, [times], [cb])](#bottradevillagerinstance-tradeindex-times-cb)
       - [bot.setCommandBlock(pos, command, track_output, [mode])](#botsetcommandblockpos-command-track_output-mode)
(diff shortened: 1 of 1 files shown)


wvffle (approved):

docs/api.md
```diff
@@ -1637,6 +1638,55 @@ Example `options` argument:
 options.mode can have 3 values: 0 (SEQUENCE), 1 (AUTO), 2 (REDSTONE)
 All options attributes are false by default, except mode which is 2 (as to replicate the default command block in Minecraft).
 
+#### bot.supportFeature(name)
+
+This can be used to check is a specific feature is available in the current Minecraft version. This is usually only required for handling version-specific functionality.
+
+Available Features:
```
rom1504, line 1645: I think it's best if we put a link to the json file here and don't copy paste its content in the doc (that would increase maintenance chores too much)

rom1504 (changes requested):

docs/api.md
TheDudeFromCI: Fixed

rom1504 (approved):

PR #1370, opened 2020-10-12
Add ability to call openChest on shulker boxes

efunneko:
With this PR, placed shulkers boxes can be opened as if they are chests

### DIFF
--- a/lib/plugins/chest.js
+++ b/lib/plugins/chest.js
@@ -12,7 +12,9 @@ function inject (bot, { version }) {
       assert.ok(chestToOpen.type === mcData.blocksByName.chest.id ||
                 chestToOpen.type === mcData.blocksByName.ender_chest.id ||
                 chestToOpen.type === mcData.blocksByName.trapped_chest.id ||
-                (mcData.blocksByName.barrel && chestToOpen.type === mcData.blocksByName.barrel.id))
+                (mcData.blocksByName.barrel && chestToOpen.type === mcData.blocksByName.barrel.id) ||
+                (mcData.blocks[chestToOpen.type].name.endsWith('_shulker_box')))
+
       chest = bot.openBlock(chestToOpen, Chest)
     } else if (chestToOpen.constructor.name === 'Entity') {
       assert.strictEqual(chestToOpen.entityType, mcData.entitiesByName.chest_minecart.id)
--- a/lib/plugins/chest.js
+++ b/lib/plugins/chest.js
@@ -13,7 +13,7 @@ function inject (bot, { version }) {
                 chestToOpen.type === mcData.blocksByName.ender_chest.id ||
                 chestToOpen.type === mcData.blocksByName.trapped_chest.id ||
                 (mcData.blocksByName.barrel && chestToOpen.type === mcData.blocksByName.barrel.id) ||
-                mcData.blocks[chestToOpen.type].name.endsWith('_shulker_box'))
+                mcData.blocks[chestToOpen.type].name.endsWith('shulker_box'))
 
       chest = bot.openBlock(chestToOpen, Chest)
     } else if (chestToOpen.constructor.name === 'Entity') {
--- a/lib/plugins/chest.js
+++ b/lib/plugins/chest.js
@@ -14,7 +14,6 @@ function inject (bot, { version }) {
                 chestToOpen.type === mcData.blocksByName.trapped_chest.id ||
                 (mcData.blocksByName.barrel && chestToOpen.type === mcData.blocksByName.barrel.id) ||
                 mcData.blocks[chestToOpen.type].name.endsWith('shulker_box'))
-
       chest = bot.openBlock(chestToOpen, Chest)
     } else if (chestToOpen.constructor.name === 'Entity') {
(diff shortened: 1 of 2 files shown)


TheDudeFromCI (approved):

lib/plugins/chest.js
```diff
@@ -12,7 +12,9 @@ function inject (bot, { version }) {
       assert.ok(chestToOpen.type === mcData.blocksByName.chest.id ||
                 chestToOpen.type === mcData.blocksByName.ender_chest.id ||
                 chestToOpen.type === mcData.blocksByName.trapped_chest.id ||
-                (mcData.blocksByName.barrel && chestToOpen.type === mcData.blocksByName.barrel.id))
+                (mcData.blocksByName.barrel && chestToOpen.type === mcData.blocksByName.barrel.id) ||
+                (mcData.blocks[chestToOpen.type].name.endsWith('_shulker_box')))
```
wvffle, line 16: there's no point in having the parentheses here.

wvffle (changes requested):

lib/plugins/chest.js
efunneko: You are correct. I have removed them

lib/plugins/chest.js
```diff
@@ -12,7 +12,9 @@ function inject (bot, { version }) {
       assert.ok(chestToOpen.type === mcData.blocksByName.chest.id ||
                 chestToOpen.type === mcData.blocksByName.ender_chest.id ||
                 chestToOpen.type === mcData.blocksByName.trapped_chest.id ||
-                (mcData.blocksByName.barrel && chestToOpen.type === mcData.blocksByName.barrel.id))
+                (mcData.blocksByName.barrel && chestToOpen.type === mcData.blocksByName.barrel.id) ||
+                mcData.blocks[chestToOpen.type].name.endsWith('_shulker_box'))
```
imharvol, line 16: If I'm not wrong, it should do something similar as with the barrel (check that the barrel does exist in mcData.blocksByName), otherwise in versions where the shulker box doesn't exist, it would throw saying that undefined.name doesn't exist.

imharvol (changes requested):

lib/plugins/chest.js
efunneko: I did actually have this there in an earlier version, but someone beat me to a PR, which omitted it. I followed their lead and got rid of it in my version, since I figured they knew better. Shulker boxes were added in version 1.1 (about 3 years ago). Does mineflayer work for versions older than that? If so, then it should be fixed.

imharvol: Yes, it suports versions since 1.8

TheDudeFromCI: I think the above implementation indirectly addresses that, as the block name cannot end with "_shulker_box" unless shulker boxes exist. So this text should work correctly in all older versions.

imharvol: I think you're right, sorry!
I got confused thinking that we were always looking for the id of the shulker box

wvffle: then I mark this comment as resolved

imharvol, line 16: Probably should be `'shulker_box'` instead of `'_shulker_box'`, as since 1.13 there's a `'shulker_box'` this can be seen in https://minecraft-data.prismarine.js.org/?v=1.13&d=items

I made some checks for this: https://github.com/imharvol/mineflayer/tree/shulker_tests

imharvol (changes requested):

lib/plugins/chest.js
TheDudeFromCI: Block IDs are separate from Item IDs as of 1.13. So the `mcData.blocks` will never contain that item at least in 1.13+. Does this occur in 1.11 or 1.12?

TheDudeFromCI: Update, even if it does exist in 1.12 and 1.11 with the same ID, it'll still be in the item for and not the block form. So `block.type` will still never return value value, even in those versions. So even then, it's safe.

imharvol: The link is a bit misleading, I meant blocks not items. See for example https://minecraft-data.prismarine.js.org/?v=1.13&d=blocks, there are the shulkers with color, but there's one with no color (`shulker_box`) and thus no underscore, so bot.openChest() wouldn't think it's a chest
I think it's easier to see it in the tests. If you run them as it is right now, it fails in versions since 1.13, if you remove the underscore (`shulker_box`) then it passess

TheDudeFromCI: Ah, okay. I see what you mean. Yeah, the underscore should be removed then.

efunneko: The underscore has been sent to the void.

imharvol (approved):
Passes tests in https://github.com/imharvol/mineflayer/tree/shulker_tests :+1:

rom1504 (approved):

rom1504:
@imharvol can you open your pr here ?

imharvol:
Yeah, I was thinking of that, I'll work on them more and then PR them. As of now I think they are too simple

PR #1371, opened 2020-10-12
Add bot.entityInSight()

goncharovchik:
bot.entityInSight() outputs entity in sight

### DIFF
--- a/lib/plugins/ray_trace.js
+++ b/lib/plugins/ray_trace.js
@@ -40,7 +40,7 @@ function inject (bot) {
         if (entity === bot.entity) {
           continue
         }
-        const dist = entity.position.offset(0, entity.height/2, 0).distanceSquared(cursor)
+        const dist = entity.position.offset(0, entity.height / 2, 0).distanceSquared(cursor)
         if ((dist <= 0.5) && (cursor.distanceTo(bot.entity.position.offset(0.5, 0.5, 0.5)) <= 5)) {
           const sightEntity = entity
           return sightEntity
--- a/lib/plugins/ray_trace.js
+++ b/lib/plugins/ray_trace.js
@@ -26,7 +26,6 @@ function inject (bot) {
   const rayTraceEntity = (maxSteps = 256, vectorLength = 5 / 16) => {
     const { height, position, yaw, pitch } = bot.entity
     const cursor = position.offset(0, height, 0)
-    
     const x = -Math.sin(yaw) * Math.cos(pitch)
     const y = Math.sin(pitch)
     const z = -Math.cos(yaw) * Math.cos(pitch)
--- a/lib/plugins/ray_trace.js
+++ b/lib/plugins/ray_trace.js
@@ -23,7 +23,7 @@ function inject (bot) {
       }
     }
   }
-  const rayTraceEntity = (maxSteps = 256,vectorLength = 5 / 16) => {
+  const rayTraceEntity = (maxSteps = 256, vectorLength = 5 / 16) => {
     const { height, position, yaw, pitch } = bot.entity
     const cursor = position.offset(0, height, 0)
     
@@ -42,7 +42,7 @@ function inject (bot) {
           continue
         }
         const dist = entity.position.distanceSquared(cursor)
-        if ((dist <= 0.5) && (cursor.distanceTo(bot.entity.position.offset(0.5,0.5,0.5)) <= 5)) {
+        if ((dist <= 0.5) && (cursor.distanceTo(bot.entity.position.offset(0.5, 0.5, 0.5)) <= 5)) {
           const sightEntity = entity
           return sightEntity
         }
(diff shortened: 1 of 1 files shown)


goncharovchik:
oh, I'm blunt. Not added to documentation

lib/plugins/ray_trace.js
```diff
@@ -23,6 +23,32 @@ function inject (bot) {
...
+    for (let i = 0; i < maxSteps; ++i) {
+      cursor.add(step)
+
+      // Checking for entity
+      for (const entity of Object.values(bot.entities)) {
+        if (entity === bot.entity) {
+          continue
+        }
+        const dist = entity.position.distanceSquared(cursor)
+        if ((dist <= 0.5) && (cursor.distanceTo(bot.entity.position.offset(0.5, 0.5, 0.5)) <= 5)) {
```
wvffle, line 44: arent you comparing to the entity foot position?

TheDudeFromCI: Might get a better approximation by using `entity.position.offset(0, entity.height/2, 0).distanceSquared(cursor)`
This equation assumes that all entities are a sphere with a radius of 0.7m, though, which is pretty inaccurate already.

TheDudeFromCI (changes requested):
Test against entity bounding box instead of distance from entity position.

goncharovchik:
> Test against entity bounding box instead of distance from entity position.

Thanks, fixed

goncharovchik:
> Using `distance` and `distanceSquared` will treat all entities like spheres. In Minecraft, entities are represented as axis-aligned bounding boxes.
> 
> The bounding box of an entity can be tested using:
> 
> ```js
> function isPointInEntity (cursor, entity) {
>   const entityPos = entity.position.offset(-entity.width / 2, 0, -entity.width / 2)
>   return cursor.x >= entityPos.x && cursor.x < entityPos.x + entity.width &&
>          cursor.y >= entityPos.y && cursor.y < entityPos.y + entity.height &&
>          cursor.z >= entityPos.z && cursor.z < entityPos.z + entity.width
> }
> ```

hmmm, doesn't work

goncharovchik:
Will there be mergers?

wvffle:
@TheDudeFromCI described you how to check if a point is inside an entity. Now simply call this function for every step the crusor takes. If the function returns true then return the entity.

With these changes we'll accept them and merge.

goncharovchik:
> @TheDudeFromCI described you how to check if a point is inside an entity. Now simply call this function for every step the crusor takes. If the function returns true then return the entity.
> 
> With these changes we'll accept them and merge.

But this function does not output anything, although the bot looks at the armor stand

goncharovchik:
> > > @TheDudeFromCI described you how to check if a point is inside an entity. Now simply call this function for every step the crusor takes. If the function returns true then return the entity.
> > > With these changes we'll accept them and merge.
> > 
> > 
> > But this function does not output anything, although the bot looks at the armor stand
> 
> Replace
> 
> ```js
> const dist = entity.position.distanceSquared(cursor)
> if ((dist <= 0.5) && (cursor.distanceTo(bot.entity.position.offset(0.5, 0.5, 0.5)) <= 5)) {
> ```
> 
> with
> 
> ```js
> if (isPointInEntity(cursor, entity)) {
> ```

изображение
I made this condition output to the console "Matched", but as you can see it does not output anything

rom1504:
seems you'd need to debug it then

goncharovchik:
> seems you'd need to debug it then

how exactly?

rom1504:
add console.log for the variables, check why it doesn't work and fix it

goncharovchik:
> add console.log for the variables, check why it doesn't work and fix it

I noticed that entity.width = undefined. That's why this function doesn't work. I don't know how to fix it

rom1504:
can you rebase on master and try again ?

goncharovchik:
> can you rebase on master and try again ?

I tried it, and this function consistently displays undefined, after debugging I did not find any reason. I don't even know what the problem is

goncharovchik:
> > > can you rebase on master and try again ?
> > 
> > I tried it, and this function consistently displays undefined, after debugging I did not find any reason. I don't even know what the problem is
> 
> Rebase your PR to the latest version of Mineflayer. The bug was patched.
> 
> Also make sure to update all node modules to the latest, if you haven't already.

I updated but same story happens? Can you try to run this code on your computer?

lib/plugins/ray_trace.js
```diff
@@ -23,6 +23,32 @@ function inject (bot) {
...
+
+      // Checking for entity
+      for (const entity of Object.values(bot.entities)) {
+        if (entity === bot.entity) {
+          continue
+        }
+        const dist = entity.position.offset(0, entity.height / 2, 0).distanceSquared(cursor)
+        if ((dist <= 0.5) && (cursor.distanceTo(bot.entity.position.offset(0.5, 0.5, 0.5)) <= 5)) {
+          const sightEntity = entity
+          return sightEntity
```
NotSugden, line 46:
```suggestion
          return entity
```
whats the use in a variable here if its only going to be returned instantly

rom1504:
stale, closing, feel free to re-open if you want to work on this

Issue #1374, opened 2020-10-14
Added better warnings for unsupported Minecraft versions


TheDudeFromCI:
## Is your feature request related to a problem? Please describe.
There are a lot of issues and confusion revolving around the supported versions of Minecraft. Currently, the major version of Minecraft is properly tested on supported version, but not the tested version.

## Describe the solution you'd like
Test the bot version again the tested version in addition to the supported version.

## Describe alternatives you've considered
None.

## Additional context
The current implementation only tests the bot version to see if it exists in supportversion. This provides misleading results for major versions that are supported but not minor versions. (For example, 1.16 is supported but 1.16.3 is not.)
https://github.com/PrismarineJS/mineflayer/blob/3cff9d799e40f8339302b72861b457883b0836fb/index.js#L114

I think that in addition to validating against the supported version, it would be useful to also check against the tested version as well. If the version has a minor version above the tested, an error would also be thrown.
https://github.com/PrismarineJS/mineflayer/blob/3cff9d799e40f8339302b72861b457883b0836fb/lib/version.js#L3

1.16.2 and 1.16.3 support will be released very shortly, but this would offer a nice way to prevent this kind of issue from happening again in future versions.

PR #1375, opened 2020-10-14
Compare version against tested version.

TheDudeFromCI:
Closes https://github.com/PrismarineJS/mineflayer/issues/1374

### DIFF
--- a/index.js
+++ b/index.js
@@ -36,6 +36,7 @@ const plugins = {
   time: require('./lib/plugins/time'),
   villager: require('./lib/plugins/villager')
 }
+
 const supportedVersions = require('./lib/version').supportedVersions
 const testedVersions = require('./lib/version').testedVersions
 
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -115,6 +129,11 @@ class Bot extends EventEmitter {
       if (supportedVersions.indexOf(version.majorVersion) === -1) {
         throw new Error(`Version ${version.minecraftVersion} is not supported.`)
       }
+
+      if (!testedVersions.find(v => compareVersions(v, version.minecraftVersion))) {
```
wvffle, line 133: .some

wvffle (changes requested):

wvffle:
Maybe add a note to the error about what is the most recent version that we support?

wvffle:
What should `compareVersions` do? I mean, should it check if the version is already tested or what? check if version is older or younger? There should be better name for such function

index.js
```diff
@@ -86,6 +86,20 @@ function createBot (options = {}) {
   return bot
 }
 
+function compareVersions (testedVersion, currentVersion) {
+  testedVersion = testedVersion.split('.')
+  currentVersion = currentVersion.split('.')
+
+  if (testedVersion.length === 2) testedVersion.push(0)
```
extremeheat, line 93: Note the compare function here won’t work with snapshots, or pre-release versions and might break in the future with hyphen releases or other version strings that might not conform with int.int.int.

TheDudeFromCI: Ah, that's true

rom1504: this should be fixed
you can use the protocol version from mcdata to make a range comparison version checking function

rom1504 (changes requested):

wvffle (approved):

rom1504 (approved):

Issue #1383, opened 2020-10-20
All non-player entities have 0/undefined height and width.


TheDudeFromCI:
- [X] The [FAQ](https://github.com/PrismarineJS/mineflayer/blob/master/docs/FAQ.md) doesn't contain a resolution to my issue 

## Versions
 - mineflayer: 2.31.0
 - server: vanilla/spigot/paper 1.16.1
 - node: 12.18.3

## Detailed description of a problem
For all non-player entities, the entity height always remains at 0.
For all entities, including players, the width always remains undefined.

## What did you try yet?
N/A

Did you try any method from the API?
There are no relevant API methods.

Did you try any example? Any error from those?
```js
bot.on('physicTick', () => {
  const entity = bot.nearestEntity()
  if (entity) {
    console.log(entity.name)
    console.log(entity.height)
    console.log(entity.width)
  }
})
```

By running the above example, it shows that all entities outside of players never have a height or width defined.

## Your current code
```js
const mineflayer = require('mineflayer')

const bot = mineflayer.createBot({
    host: process.argv[2],
    port: process.argv[3],
    username: process.argv[4],
    password: process.argv[5]
})

bot.on('physicTick', () => {
  const entity = bot.nearestEntity()
  if (entity) {
    console.log(entity.name)
    console.log(entity.height)
    console.log(entity.width)
  }
})
```

## Expected behavior
entity.height and entity.width should always return the values that are defined in mcData.

## Additional context
Players have their height hardcoded and are also not loaded from mcData.

PR #1384, opened 2020-10-20
Added entity height and width.

TheDudeFromCI:
Closes https://github.com/PrismarineJS/mineflayer/issues/1383

Also pushed a PR (https://github.com/PrismarineJS/prismarine-entity/pull/7) to prismarine-entity to add the default height and width flags. Once that and this PR are merged, height and width should be defined for all entities where possible.


rom1504:
https://minecraft-data.prismarine.js.org/?d=protocol#toClient_named_entity_spawn

No it doesn't. I'm pretty sure this packet is only for players so type
doesn't make sense

rom1504 (approved):

PR #1390, opened 2020-10-22
optimize support feature

rom1504:


lib/supportFeature.js
```diff
@@ -1,22 +1,27 @@
...
-
-  return -1
-}
+const versionList = require('minecraft-data').versions.pc
+const versionToIndex = Object.fromEntries(versionList.map((version, i) => [version.minecraftVersion, versionList.length - i]))
+const nameToFeature = Object.fromEntries(features.map(feature => [feature.name, feature]))
 
 module.exports = (feature, minecraftVersion) => {
-  const f = features.find(f => f.name === feature)
+  const f = nameToFeature[feature]
```
wvffle, line 8: Can't we just have:
```js
module.exports = (featureName, minecraftVersion) => {
  const feature = nameToFeature[featureName]
```
Having a single character variable is not descriptive and commonly seen as bad pattern

wvffle (changes requested):

lib/supportFeature.js
rom1504: yes fixed

PR #1399, opened 2020-10-23
Emit end event when calling bot.end

ARandomAssGuy:
idk this should work lmao

### DIFF
--- a/index.js
+++ b/index.js
@@ -134,5 +134,6 @@ class Bot extends EventEmitter {
 
   end () {
     this._client.end()
+    this.emit('end')
   }
 }


imharvol:
`'end'` event is emitted by [PrismarineJS/node-minecraft-protocol](https://github.com/PrismarineJS/node-minecraft-protocol):
https://github.com/PrismarineJS/mineflayer/blob/43dfe673f1532bf07a8ebda85d89ca9df679e6bb/index.js#L109-L111

 So if there's a problem with it, I think it should be handled there:
https://github.com/PrismarineJS/node-minecraft-protocol/blob/de731b3e1a443d064949f5a5cac91087572727d3/src/client.js#L136-L144
https://github.com/PrismarineJS/node-minecraft-protocol/blob/de731b3e1a443d064949f5a5cac91087572727d3/src/client.js#L160-L162

rom1504:
is this fixing anything @ARandomAssGuy ?

ARandomAssGuy:
i was trying to fix this. @rom1504

rom1504:
does this actually happen?

ARandomAssGuy:
I have test on a 1.12.2 and 1.16.1 server, doesn't happen. However, the person did say they were on a non-vanilla server, which might be the issue. I asked them for the ip address to see if i can reproduce. In hindsight,  i should have asked them first, but im kinda retarded so yeah sorry about that

ARandomAssGuy:
Screen Shot 2020-10-24 at 5 35 53 PM
yeah alr im gonna close this before i humiliate myself more

PR #1402, opened 2020-10-24
Update block_actions.js

SpikeThatMike:


rom1504 (approved):

rom1504:
can you run npm run fix ?

lib/plugins/block_actions.js
```diff
@@ -53,6 +53,7 @@ function inject (bot, { version }) {
   bot._client.on('block_action', (packet) => {
     const pt = new Vec3(packet.location.x, packet.location.y, packet.location.z)
     const block = bot.blockAt(pt)
+    if(block === null) return;
```
wvffle, line 56: Add newlines around the if. That way it'd be more readable.

wvffle: I'd also add `{` and `}`

wvffle (changes requested):

rom1504:
can you run npm run fix then commit the result?

rom1504 (approved):

Issue #1400, opened 2020-10-24
bot.sleep or sleeper.js don't work


DrMoraschi:
- [x] The [FAQ](https://github.com/PrismarineJS/mineflayer/blob/master/docs/FAQ.md) doesn't contain a resolution to my issue 

## Versions
 - mineflayer: 2.32.0
 - server: vanilla/spigot/paper 1.16.1
 - node: 12.18.3

## Detailed description of a problem
I run sleeper.js, and everytime i tell it to go to sleep(bed already placed in front of him) he sends "I can't sleep: it's not night". I think this is a problem of bot.sleep.

## What did you try yet?

Did some tweaking on sleeper.js but it seems like the problem is bot.sleep

## Expected behavior
Make the bot sleep.

imharvol:
Yeah, makes sense! Related issue: #1398

DrMoraschi:
> Yeah, makes sense! Related issue: #1398

Yeah, saw this issue before, didn't know if I had to write the issue there in the comments or not, so I created a new issue.

PR #1407, opened 2020-10-27
Fix bed packet

### LINKED ISSUE #1398: Find more reliable ways to know if the bot is sleeping and add tests
The 'bed' event has been removed from Minecraft [since 1.14](https://wiki.vg/Protocol_History#19w08a). That event was used here: https://github.com/PrismarineJS/mineflayer/blob/1b1907d48d4eaf5be990a1b53dd17cac18b11c53/lib/plugins/entities.js#L106-L111

And so `bot.isSleeping` never gets updated:
https://github.com/PrismarineJS/mineflayer/blob/1b1907d48d4eaf5be990a1b53dd17cac18b11c53/lib/plugins/bed.js#L166-L171

Which then blocks the players from running `bot.wake()`
https://github.com/PrismarineJS/mineflayer/blob/1b1907d48d4eaf5be990a1b53dd17cac18b11c53/lib/plugins/bed.js#L72-L83

I think we should find a reliable way to detect if the bot is sleeping. 
After a bit of testing I think this would be one option: Listening for  [Entity Metadata event](https://wiki.vg/Protocol#Entity_Metadata):
```json
{
  "entityId": 41,
  "metadata": [
    {
      "key": 6,
      "type": 18,
      "value": 2
    },
    {
      "key": 13,
      "type": 10,
      "value": {
        "x": -273,
        "z": 85,
        "y": 4
      }
    }
  ]
}
```

I think the ['wake' event](https://github.com/PrismarineJS/mineflayer/blob/1b1907d48d4eaf5be990a1b53dd17cac18b11c53/lib/plugins/entities.js#L113-L118) is being emitted correctly, as I think there have not been modifications to that part of the protocol. But we definetely need to add external tests for all of this. I'll work on this when I have some time :smile:

imharvol:
Closes #1398 and #1400

### DIFF
--- a/lib/plugins/entities.js
+++ b/lib/plugins/entities.js
@@ -359,6 +359,13 @@ function inject (bot, { version }) {
       bot.emit('itemDrop', entity)
     }
 
+    const pose = packet.metadata.find(e => e.type === 18)
+    if (pose) {
+      if (pose.value === 2) {
+        bot.emit('entitySleep', entity)
+      }
+    }
+
     const bitField = packet.metadata.find(p => p.key === 0)
     if (bitField === undefined) {
       return
--- a/lib/plugins/entities.js
+++ b/lib/plugins/entities.js
@@ -360,10 +360,8 @@ function inject (bot, { version }) {
     }
 
     const pose = packet.metadata.find(e => e.type === 18)
-    if (pose) {
-      if (pose.value === 2) {
-        bot.emit('entitySleep', entity)
-      }
+    if (pose && pose.value === 2) {
+      bot.emit('entitySleep', entity)
     }
 
     const bitField = packet.metadata.find(p => p.key === 0)
--- a/test/externalTests/bed.js
+++ b/test/externalTests/bed.js
@@ -12,7 +12,6 @@ module.exports = () => (bot, done) => {
 
   bot.test.callbackChain([
     (cb) => { // Put the bed
-      bot.chat(`/time set ${midnight}`)
       if (bot.supportFeature('setBlockUsesMetadataNumber', bot.version)) {
         bot.chat(`/setblock ${bedPos1.toArray().join(' ')} ${bedItem.name} 0`) // Footer
         bot.chat(`/setblock ${bedPos2.toArray().join(' ')} ${bedItem.name} 8`) // Head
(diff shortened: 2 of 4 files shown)


imharvol:
I'm not entirely sure about this:
https://github.com/PrismarineJS/mineflayer/blob/6d54b7bfc5e49e9988661f7f7b2620dd71b8af96/lib/plugins/entities.js#L106-L111

Should I create a feature in features.json that contains which versions 'bed' packet works and then put that code inside?
```js
if (bot.supportFeature('sleepPacket', bot.version)) {
  bot._client.on('bed', (packet) => { 
    // use bed 
    entity.position.set(packet.location.x, packet.location.y, packet.location.z) 
    bot.emit('entitySleep', entity) 
  }) 
}

File: lib/plugins/entities.js [hunk 1]
```diff
@@ -359,6 +359,13 @@ function inject (bot, { version }) {
       bot.emit('itemDrop', entity)
     }
 
+    const pose = packet.metadata.find(e => e.type === 18)
+    if (pose) {
+      if (pose.value === 2) {
+        bot.emit('entitySleep', entity)
+      }
+    }
```
NotSugden, line 367:
```suggestion
    if (pose && pose.value === 2) {
      bot.emit('entitySleep', entity)
    }
```

imharvol:
It's not quite clear to me why test are failing. Anyone knows?

test/externalTests/bed.js
```diff
@@ -0,0 +1,44 @@
...
+        bot.chat(`/setblock ${bedPos1.toArray().join(' ')} ${bedItem.name}[part=foot]`)
+        bot.chat(`/setblock ${bedPos2.toArray().join(' ')} ${bedItem.name}[part=head]`)
+      }
+      setTimeout(cb, 1000) // Enough time for the game-time and the block to be updated
+    },
+    (cb) => { // Sleep
+      assert(!bot.isSleeping)
+      bot.sleep(bot.blockAt(bedPos1), (err) => {
+        assert.ifError(err)
+        setTimeout(cb, 500) // Enough time for bot.isSleeping to be updated
```
rom1504, line 29: this will create flaky tests, can you use an event instead ?

rom1504 (changes requested):

test/externalTests/bed.js
imharvol: Pretty dumb of me not to use 'sleep' and 'wake' events so they're tested too

rom1504 (approved):

PR #1410, opened 2020-10-28
add objectUUID e.g. for items

Rob9315:
for entities like items, sent with 'spawn_entity', objectUUID is sent instead of entityUUID. this pr sets the entity.uuid to objectUUID if available


Rob9315:
i forgot to run chrecks beforehand

PR #1411, opened 2020-10-28
add uuid by objectUUID as it was missing

Rob9315:
I tried reading the UUID of the entity of a dropped Item and it was undefined. this should fix that

### DIFF
--- a/lib/plugins/entities.js
+++ b/lib/plugins/entities.js
@@ -189,6 +189,7 @@ function inject (bot, { version }) {
     }
 
     entity.uuid = packet.entityUUID
+    if (packet.objectUUID) entity.uuid = packet.objectUUID
     entity.yaw = conv.fromNotchianYawByte(packet.yaw)
     entity.pitch = conv.fromNotchianPitchByte(packet.pitch)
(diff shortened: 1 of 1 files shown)


lib/plugins/entities.js
```diff
@@ -189,6 +189,7 @@ function inject (bot, { version }) {
     }
 
     entity.uuid = packet.entityUUID
+    if (packet.objectUUID) entity.uuid = packet.objectUUID
```
wvffle, line 192:
```js
    if (packet.objectUUID) {
      entity.uuid = packet.objectUUID
    }
```

also move it either above or below the assigments with newlines around the if

wvffle (changes requested):

Rob9315:
so now I realized that the "spawn_entity" packet does not have an entityUUID property, it is always objectUUID so now I removed the if statement and just replaced the entityUUID by objectUUID

wvffle:
Are you sure that previous versions have `objectUUID` instead of `entityUUID` as well?

rom1504 (approved):

PR #1412, opened 2020-10-29
Added better handling for automatic dependency resolution between plugins

TheDudeFromCI:
In sitations where Plugin A relies on Plugin B to work, it's useful for plugin A to automatically trigger plugin B to load if the user does not specify this directly. This allows for better API compatibility. (A plugin can add a new plugin dependency to the bot without breaking existing bots that don't load plugin B by default.)

This also makes it much easier to load plugin trees that contain several nested plugins, all by simply loading the parent plugin.

This PR works by safely allowing the `bot.loadPlugin` to be called more than once without actually loading the plugin twice.

Bot:
```js
const { pluginA } = require('mineflayer-plugin-a')
const bot = mineflayer.createBot({...})
bot.loadPlugin(pluginA)
```

Plugin A:
```js
const { pluginB } = require('mineflayer-plugin-b')

function pluginA (bot) {
  bot.loadPlugin(pluginB)

  bot.pluginA = {}
  bot.pluginA.someFunction = function () {}
}

module.exports = {
  pluginA
}
```

Using this, Plugin A can safely load Plugin B when injected. If the user already has PluginB loaded, nothing happens.

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -38,7 +38,15 @@ Enabled debug mode https://github.com/PrismarineJS/mineflayer#debug
 One way is to increase the [checkTimeoutInterval](https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/docs/API.md#mccreateclientoptions) option (to set in createBot) to an higher value (for example `300*1000` which is 5min instead of the default 30s). If you still get disconnected, you can auto reconnect using something like this example https://github.com/PrismarineJS/mineflayer/blob/master/examples/reconnector.js
 
 ### How to get the lore / text of an item ?
+
 You can use the .nbt property of an item. prismarine-nbt nbt.simplify method may be useful
 
 ### How can I send message from the console to the server?
+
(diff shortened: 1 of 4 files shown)


wvffle (commented):
What if `plugin-a` requires `plugin-c@1.0.0` and `plugin-b` requires `plugin-c@2.0.0`?

wvffle:
> Also, wouldn't this kind of issue be resolved by npm, anyway?
Don't think so.

wvffle (commented):
> The end user must choose between whether they want to use plugin a or plugin b.

In such case I agree.

wvffle (approved):
(And I wanted to approve it.)

rom1504 (changes requested):
this is using the object reference as a plugin key
can you double-check this work in all circumstances?
for example, if you do a require in 2 different files for the plugin, do you get exactly the same reference?

rom1504 (approved):

PR #1413, opened 2020-10-29
Added tool plugin to readme

TheDudeFromCI:
Created a small utility plugin for handling tool (and eventually weapon) selection.

Currently allows for selecting the fastest tool to mine a block within your inventory, and an optional flag for automatically walking to a chest in order to collect the tool if it doesn't have a suitable one in its inventory.


wvffle (approved):

rom1504:
nice
do you think these plugins would belong to PrismarineJS org or do you prefer to keep them under your name?

PR #1425, opened 2020-11-01
Removed Hawkeye dependency

TheDudeFromCI:
It seems at some point this was accidentally added as a dependency. Removed it.

Should we add plugins as dev dependencies for example bots, though? If so, we have a few of them we should add.


wvffle:
how about per example package.json?

rom1504 (approved):

rom1504:
one dir per example could work yes (like we have in node-minecraft-protocol )

PR #1431, opened 2020-11-02
Added missing types

jecsham:
Added missing types.

Why?
Typescript projects with `Strict Type-Checking Options` enabled will not be able to transpile their program if types are missing.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -45,19 +45,19 @@ export type Plugin = (bot: Bot, options: BotOptions) => void;
 
 interface BotEvents {
   chat: (
-      username: string,
-      message: string,
-      translate: string | null,
-      jsonMsg: string,
-      matches: Array<string> | null
-    ) => void;
+    username: string,
+    message: string,
+    translate: string | null,
+    jsonMsg: string,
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -166,7 +166,7 @@ export class Bot extends (EventEmitter as new () => TypedEmitter<BotEvents>) {
   scoreboards: { [name: string]: ScoreBoard };
   scoreboard: { [slot in DisplaySlot]: ScoreBoard };
   controlState: ControlStateStatus;
-  void;
+  void: void;
```
TheDudeFromCI, line 169: Curious why void is here in the original file? Does it have a purpose in this context?

jecsham: I'm not sure if it has to do with future changes

TheDudeFromCI: If so, it would be best to remove it. It can be replaced if those future changes are made.

TheDudeFromCI (approved):

wvffle (approved):

rom1504 (approved):

PR #1460, opened 2020-11-17
Open the craft window only once

imharvol:
Right now the bot opens and closes the crafting table for every craft count that it wants to do. This opens the crafting table only once for all the crafts.

### DIFF
--- a/lib/plugins/craft.js
+++ b/lib/plugins/craft.js
@@ -14,20 +14,8 @@ function inject (bot, { version }) {
       cb(new Error('recipe requires craftingTable'))
       return
     }
-    next()
-    function next (err) {
-      if (err) {
-        cb(err)
-      } else if (count > 0) {
-        count -= 1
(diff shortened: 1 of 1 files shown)


rom1504:
Did you notice any difference in term of speed ?

imharvol:
I actually expected the change to increase the speed, but it takes moreless the same time. 
I tought that opening the crafting table would take more time, but looks like that's not why the craft function is pretty slow.

rom1504:
Isn't that due to not using shift ?

imharvol:
I was crafting melons so the bot has to fill the 9 slots of the crafting table with 1 slice of melon, so probably what takes so long is the _drawing_ of the shape. I would like to take a closer look to the craft function to try to improve the speed as right now is pretty confusing to me

rom1504:
@imharvol could you please update ? due to promises being merged this need to be updated

rom1504:
this would need to be redone after promise change. Closing for now

PR #1475, opened 2020-12-02
Promisify

### LINKED ISSUE #725: promisify

ph0t0shop:
Fixes #725. All callback functions have been promisifed. All tests are passing. The only functions not promisified are the following in villager.js:

```
villager#openVillager
villager#trade
(villager#putRequirements)
(villager#deposit)
```

I would like people to think with me on what is the best way to promisify these. The reason they aren't done now is because openVillager has a callback and _also_ synchronously returns a value. This is not something supported by promises normally.

There are also a few remarks to this PR.
- While going through the code, I was wondering if [the enchant function](https://github.com/PrismarineJS/mineflayer/blob/124e8508f44ee1e18cd2ac13c59d50a30b733d7e/lib/plugins/enchantment_table.js#L54) was actually in the right order? Shouldn't the listener be registered before the 'enchant_item' call so as to not create a race condition?
- In [simple_inventory#tossStack](https://github.com/PrismarineJS/mineflayer/blob/124e8508f44ee1e18cd2ac13c59d50a30b733d7e/lib/plugins/simple_inventory.js#L26), I have slightly changed the order of operations. This can be seen [here](https://github.com/ph0t0shop/mineflayer/blob/26c55e54f8c38a066cd60843106dc0c0e5e28974/lib/plugins/simple_inventory.js#L29). It seemed to make the most sense to me to close the window only after clicking outside the inventory, not introducing a race condition.
- In [creative#setInventorySlot](https://github.com/PrismarineJS/mineflayer/blob/124e8508f44ee1e18cd2ac13c59d50a30b733d7e/lib/plugins/creative.js#L28) it doesn't check if cb is null anymore, as can be seen [here](https://github.com/ph0t0shop/mineflayer/blob/26c55e54f8c38a066cd60843106dc0c0e5e28974/lib/plugins/creative.js#L28). Calling the function on the same slot again will now _always_ fail if called before the previous callback has succeeded, even if there is no callback passed through the "callbackified" function. This is because internally the resolve function is always present, and null checks cannot be done on it. This seemed fine to me, since it would be weird behavior anyway to call this function without a callback.
- [Inventory#consume](https://github.com/PrismarineJS/mineflayer/blob/124e8508f44ee1e18cd2ac13c59d50a30b733d7e/lib/plugins/inventory.js#L71) used to _return_ an error **synchronously**. It now throws this instead, as can be seen [here](https://github.com/ph0t0shop/mineflayer/blob/26c55e54f8c38a066cd60843106dc0c0e5e28974/lib/plugins/inventory.js#L71), so it will be passed asynchronously to the callback function as an err object, or to the promise's reject function of course.
- The same as above goes for [bed#sleep](https://github.com/PrismarineJS/mineflayer/blob/124e8508f44ee1e18cd2ac13c59d50a30b733d7e/lib/plugins/bed.js#L85). The changed version is [here](https://github.com/ph0t0shop/mineflayer/blob/26c55e54f8c38a066cd60843106dc0c0e5e28974/lib/plugins/bed.js#L87). An important distinction is that a test had to be changed for this. [Where it asserted .throws() before,](https://github.com/PrismarineJS/mineflayer/blob/124e8508f44ee1e18cd2ac13c59d50a30b733d7e/test/internalTest.js#L483), it now asserts equality with the err object instead, as can be seen [here](https://github.com/ph0t0shop/mineflayer/blob/26c55e54f8c38a066cd60843106dc0c0e5e28974/test/internalTest.js#L483)
- Lastly, [inventory#consume](https://github.com/PrismarineJS/mineflayer/blob/124e8508f44ee1e18cd2ac13c59d50a30b733d7e/lib/plugins/inventory.js#L71) and [inventory#fish](https://github.com/PrismarineJS/mineflayer/blob/124e8508f44ee1e18cd2ac13c59d50a30b733d7e/lib/plugins/inventory.js#L85) have their _setImmediate_ calls removed. The new functions are respectively [here](https://github.com/ph0t0shop/mineflayer/blob/26c55e54f8c38a066cd60843106dc0c0e5e28974/lib/plugins/inventory.js#L71) and [here](https://github.com/ph0t0shop/mineflayer/blob/26c55e54f8c38a066cd60843106dc0c0e5e28974/lib/plugins/inventory.js#L90). I'm _pretty_ sure the semantics are correct, but I'd like someone to verify if possible.

All in all, the promisification allows for much cleaner code both internally and externally. I've done some slight cleaning up in the [craft](https://github.com/ph0t0shop/mineflayer/blob/26c55e54f8c38a066cd60843106dc0c0e5e28974/lib/plugins/craft.js#L9) function that is now flattened with a for loop instead of a recursive function, as well as lots of places with chained callbacks, but more can be done. [Inventory#transfer](https://github.com/ph0t0shop/mineflayer/blob/26c55e54f8c38a066cd60843106dc0c0e5e28974/lib/plugins/inventory.js#L244) for example can be greatly improved. Functionally though, I think everything works fine, and the outside API is backwards compatible with the callback versions.

### DIFF
--- a/lib/plugins/bed.js
+++ b/lib/plugins/bed.js
@@ -4,8 +4,11 @@ module.exports = inject
 
 const CARDINAL_DIRECTIONS = ['south', 'west', 'north', 'east']
 
-function noop (err) {
-  if (err) throw err
+function callbackify (f) {
+  return function (...args) {
+    const cb = args[f.length] || (() => {})
+    return f(...args).then(r => { if (cb) { cb(null, r) } return r }, err => { if (cb) { cb(err) } else throw err })
+  }
 }
 
 function inject (bot) {
@@ -69,34 +72,32 @@ function inject (bot) {
     return metadata
   }
 
-  function wake (cb) {
+  async function wake () {
     if (!bot.isSleeping) {
-      cb(new Error('already awake'))
+      throw new Error('already awake')
     } else {
       bot._client.write('entity_action', {
         entityId: bot.entity.id,
         actionId: 2,
         jumpBoost: 0
       })
-      cb()
     }
   }
 
-  function sleep (bedBlock, cb) {
-    cb = cb || noop
+  async function sleep (bedBlock) {
     if (!(bot.time.timeOfDay >= 12541 && bot.time.timeOfDay <= 23458)) {
-      cb(new Error("it's not night"))
+      throw new Error("it's not night")
     } else if (bot.isSleeping) {
-      cb(new Error('already sleeping'))
+      throw new Error('already sleeping')
     } else if (!isABed(bedBlock)) {
-      cb(new Error('wrong block : not a bed block'))
+      throw new Error('wrong block : not a bed block')
     } else {
       const botPos = bot.entity.position.floored()
       const metadata = parseBedMetadata(bedBlock)
       let headPoint = bedBlock.position
 
       if (metadata.occupied) {
-        return cb(new Error('the bed is occupied'))
+        throw new Error('the bed is occupied')
       }
 
       if (!metadata.part) { // Is foot
@@ -112,13 +113,13 @@ function inject (bot) {
             headPoint = bedBlock.position
             bedBlock = lowerBlock
           } else {
-            return cb(new Error("there's only half bed"))
+            throw new Error("there's only half bed")
           }
         }
       }
 
       if (!bot.canDigBlock(bedBlock)) {
-        return cb(new Error('cant click the bed'))
+        throw new Error('cant click the bed')
       }
 
       const clickRange = [2, -3, -3, 2] // [south, west, north, east]
@@ -134,7 +135,7 @@ function inject (bot) {
       const nwClickCorner = headPoint.offset(clickRange[1], -2, clickRange[2]) // North-West lower corner
       const seClickCorner = headPoint.offset(clickRange[3], 2, clickRange[0]) // South-East upper corner
       if (botPos.x > seClickCorner.x || botPos.x < nwClickCorner.x || botPos.y > seClickCorner.y || botPos.y < nwClickCorner.y || botPos.z > seClickCorner.z || botPos.z < nwClickCorner.z) {
-        return cb(new Error('the bed is too far'))
+        throw new Error('the bed is too far')
       }
 
       if (bot.gameMode !== 'creative' || bot.supportFeature('creativeSleepNearMobs')) { // If in creative mode the bot should be able to sleep even if there are monster nearby (starting in 1.13)
@@ -146,13 +147,13 @@ function inject (bot) {
           if (entity.kind === 'Hostile mobs') {
             const entityPos = entity.position.floored()
             if (entityPos.x <= seMonsterCorner.x && entityPos.x >= nwMonsterCorner.x && entityPos.y <= seMonsterCorner.y && entityPos.y >= nwMonsterCorner.y && entityPos.z <= seMonsterCorner.z && entityPos.z >= nwMonsterCorner.z) {
-              return cb(new Error('there are monsters nearby'))
+              throw new Error('there are monsters nearby')
             }
           }
         }
       }
 
-      bot.activateBlock(bedBlock, cb)
+      await bot.activateBlock(bedBlock)
     }
   }
 
@@ -178,7 +179,7 @@ function inject (bot) {
   })
 
   bot.parseBedMetadata = parseBedMetadata
-  bot.wake = wake
-  bot.sleep = sleep
+  bot.wake = callbackify(wake)
+  bot.sleep = callbackify(sleep)
   bot.isABed = isABed
 }
--- a/lib/plugins/bed.js
+++ b/lib/plugins/bed.js
@@ -6,7 +6,7 @@ const CARDINAL_DIRECTIONS = ['south', 'west', 'north', 'east']
 
 function callbackify (f) {
   return function (...args) {
-    const cb = args[f.length] || (() => {})
+    const cb = args[f.length]
     return f(...args).then(r => { if (cb) { cb(null, r) } return r }, err => { if (cb) { cb(err) } else throw err })
   }
 }
@@ -180,6 +180,6 @@ function inject (bot) {
 
   bot.parseBedMetadata = parseBedMetadata
   bot.wake = callbackify(wake)
-  bot.sleep = callbackify(sleep)
+  bot.sleep = sleep
   bot.isABed = isABed
 }
--- a/lib/plugins/bed.js
+++ b/lib/plugins/bed.js
@@ -1,15 +1,9 @@
 const Vec3 = require('vec3').Vec3
+const { callbackify } = require('../promise_utils')
 
 module.exports = inject
 
 const CARDINAL_DIRECTIONS = ['south', 'west', 'north', 'east']
-
-function callbackify (f) {
-  return function (...args) {
-    const cb = args[f.length]
-    return f(...args).then(r => { if (cb) { cb(null, r) } return r }, err => { if (cb) { cb(err) } else throw err })
-  }
-}
 
 function inject (bot) {
   bot.isSleeping = false
--- a/lib/plugins/bed.js
+++ b/lib/plugins/bed.js
@@ -180,6 +180,6 @@ function inject (bot) {
 
   bot.parseBedMetadata = parseBedMetadata
   bot.wake = callbackify(wake)
-  bot.sleep = sleep
+  bot.sleep = callbackify(sleep)
   bot.isABed = isABed
 }
--- a/lib/plugins/blocks.js
+++ b/lib/plugins/blocks.js
@@ -110,19 +110,21 @@ function inject (bot, { version, storageBuilder }) {
     bot.emit('chunkColumnLoad', columnCorner)
   }
 
-  function waitForChunksToLoad (cb) {
+  async function waitForChunksToLoad () {
     // check 5x5 chunks around us
     for (let x = -2; x <= 2; x++) {
       for (let y = -2; y <= 2; y++) {
         if (bot.world.getColumnAt(bot.entity.position.plus(new Vec3(x, 0, y).scaled(16))) == null) {
           // keep wait
-          return setTimeout(() => {
-            waitForChunksToLoad(cb)
-          }, 100)
+          return new Promise((resolve) =>
+            setTimeout(async () => {
+              await waitForChunksToLoad()
+              resolve()
+            }, 100)
+          )
         }
       }
     }
-    cb()
   }
 
   function getMatchingFunction (matching) {
@@ -495,7 +497,7 @@ function inject (bot, { version, storageBuilder }) {
   bot.blockAt = blockAt
   bot._updateBlockState = updateBlockState
   bot._blockEntities = blockEntities
-  bot.waitForChunksToLoad = waitForChunksToLoad
+  bot.waitForChunksToLoad = callbackify(waitForChunksToLoad)
 }
 
 function onesInShort (n) {
@@ -505,4 +507,11 @@ function onesInShort (n) {
     count = ((1 << i) & n) ? count + 1 : count
   }
   return count
+}
+
+function callbackify (f) {
+  return function (...args) {
+    const cb = args[f.length] || (() => {})
+    return f(...args).then(r => { if (cb) { cb(null, r) } return r }, err => { if (cb) { cb(err) } else throw err })
+  }
 }
--- a/lib/plugins/blocks.js
+++ b/lib/plugins/blocks.js
@@ -511,7 +511,7 @@ function onesInShort (n) {
 
 function callbackify (f) {
   return function (...args) {
-    const cb = args[f.length] || (() => {})
+    const cb = args[f.length]
     return f(...args).then(r => { if (cb) { cb(null, r) } return r }, err => { if (cb) { cb(err) } else throw err })
   }
 }
--- a/lib/plugins/blocks.js
+++ b/lib/plugins/blocks.js
@@ -116,12 +116,8 @@ function inject (bot, { version, storageBuilder }) {
       for (let y = -2; y <= 2; y++) {
         if (bot.world.getColumnAt(bot.entity.position.plus(new Vec3(x, 0, y).scaled(16))) == null) {
           // keep wait
-          return new Promise((resolve) =>
-            setTimeout(async () => {
-              await waitForChunksToLoad()
-              resolve()
-            }, 100)
-          )
+         await new Promise(resolve => setTimeout(resolve, 100))
+         await waitForChunksToLoad()
         }
       }
     }
--- a/lib/plugins/blocks.js
+++ b/lib/plugins/blocks.js
@@ -2,6 +2,7 @@ const Vec3 = require('vec3').Vec3
 const assert = require('assert')
 const Painting = require('../painting')
 const Location = require('../location')
+const { callbackify, sleep } = require('../promise_utils')
 
 const { OctahedronIterator } = require('../iterators')
 
@@ -116,7 +117,7 @@ function inject (bot, { version, storageBuilder }) {
       for (let y = -2; y <= 2; y++) {
         if (bot.world.getColumnAt(bot.entity.position.plus(new Vec3(x, 0, y).scaled(16))) == null) {
           // keep wait
-          await new Promise(resolve => setTimeout(resolve, 100))
+          await sleep(100)
           await waitForChunksToLoad()
         }
       }
@@ -503,11 +504,4 @@ function onesInShort (n) {
     count = ((1 << i) & n) ? count + 1 : count
   }
   return count
-}
-
-function callbackify (f) {
-  return function (...args) {
-    const cb = args[f.length]
-    return f(...args).then(r => { if (cb) { cb(null, r) } return r }, err => { if (cb) { cb(err) } else throw err })
-  }
 }
--- a/lib/plugins/blocks.js
+++ b/lib/plugins/blocks.js
@@ -116,8 +116,8 @@ function inject (bot, { version, storageBuilder }) {
       for (let y = -2; y <= 2; y++) {
         if (bot.world.getColumnAt(bot.entity.position.plus(new Vec3(x, 0, y).scaled(16))) == null) {
           // keep wait
-         await new Promise(resolve => setTimeout(resolve, 100))
-         await waitForChunksToLoad()
+          await new Promise(resolve => setTimeout(resolve, 100))
+          await waitForChunksToLoad()
         }
       }
     }
--- a/lib/plugins/book.js
+++ b/lib/plugins/book.js
@@ -40,12 +40,13 @@ function inject (bot, { version }) {
     bot.setQuickBarSlot(moveToQuickBar ? 0 : slot - 36)
 
     const modifiedBook = await modifyBook(moveToQuickBar ? 36 : slot, pages, author, title, signing)
-    const response = once(bot, 'setSlot:0')
     editBook(modifiedBook, signing)
-    let [, newItem] = await response
-    while (newItem.slot !== slot) {
+
+    let newItem
+    do {
       [, newItem] = await once(bot, 'setSlot:0')
     }
+    while (newItem.slot !== slot)
 
     bot.setQuickBarSlot(quickBarSlot)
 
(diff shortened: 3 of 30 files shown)


lib/plugins/blocks.js
```diff
@@ -110,19 +110,21 @@ function inject (bot, { version, storageBuilder }) {
...
           // keep wait
-          return setTimeout(() => {
-            waitForChunksToLoad(cb)
-          }, 100)
+          return new Promise((resolve) =>
+            setTimeout(async () => {
+              await waitForChunksToLoad()
+              resolve()
+            }, 100)
+          )
```
DeltaEvo, line 124:
```suggestion
         await new Promise(resolve => setTimeout(resolve, 100))
         await waitForChunksToLoad()
```

lib/plugins/bed.js
```diff
@@ -4,8 +4,11 @@ module.exports = inject
 
 const CARDINAL_DIRECTIONS = ['south', 'west', 'north', 'east']
 
-function noop (err) {
-  if (err) throw err
+function callbackify (f) {
```
DeltaEvo, line 7: Seems duplicated in a lot of place why not add it into a file and import it ?

lib/plugins/book.js
```diff
@@ -28,44 +31,44 @@ function inject (bot, { version }) {
...
+      return new Promise(resolve => {
+        async function onSetInvSlot (oldItem, newItem) {
+          if (newItem.slot !== slot) return
+          bot.removeListener('setSlot:0', onSetInvSlot)
+          bot.setQuickBarSlot(quickBarSlot)
+          await bot.moveSlotItem(36, slot)
+          resolve()
+        }
+        bot.on('setSlot:0', onSetInvSlot)
+        editBook(book, signing)
```
DeltaEvo, line 53:
```js
// at the top of the file
const { once } = require('events')

editBook(book, signing)
let oldItem
let newItem
do {
  ;[oldItem, newItem] = await once(bot, 'setSlot:0')
} while (newItem.slot !== slot)

bot.setQuickBarSlot(quickBarSlot)
await bot.moveSlotItem(36, slot)
```

lib/plugins/book.js
```diff
@@ -28,44 +31,44 @@ function inject (bot, { version }) {
...
       })
     } else {
       bot.setQuickBarSlot(slot - 36)
-      modifyBook(slot, pages, author, title, signing, (book) => {
+      const book = await modifyBook(slot, pages, author, title, signing)
+      return new Promise(resolve => {
```
DeltaEvo, line 58: Same that upper comment

lib/plugins/book.js
```diff
@@ -94,22 +97,24 @@ function inject (bot, { version }) {
...
+        }
       }
-    }
-  }
 
-  bot.writeBook = (slot, pages, cb = noop) => {
-    write(slot, pages, null, null, false, cb)
+      bot.inventory.on('windowUpdate', onUpdate)
+      bot.inventory.updateSlot(slot, book)
+    })
```
DeltaEvo, line 110:
```js
bot.inventory.updateSlot(slot, book)
while (true) {
  const [updatedSlot] == await once(bot.inventory, 'windowUpdate')
  if (slot === updatedSlot) return
}
```

lib/plugins/chat.js
```diff
@@ -110,14 +110,16 @@ function inject (bot, options) {
...
 
-    bot._client.write('tab_complete', {
-      text,
-      assumeCommand,
-      lookedAtBlock: position
+      bot._client.write('tab_complete', {
+        text,
+        assumeCommand,
+        lookedAtBlock: position
+      })
```
DeltaEvo, line 122:
```js
bot._client.write('tab_complete', {
  text,
  assumeCommand,
  lookedAtBlock: position,
})

const [{ matches }] = await once(bot._client, 'tab_complete')
return matches
```

lib/plugins/craft.js
```diff
@@ -6,42 +6,34 @@ function inject (bot, { version }) {
...
-        }
-        startClicking(window, 3, 3)
+      return new Promise(resolve => {
+        bot.once('windowOpen', async (window) => {
+          if (!window.type.startsWith('minecraft:crafting')) {
+            throw new Error('crafting: non craftingTable used as craftingTable')
+          }
+          await startClicking(window, 3, 3)
+          resolve()
+        })
```
DeltaEvo, line 30:
```suggestion
const [window] = await once(bot, 'windowOpen')
if (!window.type.startsWith('minecraft:crafting')) {
  throw new Error('crafting: non craftingTable used as craftingTable')
}
await startClicking(window, 3, 3)

```

lib/plugins/craft.js
```diff
@@ -188,26 +151,16 @@ function inject (bot, { version }) {
...
-            return
-          }
-          bot.putAway(theSlot, (err) => {
-            if (err) {
-              cb(err)
-            } else {
-              next()
-            }
-          })
+        theSlot = slotsToClick.pop()
```
DeltaEvo, line 154:
```suggestion
        let theSlot = slotsToClick.pop()
```

DeltaEvo (changes requested):
I've not done all yet but a lot of the `new Promise` can be removed using https://nodejs.org/api/events.html#events_events_once_emitter_name_options

And I don't know if you checked it but there is already a PR to promisify mineflayer https://github.com/PrismarineJS/mineflayer/pull/984, maybe this PR have some code that could be reused :)

lib/plugins/bed.js
ph0t0shop: Well, noop was duplicated as well. And actually there's a couple places where callbackify is different because cb is not the last parameter

ph0t0shop:
I looked at the other PR indeed, but I saw only 2 changed files. And there were merging issues so I decided to redo it from scratch

lib/plugins/chat.js
ph0t0shop: That doesn't work, because the listener needs to be registered _before_ triggering it otherwise you'll introduce a race condition. But if you put it after, it will await infinitely because it never gets triggered.

lib/plugins/bed.js
DeltaEvo: For the different callbackify, it should be possible to refactor them into one by adding some extra optionals parameters like the position of the callback

lib/plugins/craft.js
ph0t0shop: theSlot is initialized above

lib/plugins/bed.js
ph0t0shop: Look at digging#callbackify though, I don't think it's possible to generalize that

lib/plugins/blocks.js
ph0t0shop: Semantically it's the same of course but I agree it looks cleaner. Keep in mind (in general) that I tried to change semantics as little as possible in the beginning, stuff like this can also easily be done later. But I agree in this case this is a good improvement

lib/plugins/book.js
ph0t0shop: Good one. it's only missing the return value which should be `book`. I think it should even be possible to do `await bot.inventory.once('windowUpdate')` instead.

lib/plugins/craft.js
ph0t0shop: Good one as well

lib/plugins/book.js
ph0t0shop: This one introduces a race condition as well since the listener should be registered before executing the action

lib/plugins/book.js
ph0t0shop: Same goes for this one

lib/plugins/chat.js
DeltaEvo: You can't really have a race condition, since having a race condition would mean that you have send & received the packet in the same event loop turn, it's impossible but if you are really bothered by that you can just do

```js
const response = once(bot._client, 'tab_complete')

bot._client.write('tab_complete', {
  text,
  assumeCommand,
  lookedAtBlock: position,
})

const [{ matches }] = await response
return matches
```

lib/plugins/book.js
DeltaEvo: Same as https://github.com/PrismarineJS/mineflayer/pull/1475#discussion_r534322910
But again you can write this like

```js
// at the top of the file
const { once } = require('events')

const response = once(bot, 'setSlot:0')
editBook(book, signing)
let [oldItem, newItem] = await response
while (newItem.slot !== slot) {
  ;[oldItem, newItem] = await once(bot, 'setSlot:0')
}

bot.setQuickBarSlot(quickBarSlot)
await bot.moveSlotItem(36, slot)
```

lib/plugins/chat.js
ph0t0shop: That is better, thanks. I never want to assume stuff like that, I'm a bit paranoid about this. Look at `inventory#clickWindow` for example, it definitely requires this structure.

lib/plugins/bed.js
DeltaEvo: I think a better solution for digging is just to not use a callbackify function and do
```js
  const callbackDig = util.callbackify(dig)
  bot.dig = (block, forceLook, cb) => {
    if (typeof forceLook === 'function') {
      cb = forceLook
      forceLook = true
    }
    cb = cb || noop
    callbackDig(block, forceLook, cb)
  }
```

lib/plugins/book.js
DeltaEvo: `await bot.inventory.once('windowUpdate')` does not do the same thing the once on the prototype of the EventEmiiter dosen't return a promise see https://nodejs.org/api/events.html#events_emitter_once_eventname_listener

lib/plugins/chat.js
DeltaEvo: I understand, but the way nodejs works you can't have this kind of race condition, but since not everyone is aware of how it works it might be better to write the code without taking this into account so everyone agrees on the code

ph0t0shop: I know, but what if later on that code is changed to a shortcut version or something that directly emits the event, I think it will make it fail, right? Or am I wrong in this assumption?

DeltaEvo: Yep in this case it will fail since emit is synchronous but since it we are dealing with protocol I think it is safe, but yeah it might not work for other cases

lib/plugins/inventory.js
```diff
@@ -68,29 +68,39 @@ function inject (bot, { version, hideErrors }) {
...
 
     activateItem()
 
-    lastEating = (err) => setImmediate(cb, err)
+    return new Promise((resolve, reject) => {
+      lastEating = (err) => {
+        if (err) reject(err)
+        else resolve()
+      }
+    })
```
wvffle, line 87: Can be rewritten as following to ommit new promise:
```js
let eatingState = false

bot._client.on('entity_status', (packet) => {
  if (packet.entityId === bot.entity.id && packet.entityStatus === 9 && eatingState) {
    bot.emit('eatingDone')
  }
})

async function consume () {
  if (eatingState) {
    throw ...
  }

  if (bot.food === 20) {
    throw ...
  }

  eatingState = true
  await once(bot, 'eatingDone')
  eatingState = false
}
```

wvffle: That's actually what we've been discussing quite a while ago on discord when I was working on #984. Same applies for a couple of other methods

lib/plugins/book.js
ph0t0shop: resolved in 43ba7aa

wvffle (changes requested):
There are many `new Promise` that are unnecessary. Try to avoid them.

I'd like to take a closer look at this PR but the least busy day this week is Sunday.

lib/plugins/book.js
ph0t0shop: resolved in 43ba7aa

lib/plugins/chat.js
ph0t0shop: resolved in 43ba7aa

lib/plugins/inventory.js
ph0t0shop: Yeah, I've been going around and fixing those, I'll keep doing that

ph0t0shop: Hmm though looking at this specific example, I'm not sure if it's correct. If consume() is called while another is running, the _other_ consume should be rejected, not the new one.

rom1504:
Looks good overall, I'll comment on details later.
Here are some big missing things:
* The doc : we should make it clear how to use promise or callback
* The examples: they should all use promises to show how people should build things

At a later stage (after this pr is merged) to make transition easier, it will be good to do the same kind of PR for the most used mineflayer plugins

ph0t0shop:
Sure, but this is just the code PR. Since it's not a breaking change, docs/examples aren't _immediately_ necessary and could be gradually introduced. And how to use callbacks should already be in the docs, no?

lib/plugins/bed.js
ph0t0shop: I will look at this later

rom1504:
True we don't have to change it all in this PR, however I bet that adapting the example will give some insight even for the main code. Same for adapting the documentation.
So I think it'd be worth it to adapt at least a few examples to try it out.

lib/plugins/bed.js
ph0t0shop: For now I did already extract the functions into a module, I have left the two exceptions as-is

ph0t0shop:
There are just 3 remaining `new Promise` constructs. They are in 
- `digging#dig`
- `inventory#consume`
- `inventory#fish`

All 3 of these require the code to be able to reject _during_ the promise handling, so I believe these will need to stay exposed. However, they could be prettified I think

ph0t0shop:
Like I asked in the discord, I'm still unclear on the following:
What is the verdict with relation to situations where you do
```
bot._client.write('blablabla', {
  data
})
```
and want a listener for the response? Is it okay to register the listener directly after? Or should it definitely be registered before? I'm worried about race conditions where it gets immediately triggered, but registering it directly after allows for much cleaner code. It will enable some more improvements, like in `book#write`, `book#modifyBook`, `chat#tabComplete` and `inventory#clickWindow`

lib/plugins/inventory.js
wvffle: You have the overall idea though

wvffle: `once` fails when `error` emitted so you'd have to do a new emitter for that. Both consume and fish can share the emitter:
```js
const state = new EventEmitter()
state.eating = false

bot._client.on('entity_status', (packet) => {
  if (packet.entityId === bot.entity.id && packet.entityStatus === 9 && state.eating) {
    state.emit('eatingDone', true)
  }
})

async function consume () {
  if (state.eating) {
    state.emit('eatingDone', false)
  }

  if (bot.food === 20) {
    throw ...
  }

  state.eating = true
  if (!await once(state, 'eatingDone')) {
    throw ...
  }
  state.eating = false
}
```

lib/plugins/inventory.js
```diff
@@ -37,72 +39,77 @@ function inject (bot, { version, hideErrors }) {
...
 
-    return new Promise((resolve, reject) => {
-      lastEating = (err) => {
-        if (err) reject(err)
-        else resolve()
-      }
-    })
+    const [eatingDone, err] = await once(inventoryEvents, 'eatingDone')
+
+    if (!eatingDone) {
```
wvffle, line 88:
```js
const [err] = ...
if (err) {
  ...
}
```
And skip the boolean when emitting

lib/plugins/inventory.js
```diff
@@ -22,10 +22,12 @@ function inject (bot, { version, hideErrors }) {
     bobberId = mcData.entitiesByName.fishing_bobber.id
   }
 
+  const inventoryEvents = new EventEmitter()
```
wvffle, line 25: You can share this eventemitter with digging.js

I'd name it something like `bot._internalEvents`

wvffle (changes requested):

rom1504:
For all the exposed function that now return a promise, could you indicate it in the doc ? (API.md)
if the promise contain a value, also indicate it

ph0t0shop:
> 
> 
> For all the exposed function that now return a promise, could you indicate it in the doc ? (API.md)
> if the promise contain a value, also indicate it

done

lib/plugins/digging.js
```diff
@@ -10,81 +11,81 @@ function inject (bot) {
...
-        clearInterval(swingInterval)
-        clearTimeout(waitTimeout)
-        swingInterval = null
-        waitTimeout = null
-        bot.targetDigBlock = null
-        bot.lastDigTime = performance.now()
-        bot.emit('diggingCompleted', newBlock)
-        cb()
-      }
+    const [diggingDone] = await once(bot._internalEvents, `diggingDone:${block.position}`)
```
rom1504, line 49: can you explain why we need such an event whereas we didn't need one before ?

lib/plugins/inventory.js
```diff
@@ -20,10 +22,11 @@ function inject (bot, { version, hideErrors }) {
     bobberId = mcData.entitiesByName.fishing_bobber.id
   }
 
+  bot._internalEvents.eating = false
```
rom1504, line 25: why is this stored in this _internalEvents thing? it seems it wasn't needed before

lib/plugins/inventory.js
```diff
@@ -35,62 +38,77 @@ function inject (bot, { version, hideErrors }) {
...
-    if (!lastFloat || !lastFishing) return
+    if (!lastFloat || !bot._internalEvents.fishing) return
 
     const pos = lastFloat.position
     if (packet.particleId === 4 && packet.particles === 6 && pos.distanceTo(new Vec3(packet.x, pos.y, packet.z)) <= 1.23) {
       activateItem()
       lastFloat = undefined
-      setImmediate(lastFishing)
-      lastFishing = undefined
+      bot._internalEvents.emit('fishingDone')
```
rom1504, line 59: is it not possible to store a promise here instead of using an event ?

rom1504 (changes requested):
Thanks for working on this!
Overall it looks good.

I think bot._internalEvents is an antipattern:
* it keeps hidden and misspecified state that is completely global
* that makes plugins more interdependent

Plus the places it is used seem incorrect: an event seems like the wrong information in those cases as you want to listen for a specific end of a task rather than a general "that kind of thing happened here".
Storing the promise and doing an await on it seems like the cleaner way to do things.

I left specific comments about it.

rom1504:
I'd appreciate if you could also have a look @TheDudeFromCI and @Karang as this is a major change for mineflayer.

lib/plugins/digging.js
ph0t0shop: I agree with your sentiment - please look at @wvffle's changes he requested: https://github.com/PrismarineJS/mineflayer/pull/1475#pullrequestreview-543115315 . It was actually like you suggested before: storing the promise and doing an await, but it was deemed unwanted. There is also discussion about this in the discord

lib/plugins/inventory.js
```diff
@@ -39,77 +38,71 @@ function inject (bot, { version, hideErrors }) {
...
-    inventoryEvents.fishing = true
+    fishingTask = createTask()
 
     activateItem()
 
-    const [err] = await once(inventoryEvents, 'fishingDone')
-
-    if (err) {
+    try {
+      await fishingTask.promise
```
rom1504, line 101: isn't it the same thing to remove that try?

rom1504:
examples that would be great to try manually:
* [ ] inventory
* [ ] digger
* [ ] jumper
* [ ] chest

they have the most important mineflayer features

ph0t0shop:
I have tested all of those, however only 3 of them had to be changed for the promisify (jumper didn't, and for digger only 1 statement). It works perfectly.

rom1504 (approved):

PR #1477, opened 2020-12-03
Added mcData reference QOL

TheDudeFromCI:
I thought that it would be useful to initalize a mcData instance alongwith the bot on startup. Since Minecraft-Data is loaded by the bot on startup anyway, why not attach it as a property for quality of life. This way you don't have to redefine mcData whenever you want to use it. (Which is quite often.)

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -103,6 +103,7 @@
       - [bot.game.gameMode](#botgamegamemode)
       - [bot.game.hardcore](#botgamehardcore)
       - [bot.game.maxPlayers](#botgamemaxplayers)
+      - [bot.mcData](#botmcdata)
     - [bot.player](#botplayer)
       - [bot.players](#botplayers)
       - [bot.isRaining](#botisraining)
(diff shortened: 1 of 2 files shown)


wvffle (changes requested):
Examples and pretty much every reference to mcData should be updated. But most importantly, the examples.

rom1504:
yeah I'm not sure we want to do this. Bundling the libs inside mineflayer make it harder to understand from users that they are separate libs.
This choice should be made once for all libs used by mineflayer

rom1504:
I mean all the prismarine-*

rom1504:
For example for prismarine-block, we currently do Block = require ("prismarine-block")(bot.version)
At a bunch of places
We could have a bot.Block to avoid that

In fact that's how it used to be (except it was mineflayer.Block because there was no multi version support)

However the reason I think we should not do it is because the direction I think we should go towards is making mineflayer *less* of a monolith. Ie I'd like to encourage people as much as possible to use other libs independently. Exposing them through mineflayer seems to go in the opposite direction.

Does that make sense to you ?

PR #1484, opened 2020-12-05
Promisify Villager & Trader

ph0t0shop:
Aiming to promisify the last villager stuff, including api, ts bindings, and test

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1703,8 +1703,11 @@ you are opening.
 #### bot.openVillager(villagerEntity)
 
 Returns an `Villager` instance which represents the trading window you are opening.
+You can listen to the `ready` event on this `Villager` to know when it's ready
 
 #### bot.trade(villagerInstance, tradeIndex, [times], [cb])
+
+This function also returns a `Promise`, with `void` as its argument upon completion.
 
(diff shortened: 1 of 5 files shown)


docs/api.md
```diff
@@ -1703,9 +1703,12 @@ you are opening.
 #### bot.openVillager(villagerEntity)
 
 Returns an `Villager` instance which represents the trading window you are opening.
+You can listen to the `ready` event on this `Villager` to know when it's ready
 
 #### bot.trade(villagerInstance, tradeIndex, [times], [cb])
 
+This function also returns a `Promise`, with `void` as its argument upon completion.
```
wvffle, line 1710: There are many javascript developers that are unfamiliar with void concept.

wvffle (changes requested):

docs/api.md
wvffle: And it's not an argument, rather a resolved value.

ph0t0shop: I just mean that it's a void promise, there is no argument type

ph0t0shop: All the other promise functions also contain this line

rom1504 (approved):

PR #1485, opened 2020-12-05
Trying out for await instead of normal listeners

rom1504:


rom1504:
It's not particularly better in these examples
It could be useful for internal plugin code

PR #1486, opened 2020-12-06
fix incorrect handling of username definition

rom1504:
The right place to set the username mineflayer side is the login packet.
At this point both in offline and online mode, all information about the username is known.

The previous behavior relied on an incorrect assumption that the username definition was always available in the same event loop iteration than createClient call.


rom1504:
more explanation there https://github.com/PrismarineJS/node-minecraft-protocol/pull/787#issuecomment-739521652

PR #1487, opened 2020-12-06
Fix looking behavior

ph0t0shop:
See https://github.com/PrismarineJS/mineflayer-pathfinder/issues/81

### DIFF
--- a/lib/plugins/physics.js
+++ b/lib/plugins/physics.js
@@ -198,7 +198,7 @@ function inject (bot) {
   let lookingTask = createDoneTask()
 
   bot.on('move', () => {
-    if (Math.abs(deltaYaw(bot.entity.yaw, lastSentYaw)) < 0.001 && !lookingTask.done) {
+    if (!lookingTask.done && Math.abs(deltaYaw(bot.entity.yaw, lastSentYaw)) < 0.001) {
       lookingTask.finish()
     }
   })
--- a/lib/plugins/physics.js
+++ b/lib/plugins/physics.js
@@ -3,8 +3,8 @@ const assert = require('assert')
 const math = require('../math')
 const conv = require('../conversions')
 const { performance } = require('perf_hooks')
-const { callbackify } = require('../promise_utils')
-const { once } = require('events')
+const { callbackify, createDoneTask } = require('../promise_utils')
+const { on } = require('events')
 
 const { Physics, PlayerState } = require('prismarine-physics')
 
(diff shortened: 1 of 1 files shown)


lib/plugins/physics.js
```diff
@@ -196,26 +195,37 @@ function inject (bot) {
...
     }
 
-    do {
-      await once(bot, 'move')
-    } while (Math.abs(deltaYaw(bot.entity.yaw, lastSentYaw)) >= 0.001)
+    await lookingTask.promise
   })
 
-  bot.lookAt = callbackify(async (point, force) => {
+  bot.lookAt = (point, force, cb) => {
```
rom1504, line 222: why did you add back the cb ?

lib/plugins/physics.js
```diff
@@ -196,26 +195,37 @@ function inject (bot) {
     })
   }
 
+  let lookingTask = createDoneTask()
+
+  bot.on('move', () => {
+    if (Math.abs(deltaYaw(bot.entity.yaw, lastSentYaw)) < 0.001 && !lookingTask.done) {
```
rom1504, line 201: better put the lookingTask.done first in the condition so it doesn't do the computation if not needed

rom1504:
left 2 comments, but overall it looks good, I like the use of the task here

lib/plugins/physics.js
ph0t0shop: Because this is just the lookAt function, it is a proxy function for the look function. All the functions still take cb, but this one I made it just return the promise of look() after calculating those values (yaw and pitch). Otherwise it needed an await in there.

lib/plugins/physics.js
ph0t0shop: You're right... short-circuiting for the win

lib/plugins/physics.js
rom1504: what's the problem for having an await ?

ph0t0shop: Not really a problem, this just seemed the most pragmatic, it being a proxy function it doesn't really need to be async.

rom1504:
@ph0t0shop to clarify, https://gist.github.com/rom1504/6a9127764b43e38569e704f94ac99908 will always print a b c

ph0t0shop:
@rom1504 yeah I understand that now, it just seems unnecessary wrapping a promise inside another promise. But we can do it if you want

lib/plugins/physics.js
rom1504: the problem with doing it that way is that it will make it harder to remove callbacks the day we want to
I think it's better in mineflayer internals to never call functions with their callbacks
and to callbackify them from the outside
That way removing the callback can be done simply by removing the callbackify, nothing else

ph0t0shop: Ok fine, done

rom1504 (approved):

PR #1491, opened 2020-12-10
Fixed && Improved blockFind function with useExtraInfo = true

magicaltoast:
Basically, it's correcting the previous behavior with a boolean value as an argument to expected (Matching function get extra info)

And adds two-stage matching to allow e.g. matching signs with specific text

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1333,7 +1333,9 @@ Finds the closest blocks from the given point.
  * `options` - Options for the search:
    - `point` - The start position of the search (center). Default is the bot position.
    - `matching` - A function that returns true if the given block is a match. Also supports this value being a block id or array of block ids.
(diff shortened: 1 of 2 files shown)


Karang:
Sounds good. Could you document the new function behavior in https://github.com/PrismarineJS/mineflayer/blob/master/docs/api.md ?

magicaltoast:
@Karang 
Yes, sure!
Maybe do you know how to edit this commit? Or should I just create a new one with updated docs?

Karang:
> @Karang
> Yes, sure!
> Maybe do you know how to edit this commit? Or should I just create a new one with updated docs?

You can just add a new commit on your branch, it will be added to this PR automatically.

magicaltoast:
@Karang docs updated

TheDudeFromCI (approved):

rom1504 (approved):

PR #1493, opened 2020-12-10
Make the ChatMessage.toAnsi:lang argument optional

Antonio32A:
According to the example ansi.js on line 21 (https://github.com/PrismarineJS/mineflayer/blob/4afb69d83fc58008c4d96bcde78fbca4954dcc8e/examples/ansi.js#L21), the `lang` argument should be optional, similar to `toMotd`.
If I'm incorrect, feel free to correct me.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -437,7 +437,7 @@ export class ChatMessage {
 
   toMotd(lang?: { [key: string]: string }): string;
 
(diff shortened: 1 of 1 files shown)


SiebeDW (approved):
LGTM

wvffle (approved):

rom1504 (approved):

PR #1528, opened 2020-12-23
[Quality Of Life] Added waitForTicks() Function

TheDudeFromCI:
This is a simple promise-based function that allows for waiting for a given number of physics ticks before continuing. While setTimeout exists, it's nice to also have this since it respects the physics timer of the bot. If the physics speed is turned up or down, this timer respects that. In addition, it also respects bot lag spikes as well, which is nice for certain situations regarding very specific physics inputs.

It's also worth noting that a lot of events in Minecraft make more sense to calculate based on the tick speed, so this just balances that out nicely.

Usage:
```js
async function afkCobblestoneFarm(bot, blockPos) {
  for (let i = 0; i < 1000; i++) {
    await bot.dig(bot.blockAt(blockPos))
    await bot.waitForTicks(100)
  }
}
```


docs/api.md
```diff
@@ -1738,6 +1739,10 @@ This can be used to check is a specific feature is available in the current Mine
 
 The list of available features can be found inside the [./lib/features.json](https://github.com/PrismarineJS/mineflayer/blob/master/lib/features.json) file.
 
+#### bot.waitForTicks(ticks)
+
+This is a promise-based function that waits for a given number of in-game ticks to pass before continuing. This is useful for quick timers that need to function with specific timing, regardless of the given physics tick speed of the bot. This is similar to the standard Javascript setTimeout function, but runs on the physics timer of the bot specifically.
```
wvffle, line 1744: What about callback users?

TheDudeFromCI: Oh, I thought callbacks were only being maintained for backwards compatibility? I can add it if we intend to maintain both.

extremeheat: I don’t see the need for a callback here. They’re primarily useful when you need to callback multiple times. Most code should update away from callbacks to async/promises as they’re easier to read  and use less nesting. The existing callbacks could be deprecated and eventually removed at some point in the future.

wvffle: > I thought callbacks were only being maintained for backwards compatibility?

That's why I commented and not requested changes :D I didn't know if you guys decided on that.

> I don’t see the need for a callback here.

I can understand the basic argument of moving away from callbacks, though currently we are supporting both approaches.
If we add new features without the callbacks, we (to some degree) are forcing callback users to mix both things.

> Most code should update away from callbacks to async/promises as they’re easier to read and use less nesting

Most developers don't know how to use promises the right way and gain performance from this. Nesting is nothing bad and it's easy to read if you know how to style your code. 

> The existing callbacks could be deprecated and eventually removed at some point in the future.

Adding a new callback function does not prevent that.

extremeheat: > Most developers don't know how to use promises the right way and gain performance from this. Nesting is nothing bad and it's easy to read if you know how to style your code.

Yeah, I get that. Just that intermixing promises and callbacks is kind of odd. If you're using promises/async in your code, and want to use something with a function callback that means you may have to wrap around that code like this to wait for it: 
```js
async function dig(block, look) { return await new Promise(res => bot.dig(block, look, (...args) => res(...args))) }
```
If dig were to return a promise, all you need is `bot.dig().then(...)` versus the callback form of `bot.dig(() => ...)`. The point is it's easier to handle promises if you're using callbacks vs. handling callbacks when you're using promises/async. But yes, since this supports both promise/cb, it's just dependent on whether or not people should be encouraged to switch to promises.

TheDudeFromCI: Alright, then I'll add the callback for the sake of completeness, but maybe in a 3.0.0 _way_ in the future we can remove all callbacks.

wvffle: > but maybe in a 3.0.0 way in the future

Good luck with 3.0.0, haha. I'd make the cut and go to the new major version way back then ;p

Karang: Callback users still have the option to callbackify the function themselves. But imo there is no need to encourage it.

TheDudeFromCI: Since this seems to be a bigger discussion, I'm moving this topic to a dedicated thread to decide the path to take for all future APIs. https://github.com/PrismarineJS/mineflayer/discussions/1532

index.js
```diff
@@ -40,7 +40,8 @@ const plugins = {
   spawn_point: require('./lib/plugins/spawn_point'),
   tablist: require('./lib/plugins/tablist'),
   time: require('./lib/plugins/time'),
-  villager: require('./lib/plugins/villager')
+  villager: require('./lib/plugins/villager'),
+  util: require('./lib/plugins/util')
```
rom1504, line 44: util is a name that should be avoided
this could go in physics or time files

TheDudeFromCI: Okay. What should I call it? Helper?

rom1504: helper has the same issue as util: it can contain anything and everything
why not put this function in physic.js as this is where physicTick is defined ?

TheDudeFromCI: Ooh, I misunderstood you. Okay, thank makes sense. At first I thought you said there was a file called util that contained physics based utility functions.

I'll move it to physics, sure.

TheDudeFromCI: Done.

rom1504 (approved):

PR #1533, opened 2020-12-24
Better calculation of digging range.

goncharovchik:
Before the change, the bot that used `bot.canDigBlock()` only dig 4 blocks up, but the player digs 5 up. I tested the function in different directions and achieved a perfect calculation.

### DIFF
--- a/lib/plugins/digging.js
+++ b/lib/plugins/digging.js
@@ -98,7 +98,7 @@ function inject (bot) {
   }
 
   function canDigBlock (block) {
-    return block && block.diggable && block.position.offset(0.5, 0.5, 0.5).distanceTo(bot.entity.position.offset(0, 2, 0)) <= 5.2
+    return block && block.diggable && block.position.offset(0.5, 0.5, 0.5).distanceTo(bot.entity.position.offset(0, 1.65, 0)) <= 5.1
   }
 
   function digTime (block) {
--- a/lib/plugins/digging.js
+++ b/lib/plugins/digging.js
@@ -98,7 +98,7 @@ function inject (bot) {
   }
 
   function canDigBlock (block) {
-    return block && block.diggable && block.position.offset(0.5, 0.5, 0.5).distanceTo(bot.entity.position) < 6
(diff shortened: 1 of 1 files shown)


lib/plugins/digging.js
```diff
@@ -98,7 +98,7 @@ function inject (bot) {
   }
 
   function canDigBlock (block) {
-    return block && block.diggable && block.position.offset(0.5, 0.5, 0.5).distanceTo(bot.entity.position) < 6
+    return block && block.diggable && block.position.offset(0.5, 0.5, 0.5).distanceTo(bot.entity.position.offset(0, 2, 0)) <= 5.2
```
TheDudeFromCI, line 101: Use `.offset(0, 1.65, 0)` to use the player's eye height.

TheDudeFromCI (changes requested):

TheDudeFromCI (approved):

wvffle (approved):

rom1504:
Is there any external reference about this ?
This will reduce the digging distance for block that are below, is it expected ?

goncharovchik:
> This implementation is more accurate to vanilla. The old implementation was a poor approximation.

I completely agree. If I'm not mistaken, the player digs 2-3 blocks and presumably a bot with such a change will be ± also

rom1504:
ok, would still be great to find any reference confirming this is how it's implemented in vanilla

ph0t0shop:
Still not completely accurate, since the client ray casts to the block _face_, not just the block. This also means, when looking along the edge of the block, a part of a block face might be within range, while another part of the same face might not be.

Karang:
To be accurate it should use the raycast iterator, but this is already better than what was before

goncharovchik:
> Still not completely accurate, since the client ray casts to the block _face_, not just the block. This also means, when looking along the edge of the block, a part of a block face might be within range, while another part of the same face might not be.


hmm, well, it seems that I did 5.1 for this, and not just 5

goncharovchik:
> To be accurate it should use the raycast iterator, but this is already better than what was before

It seems to me that there may be problems with this. Although. I'll try later

PR #1536, opened 2020-12-25
Emit death once

extremeheat:
Fixes an issue where death is emitted after bot is already dead (untested in-game, so someone should check to verify)

### DIFF
--- a/lib/plugins/health.js
+++ b/lib/plugins/health.js
@@ -20,8 +20,10 @@ function inject (bot) {
     bot.foodSaturation = packet.foodSaturation
     bot.emit('health')
     if (bot.health <= 0) {
-      bot.isAlive = false
-      bot.emit('death')
+      if (bot.isAlive) {
+        bot.isAlive = false
+        bot.emit('death')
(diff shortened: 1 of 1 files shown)


lib/plugins/health.js
```diff
@@ -20,8 +20,10 @@ function inject (bot) {
     bot.foodSaturation = packet.foodSaturation
     bot.emit('health')
     if (bot.health <= 0) {
+      if (bot.isAlive) {
+        bot.emit('death')
```
TheDudeFromCI, line 24: `bot.isAlive = false` should be called before emitting the death event.

TheDudeFromCI (changes requested):

extremeheat:
It was discussed on discord whether or not 'client_command' should be resent twice, since it hasn't caused issues so far, we can leave it as-is. Maybe someone can look into the vanilla implementation and see how it's implemented there later.

TheDudeFromCI (approved):

wvffle (approved):

Issue #1526, opened 2020-12-23
bot.openFurnace doesn't work with Smokers and Blast Furnaces


SoLegendary:
- [x] The [FAQ](https://github.com/PrismarineJS/mineflayer/blob/master/docs/FAQ.md) doesn't contain a resolution to my issue 

## Versions
 - mineflayer: 2.37.1
 - server: vanilla 1.16.3
 - node: 6.14.8

## Detailed description of a problem

bot.openFurnace(furnaceBlock) doesn't work with specialised types of furnaces (blast furnaces and smokers)

## What did you try yet?

Placed down regular furnace, smoker and blast furnace but only the regular furnace returns a valid response while the others give the below error.

**EDIT: I can do a very hacky fix by manually renaming the Block object to 'furnace' from 'smoker' or 'blast_furnace' but this really isn't such a great long term solution**

**EDIT 2: This works to trigger the furnace.on() events but any methods like adding items/fuel won't work (no error but bot does nothing)**

## Your current code
```js
let furnaceBlock = bot.findBlock({
  matching: mcdata.blocksByName.blast_furnace.id,
  count: 1,
  maxDistance: 4,
})
let furnace = bot.openFurnace(furnaceBlock)
```
Something went wrong processing a chat command:
AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:
```js
  assert.ok(furnaceBlock.name === 'furnace' || furnaceBlock.name === 'lit_furnace')
```

## Expected behavior
Expected it to return a Furnace object with all its usual methods like add fuel, check time etc.

## Additional context
I don't really mind if it fails to recognise what kind of items can be put in or not (eg. trying to use food in a blast furnace), I can easily handle that in my own code. I just want to be able to use them like regular furnaces.

rom1504:
fixed

PR #1538, opened 2020-12-26
Allow opening of smoker/blast furnace

ArcticZeroo:
This PR adds the ability to use `bot.openFurnace` on the smoker and blast furnace type furnaces. This currently does not validate items being put into their correct furnaces (e.g. only food into the smoker, only ores into the blast furnace), but that seems to be the case for the base furnace as well.

External tests also now test for the ability to open all types of furnaces rather than just the base furnace. 

I also noticed that tasks could be cancelled more than once, and if task.finish is called after task.cancel, the promise will throw an error because it has already rejected. This is now not possible behavior.

This PR is "complete", but I was unable to run local tests on my windows machine, so I can't verify whether the changes to the test work.


rom1504:
tests are not passing

ph0t0shop:
Just to link them, this would fix #1526.

ArcticZeroo:
Oh whoops, totally forgot to link to the issue, thanks @ph0t0shop 

Regarding the tests, @TheDudeFromCI I was seeing a bunch of errors both locally and on gitpod that rom thought may have been due to an incorrect java version. When I get a some free time, I will see if  I can get local tests to start running and diagnose the issue.

ArcticZeroo:
FYI, waiting for https://github.com/PrismarineJS/mineflayer/pull/1586/ to land until I look into these test failures

rom1504:
https://github.com/PrismarineJS/mineflayer/pull/1586 is now merged, please rebase

Karang:
done in #1640

PR #1540, opened 2020-12-28
Gravity is 0.08 and not null

SWEDCODER:
This small little change fixes stuff when you call bot.creative.stopFlying(); and that you had a previously (even stopped) goal with mineflayer-pathfinder,
going to a goal again would not work at all after calling bot.creative.stopFlying();

### DIFF
--- a/lib/plugins/creative.js
+++ b/lib/plugins/creative.js
@@ -2,8 +2,7 @@ const assert = require('assert')
 const Vec3 = require('vec3').Vec3
 const { callbackify, sleep } = require('../promise_utils')
 const { once } = require('events')
-const { Physics } = require("prismarine-physics")
-
+const { Physics } = require('prismarine-physics')
 
 module.exports = inject
 
@@ -11,6 +10,7 @@ function inject (bot, { version }) {
   const Item = require('prismarine-item')(version)
   const mcData = require('minecraft-data')(bot.version)
   const constantGravity = new Physics(mcData).gravity
+  
   // these features only work when you are in creative mode.
   bot.creative = {
     setInventorySlot: callbackify(setInventorySlot),
@@ -44,7 +44,6 @@ function inject (bot, { version }) {
     creativeSlotsUpdates[slot] = false
   }
 
- 
   const flyingSpeedPerUpdate = 0.5
 
   // straight line, so make sure there's a clear path.
--- a/lib/plugins/creative.js
+++ b/lib/plugins/creative.js
@@ -41,7 +41,7 @@ function inject (bot, { version }) {
     creativeSlotsUpdates[slot] = false
   }
 
-  let normalGravity = 0.08;
+  let normalGravity = 0.08
   const flyingSpeedPerUpdate = 0.5
 
   // straight line, so make sure there's a clear path.
--- a/lib/plugins/creative.js
+++ b/lib/plugins/creative.js
@@ -10,7 +10,6 @@ function inject (bot, { version }) {
   const Item = require('prismarine-item')(version)
   const mcData = require('minecraft-data')(bot.version)
   const constantGravity = new Physics(mcData).gravity
-  
   // these features only work when you are in creative mode.
   bot.creative = {
     setInventorySlot: callbackify(setInventorySlot),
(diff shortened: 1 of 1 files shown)


lib/plugins/creative.js
```diff
@@ -41,7 +41,7 @@ function inject (bot, { version }) {
     creativeSlotsUpdates[slot] = false
   }
 
-  let normalGravity = null
+  let normalGravity = 0.08
```
wvffle, line 44: After the change line L75 is never executed. Also why exactly 0.08?

wvffle (changes requested):

rom1504:
https://github.com/PrismarineJS/mineflayer/blob/d3401a6f9c95db2de2857a7dfa5a21778bc6ff2e/lib/plugins/creative.js#L75

bot.physics.gravity should be used

in what case is it not used today?

Karang:
This normalGravity variable shouldn't exist at all. It should use `bot.physics.gravity` directly.

SWEDCODER:
When you debug bot.physics.gravity the whole time it is 0.08, when you call the function to stop flying though, it sets to null and messes some stuff with mineflayer-pathfinder causing it to not work at all.

wvffle (changes requested):
As @TheDudeFromCI said

Karang:
I think this PR is fine. Its not easy to not hardcode the value because the only place where it is available `bot.physics.gravity` is the variable that has to be set...

The flying logic should be refactored and put in prismarine-physics. Because setting the gravity isn't enough to cover all cases (different gravity in liquids for instance).

If implemented in physics it would be closer to vanilla and easier to work with, but IMO that's outside the scope of this PR.

SWEDCODER:
Ok the var now comes from prismarine-physics directly, (sorry for the multiple commits too, forgot about eslint)

lib/plugins/creative.js
```diff
@@ -2,12 +2,14 @@ const assert = require('assert')
...
 const { once } = require('events')
+const { Physics } = require('prismarine-physics')
 
 module.exports = inject
 
 function inject (bot, { version }) {
   const Item = require('prismarine-item')(version)
-
+  const mcData = require('minecraft-data')(bot.version)
+  const constantGravity = new Physics(mcData).gravity
```
Karang, line 12: Physics is not a constructor but a regular function, new is not needed. This is a bit over-complicated, but I'm fine with it as it should be temporary.

SWEDCODER: Oh oops my bad, is minecraft-data installed by default too btw ? because that'd be a problem if it doesn't come in the package lol,

SWEDCODER: I don't see how it is over-complicated though, 
since you asked me to import prismarine js, the most obvious way was to call the function to get gravity from it, how would you have done this ?

Karang: It's over complicated to create a new physic engine just to get a constant. But as I said, I'm fine with it. At the moment there isn't a nice way of doing it, we'll need to rework a bit prismarine-physics.

lib/plugins/creative.js
```diff
@@ -2,12 +2,14 @@ const assert = require('assert')
...
 const { once } = require('events')
+const { Physics } = require('prismarine-physics')
 
 module.exports = inject
 
 function inject (bot, { version }) {
   const Item = require('prismarine-item')(version)
-
+  const mcData = require('minecraft-data')(version)
+  const constantGravity = Physics(mcData).gravity
```
wvffle, line 12: Can't we just do that:

```js
  const constantGravity = bot.physics.gravity
```

Or set it in physics plugin? Like `bot._constantGravity` and then use it in `stopFlying`?

SWEDCODER: Ok I'll just do that then

rom1504:
tests are failing

lib/plugins/creative.js
wvffle: It was a comment for a discussion, not a request to do changes :)

Simply setting `constantGravity` to `bot.physics.gravity` is actually kind of dangerous, as physics plugin may not load before creative. In such case, it'll throw an exception that bot.physics is undefined. The second approach is a bit better but still should be discussed.

rom1504:
please reopen if you want to finish it

PR #1542, opened 2020-12-28
Move iterators to pworld

rom1504:


rom1504:
I also want to move findBlock but for this I need to deprecate blockAt (part of https://github.com/PrismarineJS/mineflayer/issues/334#issuecomment-670987395 )

Issue #1497, opened 2020-12-11
bot.time has negative values


wvffle:
- [x] The [FAQ](https://github.com/PrismarineJS/mineflayer/blob/master/docs/FAQ.md) doesn't contain a resolution to my issue 

## Versions
 - mineflayer: 2.37.1
 - server: vanilla/spigot/paper 1.15.2
 - node: 15.2.0

## Detailed description of a problem
The `bot.time.day`, `bot.time.time`, `bot.time.timeOfDay` and `bot.time.moonPhase` are negative values when connected to a Spigot server which prevents going to sleep

## What did you try yet?
Nothing
## Your current code
```js
console.log(bot.time)
```

## Expected behavior
The values should be positive

wvffle:
I just tested it on vanilla 1.16.4 and the same issue happened

TheDudeFromCI:
What plugin is it? The implementation might point to the exact cause.

Naomi-alt:
From my limited testing, it seems to be something to do with the gamerule `doDaylightCycle`?  
When the rule is true, everything works fine, but with it false I get negative numbers.

ph0t0shop:
@Naomi-alt absolutely, I noticed the exact same. That is an issue even on vanilla servers, but there might be more issues.

Naomi-alt:
I did some more testing, and maybe I'm not understanding it correctly, but this is my take on it.

If you set the time to noon using `/time set noon` the time packet will be `[ 0, 6000 ]`, so far so good.  
But then when you disable the `doDaylightCycle` the time packet will be `[ -1, -6000 ]`

[wiki.vg](https://wiki.vg/Protocol#Time_Update) say this is because a negative number means that the sun isn't moving.   (dayLightyCycle = false) and the time of day will be the absolute value of the negative number.  
And I'm guessing the first -1 is because the protocol is using [Two's complement](https://en.wikipedia.org/wiki/Signed_number_representations#Two's_complement) to encode the numbers.

So now somebody who actually knows what they're doing should improve the `longToNumber` to also work with negative numbers.  
I couldn't figure out a nice way of doing it.

ph0t0shop:
@Naomi-alt those packets you listed, are they the parsed packets that mineflayer receives already? Or are they simply an example you constructed, of how the server sends them and how mineflayer _should_ receive them instead?

Naomi-alt:
[...]

PR #1543, opened 2020-12-29
Fix for not handling negative numbers in time plugin (#1497)

Naomi-alt:
Add missing initial values

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -126,6 +126,7 @@
       - [bot.food](#botfood)
       - [bot.foodSaturation](#botfoodsaturation)
       - [bot.physics](#botphysics)
+      - [bot.time.doDaylightCycle](#bottimedodaylightcycle)
       - [bot.time.time](#bottimetime)
       - [bot.time.timeOfDay](#bottimetimeofday)
       - [bot.time.day](#bottimeday)
@@ -934,6 +935,10 @@ saturation of 5.0. Eating food increases the saturation as well as the food bar.
 
 Edit these numbers to tweak gravity, jump speed, terminal velocity, etc.
 Do this at your own risk.
+
+### bot.time.doDaylightCycle
+
+Whether or not the gamerule doDaylightCycle is true or false.
 
 #### bot.time.time
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -127,11 +127,13 @@
       - [bot.foodSaturation](#botfoodsaturation)
       - [bot.physics](#botphysics)
       - [bot.time.doDaylightCycle](#bottimedodaylightcycle)
+      - [bot.time.bigTime](#bottimebigtime)
       - [bot.time.time](#bottimetime)
       - [bot.time.timeOfDay](#bottimetimeofday)
       - [bot.time.day](#bottimeday)
       - [bot.time.isDay](#bottimeisday)
       - [bot.time.moonPhase](#bottimemoonphase)
+      - [bot.time.bigAge](#bottimebigage)
       - [bot.time.age](#bottimeage)
       - [bot.quickBarSlot](#botquickbarslot)
       - [bot.inventory](#botinventory)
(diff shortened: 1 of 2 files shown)


lib/plugins/time.js
```diff
@@ -2,20 +2,45 @@ module.exports = inject
...
 function longToNumber (arr) {
-  return arr[1] + 4294967296 * arr[0]
+  const str = (arr[0] >>> 0).toString(2) + (arr[1] >>> 0).toString(2)
+  if (str[0] === '1') {
+    // If the first character is a 1, it means the number is negative.
+    // To parse the number we first need to split the string to an array.
+    // Then we can flip each bit, and join it to a string again.
+    // Then the string is parsed to an int using base 2.
+    // And lastly the result is negated and subtracted by 1.
+    return parseInt(str.split('').map(bit => 1 - bit).join(''), 2) * -1 - 1
```
wvffle, line 42: Just a suggestion:
```js
return -1 - parseInt(str, 2) ^ 2**str.length - 1
```

Naomi-alt: Hmm, it doesn't seem to work correctly.

I'm starting to think that `parseInt` somehow doesn't work bit such large numbers?  
When I do `console.log(parseInt('1111111111111111111111111111111111111111111111111111111111111111', 2))` (64 1s) I should just get 2^64 right?  
But instead I get 18446744073709552000 which is `18446744073709552000 - 2^64 = 384` off.

2^64 = 18446744073709551616, so the last 3 digits are rounded up it seems.

Karang: That's because the largest representable integer in js is `2^53-1` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER

Naomi-alt: Welp, that makes sense then.

So I guess the best way to parse the number is to just flip it and parse it as a positive number first, like it is now?  
That way the inaccuracy at least doesn't show up with smaller numbers.

extremeheat: The ideal way to fix issues like this I think is to read 64bit integers into BigInts in protodef :)

In order to know if the number is negative, all you need is the upper half of the integer, you don’t need to go through the process of merging them. If you do want to merge them, do it the right way and use a BigInt.

extremeheat: untested, but something like this should be able to read it as a signed BigInt: 
`BigInt.asIntN(64, (BigInt(upper) << 32n)) | BigInt(lower)`

Naomi-alt: Using a BigInt seems to works, but it does mean that all the values (bot.time.time etc) also need to be a BigInt, as converting it to a normal number wouldn't make sense right?  
I don't know whether using a BigInt for those values is a good thing or not?

extremeheat: This would be an API break, but IMO if it’s unlikely that the value will ever get that high into the 64 bit integer other than negative numbers, it’s not really necessary for the breaking change. Because of the way JS works, numbers and BigInts cannot interact with each other without being converted. Maybe an additional .tickTime or something similar could be added to store the precise tick time if the numbers can get really big.

Naomi-alt: Yeah that would seem like the most sensible option.

I've implemented the changes and also added something to the docs to explain when to use which value.

Let me know if it needs some improvements.

wvffle (approved):
Awesome. Can we have tests for this? just in case

docs/api.md
```diff
@@ -935,9 +938,22 @@ saturation of 5.0. Eating food increases the saturation as well as the food bar.
...
+
+This value is of type BigInt and is accurate even at very large values. (more than 2^51 - 1 ticks)
+
 #### bot.time.time
 
-Total time of the world since day 0.
+The total numbers of ticks since day 0.
+
+Because the Number limit of Javascript is at 2^51 - 1 bot.time.time becomes inaccurate higher than this limit and the use of bot.time.bigTime is recommended.  
+Realistically though you'll probably never need to use bot.time.bigTime as it will only reach 2^51 - 1 ticks naturally after ~14280821 real years.
```
rom1504, line 956: so why is bigTime exposed then ?

rom1504 (changes requested):

docs/api.md
Naomi-alt: My thought was that in a normal situation you'd probably never need it.  
But if someone was working with some highly modified server that messed with the time a lot, it might reach the point where it becomes inaccurate.  
So I guess I want to give to option if someone ever needs it?

rom1504 (approved):

PR #1554, opened 2020-12-31
add auth as an option for createbot

u9g:


docs/api.md
```diff
@@ -765,6 +765,7 @@ Create and return an instance of the class bot.
  * password : can be omitted (if the tokens are also omitted then it tries to connect in offline mode)
  * host : default to localhost
  * version : default to automatically guessing the version of the server. Example of value : "1.12.2"
+ * auth : default to mojang
```
rom1504, line 768: Can you mention it can be microsoft?

rom1504 (changes requested):

docs/api.md
u9g: added

rom1504 (approved):

PR #1568, opened 2021-01-04
avoid typescript compile errors in strict mode

Rob9315:
the constructor of the "interface" Bot had no implicit return type, but should have been Bot


rom1504:
Ah but that's incorrect, there is actually no constructor. It should simply be removed from the interface

rom1504 (approved):

PR #1586, opened 2021-01-12
Promisify testCommon  

ArcticZeroo:
This PR replaces all `bot.test.*` methods which took callbacks with methods that now return promises instead. Support for callbacks is still retained however, using `callbackify`. Because of this, I have chosen not to promisify all the tests themselves for now. `bot.test.callbackChain` is unmodified as well, so all existing tests still work. 

The main benefit of this is to allow us to convert each test individually if desired (as they're modified in the future), rather than needing to sit down for hours on end to tediously convert each test from callbacks to promises.

### DIFF
--- a/lib/promise_utils.js
+++ b/lib/promise_utils.js
@@ -58,7 +58,7 @@ function onceWithCleanup (emitter, event, { timeout = 0, checkCondition = undefi
   const task = createTask()
 
   const onEvent = (...data) => {
-    if (checkCondition && !checkCondition(...data)) {
+    if (typeof checkCondition === 'function' && !checkCondition(...data)) {
       return
     }
 
@@ -67,9 +67,9 @@ function onceWithCleanup (emitter, event, { timeout = 0, checkCondition = undefi
 
   emitter.addListener(event, onEvent)
 
-  if (timeout > 0) {
+  if (typeof timeout === 'number' && timeout > 0) {
     // For some reason, the call stack gets lost if we don't create the error outside of the .then call
-    const timeoutError = new Error(`Event ${event} did not fire within timeout of ${timeout}ms ${checkCondition && checkCondition.name}`)
+    const timeoutError = new Error(`Event ${event} did not fire within timeout of ${timeout}ms`)
     sleep(timeout).then(() => {
       if (!task.done) {
         task.cancel(timeoutError)
@@ -83,21 +83,12 @@ function onceWithCleanup (emitter, event, { timeout = 0, checkCondition = undefi
 }
 
 function withTimeout (promise, timeout) {
-  return new Promise((resolve, reject) => {
-    const timeoutHandle = setTimeout(() => reject(new Error('Promise timed out.')), timeout)
-    promise
-      .finally(() => {
-        clearTimeout(timeoutHandle)
-      })
-      .then((result) => {
-        resolve(result)
-        return result
-      })
-      .catch((err) => {
-        reject(err)
-        return err
-      })
-  })
+  return Promise.race([
+    promise,
+    sleep(timeout).then(() => {
+      throw new Error('Promise timed out.')
+    })
+  ])
 }
 
 module.exports = {
--- a/lib/promise_utils.js
+++ b/lib/promise_utils.js
@@ -1,7 +1,10 @@
 function callbackify (f) {
   return function (...args) {
     const cb = args[f.length]
-    return f(...args).then(r => { if (cb) { cb(undefined, r) } return r }, err => { if (cb) { cb(err) } else throw err })
+    return f(...args).then(r => {
+      if (cb) { cb(undefined, r) }
+      return r
+    }, err => { if (cb) { cb(err) } else throw err })
   }
 }
 
@@ -15,12 +18,16 @@ function createTask () {
   }
   task.promise = new Promise((resolve, reject) => {
     task.cancel = (err) => {
-      reject(err)
-      task.done = true
+      if (!task.done) {
+        task.done = true
+        reject(err)
+      }
     }
-    task.finish = () => {
-      resolve()
-      task.done = true
+    task.finish = (result) => {
+      if (!task.done) {
+        task.done = true
+        resolve(result)
+      }
     }
   })
   return task
@@ -36,9 +43,66 @@ function createDoneTask () {
   return task
 }
 
+/**
+ * Similar to the 'once' function from the 'events' module, but allows you to add a condition for when you want to
+ * actually handle the event, as well as a timeout. The listener is additionally removed if a timeout occurs, instead
+ * of with 'once' where a listener might stay forever if it never triggers.
+ * Note that timeout and checkCondition, both optional, are in the third parameter as an object.
+ * @param emitter - The event emitter to listen to
+ * @param event - The name of the event you want to listen for
+ * @param [timeout=0] - An amount, in milliseconds, for which to wait before considering the promise failed. <0 = none.
+ * @param [checkCondition] - A function which matches the same signature of an event emitter handler, and should return something truthy if you want the event to be handled. If this is not provided, all events are handled.
+ * @returns {Promise} A promise which will either resolve to an *array* of values in the handled event, or will reject on timeout if applicable. This may never resolve if no timeout is set and the event does not fire.
+ */
+function onceWithCleanup (emitter, event, { timeout = 0, checkCondition = undefined } = {}) {
+  const task = createTask()
+
+  const onEvent = (...data) => {
+    if (checkCondition && !checkCondition(...data)) {
+      return
+    }
+
+    task.finish(data)
+  }
+
+  emitter.addListener(event, onEvent)
+
+  if (timeout > 0) {
+    sleep(timeout).then(() => {
+      if (!task.done) {
+        task.cancel(new Error(`Event ${event} did not fire within timeout of ${timeout}ms`))
+      }
+    })
+  }
+
+  task.promise.finally(() => emitter.removeListener(event, onEvent))
+
+  return task.promise
+}
+
+function withTimeout (promise, timeout) {
+  return new Promise((resolve, reject) => {
+    const timeoutHandle = setTimeout(() => reject(new Error('Promise timed out.')), timeout)
+    promise
+      .finally(() => {
+        clearTimeout(timeoutHandle)
+      })
+      .then((result) => {
+        resolve(result)
+        return result
+      })
+      .catch((err) => {
+        reject(err)
+        return err
+      })
+  })
+}
+
 module.exports = {
   callbackify,
   sleep,
   createTask,
-  createDoneTask
+  createDoneTask,
+  onceWithCleanup,
+  withTimeout
 }
--- a/lib/promise_utils.js
+++ b/lib/promise_utils.js
@@ -68,9 +68,11 @@ function onceWithCleanup (emitter, event, { timeout = 0, checkCondition = undefi
   emitter.addListener(event, onEvent)
 
   if (timeout > 0) {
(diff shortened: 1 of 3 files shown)


test/externalTests/plugins/testCommon.js
```diff
@@ -215,60 +204,73 @@ function inject (bot) {
...
+        if (timeout) clearTimeout(timeout)
+        console.log('kill process ' + child.pid)
 
-      child.once('close', (code) => {
-        console.log('close requested ' + code)
-        cb(err)
-      })
-      process.kill(child.pid, 'SIGTERM')
-    }
+        bot.removeListener('chat', onChatMessage)
```
ArcticZeroo, line 262: Listener was previously not being removed, and I was noticing issues with listener limits being hit in the test output

test/externalTests/plugins/testCommon.js
```diff
@@ -77,136 +78,124 @@ function inject (bot) {
...
-    ], cb)
+  async function resetState () {
+    await becomeCreative()
+    await clearInventory()
+    bot.creative.startFlying()
+    await teleport(new Vec3(0, 4, 0))
+    await bot.waitForChunksToLoad()
+    await resetBlocksToSuperflat()
+    await sleep(1000)
+    await clearInventory()
```
wvffle, line 123: You can use `Promise.all` to join stuff that can be executed in parallel

test/externalTests/plugins/testCommon.js
```diff
@@ -77,136 +78,124 @@ function inject (bot) {
...
-          bot.removeListener('message', onMessage)
-          clearTimeout(timeOut)
-          return cb()
-        case 'commands.generic.permission':
-          sayEverywhere('ERROR: I need to be an op (allow cheats).')
-          bot.removeListener('message', onMessage)
-        // at this point we just wait forever.
-        // the intention is that someone ops us while we're sitting here, then you kill and restart the test.
+  async function setCreativeMode (value) {
+    return new Promise((resolve) => {
```
wvffle, line 136: rather than having new Promise, use something like this:

```js
while (true) {
  const { translate } = once(bot, 'message')
  if (translate == 'gameMode.changed') {
    // ...
    return
  }
}
```

also, to timeout, simply use `Promise.race`

test/externalTests/plugins/testCommon.js
```diff
@@ -77,136 +78,124 @@ function inject (bot) {
...
+  async function teleport (position) {
     bot.chat(`/tp ${bot.username} ${position.x} ${position.y} ${position.z}`)
-    bot.on('move', function onMove () {
-      if (bot.entity.position.distanceTo(position) < 0.9) {
-        // close enough
-        bot.removeListener('move', onMove)
-        cb()
-      }
+
+    return new Promise(resolve => {
```
wvffle, line 191: doable with `once` function

test/externalTests/plugins/testCommon.js
wvffle: You can import `once` from events

test/externalTests/plugins/testCommon.js
```diff
@@ -215,60 +204,73 @@ function inject (bot) {
...
+    return bot.creative.flyTo(bot.entity.position.plus(delta))
   }
 
-  function tellAndListen (to, what, listen, done) {
-    function chatHandler (username, message) {
-      if (username === to && listen(message)) {
-        bot.removeListener('chat', chatHandler)
-        done()
+  async function tellAndListen (to, what, listen) {
+    return new Promise(resolve => {
```
wvffle, line 212: doable with `once`

test/externalTests/plugins/testCommon.js
```diff
@@ -215,60 +204,73 @@ function inject (bot) {
...
-    function joinHandler (message) {
-      if (message.json.translate === 'multiplayer.player.joined') {
-        bot.removeListener('message', joinHandler)
-        childBotName = message.json.with[0].insertion
-        bot.chat(`/tp ${childBotName} 50 4 0`)
-        setTimeout(() => {
-          bot.chat('loaded')
-        }, 5000)
+  async function runExample (file, run) {
+    return new Promise((resolve, reject) => {
```
wvffle, line 225: doable with `once`

wvffle (changes requested):

test/externalTests/plugins/testCommon.js
ArcticZeroo: I'm aware, but I simply ported the code from the callback version. It was not clear which actions can be parallel. As there are two clearInventory calls, there is clearly a defined order

test/externalTests/plugins/testCommon.js
ArcticZeroo: The event is not listening for one message, it is listening for the first which fits the pattern. Once is not applicable.

test/externalTests/plugins/testCommon.js
ArcticZeroo: ... Why would you do that though? That looks way worse

ArcticZeroo: OK, I changed my mind after thinking about it. I don't think Promise.race is a good solution though, since you cannot cancel the event listeners. Will probably make a helper function of some sort

test/externalTests/plugins/testCommon.js
wvffle:
```js
    await becomeCreative()
    bot.creative.startFlying()
    await Promise.all([
      clearInventory(),
      teleport(new Vec3(0, 4, 0))
    ])
    await bot.waitForChunksToLoad()
    await resetBlocksToSuperflat()
    await sleep(1000)
    await clearInventory()
```

or
```js
    await becomeCreative()
    bot.creative.startFlying()
    await Promise.all([
      clearInventory(),
      teleport(new Vec3(0, 4, 0)),
      bot.waitForChunksToLoad()
    ])
    await resetBlocksToSuperflat()
    await sleep(1000)
    await clearInventory()
```

lib/promise_utils.js
```diff
@@ -36,9 +43,68 @@ function createDoneTask () {
...
+ * @param event - The name of the event you want to listen for
+ * @param [timeout=0] - An amount, in milliseconds, for which to wait before considering the promise failed. <0 = none.
+ * @param [checkCondition] - A function which matches the same signature of an event emitter handler, and should return something truthy if you want the event to be handled. If this is not provided, all events are handled.
+ * @returns {Promise} A promise which will either resolve to an *array* of values in the handled event, or will reject on timeout if applicable. This may never resolve if no timeout is set and the event does not fire.
+ */
+function onceWithCleanup (emitter, event, { timeout = 0, checkCondition = undefined } = {}) {
+  const task = createTask()
+
+  const onEvent = (...data) => {
+    if (checkCondition && !checkCondition(...data)) {
```
wvffle, line 61:
```js
if (checkCondition instanceof Function && !checkCondition(...data)) {
```

lib/promise_utils.js
```diff
@@ -36,9 +43,68 @@ function createDoneTask () {
...
+    }
+
+    task.finish(data)
+  }
+
+  emitter.addListener(event, onEvent)
+
+  if (timeout > 0) {
+    // For some reason, the call stack gets lost if we don't create the error outside of the .then call
+    const timeoutError = new Error(`Event ${event} did not fire within timeout of ${timeout}ms ${checkCondition && checkCondition.name}`)
```
wvffle, line 72: use instanceof, same as above

lib/promise_utils.js
```diff
@@ -36,9 +43,68 @@ function createDoneTask () {
...
+      })
+      .then((result) => {
+        resolve(result)
+        return result
+      })
+      .catch((err) => {
+        reject(err)
+        return err
+      })
+  })
```
wvffle, line 100:
```js
return Promise.race([
  promise,
  sleep(timeout).then(() => { throw new Error('Promise timed out.') })
])
```

lib/promise_utils.js
```diff
@@ -36,9 +43,68 @@ function createDoneTask () {
...
+function withTimeout (promise, timeout) {
+  return new Promise((resolve, reject) => {
+    const timeoutHandle = setTimeout(() => reject(new Error('Promise timed out.')), timeout)
+    promise
+      .finally(() => {
+        clearTimeout(timeoutHandle)
+      })
+      .then((result) => {
+        resolve(result)
+        return result
```
wvffle, line 94: Why are you returning result or error there?

lib/promise_utils.js
```diff
@@ -36,9 +43,68 @@ function createDoneTask () {
...
+    sleep(timeout).then(() => {
+      if (!task.done) {
+        task.cancel(timeoutError)
+      }
+    })
+  }
+
+  task.promise.finally(() => emitter.removeListener(event, onEvent))
+
+  return task.promise
```
wvffle, line 82: Wouldn't that work?
```js
return Promise.race([
  task.promise,
  sleep(timeout).then(() => {
    if (!task.done) {
      task.cancel(timeoutError)
    }
          
    return task.promise
  })
]).finally(() => emitter.removeListener(event, onEvent))
```

test/externalTests/plugins/testCommon.js
```diff
@@ -77,136 +80,116 @@ function inject (bot) {
...
-          bot.removeListener('message', onMessage)
         // at this point we just wait forever.
         // the intention is that someone ops us while we're sitting here, then you kill and restart the test.
       }
       // console.log("I didn't expect this message:", jsonMsg);
+      return false
     }
-    bot.on('message', onMessage)
+
+    // For whatever reason, timeout = success, and there is no other failure condition in this promise
```
wvffle, line 156: I guess, that's not intended. Should probably throw an error..

wvffle (changes requested):

lib/promise_utils.js
wvffle: You may need to return task.promise from finally

test/externalTests/plugins/testCommon.js
ArcticZeroo: If I throw an error, tests start failing. I would agree this should throw an error, but IMO it is out of scope of this PR as this is the same behavior that has always been here.

lib/promise_utils.js
ArcticZeroo: In case anything else attaches to this promise, result/error need to be returned so that they can still use the result. Though there might not be any use cases for it in this PR, it is possible that someone may want to timeout on a promise in one place but handle it in parallel in another function.

lib/promise_utils.js
ArcticZeroo: For some reason I was thinking this would cause an unhandled rejection (but it doesn't), will do

lib/promise_utils.js
ArcticZeroo: It might work, but that sleep.then is really hard to read IMO. I think the current code is fine, it's pretty similar

lib/promise_utils.js
ArcticZeroo: instanceof what? you mean typeof?

lib/promise_utils.js
ArcticZeroo: The finally callback also doesn't take any params

wvffle: > is really hard to read IMO.

It's how you should use promises correctly

wvffle: If we'd return task in `task.cancel`, then it could be written as:

```
sleep(timeout).then(() => task.cancel(timeoutError).promise)
```

ArcticZeroo: > It's how you should use promises correctly

Promises don't expose done state, so I don't really think there is a "correct" way to do this.

rom1504 (approved):

PR #1588, opened 2021-01-15
added look event

Bayugoon:
When ever a look packet is sent to the server the bot will emit 'look' with an object with the old pitch/yaw


Bayugoon:
I need to make sure onGround still works

rom1504:
Can be done in this pr

rom1504 (changes requested):
can you add this new look event to the doc ?

PR #1613, opened 2021-01-24
Simplified inventorytest

u9g:
removed repeating code

### DIFF
--- a/test/externalTests/exampleInventory.js
+++ b/test/externalTests/exampleInventory.js
@@ -1,5 +1,4 @@
 const assert = require('assert')
-// const { callbackify } = require('util')
 
 const tests = [
   {
--- a/test/externalTests/exampleInventory.js
+++ b/test/externalTests/exampleInventory.js
@@ -62,6 +62,7 @@ module.exports = () => (bot, done) => {
     bot.chat('/give inventory iron_ore 64')
     bot.chat('/give inventory diamond_boots 1')
     const inventoryTest = Object.values(tests).map(testData => generateTest(testData, cb))
+    inventoryTest.push(removeCraftingTable)
     // start test
     setTimeout(() => {
       bot.test.callbackChain(inventoryTest, cb)
@@ -69,6 +70,10 @@ module.exports = () => (bot, done) => {
 
     function generateTest ({ command, wantedMessage }, cb) {
       return () => bot.test.tellAndListen(ign, command, makeListener(wantedMessage), cb)
+    }
+    function removeCraftingTable () {
+      bot.chat('/setblock ~2 ~ ~2 air')
+      return true
     }
   }, done)
 }
--- a/test/externalTests/exampleInventory.js
+++ b/test/externalTests/exampleInventory.js
@@ -62,7 +62,6 @@ module.exports = () => (bot, done) => {
     bot.chat('/give inventory iron_ore 64')
     bot.chat('/give inventory diamond_boots 1')
     const inventoryTest = Object.values(tests).map(testData => generateTest(testData, cb))
-    inventoryTest.push(removeCraftingTable)
     // start test
     setTimeout(() => {
       bot.test.callbackChain(inventoryTest, cb)
@@ -70,10 +69,6 @@ module.exports = () => (bot, done) => {
 
     function generateTest ({ command, wantedMessage }, cb) {
       return () => bot.test.tellAndListen(ign, command, makeListener(wantedMessage), cb)
-    }
-    function removeCraftingTable () {
-      bot.chat('/setblock ~2 ~ ~2 air')
-      return true
     }
   }, done)
 }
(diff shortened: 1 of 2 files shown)


u9g:
@TheDudeFromCI

test/externalTests/exampleInventory.js
```diff
@@ -1,38 +1,88 @@
...
       bot.test.callbackChain(inventoryTest, cb)
     }, 2000)
+
+    function generateTest ({ command, wantedMessage }, cb) {
+      return () => bot.test.tellAndListen(ign, command, makeListener(wantedMessage), cb)
+    }
+    function removeCraftingTable () {
+      bot.chat('/setblock ~2 ~ ~2 air')
+      return true
+    }
```
imharvol, line 77: Probably no need to remove the block after the test
https://github.com/PrismarineJS/mineflayer/blob/18d213b88a92accffd6e2cba7b18cdbd8b3c4a52/test/externalTests/plugins/testCommon.js#L68-L111

u9g: I undid the commit, it was just a quick attempt to make the tests pass

imharvol: Yeah, I've had some problems with the fishing tests apparently failing randomly in the past

u9g: I don't think it's failing, on local system, tests passed but fishing tests take an excruciating amount of time: 
`√ fishing (114777ms)`
compared to
`√ exampleInventory (7929ms)`

extremeheat: Looking at it. For one, I think there should probably be a timeout on createTask to avoid an indefinite hang.

extremeheat:
Example of how fishing logic could work, with cancelable promises (untested):
```js
async function fish() {
   if (bot._fishing) {
      bot._fishing.cancel(Error('fish called again'))
   }
   bot._fishing = createTask(async (res, rej) => {
      await activateItem() // throw bobber
      const bobber = await bot.once('spawn_entity_boober')
      bot._fishingBobber = bobber
      bot.on('partcies', (data) => {
         if (particles_are_bubbles_and_close_to_fishingBobber) res(has_fish)
      })
   }, { timeout: 60 * 1000 }, function onTimeout() {
      if (bot._fishingBobber) {
         await activateItem() // pull back bobber
      }
      delete bot._fishing
   })
   return bot._fishing
}

class Task extends Promise {
   constructor(handler) {
      let cancel
      super((res, rej) => {
         handler(res, rej)
         cancel = (...args) => rej(...args)
      })
      this.cancel = cancel
      // only needed if we want to know if promise has been resolved:
      for (const k of ['then', 'catch', 'finally']) {
         this[k] = (...args) => {
            this.done = true
            return super[k](...args)
         }
      }
   }
}

function createTask(handler, opts, onComplete) {
   const complete = async () => { await onComplete(); }
   return new Task((res, rej) => {
      if (opts?.timeout) setTimeout(rej, opts?.timeout)
      handler(res, rej)
   }).then(complete, complete)
}
```

rom1504:
This isn't ready for review, right ?

u9g:
> This isn't ready for review, right ?

I would be ready to merge, but the problem is that fishing test fails on timeout which is a test AFTER inventory test, i couldn't figure it out so I just left it up

test/externalTests/exampleInventory.js
```diff
@@ -1,38 +1,83 @@
...
+    bot.chat('/give inventory iron_ore 64')
+    bot.chat('/give inventory diamond_boots 1')
+    const inventoryTest = Object.values(tests).map(testData => generateTest(testData, cb))
+    // start test
     setTimeout(() => {
       bot.test.callbackChain(inventoryTest, cb)
     }, 2000)
+
+    function generateTest ({ command, wantedMessage }, cb) {
+      return () => bot.test.tellAndListen(ign, command, makeListener(wantedMessage), cb)
```
imharvol, line 71: The function it returns should receive a callback (which is passed by `bot.test.callbackChain()`), see for example https://github.com/PrismarineJS/mineflayer/blob/d792dbf44aeda53749e76c5689107492cead09d7/test/externalTests/bed.js#L13-L53

imharvol (changes requested):
The problems are due the callbacks which are a bit missleading.

As of the fishing test I don't know why it could be failing, but I'm pretty sure that after you fix this test, the fishing test will start working again.

test/externalTests/exampleInventory.js
Karang: 114s for the test seems indeed too long, according to wiki the time should be between 5s and 30s. This seems to indicate there is a problem with the catch detection. Once the issue is fixed, we could look into speeding the test by giving the bot a fishing rod with lure enchantment or making it rain (as it speed up by 20%, but i think we already do that).

u9g:
waiting on https://github.com/PrismarineJS/mineflayer/pull/1640 to continue

Karang:
duplicate of #1650

PR #1614, opened 2021-01-24
used promises instead of cb's

u9g:


rom1504 (changes requested):
can you use async await instead ?

PR #1617, opened 2021-01-24
added null or undefined check

u9g:
in the api doc here: https://github.com/PrismarineJS/mineflayer/blob/master/docs/api.md#botequipitem-destination-callback it says that equipitem should default to hand, but now, it doesn't, so this check would add that


Karang (approved):
LGTM

wvffle (approved):

PR #1619, opened 2021-01-26
Remake bot.fishing()

u9g:


lib/plugins/inventory.js
```diff
@@ -641,6 +602,81 @@ function inject (bot, { version, hideErrors }) {
     bot.emit(`setWindowItems:${window.id}`)
   })
 
+  async function fish () {
+    let bobberDestroyed, getBobber, onParticles
+
+    async function onFishingDone () {
```
extremeheat, line 608: add an err parameter here and if you get an error, throw it back otherwise fish() will succeed even if it fails to get anything

rom1504 (changes requested):
Can you explain in the PR description what is the objective here?

PR #1621, opened 2021-01-26
Fishing refactor

extremeheat:
Fishing now returns a promise that can be canceled. Also adds a timeout of 60 seconds to fish() to cancel if no fishes were found based on the timeout on the wiki. It throws on a timeout and entity despawn now, so these can be handled by the user now (for example moving away somewhere else to find fish).

### DIFF
--- a/examples/fisherman.js
+++ b/examples/fisherman.js
@@ -55,11 +55,10 @@ async function startFishing () {
     return bot.chat(err.message)
   }
 
-  nowFishing = true
   bot.on('playerCollect', onCollect)
 
   try {
-    await bot.fish()
+    nowFishing = await bot.fish()
   } catch (err) {
     bot.chat(err.message)
   }
@@ -70,7 +69,7 @@ function stopFishing () {
   bot.removeListener('playerCollect', onCollect)
 
   if (nowFishing) {
-    bot.activateItem()
+    nowFishing.cancel()
   }
 }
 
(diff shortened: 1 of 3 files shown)


rom1504:
What are bugs in the current implementation that this is fixing?

extremeheat:
The purpose of the pr was to make the fishing logic a bit easier to flow/debug. In terms of changes,
* fishing times out on failure, this way if the bot throws somewhere where it can't catch fish, or throwing the bobber fails, the fishing fails as mentioned
* adds a way to cancel fishing that checks if an active bobber was successfully thrown before calling `activateItem()`

```js
const task = bot.fish((err) => {
  if (err) { console.error(err) }
})
// some other code
task.cancel(Error('reason'))

// vs
let fishing = true
bot.fish(() => {
  fishing = false
})
//...
if (fishing) activateItem() // <- can fail
```

also enables behaviors like this:
```js
let keepFishing = true
let task
while (keepFishing) {
    try {
      task = await bot.fish()
      // check for entity spawn and collect fish
    } catch (e) {
        // move away, check rod, etc.
    }
}
keepFishing = true
task.cancel() // can safely cancel no matter state of the fishing rod
```

rom1504 (changes requested):
I'm not so convinced to extend promise in that way.
Why not instead have a bot.stopFish() ?
It is not possible to fish 2 things at once, so it makes sense mineflayer would handle internally the fishing/not fishing status

extremeheat:
ok, this was primarily intended to help debug/fix the fishing tests failing so I’ll close it

PR #1623, opened 2021-01-27
Add support for Russian usernames in chat

aoleg94:
Other languages can be added in similar way, though XRegexp would be an ultimate solution for this.


rom1504 (changes requested):
maybe we should just use a `.` ?

rom1504:
reopen if you want to do it

PR #1624, opened 2021-01-27
example for bot with many accs from external file

u9g:
One of the most common ways to hold minecraft accounts is in a "username:password" text file, which this example shows how to use


Karang:
Considering that mojang accounts would soon be migrated to ms accounts, wouldn't it be better to wait and add the example for ms accounts ? Otherwise we'll soon have to deprecate this example.

rom1504:
I think it's exactly the same with ms account, just adding a `auth:microsoft` key

rom1504 (approved):

Issue #1626, opened 2021-01-27
mineflayer crashing if someone interacts with blocks nearby


SwiftFizz:
## Versions
 - mineflayer: 2.41.0
 - server: spigot/paper 1.12.2
 - node: v14.15.4
## Description of the problem

I play on a server where there is a plugin called Orebfuscator that turns ores, chests and other things in to stone when you aren't close to it, to stop X-Ray cheats.

The issue is that if somebody opens a chest in the bots render distance the bot crashes.
I think this is because mineflayer is trying to get the chest type however it is a stone block and there is no chest to get the type of.

```
C:\Users\Samuel\Desktop\Snitch\node_modules\minecraft-protocol\src\transforms\framing.js:67
          } else { throw e }
                   ^
TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at getChestType (C:\Users\Samuel\Desktop\Snitch\node_modules\mineflayer\lib\plugins\block_actions.js:39:45)
    at Client.<anonymous> (C:\Users\Samuel\Desktop\Snitch\node_modules\mineflayer\lib\plugins\block_actions.js:69:27)
    at Client.EventEmitter.emit (domain.js:467:12)
    at FullPacketParser.<anonymous> (C:\Users\Samuel\Desktop\Snitch\node_modules\minecraft-protocol\src\client.js:91:12)
    at FullPacketParser.emit (events.js:315:20)
    at FullPacketParser.EventEmitter.emit (domain.js:467:12)
    at addChunk (C:\Users\Samuel\Desktop\Snitch\node_modules\readable-stream\lib\_stream_readable.js:298:12)
    at readableAddChunk (C:\Users\Samuel\Desktop\Snitch\node_modules\readable-stream\lib\_stream_readable.js:280:11)
PS C:\Users\Samuel\Desktop\Snitch> node index.js
```

SwiftFizz:
Fixed the issue by adding a null check at the start of the code 
```
 function getChestType (chestBlock) { // Returns 'single', 'right' or 'left'
    if (bot.supportFeature('doesntHaveChestType')) {
      const facing = parseChestMetadata(chestBlock).facing

      // We have to check if the adjacent blocks in the perpendicular cardinals are the same type
      if(facing == null) return;
      const perpendicularCardinals = Object.keys(FACING_MAP[facing])
      for (const cardinal of perpendicularCardinals) {
        const cardinalOffset = CARDINALS[cardinal]
[...]

PR #1633, opened 2021-01-29
Fixed crash when opening chest

SwiftFizz:
https://github.com/PrismarineJS/mineflayer/issues/1626

### DIFF
--- a/lib/plugins/block_actions.js
+++ b/lib/plugins/block_actions.js
@@ -35,7 +35,7 @@ function inject (bot, { version }) {
     if (bot.supportFeature('doesntHaveChestType')) {
       const facing = parseChestMetadata(chestBlock).facing
 
-      if(facing == null) return `single`
+      if(!facing) return 'single'
       // We have to check if the adjacent blocks in the perpendicular cardinals are the same type
       const perpendicularCardinals = Object.keys(FACING_MAP[facing])
       for (const cardinal of perpendicularCardinals) {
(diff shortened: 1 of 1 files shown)


lib/plugins/block_actions.js
```diff
@@ -35,6 +35,7 @@ function inject (bot, { version }) {
     if (bot.supportFeature('doesntHaveChestType')) {
       const facing = parseChestMetadata(chestBlock).facing
 
+      if(facing == null) return;
```
imharvol, line 38: Probably better to do `return 'single'` , otherwise `'chestLidMove'` won't be emitted in these cases

SwiftFizz: Using strict equals causes the original crash to still happen, added return `single` and removed the ; however it didn't pass.

imharvol: Then it's because `facing` is `undefined`

You can use `if(!facing) return 'single'` to catch it when it's undefined/null

rom1504:
can you rebase to see if it fixes tests ?

Karang:
> can you rebase to see if it fixes tests ?

And probably fix the linter errors

Karang:
@SwiftFizz can you run `npm run fix` to fix the linter errors, that's all that is needed to merge

Karang:
done in https://github.com/PrismarineJS/mineflayer/pull/1687/commits/fe8174d660c81c94e60a80bb0af568205270a821

PR #1637, opened 2021-01-30
Set createBot default auth option to 'mojang'

DecentralisedTech:
Set createBot default auth option to 'mojang'


u9g:
This should have already been mojang, it says mojang in all the docs.

rom1504:
This isn't needed. Undefined is already mojang

PR #1640, opened 2021-02-02
Simplify windows API and promisify tests

Karang:
A much needed refactor.

* Fixed fishing test (cannot force the look, otherwise it doesn't wait for the server to be aware the head direction changed)
* Removed Chest, EnchantmentTable, Furnace, Dispenser and Villager classes (they all are Windows now)
* Dispensers are now handled by the same code as other containers, hopper too (they were missing)
* There is now only 2 events signaling a slot update ("updateSlot" and "updateSlot:slotId" of the Window class) (before there was: "setSlot", "setSlot:windowId", "windowUpdate", "updateSlot", on 3 different eventEmitter (and not all of them were working properly))
* All windows (present and future) now have a withdraw and deposit function
* openBlock and openEntity are now promisified
* moved fish() from inventory to its own file
* promisified all external tests

Depends on https://github.com/PrismarineJS/prismarine-windows/pull/19

If someone have time to check the examples and fix them, or tell me the ones that are failing, it would be helpful.


Karang:
The scoreboard test is failing randomly and I cant reproduce it offline. I disabled it and consider it outside of the scope of this PR.

rom1504:
Does this introduce breaking changes ?

Karang:
> Does this introduce breaking changes ?

Yes, its needed to fix broken functionnalities. The main breaking change is the openBlock / openEntity that needs to be awaited.

Before the pattern was:
```js
const chest = bot.openBlock(block, Chest)// we had to give the wrapper class, and the function returned immediately with an instance of Chest
await once(chest, 'open') // we had to wait the window was open to use it
```
Now its:
```js
const chest = await bot.openBlock(block) // no need to give a wrapper class, return when the window is ready
```

The second breaking change is the removal of the various setSlot events, there is now only 1 event to listen to. This is a consequence of the removal of one of the event emitter. But those events were mainly used internally and I found that some of them weren't triggered properly anyway.

Karang:
To be noted: for chests, the user-facing api is `openChest` and this one already returned a promise so the behavior didn't change. `openBlock` was also exposed, but its less likely that users were using it directly.

rom1504:
Breaking changes are ok but then we need to :
* Check which PrismarineJS org plugins we break
* Release a new major version
* be sure we don't need to rebreak this part of mineflayer soon

rom1504:
released pwindows as 2.0.0, update to that here please

rom1504 (approved):
LGTM
It's breaking, will release as new major

EvModder:
Note that this commit broke a few things.
Mainly, it means mineflayer can no longer open ender_chests.
It also makes the error message on chest.js L21 much less accurate, since it can get triggered even when a valid (but non-container) block gets passed as an argument, which makes debugging more annoying

PR #1647, opened 2021-02-06
Optional skip of liquids

goncharovchik:
Optional skip of liquids in blockAtCursor()

### DIFF
--- a/lib/plugins/ray_trace.js
+++ b/lib/plugins/ray_trace.js
@@ -37,7 +37,7 @@ module.exports = (bot) => {
           return block
         }
       } else {
-        if (block && block.type !== 0 && block.name !== 'water' && block.name !== 'lava') {
+        if (block && block.boundingBox !== 'empty') {
           block.face = pos.face
           return block
         }
--- a/lib/plugins/ray_trace.js
+++ b/lib/plugins/ray_trace.js
@@ -17,7 +17,7 @@ module.exports = (bot) => {
   }
 
   bot.blockAtCursor = (maxDistance = 256, liquid = true) => {
-    if (typeof maxDistance == 'boolean') {
+    if (typeof maxDistance === 'boolean') {
       liquid = maxDistance
       maxDistance = 256
     }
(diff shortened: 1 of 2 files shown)


lib/plugins/ray_trace.js
```diff
@@ -27,9 +31,16 @@ module.exports = (bot) => {
     while (pos) {
       const block = bot.blockAt(new Vec3(pos.x, pos.y, pos.z))
       // TODO: Check boundingBox of block
-      if (block && block.type !== 0) {
-        block.face = pos.face
-        return block
+      if (liquid) {
+        if (block && block.type !== 0) {
```
Karang, line 35: In what cases do we need to stop in a liquid ?
If instead of `block.type !== 0` we test for `block.boundingBox !== 'empty'`it will include liquids and plants (and also cave air)

goncharovchik: > In what cases do we need to stop in a liquid ?
> If instead of `block.type !== 0` we test for `block.boundingBox !== 'empty'`it will include liquids and plants (and also cave air)

To collect in a bucket. Is not it?

Karang: ok. good point

lib/plugins/ray_trace.js
```diff
@@ -16,7 +16,11 @@ module.exports = (bot) => {
     if (block) return block
   }
 
-  bot.blockAtCursor = (maxDistance = 256) => {
+  bot.blockAtCursor = (maxDistance = 256, liquid = true) => {
+    if (typeof maxDistance === 'boolean') {
```
rom1504, line 20: Can we not do that ? It's confusing for users

goncharovchik: What do you mean?

Karang:
With the matcher function introduced here https://github.com/PrismarineJS/mineflayer/blob/a492e30314b55c617e366a803f7f7a996be2b281/lib/plugins/ray_trace.js#L18 this could be left to the user to implement or we could provide a separate function like `liquidAtCursor`

goncharovchik:
> With the matcher function introduced here https://github.com/PrismarineJS/mineflayer/blob/a492e30314b55c617e366a803f7f7a996be2b281/lib/plugins/ray_trace.js#L18 this could be left to the user to implement or we could provide a separate function like `liquidAtCursor`

Oh, so this PR can be closed?

rom1504:
Ah yeah ok
That's not what I meant
I just meant that having an optional argument in the same position that changes behavior based on its type is error prone

PR #1648, opened 2021-02-06
Added a option to dig that changes the face to bot uses to mine a block

IceTank:
Most servers with anti cheat, don't allow you to mine blocks if you dont have direct line of sight to the block you want to mine. The bot mines blocks using the top face of the block by default while looking at the center of the block. 
With this change to bot can mine blocks using different faces and also change its view to be in line of sight with the block face it wants to mine. 
The raycast option tries to look for faces that are visible to the bot and chooses a valid face automatically.

### DIFF
--- a/lib/plugins/digging.js
+++ b/lib/plugins/digging.js
@@ -1,7 +1,7 @@
 const nbt = require('prismarine-nbt')
 const { performance } = require('perf_hooks')
 const { createDoneTask, createTask } = require('../promise_utils')
-const { RaycastIterator, BlockFace } = require('prismarine-world').iterators
+const BlockFaces = require('prismarine-world').iterators.BlockFace
 const { Vec3 } = require('vec3')
 
 module.exports = inject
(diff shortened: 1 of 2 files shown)


lib/plugins/digging.js
```diff
@@ -13,22 +15,104 @@ function inject (bot) {
...
+          const targetPos = block.position.offset(0.5 + (i === 'x' ? visibleFaces[i] * 0.5 : 0), 0.5 + (i === 'y' ? visibleFaces[i] * 0.5 : 0), 0.5 + (i === 'z' ? visibleFaces[i] * 0.5 : 0))
+          const startPos = bot.entity.position.offset(0, bot.entity.height, 0)
+          const rayIterator = new RaycastIterator(startPos, targetPos.clone().subtract(startPos).normalize(), 5) // overshoot a bit for more reliable raycasting
+          let rayBlock = rayIterator.next()
+          let last = rayBlock
+          while (rayBlock) {
+            if (rayBlock.x === block.position.x && rayBlock.y === block.position.y && rayBlock.z === block.position.z) {
+              if (last) {
+                console.info('Valid face', targetPos, last.face)
+                // is the face value switched for rayIterator.next() ? Should be the face it entered not exited ?
```
IceTank, line 69: To get the right face of the block I want to hit I had to use the last block I entered with the raycast. Im not sure if this is a bug in the Raycasting iterator or if my code is wrong in some way?

lib/plugins/digging.js
```diff
@@ -13,22 +15,104 @@ function inject (bot) {
...
+              closest = validFaces[i]
+              distSqrt = cDist
+            }
+          }
+          await bot.lookAt(closest.targetPos, forceLook)
+          diggingFace = closest.face
+        } else {
+          // Block is obstructed return error?
+          // Try to mine the block anyway
+          await bot.lookAt(block.position.offset(0.5, 0.5, 0.5), forceLook)
```
IceTank, line 104: It would maybe be useful to return an error if there is no valid face

lib/plugins/digging.js
```diff
@@ -1,6 +1,8 @@
 const nbt = require('prismarine-nbt')
 const { performance } = require('perf_hooks')
 const { createDoneTask, createTask } = require('../promise_utils')
+const { RaycastIterator, BlockFace } = require('prismarine-world').iterators
+const Vec3 = require('vec3')
```
Karang, line 5: `const {Vec3} = require('vec3')`
to get the constructor

lib/plugins/digging.js
```diff
@@ -1,6 +1,8 @@
 const nbt = require('prismarine-nbt')
 const { performance } = require('perf_hooks')
 const { createDoneTask, createTask } = require('../promise_utils')
+const { RaycastIterator, BlockFace } = require('prismarine-world').iterators
```
Karang, line 4: once https://github.com/PrismarineJS/prismarine-world/pull/58 is merged you can use `bot.world.raycast()`

IceTank: Does this new raycasting still return every block within the raycasting path?

Karang: it returns the end block but test every block shapes

IceTank:
I changed dig with the raycast option to return an error if the block is not visible. I figured if you use raycasting specifially you should know when the block is opstructed. 
The current implementation does not consider blocks that have a none full block bounding box. So the bot might not acctually look at the blocks bounding box while mining but at the full block bounding box.

Karang (approved):
LGTM

PR #1649, opened 2021-02-09
Place block and Raycast improvements

Karang:
* inventory.js is already too big, so i moved placeBlock to its own plugin file
* added a cursor option to placeBlock so we can select between top or bottom stairs, slabs, etc..
* use pworld's improved raycasting

### DIFF
--- a/lib/plugins/place_block.js
+++ b/lib/plugins/place_block.js
@@ -1,0 +1,88 @@
+const assert = require('assert')
+const { once } = require('events')
+const { callbackify } = require('../promise_utils')
+
+module.exports = inject
+
+function inject (bot, { version }) {
+  const Item = require('prismarine-item')(version)
+
+  async function placeBlock (referenceBlock, faceVector) {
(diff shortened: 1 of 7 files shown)


lib/plugins/place_block.js
```diff
@@ -0,0 +1,97 @@
...
+      throw new Error(`No block has been placed : the block is still ${oldBlock.name}`)
+    }
+  }
+
+  async function placeBlock (referenceBlock, faceVector) {
+    await placeBlockWithOptions(referenceBlock, faceVector, {})
+  }
+
+  bot.placeBlock = callbackify(placeBlock)
+  bot._placeBlockWithOptions = placeBlockWithOptions
```
rom1504, line 79: What's the plan regarding this semi exposed function ?

Karang: I'm not sure about the api yet, but this feature is needed. I thought we could have it as part of the unstable api for a while and see how it goes.

rom1504 (approved):
Ok
We can add this options arg optionally if it makes sense in the future

PR #1650, opened 2021-02-09
removed repeating code and added more tests

u9g:


test/externalTests/exampleInventory.js
```diff
@@ -1,29 +1,81 @@
...
-        assert.fail(`Unexpected message: ${message}`) // error
-      }
-      return true // stop listening
-    })
+    const testFuncs = tests.map(test => makeTest(test.command, test.wantedMessage))
+    for await (const test of testFuncs) {
+      await test()
+      await bot.test.wait(100)
+    }
+    await Promise.allSettled(tests)
```
extremeheat, line 65: `tests` is not an array of promises

test/externalTests/exampleInventory.js
```diff
@@ -1,29 +1,81 @@
...
-      }
-      return true // stop listening
-    })
-    await bot.test.tellAndListen(name, 'toss 64 dirt', (message) => {
-      if (!message.startsWith('tossed 64 x dirt')) {
-        assert.fail(`Unexpected message: ${message}`) // error
-      }
-      return true // stop listening
-    })
+    const testFuncs = tests.map(test => makeTest(test.command, test.wantedMessage))
```
extremeheat, line 60: you can simplify this into one for loop:
```js
for (const { command, wantedMessage } of tests) {
    await bot.test.tellAndListen(name, command, (message) => {
        if (!message.startsWith(wantedMessage)) {
            assert.fail(`Unexpected message: ${message}`) // error
        }
        return true // stop listening
    })
    await bot.test.wait(2000)
}
```

test/externalTests/exampleInventory.js
u9g: fixed

test/externalTests/exampleInventory.js
u9g: I could but this heavily hurts readability imo

imharvol:
Does using `for await ... of` instead of `for ... of` make any different? I  have never heard of it, and it doesn't seem to be any difference

Remember that as the crafting table is placer far from `flatbot` you should remove it manually. You can just add `bot.chat('/setblock 52 4 0 crafting_table')` after the loop. Other than that everything seems fine

u9g:
> Does using `for await ... of` instead of `for ... of` make any different? I have never heard of it, and it doesn't seem to be any difference
> 
> Remember that as the crafting table is placer far from `flatbot` you should remove it manually. You can just add `bot.chat('/setblock 52 4 0 crafting_table')` after the loop. Other than that everything seems fine

I think I'm going to change that to bot.test.placeblock, then it should autoremove right?

Karang:
> I think I'm going to change that to bot.test.placeblock, then it should autoremove right?

There are no placeBlock function with autoremove.

rom1504:
please rebase on master (git fetch && git rebase origin/master && git push --force origin your_branch )

PR #1658, opened 2021-02-11
Added bot.findBlocks2

u9g:
I feel like this was needed. It's simpler to understand for people wanting to understand how findblocks works, easier to use, more options, and safer

### DIFF
--- a/lib/plugins/blocks.js
+++ b/lib/plugins/blocks.js
@@ -227,7 +227,7 @@ function inject (bot, { version, storageBuilder }) {
     return bot.blockAt(blocks[0])
   }
 
-  bot.findBlocks2 = (opts) => {
+  bot.findBlocks2 = async (opts) => {
     assert(
       typeof opts.wantedBlock === 'number' || typeof opts.wantedBlocks === 'object',
       Error('wantedBlock has to be an id of a block or wantedBlocks has to be an array of block ids')
@@ -279,8 +279,8 @@ function inject (bot, { version, storageBuilder }) {
     // filter for wanted blocks
     let matcher = opts.matcher
     if (matcher === undefined) {
-      matcher = (currentBlock) => {
-        const { type } = bot.world.getBlock(currentBlock)
+      matcher = async (currentBlock) => {
+        const { type } = await bot.world.getBlock(currentBlock)
         return internalWantedBlocks.includes(type)
       }
     }
--- a/lib/plugins/blocks.js
+++ b/lib/plugins/blocks.js
@@ -227,6 +227,73 @@ function inject (bot, { version, storageBuilder }) {
     return bot.blockAt(blocks[0])
   }
 
+  bot.findBlocks2 = (opts) => {
+    assert(
+      typeof opts.wantedBlock === 'number' || typeof opts.wantedBlocks === 'object',
+      Error('wantedBlock has to be an id of a block or wantedBlocks has to be an array of block ids')
+    )
+    assert(
+      typeof opts.wantedBlock !== 'number' || typeof opts.wantedBlocks !== 'object',
+      Error('Either set wantedBlock or wantedBlocks, not both')
+    )
+    assert(
+      typeof opts.matcher === 'function' || opts.matcher === undefined,
+      Error('Matcher must be a function')
+    )
+    assert(
+      opts.maxMatches > 0 || opts.maxMatches === undefined,
+      Error('Max blocks must be >0')
+    )
+    assert(
+      opts.point instanceof Vec3 || opts.point === undefined,
+      Error('Start point must be a vec3 point')
+    )
+    const internalWantedBlocks = opts.wantedBlocks || [opts.wantedBlock]
+    const searchRadius = opts.searchRadius || 16
+
+    const foundBlocks = []
+    const { x: originX, y: originY, z: originZ } = (opts.point || bot.entity.position).floored()
+    // start and end positions
+    const startX = originX - searchRadius
+    let startY = originY - searchRadius
+    const startZ = originZ - searchRadius
+    const endX = originX + searchRadius
+    let endY = originY + searchRadius
+    const endZ = originZ + searchRadius
+    // if set by user, change start and end pos
+    if (opts.toBedrock) startY = 0
+    if (opts.toSkyLimit) endY = 256
+    // find blocks
+    for (let x = startX; x < endX; x++) {
+      for (let y = startY; y < endY; y++) {
+        for (let z = startZ; z < endZ; z++) {
+          const currBlock = new Vec3(x, y, z)
+          foundBlocks.push(currBlock)
+        }
+      }
+    }
+    // filter for wanted blocks
+    let matcher = opts.matcher
+    if (matcher === undefined) {
+      matcher = (currentBlock) => {
+        const { type } = bot.world.getBlock(currentBlock)
+        return internalWantedBlocks.includes(type)
+      }
+    }
+    let wantedBlocks = foundBlocks.filter(matcher)
+    // sort blocks by distance from bot
+    const botLocation = bot.entity.position.floored()
+    wantedBlocks = wantedBlocks.sort((a, b) => {
+      return a.distanceTo(botLocation) - b.distanceTo(botLocation)
+    })
+    // splice the number of matches the user wanted
+    if (opts.maxMatches) {
+      wantedBlocks = wantedBlocks.splice(0, opts.maxMatches)
+    }
+    wantedBlocks = wantedBlocks.map(blockAt)
+    return wantedBlocks
+  }
+
   function blockAt (absolutePoint, extraInfos = true) {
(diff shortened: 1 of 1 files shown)


extremeheat:
Until recently there were iterators in mineflayer, they were seperated to prismarine-world. The manhattan or octahedron iterators are much more efficient for this use case: https://github.com/PrismarineJS/prismarine-world/blob/master/src/iterators.js

u9g:
# Examples that use all options:
```js
function test1 () {
  const mcData = require('minecraft-data')(bot.version)
  const t0 = performance.now()
  const block = bot.findBlocks2({
    wantedBlock: mcData.blocksByName.bedrock.id,
    searchRadius: 16,
    toSky: true,
    toBedrock: true,
    point: bot.entity.position.offset(0, 1, 0),
    maxMatches: 10,
    matcher: (currentBlock) => {
      const { type } = bot.world.getBlock(currentBlock)
      return type === 0
    }
  })
  const t1 = performance.now()
  bot.chat(`The bot took ${(t1 - t0) / 1000} seconds`)

  console.log('returned: ')
  console.log(block)
}

function test2 () {
  const mcData = require('minecraft-data')(bot.version)
  const t0 = performance.now()
  const block = bot.findBlocks2({
    wantedBlocks: [mcData.blocksByName.bedrock.id, mcData.blocksByName.dirt.id],
    searchRadius: 16,
    toSky: true,
    toBedrock: true,
    point: bot.entity.position.offset(0, 1, 0),
    maxMatches: 10,
    matcher: (currentBlock) => {
      const { type } = bot.world.getBlock(currentBlock)
      return type === 0
    }
  })
  const t1 = performance.now()
  bot.chat(`The bot took ${(t1 - t0) / 1000} seconds`)

  console.log('returned: ')
  console.log(block)
}
```

u9g:
> Until recently there were iterators in mineflayer, they were seperated to prismarine-world. The manhattan or octahedron iterators are much more efficient for this use case: https://github.com/PrismarineJS/prismarine-world/blob/master/src/iterators.js

I understand that, that's why I didn't just delete the old findblocks, but the one I made is safer, has more options, and is clearer to understand than either of those.

extremeheat:
Yes, but your current method is very inefficient since it starts from corners. For loops block the node process and getBlock calls aren't cheap, so it's a good idea to be efficient where possible. For example, doing getBlocks on ore positions while pathfinding is inefficient because the first searched nodes are the ones furthest away from the bot. So it would be a good idea to use a more efficient iterator and maybe turn this to async.

u9g:
> Yes, but your current method is very inefficient since it starts from corners. For loops block the node process and getBlock calls aren't cheap, so it's a good idea to be efficient where possible. For example, doing getBlocks on ore positions while pathfinding is inefficient because the first searched nodes are the ones furthest away from the bot. So it would be a good idea to use a more efficient iterator and maybe turn this to async.

I awaited getBlock call, the intention for findblocks2 isnt to overwrite pathfinding, it's to be a beginner-friendly api that's more safe and has more options

extremeheat:
I agree with you that smaller indexes won't make a difference, but here's an example:

* Log directly in front of bot
* Bot is written to findBlocks2() on logs with a radius of 64 blocks (4 chunks)
* Around (64^3)/2=131072 nodes will be visited

Where as in a octahedron search for example, the nodes can come down to <=10. (You don't need A* or anything, I'm just talking about iterators)

For an example, I think it's ok but I think it's a good idea if in mineflayer to make it efficient if we have the iterators

u9g:
> I agree with you that smaller indexes won't make a difference, but here's an example:
> 
> * Log directly in front of bot
> * Bot is written to findBlocks2() on logs with a radius of 64 blocks (4 chunks)
> * Around (64^3)/2=131072 nodes will be visited
> 
> Where as in a octahedron search for example, the nodes can come down to <=10. (You don't need A* or anything, I'm just talking about iterators)
> 
> For an example, I think it's ok but I think it's a good idea if in mineflayer to make it efficient if we have the iterators

I understand that, just say that if you want >32 blocks in each direction, use findBlocks instead of findBlocks2. I just feel like there are not that many players looking for >=2 chunks in each direction from sky to bedrock. Also, maybe I did it wrong but findBlocks only allows blocks that the player can see.

Karang:
What you wrote is the implementation that findBlocks replaced. We had very good reasons not to use this naive implementation as it was way too slow for real time usage (in some cases it could take multiple seconds to find a block https://github.com/PrismarineJS/mineflayer/pull/1035). I suggest you benchmark your code (in real scenarios) to see for yourself.

Adding yet an other function that does the same thing will confuse beginners even more as they will never know which one to use in what conditions.

For your function to be fast, it requires the user to know the range in which the block is (ie. it needs to fine tune the parameters searchRadius, toSky, toBedrock), which defeat the point of the function to find a block you don't know where it is (in the current function you can set the default radius to 256 and not worry about it).

Appart from the added toSky, toBedrock parameters, the api is identical to the current function, how is it simpler to use ? If you have some suggestions to improve the api, you can improve the interface of the current findBlock, but the implementation will have to stay the same.

lib/plugins/blocks.js
```diff
@@ -227,6 +227,77 @@ function inject (bot, { version, storageBuilder }) {
...
+    )
+    assert(
+      opts.point instanceof Vec3 || opts.point === undefined,
+      Error('Start point must be a vec3 point')
+    )
+    const internalWantedBlocks = opts.wantedBlocks || [opts.wantedBlock]
+    const searchRadius = opts.searchRadius || 16
+
+    const foundBlocks = []
+    const { x: originX, y: originY, z: originZ } = (opts.point || bot.entity.position).floored()
```
Karang, line 259: `floored()` creates an unnecessary temporary Vec3 instance. Creating temporary objects is bad for performances.

lib/plugins/blocks.js
```diff
@@ -227,6 +227,77 @@ function inject (bot, { version, storageBuilder }) {
...
+    let endY = originY + searchRadius
+    const endZ = originZ + searchRadius
+    // if set by user, change start and end pos
+    if (opts.toBedrock) startY = 0
+    if (opts.toSkyLimit) endY = 256
+    // find blocks
+    for (let x = startX; x < endX; x++) {
+      for (let y = startY; y < endY; y++) {
+        for (let z = startZ; z < endZ; z++) {
+          const currBlock = new Vec3(x, y, z)
```
Karang, line 274: Even worse, you are creating a lot of temporary Vec3, the proper way to iterate over a region is to reuse the cursor, as done here: https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/blocks.js#L196

lib/plugins/blocks.js
```diff
@@ -227,6 +227,77 @@ function inject (bot, { version, storageBuilder }) {
...
+          foundBlocks.push(currBlock)
+        }
+      }
+    }
+    // filter for wanted blocks
+    let matcher = opts.matcher
+    if (matcher === undefined) {
+      matcher = async (currentBlock) => {
+        const { type } = await bot.world.getBlock(currentBlock)
```
Karang, line 283: await not needed world is sync

lib/plugins/blocks.js
```diff
@@ -227,6 +227,77 @@ function inject (bot, { version, storageBuilder }) {
...
+        return internalWantedBlocks.includes(type)
+      }
+    }
+    let wantedBlocks = foundBlocks.filter(matcher)
```
Karang, line 287: Creating the array and then filtering it, is bad for performances, you need to test the blocks on the fly.

lib/plugins/blocks.js
```diff
@@ -227,6 +227,77 @@ function inject (bot, { version, storageBuilder }) {
...
+    // sort blocks by distance from bot
+    const botLocation = bot.entity.position.floored()
```
Karang, line 289: If you really wanted to have an array with all the positions, you could have sorted it before doing the matching, so you could stop when reaching the count.

lib/plugins/blocks.js
```diff
@@ -227,6 +227,77 @@ function inject (bot, { version, storageBuilder }) {
...
+    // start and end positions
+    const startX = originX - searchRadius
+    let startY = originY - searchRadius
+    const startZ = originZ - searchRadius
+    const endX = originX + searchRadius
+    let endY = originY + searchRadius
+    const endZ = originZ + searchRadius
+    // if set by user, change start and end pos
+    if (opts.toBedrock) startY = 0
+    if (opts.toSkyLimit) endY = 256
```
Karang, line 269: This will become obsolete in 1.17

Karang (commented):
Here are some comments about performances. You can also check the old findBlocks versions:
https://github.com/PrismarineJS/mineflayer/blob/29cf25a469f27f579552512bccf2bc3e8e80ad30/lib/plugins/blocks.js#L102
and
https://github.com/Darthfett/mineflayer-blockfinder

PR #1676, opened 2021-02-17
world events through p-world instead of mineflayer

u9g:
WIP

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -665,6 +665,23 @@ Create and return an instance of the class bot.
 
 A sync representation of the world. Check the doc at http://github.com/PrismarineJS/prismarine-world
 
+#### Events:
+
+#### "blockUpdate" (oldBlock, newBlock)
+
+Fires when a block updates. Both `oldBlock` and `newBlock` provided for
+comparison.
+
+Note that `oldBlock` may be `null`.
+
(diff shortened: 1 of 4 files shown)


rom1504:
seems this fails

rom1504:
nice it works
@u9g since the goal is to deprecate using bot level events and encourage people to use world events, could you adapt the mineflayer doc to indicate this?

docs/api.md
```diff
@@ -665,6 +665,23 @@ Create and return an instance of the class bot.
 
 A sync representation of the world. Check the doc at http://github.com/PrismarineJS/prismarine-world
 
+#### Events:
```
rom1504, line 668: not sure if we should duplicate pworld doc here, but I guess let's let it there for now

package.json
```diff
@@ -31,7 +31,7 @@
     "prismarine-physics": "^1.0.4",
     "prismarine-recipe": "^1.1.0",
     "prismarine-windows": "^2.0.0",
-    "prismarine-world": "^3.3.1",
+    "prismarine-world": "PrismarineJS/prismarine-world#block_events",
```
rom1504, line 34: can you change back to ^3.4.0 ?

rom1504 (approved):

rom1504 (approved):

PR #1677, opened 2021-02-17
add bot.helditem to doc

u9g:
bot.updatehelditem isn't needed anymore, as it's updated on it's own

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -681,6 +681,10 @@ Use this to find out your own name.
 
 Coordinates to the main spawn point, where all compasses point to.
 
+#### bot.heldItem
+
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -1694,10 +1698,6 @@ This function also returns a `Promise`, with `void` as its argument upon complet
 
 Move an item from `sourceSlot` to `destSlot` in the current window.
 
-#### bot.updateHeldItem()
```
Karang, line 1697: updateHeldItem is still exposed https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/inventory.js#L525
it should be removed from the code before than from  the documentation

PR #1680, opened 2021-02-18
Refactor crafting tests

imharvol:
In my opinion they are confusing and is missing checks, for example, it doesn't try to craft in a crafting table, it just does in the inventory, it doesn't check the amount of crafted items, or that the exceptions are thrown correctly.

For now I'll mark it as a draft. I already encountered what I believe could be a bug (that's why tests are failing). The craft function seems to craft 1 less item that requested.
A quick fix would be to use `<=` in:
https://github.com/PrismarineJS/mineflayer/blob/15096d8a8e246295ff16373967660d6a70c6adf0/lib/plugins/craft.js#L17-L19
Of course that doesn't look like the right solution

### DIFF
--- a/test/externalTests/crafting.js
+++ b/test/externalTests/crafting.js
@@ -1,52 +1,28 @@
-const { Vec3 } = require('vec3')
+const assert = require('assert')
+const { once } = require('events')
 
 module.exports = () => async (bot) => {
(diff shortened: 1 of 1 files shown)


rom1504:
this is failing

leumasme:
> this is failing

did- did you read what he wrote?

rom1504:
please reopen if you want to work on it again

PR #1687, opened 2021-02-20
Fix wait for ticks && Expose bot.physicEnabled

Karang:
* `this.` cannot be used in the promise scope @TheDudeFromCI 
* exposed physicEnabled so that users don't have to disable the physic plugin if they want to switch off physic
* implemented https://github.com/PrismarineJS/mineflayer/pull/1633 changes with linter fix

### DIFF
--- a/lib/plugins/physics.js
+++ b/lib/plugins/physics.js
@@ -35,7 +35,7 @@ function inject (bot) {
   let lastSentYaw = null
   let doPhysicsTimer = null
   let lastPhysicsFrameTime = null
-  let physicEnabled = false
+  bot.physicEnabled = false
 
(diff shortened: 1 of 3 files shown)


lib/plugins/physics.js
```diff
@@ -253,7 +253,7 @@ function inject (bot) {
     }
     sendPacketPositionAndLook(pos, newYaw, newPitch, bot.entity.onGround)
 
-    physicEnabled = true
```
rom1504, line 256: Can you add it to API.md ?

Karang: done

rom1504 (approved):

PR #1699, opened 2021-02-21
Add event to blocks which was placed by the bot

CyberPatrick:


lib/plugins/place_block.js
```diff
@@ -73,6 +73,8 @@ function inject (bot, { version }) {
 
     if (oldBlock.type === newBlock.type) {
       throw new Error(`No block has been placed : the block is still ${oldBlock.name}`)
+    } else {
+      bot.emit('blockPlaced', [oldBlock, newBlock])
```
Karang, line 77: You don't need to put the parameters between brackets, unless you want them to be a single array parameter.

Karang (changes requested):
One comment, otherwise looks good to me. Can you add the event to api.md ?

lib/plugins/place_block.js
CyberPatrick: Ok

Karang (approved):
LGTM

PR #1703, opened 2021-02-21
setQuickBarSlot Small fix

CyberPatrick:
If slot value will be an not empty string, it will not throw an error.


u9g:
I don't agree with this, it says in the typings this should be a number, and there are no other ambiguous functions like this. https://github.com/PrismarineJS/mineflayer/blob/229e9d559f170227e0d1b8becaadbeb878fcb721/index.d.ts#L305 If you want this, I feel like you should just cast it yourself.

Karang:
Yes this function shouldn't accept anything else than a number.

PR #1716, opened 2021-02-26
Removing unnecessary package sending.

goncharovchik:
Removing unnecessary package sending. This is not vanilla, the bot does not pass many checks because of this part of code. After deleting it, the bot passes all checks.

### DIFF
--- a/lib/plugins/physics.js
+++ b/lib/plugins/physics.js
@@ -140,10 +140,6 @@ function inject (bot) {
       sendPacketPosition(position, onGround)
     } else if (lookUpdated && bot.isAlive) {
       sendPacketLook(yaw, pitch, onGround)
-    } else if (performance.now() - lastSent.time >= 1000) {
-      // Send a position packet every second, even if no update was made
-      sendPacketPosition(position, onGround)
-      lastSent.time = performance.now()
     } else if (bot.supportFeature('positionUpdateSentEveryTick') && bot.isAlive) {
       // For versions < 1.12, one player packet should be sent every tick
       // for the server to update health correctly


Karang:
Tests are not passing and I'm not sure removing this is correct. The extra packet that was added when experimenting with the fall check is this one: https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/physics.js#L252 maybe now that we fixed the airdrag this isn't needed anymore.

goncharovchik:
> Tests are not passing and I'm not sure removing this is correct. The extra packet that was added when experimenting with the fall check is this one: https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/physics.js#L252 maybe now that we fixed the airdrag this isn't needed anymore.

But the physics test passed? How can you be sure that this is the right solution?

Karang:
> > Tests are not passing and I'm not sure removing this is correct. The extra packet that was added when experimenting with the fall check is this one: https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/physics.js#L252 maybe now that we fixed the airdrag this isn't needed anymore.
> 
> But the physics test passed? How can you be sure that this is the right solution?
> In addition, the client that passes all the checks does not have this package https://github.com/ORelio/Minecraft-Console-Client/.

Then maybe it isn't needed for every versions. We should identify which versions are sending this extra packets.

Karang:
According to https://wiki.vg/Protocol#Player_Movement and verified with a proxy (on 1.16.4), the vanilla client does send a player position packet every second if the player didn't move. So maybe the behavior is wrong on old versions alone. What version are you testing on ?

shketov:
> According to https://wiki.vg/Protocol#Player_Movement and verified with a proxy (on 1.16.4 and 1.12.2), the vanilla client does send a player position packet every second if the player didn't move. So maybe the behavior is wrong on old versions alone. What version are you testing on ?

we checked it in 1.12.2

goncharovchik:
> Tests are not passing and I'm not sure removing this is correct. The extra packet that was added when experimenting with the fall check is this one: https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/physics.js#L252 maybe now that we fixed the airdrag this isn't needed anymore.

    
I will temporarily close this PR. I will find the best way out in the future.

PR #1717, opened 2021-02-27
Changes to internal Physics Plugin

leumasme:
This Pull 
- Renames bot.physicEnabled to bot.physicsEnabled, including Documentation
- Adds an alias getter/setter property for physicEnabled to make the above change non-breaking
- Adds the physicsTick event that is identical to the physicTick event + typescript types
- Add typescript types for physicsEnabled

'physic' without S is not a valid word here, 'physics' is already considered the singular and may not be written without s.
(see other nouns this applies to such as mathematics, economics, politics...)

What could be done in a future major release:
- Remove the physicEnabled alias for physicsEnabled
- Remove the physicTick event alias for physicsEnabled

These would make this pull a potentially breaking change which is why i left this functionality in for now.

### DIFF
--- a/lib/plugins/physics.js
+++ b/lib/plugins/physics.js
@@ -14,7 +14,7 @@ const PI_2 = Math.PI * 2
 const PHYSICS_INTERVAL_MS = 50
 const PHYSICS_TIMESTEP = PHYSICS_INTERVAL_MS / 1000
 
-function inject (bot) {
+function inject (bot, { physicsEnabled }) {
   const mcData = require('minecraft-data')(bot.version)
 
   const world = { getBlock: (pos) => { return bot.blockAt(pos, false) } }
@@ -35,7 +35,8 @@ function inject (bot) {
   let lastSentYaw = null
   let doPhysicsTimer = null
   let lastPhysicsFrameTime = null
-  bot.physicsEnabled = false
+  let shouldUsePhysics = false
+  bot.physicsEnabled = physicsEnabled ?? true
   Object.defineProperty(bot, 'physicEnabled', {
     set: (d) => { bot.physicsEnabled = d },
     get: () => bot.physicsEnabled
@@ -62,7 +63,7 @@ function inject (bot) {
     timeAccumulator += deltaSeconds
 
     while (timeAccumulator >= PHYSICS_TIMESTEP) {
-      if (bot.physicsEnabled) {
+      if (bot.physicsEnabled && shouldUsePhysics) {
         physics.simulatePlayer(new PlayerState(bot, controlState), world).apply(bot)
         bot.emit('physicsTick')
         bot.emit('physicTick') // Deprecated, only exists to support old plugins. May be removed in the future
@@ -258,7 +259,7 @@ function inject (bot) {
     }
     sendPacketPositionAndLook(pos, newYaw, newPitch, bot.entity.onGround)
 
-    bot.physicsEnabled = true
+    shouldUsePhysics = true
     bot.entity.timeSinceOnGround = 0
     lastSentYaw = bot.entity.yaw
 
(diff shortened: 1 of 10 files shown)


leumasme:
Made the exposed physicsEnabled property not correspond to the internal state that controls wether or not physics should be used, for example while riding.
This stops the physicsEnabled state from resetting whenever the bot is teleported or stops riding / exits a boat
Also added an physicsEnabled option to createBot which sets the initial value for the property, as the physicsEnabled property only becomes available after the plugins have loaded, so you can not change it right after creating the bot which is a bit annoying, however wether or not this is actually useful is questionable.

lib/plugins/physics.js
```diff
@@ -35,7 +35,12 @@ function inject (bot) {
   let lastSentYaw = null
   let doPhysicsTimer = null
   let lastPhysicsFrameTime = null
-  bot.physicEnabled = false
+  let shouldUsePhysics = false
+  bot.physicsEnabled = physicsEnabled ?? true
+  Object.defineProperty(bot, 'physicEnabled', {
```
rom1504, line 40: you can just remove that, this was added very recently no need to keep compat

extremeheat:
What's the benefit of the PR? The physicEnabled option can be renamed but I'm not sure it's worthwhile to rename the event, it's widely used and would be a breaking change to rename it. You could emit both events at a small cost, but I don't think it's worthwhile.

leumasme:
> What's the benefit of the PR? The physicEnabled option can be renamed but I'm not sure it's worthwhile to rename the event, it's widely used and would be a breaking change to rename it. You could emit both events at a small cost, but I don't think it's worthwhile.

As it currently is, both events are emitted, with the misspelled one removed from docs, which may then be removed in a future major release.
I don't think that it's acceptable to simply leave these things misspelled - even rom himself has accidentally spelled the property name correctly according to grammar but incorrectly according to the actual API (with s) in the release notes; if even he does this by accident, it will probably cause more confusion and issues amongst users.

lib/plugins/physics.js
leumasme: removed in 7d3c223

lib/plugins/physics.js
```diff
@@ -14,7 +14,7 @@ const PI_2 = Math.PI * 2
 const PHYSICS_INTERVAL_MS = 50
 const PHYSICS_TIMESTEP = PHYSICS_INTERVAL_MS / 1000
 
-function inject (bot) {
+function inject (bot, { physicsEnabled }) {
```
extremeheat, line 17: where does `physicsEnabled` get set?

leumasme: The second argument to a plugins' Inject function is always the options object passed to createBot, as called in https://github.com/PrismarineJS/mineflayer/blob/HEAD/lib/plugin_loader.js#L41 . By destructuring like this, we are only taking the physicsEnabled property from the options object.
See https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/blocks.js#L23 for similar usage in an existing plugin

extremeheat (approved):

lib/plugins/physics.js
```diff
@@ -35,7 +35,8 @@ function inject (bot) {
   let lastSentYaw = null
   let doPhysicsTimer = null
   let lastPhysicsFrameTime = null
-  bot.physicEnabled = false
+  let shouldUsePhysics = false
```
rom1504, line 38: why did you introduce shouldUsePhysics ? it's duplicated with bot.physicsEnabled

leumasme: Before this commit, physicsEnabled represendet both the internal state of the physics plugin, that disabled physics when riding a horse, using a boat etc; but at the same time was also the exposed property. This meant, that whenever the player got a position(teleport) packet from the server (which is pretty common), or dismounted a boat/horse, physicsEnabled would reset to true.
By introducing shouldUsePhysics for the internal state and only using physics when both the internal state and the exposed toggle are true, physics will only be enabled when they are usable (e.g. not riding), and also not explicitly disabled by the user. See further down, where physicsEnabled was always set to true when the bot got a position packet; this (among other occourances) now sets shouldUsePhysics instead.

rom1504 (approved):
Ok. This is better than before

However this makes me think we should not allow disabling physics at all.

PR #1721, opened 2021-02-28
fix newListener and removeListener stacking on world change - don't know if this has any unintended side effects

### LINKED ISSUE #1719: newListener and removeListener stacking up on world change

## Versions
 - mineflayer: 3.2.0
 - server: paper 1.16.4
 - node: 14.15.4

On world change, listeners called "newListener" and "removeListener" are added. They keep adding up on every world change until it prints an "MaxListenersExceededWarning" to my console. It only happens on world change and dimension change.
The warning is only printed once, however listeners are continuously added.
## What did you try yet?
I tried using bot.removeEventListener('newListener), however that function was not defined.

Did you try any method from the API?
I did not.
Did you try any example? Any error from those?
I used looker.js with some modifications

## Your current code
```js

if (process.argv.length < 4 || process.argv.length > 6) {
  console.log('Usage : node echo.js <host> <port> [<name>] [<password>]')
  process.exit(1)
}

  host: process.argv[2],
  port: parseInt(process.argv[3]),
  username: process.argv[4] ? process.argv[4] : 'looker',
  password: process.argv[5],
  version: '1.16.4'
})

setInterval(() => {
  console.log(bot._events)
}, 5000)

bot.on('spawn', () => {
  bot.chat('uwu')
})
bot.once('spawn', function () {
  setInterval(() => {
    const entity = bot.nearestEntity()
    if (entity !== null) {
      if (entity.type === 'player') {
        bot.lookAt(entity.position.offset(0, 1.6, 0))
      } else if (entity.type === 'mob') {
        bot.lookAt(entity.position)
      }
    }
  }, 50)
})

```

## Expected behavior
I expect to be only one "newListener" and "removeListener" and not have them stack up over time.
## Additional context
This seems to only happem on world change, or dimension change.
Running bot.quit and relogging into the server resets it.

Easiest way to reproduce it:
1. Enter nether portal
2. Teleport bot to you
3. Loop

Might help:
https://gist.github.com/U5B/2243710e2c9ccd768d8ecdd375a86d48

U5B:
L456 and L450 I edited

I do not know if this is the correct edit so I hope that someone can give some insight into this.

Fixes #1719

### DIFF
--- a/lib/plugins/blocks.js
+++ b/lib/plugins/blocks.js
@@ -458,9 +458,9 @@ function inject (bot, { version, storageBuilder }) {
     startListenerProxy()
   })
 
-  let listenerState = false
+  let listenerState
   function startListenerProxy () {
-    if (listenerState === true) {
+    if (listenerState) {
       bot.removeAllListeners('removeListener')
     }
     // standardized forwarding
@@ -470,8 +470,8 @@ function inject (bot, { version, storageBuilder }) {
     }
     // custom forwarder for custom events
     const blockUpdateRegex = /blockUpdate:\(-?\d+, -?\d+, -?\d+\)/
-    if (listenerState === false) {
-      bot.on('newListener', (event, listener) => {
+    if (!listenerState) {
+      listenerState = bot.on('newListener', (event, listener) => {
         if (blockUpdateRegex.test(event)) {
           bot.world.on(event, listener)
         }
@@ -482,9 +482,6 @@ function inject (bot, { version, storageBuilder }) {
         bot.world.on(event, listener)
       }
     })
-    if (listenerState === false) {
-      listenerState = true
-    }
   }
 
   bot.findBlock = findBlock
--- a/lib/plugins/blocks.js
+++ b/lib/plugins/blocks.js
@@ -458,7 +458,7 @@ function inject (bot, { version, storageBuilder }) {
     startListenerProxy()
   })
 
-  let listenerState
+  let listenerState = undefined
   function startListenerProxy () {
     if (listenerState) {
       bot.removeAllListeners('removeListener')
@@ -470,7 +470,7 @@ function inject (bot, { version, storageBuilder }) {
     }
     // custom forwarder for custom events
     const blockUpdateRegex = /blockUpdate:\(-?\d+, -?\d+, -?\d+\)/
-    if (!listenerState) {
+    if (listenerState === undefined) {
       listenerState = bot.on('newListener', (event, listener) => {
         if (blockUpdateRegex.test(event)) {
           bot.world.on(event, listener)
--- a/lib/plugins/blocks.js
+++ b/lib/plugins/blocks.js
@@ -461,9 +461,7 @@ function inject (bot, { version, storageBuilder }) {
   let listenerState = false
   function startListenerProxy () {
     if (listenerState === true) {
-      bot.removeAllListeners('newListener')
       bot.removeAllListeners('removeListener')
-      listenerState = false
     }
     // standardized forwarding
     const forwardedEvents = ['blockUpdate', 'chunkColumnLoad', 'chunkColumnUnload']
@@ -472,17 +470,21 @@ function inject (bot, { version, storageBuilder }) {
     }
     // custom forwarder for custom events
     const blockUpdateRegex = /blockUpdate:\(-?\d+, -?\d+, -?\d+\)/
-    bot.on('newListener', (event, listener) => {
-      if (blockUpdateRegex.test(event)) {
-        bot.world.on(event, listener)
-      }
-    })
+    if (listenerState === false) {
+      bot.on('newListener', (event, listener) => {
+        if (blockUpdateRegex.test(event)) {
+          bot.world.on(event, listener)
+        }
+      })
+    }
     bot.on('removeListener', (event, listener) => {
       if (blockUpdateRegex.test(event)) {
         bot.world.on(event, listener)
       }
     })
-    listenerState = true
+    if (listenerState === false) {
+      listenerState = true
+    }
   }
 
   bot.findBlock = findBlock
--- a/lib/plugins/blocks.js
+++ b/lib/plugins/blocks.js
@@ -447,16 +447,14 @@ function inject (bot, { version, storageBuilder }) {
     }
   }
   bot._client.on('login', (packet) => {
-    if (bot.world) return
     dimension = packet.dimension
     bot.world = new World(null, storageBuilder ? storageBuilder({ version: bot.version, worldName: dimensionToFolderName(packet.dimension) }) : null).sync
     startListenerProxy()
   })
   bot._client.on('respawn', (packet) => {
-    if (dimension === packet.dimension || bot.world) return
+    if (dimension === packet.dimension) return
     dimension = packet.dimension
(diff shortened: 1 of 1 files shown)


U5B:
This fix does "fix" it but it causes other hidden problems when switching worlds, such as increasing the memory usage way too much so a better fix is needed.

rom1504 (changes requested):
That's the wrong fix.

All these listeners should be created for the new world and removed from the old one.

For newListener you could indeed consider reusing the existing listener. That could be done by splitting the startListener function in 2

For the other listeners you do have to call the function on respawn

U5B:
Actually I believe this only fixes dimension changes and not the actual problem. The actual problem is that we need to remove already existing newlistener and removeListener

U5B:
To be honest, I do not understand this code enough for me to fix it. So I would like some assistance with this since I am a complete noob at javascript

U5B:
hmm, if intensively switching worlds with this fix, the cpu goes from 0.10 to 10.0 and memory usage goes up for a short time.

Intesively switching worlds means that a command block is teleporting you immediately into the nether from the overworld and vice versa when you spawn .

lib/plugins/blocks.js
```diff
@@ -458,24 +458,33 @@ function inject (bot, { version, storageBuilder }) {
     startListenerProxy()
   })
 
+  let listenerState = false
```
rom1504, line 461: Why not store the listener instead ? It would be cleaner

U5B: Could you give an example of storing the listener?

Is it like putting a variable before the listener?

U5B: I believe this is what you mean.

lib/plugins/blocks.js
```diff
@@ -457,20 +456,25 @@ function inject (bot, { version, storageBuilder }) {
...
     }
     // custom forwarder for custom events
     const blockUpdateRegex = /blockUpdate:\(-?\d+, -?\d+, -?\d+\)/
-    bot.on('newListener', (event, listener) => {
-      if (blockUpdateRegex.test(event)) {
-        bot.world.on(event, listener)
-      }
-    })
+    if (listenerState === undefined) {
+      listenerState = bot.on('newListener', (event, listener) => {
```
Karang, line 472: You are using `listenerState` as a boolean, it would be more easy to understand if you made it more explicit. ie:
```js
if (!listenerState) {
   listenerState = true
   bot.on( // ...
```

Karang (commented):
One minor comment, otherwise LGTM

lib/plugins/blocks.js
Karang: Edit: just saw Rom comments, I'm gonna read the discussion you had

lib/plugins/blocks.js
```diff
@@ -457,20 +456,25 @@ function inject (bot, { version, storageBuilder }) {
     bot.world = new World(null, storageBuilder ? storageBuilder({ version: bot.version, worldName: dimensionToFolderName(packet.dimension) }) : null).sync
     startListenerProxy()
   })
-
+  let listenerState
   function startListenerProxy () {
+    if (listenerState !== undefined) {
+      bot.removeAllListeners('removeListener')
```
Karang, line 462: The `newListener` event listeners should be removed as well.

lib/plugins/blocks.js
```diff
@@ -457,20 +456,25 @@ function inject (bot, { version, storageBuilder }) {
...
       bot.world.on(event, (...args) => bot.emit(event, ...args))
     }
     // custom forwarder for custom events
     const blockUpdateRegex = /blockUpdate:\(-?\d+, -?\d+, -?\d+\)/
-    bot.on('newListener', (event, listener) => {
-      if (blockUpdateRegex.test(event)) {
-        bot.world.on(event, listener)
-      }
-    })
+    if (listenerState === undefined) {
```
Karang, line 471: if the `newListener` from the previous world is removed, you don't need this if

lib/plugins/blocks.js
Karang, line 472: you need to store the listener (the function) if you want to be able to remove this one specifically (and avoid removing all listeners)
```js
listenerState = (event, listener) => { /* the function */}
bot.on('newListener', listenerState)
```
then you can do `bot.removeListener('newListener', listenerState)`

And do the same for the removeListener event

U5B: I was experimenting with this, however it wasn't working properly with `bot.removeListener`

U5B: `js
  function startListenerProxy () {
    let listener = (event, listener) => {
      if (blockUpdateRegex.test(event)) {
        bot.world.on(event, listener)
      }
    }
    bot.removeListener('removeListener', listener)
    bot.removeListener('newListener', listener)
    // standardized forwarding
    const forwardedEvents = ['blockUpdate', 'chunkColumnLoad', 'chunkColumnUnload']
    for (const event of forwardedEvents) {
      bot.world.on(event, (...args) => bot.emit(event, ...args))
    }
    // custom forwarder for custom events
    const blockUpdateRegex = /blockUpdate:\(-?\d+, -?\d+, -?\d+\)/
    bot.on('newListener', listener)
    bot.on('removeListener', listener)
  }
`

U5B: This is the code that I was trying to use, however it doesn't seem to remove them properly

Karang: try with
```js
  let listenerState
  function startListenerProxy () {
    if (listenerState) {
      // custom forwarder for custom events
      bot.removeListener('removeListener', listenerState)
      bot.removeListener('newListener', listenerState)
    }
    // standardized forwarding
    const forwardedEvents = ['blockUpdate', 'chunkColumnLoad', 'chunkColumnUnload']
    for (const event of forwardedEvents) {
      bot.world.on(event, (...args) => bot.emit(event, ...args))
    }
    const blockUpdateRegex = /blockUpdate:\(-?\d+, -?\d+, -?\d+\)/
    listenerState = (event, listener) => {
      if (blockUpdateRegex.test(event)) {
        bot.world.on(event, listener)
      }
    }
    bot.on('newListener', listenerState)
    bot.on('removeListener', listenerState)
  }
```

U5B: Yep that code works I believe.

lib/plugins/blocks.js
```diff
@@ -457,25 +456,28 @@ function inject (bot, { version, storageBuilder }) {
     bot.world = new World(null, storageBuilder ? storageBuilder({ version: bot.version, worldName: dimensionToFolderName(packet.dimension) }) : null).sync
     startListenerProxy()
   })
-
+  let listenerState
   function startListenerProxy () {
+    if (listenerState) {
+      // remove the forwarders if they already exist to prevent performance issues
+      bot.removeListener('removeListener', listenerState)
```
rom1504, line 463: I don't understand why removing a listener does the same as adding

U5B: ask node, but if I don't specify the arguments matching the listener being removed, it fails to remove it

U5B: "listener argument must be of type function"

rom1504 (approved):

PR #1722, opened 2021-02-28
Use Nullish Coalescing Operator for Default Values

leumasme:
The Nullish Coalescing Operator (??) was introduced in Node 14 which is currently already the minimum Version that Mineflayer can be used with.
The Nullish Coalescing Operator is used like this:
`value = value ?? default`
And behaves like this:
`value = (value === undefined || value === null ? default : value)`
This is a better alternative to the OR || Operator, because using the OR operator can create bugs where a falsly value may be allowed. For example: index.js: Line 65
`options.hideErrors = options.hideErrors || true`
I do not know how noone ever noticed this, but this code can never assign false to options.hideErrors:
When omitting the option, undefined || true will result in true, as is intended
However when setting the option to false, false || true will also result in true, which makes it impossible to assign false to that value - it will always be true

### DIFF
--- a/index.js
+++ b/index.js
@@ -64,7 +64,7 @@ function createBot (options = {}) {
   options.plugins = options.plugins ?? {}
   options.hideErrors = options.hideErrors ?? true
   options.logErrors = options.logErrors ?? true
-  options.loadInternalPlugins = options.loadInternalPlugins ?? false
+  options.loadInternalPlugins = options.loadInternalPlugins ?? true
   const bot = new EventEmitter()
   bot._client = null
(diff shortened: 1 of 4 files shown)


leumasme:
Another example of this causing a bug:
command_block.js; Line 17
`options.mode = options.mode || 2 // Possible values: 0: SEQUENCE, 1: AUTO and 2: REDSTONE`
Comment indicats that 0 is a valid value; however when passing 0, it will default to 2 as 0 is nullish, so (0 || 2) == 2
This is also fixed by using ??

rom1504 (approved):

PR #1725, opened 2021-03-01
clean up/update readme

u9g:
changes: 
* remove deprecated packages: mineflayer-navigate
* update schematic project to be our own because it's more feature-rich
* remove link from cheesebot since the link is down
* comment out version & auth lines of example since they aren't needed except for special cases
* update pathfinder example link
* add ansi example
* add guard example
* add multiple-from-file example
* remove old testing suggestions

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -126,6 +126,10 @@ And you'll get a *live* view looking like this:
 
 And many mores in the [examples](https://github.com/PrismarineJS/mineflayer/tree/master/examples) folder
 
+### Modules
+
+A lot of the active development is happening inside of small npm packages which are used by mineflayer.
+
 #### The Node Way&trade;
 
 > "When applications are done well, they are just the really application-specific, brackish residue that can't be so easily abstracted away. All the nice, reusable components sublimate away onto github and npm where everybody can collaborate to advance the commons." — substack from ["how I write modules"](https://gist.github.com/substack/5075355)
--- a/docs/README.md
+++ b/docs/README.md
@@ -78,12 +78,12 @@ Without auth specified, the mojang auth style will be guessed.
 const mineflayer = require('mineflayer')
 
 const bot = mineflayer.createBot({
-  host: 'localhost', // optional
-  port: 25565,       // optional
-  username: 'email@example.com', // email and password are required only for
-  password: '12345678',          // online-mode=true servers
-  // version: false,             /* only set if you need a specific version or snapshot (ie: "1.8.9" or "1.16.5"), otherwise it's set automatically */
-  // auth: 'mojang'              /* only set if you need microsoft auth, then set this to 'microsoft' */
+  host: 'localhost',             // minecraft server ip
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -70,8 +70,8 @@ If you want to learn more, more video tutorials are [there,](https://www.youtube
 
 **Getting Started**
 
-Without version specified, the version of the server will be guessed automatically, you can set a specific one using the version option.
-For example `version:"1.8"`.
```
Karang, line 74: why remove the version example ?

docs/README.md
```diff
@@ -114,19 +114,18 @@ And you'll get a *live* view looking like this:
...
+|[ansi](https://github.com/PrismarineJS/mineflayer/blob/master/examples/ansi.js) | Display your bot's chat with all of the chat colors shown in your terminal |
+|[guard](https://github.com/PrismarineJS/mineflayer/blob/master/examples/guard.js) | Make a bot guard a defined area from nearby mobs |
+|[multiple-from-file](https://github.com/PrismarineJS/mineflayer/blob/master/examples/multiple_from_file.js) | Add a text file with accounts and have them all login |
 
 And many mores in the [examples](https://github.com/PrismarineJS/mineflayer/tree/master/examples) folder
 
-### Modules
-
-A lot of the active development is happening inside of small npm packages which are used by mineflayer.
-
```
Karang, line 129: Why remove this ?

docs/README.md
u9g: its a few lines down still there, just not really as important since mineflayer can see what version it needs, which is all that most users need

docs/README.md
u9g: There's two modules categories, one which said this, and the other with all the p-* packages

Karang: It's not the same title level, this one is the big title

u9g: reverted

docs/README.md
```diff
@@ -82,28 +82,28 @@ const bot = mineflayer.createBot({
   port: 25565,       // optional
   username: 'email@example.com', // email and password are required only for
   password: '12345678',          // online-mode=true servers
-  version: false,                 // false corresponds to auto version detection (that's the default), put for example "1.8.8" if you need a specific version
-  auth: 'mojang'      // optional; by default uses mojang, if using a microsoft account, set to 'microsoft'
+  // version: false,             /* only set if you need a specific version or snapshot (ie: "1.8.9" or "1.16.5"), otherwise it's set automatically */
+  // auth: 'mojang'              /* only set if you need microsoft auth, then set this to 'microsoft' */
```
leumasme, line 86: This isn't really consistent - Why comment out optional properties here and use multi-line comments for commenting at the end of lines, when other optional properties are not commented out and the regular single-line comment format is used everywhere?

u9g: fixed.

rom1504 (approved):

PR #1729, opened 2021-03-02
Add an option like a digging

CyberPatrick:


lib/plugins/place_block.js
```diff
@@ -22,7 +22,9 @@ function inject (bot, { version }) {
       dy = options.delta.y
       dz = options.delta.z
     }
-    await bot.lookAt(referenceBlock.position.offset(dx, dy, dz), false)
+    if (options.forceLook !== 'ignore') {
+      await bot.lookAt(referenceBlock.position.offset(dx, dy, dz), options.forceLook)
```
rom1504, line 26: You are passing here something of type string to something that expects a boolean

CyberPatrick: Options.forceLook can be: 'ignore', false, true. If it is not 'ignore' - it is boolean. https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/digging.js#L30-L40 - here the same thing.

rom1504: Is it handled by the function you're passing it too ?

CyberPatrick: No

Karang: lookAt isnt called if forceLook isnt a boolean

CyberPatrick: So i should add check? 
```
If (options.forceLook !== 'ignore' && typeof options.forceLook !== Boolean) {
    throw new Error('forceLook invalid input')
}
```
Or just in doc add info about types.

rom1504 (approved):

PR #1735, opened 2021-03-05
add a 'messagestr' event

u9g:
I think a lot of people use the message event, but don't care about the color or any other information from the message, and just instantly make it into a string using `msg.toString()` then use the input, this adds a native way to just use the stringified version of message instantly, without the boilerplate. An example of when a chataddpattern wouldn't be better is when I have to parse messages that happen in series, in this case its just cleaner to have it all in one switch, or an if else ladder. I'd argue that this isn't any more extra then emitting a different event for an actionbar message, because that too is emitted by a separate event to make it easier.


rom1504:
If this is to be used there should be:
* adaptation to doc
* use it in an example ?
* maybe a word in the FAQ ?

why don't people use chat event instead ?

u9g:
> If this is to be used there should be:
> 
> * adaptation to doc
> * use it in an example ?
> * maybe a word in the FAQ ?
> 
> why don't people use chat event instead ?

1. Done

rom1504 (approved):

PR #1742, opened 2021-03-08
created an example afk plugin

### LINKED ISSUE #1401: Add plugin creation examples
Many new contributors want to create new plugins for Mineflayer, but are either unsure how to begin or use an unexpected format.

Adding some example plugins in the examples folder might make it much easier for new contributors to create plugins using a predictable format.

N/A

## Additional context
This should make it much clearer how plugins work and how plugins are expected to be loaded and operate.

This might also be a starting point for standardizing plugin metadata. (#1244)

Zn10plays:
added a plugin directory in the examples directory  
Created an example plugin 

close #1401

I sware this works pls let me pr.

### DIFF
--- a/examples/plugins/afk.js
+++ b/examples/plugins/afk.js
@@ -8,7 +8,6 @@
 // Each plugin should have a inject function
 // This is the function that will be put in bot.loadPlugin
 function inject (bot, option) {
-
 // It is a good idea to make a namespace for your plugin
   bot.afk = {}
 
--- a/examples/plugins/afk.js
+++ b/examples/plugins/afk.js
@@ -34,9 +34,7 @@ function inject (bot, option) {
   }
 }
 
-// Most of the time plugins are stored in different files so you must export them.
 module.exports = {
-  // Here you should name your inject function what ever you want.
   afk: inject
 }
 // This is a simple plugin with two methods, start and stop.
--- a/examples/plugins/afk.js
+++ b/examples/plugins/afk.js
@@ -1,8 +1,5 @@
-// Plugins are loaded using the bot.loadPlugin() function
-// The load plugin function excepts one function as its only parameter
-// The plugin runs the function passing it two parameter
-// The former is the bot object and the latter is the bot option
-// the options contain version player info and other plugins
+// All plugins are loaded with the bot.loadPlugin function, look at documentation for more info
+// https://github.com/PrismarineJS/mineflayer/blob/master/docs/api.md#botloadpluginplugin
 // Here is an example AFK plugin
 
 // Each plugin should have a inject function
@@ -15,7 +12,7 @@ function inject (bot, option) {
   bot.afk.status = 'Idle'
 
   // All your plugin functions should be in the name space other than the logic.
-  bot.afk.start = async function () {
+  bot.afk.start = async () => {
     afkInterval = setInterval(async () => {
       if (rotation === 0) {
         await bot.look(0, 0)
@@ -28,7 +25,7 @@ function inject (bot, option) {
     bot.afk.status = 'afk'
   }
 
-  bot.afk.stop = function () {
+  bot.afk.stop = () => {
     if (afkInterval) {
       clearInterval(afkInterval)
       bot.afk.status = 'Idle'
@@ -41,9 +38,4 @@ module.exports = {
   // Here you should name your inject function what ever you want.
   afk: inject
 }
-
 // This is a simple plugin with two methods, start and stop.
-// On the start method this plugin makes the bot look from east to west in 3-second intervals,
-// On stop it stops the bot from looking constantly spinning around
-// Additionally it also has a status variable inside the name space,
-// Users can use the variable as a reference for the code.
--- a/examples/plugins/afk.js
+++ b/examples/plugins/afk.js
@@ -1,32 +1,30 @@
-function inject(bot, option) {
-    
-    //it is a good idea to make a same space for your plugin
-    bot.afk = {
-        start,
-        stop,
-    }
+function inject (bot, option) {
+  // it is a good idea to make a same space for your plugin
+  bot.afk = {
+    start,
+    stop
+  }
 
-    let afk, rotation = 0
+  let afk; let rotation = 0
 
-    async function start() {
+  async function start () {
+    afk = setInterval(async () => {
+      if (rotation === 0) {
+        await bot.look(0, 0)
+        rotation = 1
+      } else {
+        await bot.look(Math.PI, 0)
+        rotation = 0
+      }
+    }, 3000)
+  }
 
-        afk = setInterval(async ()=>{
-            if(rotation===0) {
-                await bot.look(0,0)
-                rotation=1
-            } else {
-                await bot.look(Math.PI,0)
-                rotation=0
-            }
-        },3000)
-    }
-
-    function stop() {
-        if (!afk) return 
-        clearInterval(afk)
-    }
+  function stop () {
+    if (!afk) return
+    clearInterval(afk)
+  }
 }
 
 module.exports = {
-    afk:inject
+  afk: inject
 }
--- a/examples/plugins/afk.js
+++ b/examples/plugins/afk.js
@@ -5,13 +5,14 @@
 // Each plugin should have a inject function
 // This is the function that will be put in bot.loadPlugin
 function inject (bot, option) {
+  // create a scope for you functions
   bot.afk = {}
 
   let afkInterval, rotation
-  // This namespace can be used to store some variables for user reference.
+  // This scope can be used to store some variables for user reference.
   bot.afk.status = 'Idle'
 
-  // All your plugin functions should be in the name space other than the logic.
+  // All your plugin functions should be in the scope other than the logic.
   bot.afk.start = async () => {
     afkInterval = setInterval(async () => {
       if (rotation === 0) {
--- a/examples/plugins/afk.js
+++ b/examples/plugins/afk.js
@@ -8,7 +8,6 @@
 // Each plugin should have a inject function
 // This is the function that will be put in bot.loadPlugin
 function inject (bot, option) {
-// It is a good idea to make a namespace for your plugin
   bot.afk = {}
 
   let afkInterval, rotation
--- a/examples/plugins/afk.js
+++ b/examples/plugins/afk.js
@@ -9,32 +9,29 @@ function inject (bot, option) {
   bot.afk = {}
 
   let afkInterval, rotation
-  // This scope can be used to store some variables for user reference.
-  bot.afk.status = 'Idle'
 
   // All your plugin functions should be in the scope other than the logic.
   bot.afk.start = async () => {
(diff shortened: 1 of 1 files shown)


rom1504 (changes requested):
what is it supposed to do ?
add explanation as a comment in the file

also explain how to use it

Zn10plays:
It is a plugin example.

Its main purpose is to show new creators how easy it is to make plugins.

This plugin is just an AFK plugin with two methods. afk.start() and afk.stop()

For the rest, you can see you opening the file.

examples/plugins/afk.js
```diff
@@ -0,0 +1,30 @@
+function inject (bot, option) {
+  // it is a good idea to make a same space for your plugin
+  bot.afk = {
+    start,
+    stop
+  }
+
+  let afk; let rotation = 0
```
u9g, line 8: use newlines or make this actually one line, ie: `let afk, rotation = 0;`

examples/plugins/afk.js
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  function stop () {
+    if (!afk) return
+    clearInterval(afk)
+  }
+}
+
+module.exports = {
+  afk: inject
```
u9g, line 29: wait so this does nothing? have you tested it?

examples/plugins/afk.js
```diff
@@ -0,0 +1,30 @@
...
+    stop
+  }
+
+  let afk; let rotation = 0
+
+  async function start () {
+    afk = setInterval(async () => {
+      if (rotation === 0) {
+        await bot.look(0, 0)
+        rotation = 1
```
u9g, line 14: by making rotation 0/1 you've essentially made a boolean, maybe actually use a boolean to be explicit?

examples/plugins/afk.js
```diff
@@ -0,0 +1,30 @@
...
+      } else {
+        await bot.look(Math.PI, 0)
+        rotation = 0
+      }
+    }, 3000)
+  }
+
+  function stop () {
+    if (!afk) return
```
u9g, line 23: you can make this one line by doing if(afk) clearInterval....

examples/plugins/afk.js
Zn10plays: What do you mean?

yes, I tested it. It works 1.16.5 without any issues.

u9g: why not call afk inject then if you are exporting it as inject?

rom1504:
do you intend to improve this ?

Zn10plays:
I really hate standard

examples/plugins/afk.js
```diff
@@ -0,0 +1,50 @@
...
+
+// Each plugin should have a inject function
+// This is the function that will be put in bot.loadPlugin
+function inject (bot, option) {
+// It is a good idea to make a namespace for your plugin
+  bot.afk = {}
+
+  let afkInterval, rotation
+  // This namespace can be used to store some variuabels for user reffrence.
+  bot.afk.status = 'Idel'
```
u9g, line 16: you misspelled idle

examples/plugins/afk.js
```diff
@@ -0,0 +1,50 @@
+// Plugins are loaded using the bot.loadPlugin() function
```
u9g, line 1: this information should be and is in api.md, theres no reason to rewrite a description of bot.loadplugin

examples/plugins/afk.js
```diff
@@ -0,0 +1,50 @@
...
+// The load plugin function excepts one function as its only parameter
+// The plugin runs the function passing it two parameter
+// The former is the bot object and the latter is the bot option
+// the options contain version player info and other plugins
+// Here is an example AFK plugin
+
+// Each plugin should have a inject function
+// This is the function that will be put in bot.loadPlugin
+function inject (bot, option) {
+// It is a good idea to make a namespace for your plugin
```
u9g, line 11: thats not what a namespace is, you are talking about scope, namespace doesnt exist in js

examples/plugins/afk.js
```diff
@@ -0,0 +1,50 @@
...
+  bot.afk = {}
+
+  let afkInterval, rotation
+  // This namespace can be used to store some variuabels for user reffrence.
```
u9g, line 15: mispelled variables and reference

examples/plugins/afk.js
```diff
@@ -0,0 +1,50 @@
...
+  bot.afk.status = 'Idel'
+
+  // All your plugin functions should be in the name space other than the logic.
+  bot.afk.start = async function () {
+    afkInterval = setInterval(async () => {
```
u9g, line 20: whats the point in this being async, awaiting before setting the boolean to true is pointless

examples/plugins/afk.js
```diff
@@ -0,0 +1,50 @@
...
+      if (rotation === 0) {
+        await bot.look(0, 0)
+        rotation = true
+      } else {
+        await bot.look(Math.PI, 0)
+        rotation = false
+      }
+    }, 3000)
+    bot.afk.status = 'afk'
```
u9g, line 29: imo the bot.afk.status shouldn't be polluting the bot object, it should be a variable scoped within the plugin

examples/plugins/afk.js
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  bot.afk.stop = function () {
```
u9g, line 32: () => {

examples/plugins/afk.js
```diff
@@ -0,0 +1,50 @@
...
+    if (afkInterval) {
+      clearInterval(afkInterval)
+      bot.afk.status = 'Idel'
```
u9g, line 35: mispelled idle

u9g (changes requested):

examples/plugins/afk.js
Zn10plays: Doesn't matter 
in fact, it would be better this way because function () {}
doesn't inherent anything form the higher scope other than the var

examples/plugins/afk.js
Zn10plays: This can be used as a reference which is quite convenient. 
Let's say you are AFK and you don't know that. Then you try to walk with pathfider, it would cause issues if you keep looking away from the path you are trying to walk.
So I think it is a good idea to let future devs know the importance of reffrence

examples/plugins/afk.js
Zn10plays: This way is more precise, so I don't want to change it. Think about it if the server has a lot of lag, the bloolen would be assigned too early, and if the lag continues it will cause issues with too many requests going on.

does that make any sence?

Zn10plays:
fixed the spelling :)

examples/plugins/afk.js
Zn10plays: It is hard to explain how to make a plugin without a decent understanding of how plugins are loaded.

u9g: you should just link to the api.md entry for bot.loadplugin then

examples/plugins/afk.js
u9g: why would it be better lol, thats old syntax but ig it doesnt matter

examples/plugins/afk.js
u9g: wouldn't it be better to have a getter then, so they can't mutate the status and cause problems?

examples/plugins/afk.js
u9g: no, not really, a server having 3 seconds of lag between look packets is unheardof

Moondarker: I had a 4 minute delay between pearl meeting the ground and me getting teleported there today, guess what was happening with packets. The fact that this won't stop the bot from sending look packets is a different question:
Open browser console and try this
`setInterval(async () => {setTimeout(console.log, 2500, 'this will execute every 500ms with initial delay of 3s')}, 500)`

examples/plugins/afk.js
```diff
@@ -0,0 +1,41 @@
...
+// This is the function that will be put in bot.loadPlugin
+function inject (bot, option) {
+  bot.afk = {}
+
+  let afkInterval, rotation
+  // This namespace can be used to store some variables for user reference.
+  bot.afk.status = 'Idle'
+
+  // All your plugin functions should be in the name space other than the logic.
+  bot.afk.start = async () => {
```
dada513, line 15: This doesnt need to be async

dada513 (approved):
I think its ready now

examples/plugins/afk.js
Zn10plays: this is only supposed to be an small example so I see no need for a getter.

examples/plugins/afk.js
```diff
@@ -0,0 +1,42 @@
...
+    }
+  }
+}
+
+// Most of the time plugins are stored in different files so you must export them.
+module.exports = {
+  // Here you should name your inject function what ever you want.
+  afk: inject
+}
+// This is a simple plugin with two methods, start and stop.
```
u9g, line 42: not necessary

examples/plugins/afk.js
```diff
@@ -0,0 +1,42 @@
...
+    if (afkInterval) {
+      clearInterval(afkInterval)
+      bot.afk.status = 'Idle'
+    }
+  }
+}
+
+// Most of the time plugins are stored in different files so you must export them.
+module.exports = {
+  // Here you should name your inject function what ever you want.
```
u9g, line 39: not needed

examples/plugins/afk.js
```diff
@@ -0,0 +1,42 @@
...
+
+  bot.afk.stop = () => {
+    if (afkInterval) {
+      clearInterval(afkInterval)
+      bot.afk.status = 'Idle'
+    }
+  }
+}
+
+// Most of the time plugins are stored in different files so you must export them.
```
u9g, line 37: they are always, this isnt needed

examples/plugins/afk.js
```diff
@@ -0,0 +1,42 @@
...
+
+// Each plugin should have a inject function
+// This is the function that will be put in bot.loadPlugin
+function inject (bot, option) {
+  // create a scope for you functions
+  bot.afk = {}
+
+  let afkInterval, rotation
+  // This scope can be used to store some variables for user reference.
+  bot.afk.status = 'Idle'
```
u9g, line 13: isnt the terminology of idle and afk kind of synonymous

examples/plugins/afk.js
```diff
@@ -0,0 +1,42 @@
...
+
+  // All your plugin functions should be in the scope other than the logic.
+  bot.afk.start = async () => {
+    afkInterval = setInterval(async () => {
+      if (rotation === 0) {
```
u9g, line 18: this will never be true since you treat rotation as a boolean so using === will never work against a number

rom1504:
please check u9g comments

Zn10plays:
I am working on it

rom1504 (approved):

PR #1745, opened 2021-03-08
[WIP] go back to node 12

rom1504:


TheAlan404:
But why?

rom1504:
yeah but actually not worth it, let's stay on 14

PR #1755, opened 2021-03-13
Add a spammer example

DrMoraschi:
Just saw a message in discord of someone asking for a spammer example

### DIFF
--- a/examples/spammer.js
+++ b/examples/spammer.js
@@ -1,0 +1,24 @@
+/*
+ *
+ * A bot that only spams
+ *
+ */
+const mineflayer = require('mineflayer')
+
+if (process.argv.length < 4 || process.argv.length > 6) {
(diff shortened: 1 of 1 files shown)


DrMoraschi:
Oh welp, i don't understand

Karang:
1) you need to run standard --fix to fix tests
2) I don't think we want to add this kind of example, there is more interesting stuff one can do using mineflayer, than spamming...

DrMoraschi:
Fair enough

wvffle (changes requested):
I don't think we should teach how to create malicious bots...

We already have a tutorial that explains the basics of javascript here: https://github.com/PrismarineJS/mineflayer/blob/master/docs/tutorial.md

It's lacking setInterval; you could add it. Then a person asking how to create such bot wouldn't have any questions after reading this guide.

PR #1760, opened 2021-03-14
Change transaction apology packet to match vanilla client

FeldrinH:
When server rejects an inventory transaction packet, mineflayer sends a serverbound transaction packet to apologize. I noticed that the server seemed to not accept these apologies in some cases. This PR changes the apology packet to match apology packets sent by 1.16.5 vanilla client (as observed with a minecraft packet sniffer), which seems to make the apology system more reliable.

### DIFF
--- a/lib/plugins/inventory.js
+++ b/lib/plugins/inventory.js
@@ -359,9 +359,9 @@ function inject (bot, { version, hideErrors }) {
 
     function onRejected () {
       bot._client.write('transaction', {
-        windowId: 0,
+        windowId: click.windowId,
         action: click.id,
-        accepted: false
+        accepted: true
       })
       bot.emit(`confirmTransaction${click.id}`, false)
     }


lib/plugins/inventory.js
```diff
@@ -359,9 +359,9 @@ function inject (bot, { version, hideErrors }) {
 
     function onRejected () {
       bot._client.write('transaction', {
-        windowId: 0,
+        windowId: click.windowId,
         action: click.id,
-        accepted: false
+        accepted: true
```
rom1504, line 364: why accepted true ?

FeldrinH: Accepted true was the value used by the vanilla 1.16.5 client when sending apology packets. (this was determined using [pakkit](https://github.com/Heath123/pakkit))

rom1504 (approved):
seems weird but not breaking tests

will revert if it turns out to be wrong

PR #1764, opened 2021-03-16
fix entity name

rom1504:


lib/plugins/entities.js
```diff
@@ -199,6 +209,7 @@ function inject (bot, { version }) {
   bot._client.on('spawn_entity_experience_orb', (packet) => {
     const entity = fetchEntity(packet.entityId)
     entity.type = 'orb'
+    entity.name = 'experience_orb'
```
rom1504, line 212: hardcoded it because it's not consistent per version and we don't have the id here

PR #1768, opened 2021-03-17
Create examples/advanced/

u9g:
In this pr, I added an advanced folder for examples that not everyone will use, but have their specific use, that several people may encounter


U5B (approved):
I like this.

rom1504 (approved):

rom1504:
Nice

Issue #1766, opened 2021-03-16
add bot.waitforticks to typings


u9g:
I deleted the template because theres no point, its literally just the title

PR #1769, opened 2021-03-17
add waitforticks to typings

u9g:


u9g:
closes https://github.com/PrismarineJS/mineflayer/issues/1766

rom1504 (approved):

rom1504:
Nice
I bet it's easy to do a small script that automatically check api.md and typings are aligned (just check methods and event sets are the same)

PR #1771, opened 2021-03-17
change listener to listenerAdd

U5B:
They might be conflicting each other.

Since listener was a global variable, and it might've been being passed into bot.world.on(event, listener) which would have not been correct.

The other question is L462 and L451 are cleaned up on switch.


U5B:
This probably isn't needed.

PR #1778, opened 2021-03-18
Add proxy support (untested)

BlueBurgersTDD:
This PR adds easier support for proxies. It's currently untested.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -32,6 +32,12 @@ export interface BotOptions extends ClientOptions {
   difficulty?: number;
   chatLengthLimit?: number;
   physicsEnabled?: boolean;
+  proxy?: {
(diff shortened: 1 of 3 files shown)


rom1504:
if we want to support this, it should go into https://github.com/PrismarineJS/node-minecraft-protocol

however what's the problem with the example?

BlueBurgersTDD:
> if we want to support this, it should go into https://github.com/PrismarineJS/node-minecraft-protocol
ok sure i can move it there
> however what's the problem with the example?
i've seen people not reading the example/not understanding. There are 58 issues about proxies

BlueBurgersTDD:
But ill close this PR and make a new one in https://github.com/PrismarineJS/node-minecraft-protocol

PR #1779, opened 2021-03-18
allow sleeping during rain/thunderstorms

qrvd:


qrvd:
The downside is that `bot.isRaining` is defined in the rain.js plugin, so this would make the bed.js plugin depend on the rain.js plugin

extremeheat (approved):

rom1504:
Do you have a reference (minecraft wiki maybe) that confirms that rule ? Is there any other rule for sleeping that we miss ?

Moondarker:
I thought in only works during thunderstorms?

extremeheat:
Seems like it's been in the game since they added rain https://www.minecraftforum.net/forums/minecraft-java-edition/discussion/148355-rain-and-sleeping?comment=3

Moondarker:
Adding rain and thunderstorm levels to rain.js should be pretty easy, [see reason codes 7 and 8](https://wiki.vg/Protocol#Change_Game_State)

Here's a screenshot of 1.16.4 behavior, Minecraft version where this was added is also to be determined
image

qrvd:
> Do you have a reference (minecraft wiki maybe) that confirms that rule ?
> I thought it only works during thunderstorms?
Rain was re-added to the game in beta 1.5, after beds had been implemented (in beta 1.3). (according to https://minecraft.gamepedia.com/Rain#History and https://minecraft.gamepedia.com/Bed#History)

I confirmed manually that sleeping during rain (not thunderstorm) works in both the old beta 1.5 and in the latest Minecraft version 1.16.5. The in-game text that says "You may only sleep at night or during thunderstorms" is actually misleading, since sleeping works during rainy weather as well.

> Is there any other rule for sleeping that we miss ?
In beta 1.3 (when beds were added), trying to sleep in the Nether does nothing. Trying to sleep in the Nether since beta 1.6 causes the bed to explode. Beds also explode when trying to sleep in The End, but I didn't find info about when this was added.

qrvd:
Yes, beds explode in The End (since version 1.0.0 when The End was added to the game)

Moondarker:
> I confirmed manually that sleeping during rain (not thunderstorm) works in both the old beta 1.5 and in the latest Minecraft version 1.16.5. The in-game text that says "You may only sleep at night or during thunderstorms" is actually misleading, since sleeping works during rainy weather as well.
@qrvd 

Aight, I had to hop in Minecraft to test that once again, and yes, you can not sleep during rains in 1.16.4, however right-clicking on the bed sets the respawn point regardless of weather or day time (so it should be allowed even without the rain, but I'm not sure when was this added, definitely >1.12.2)
As for earlier versions, testing it rn

extremeheat:
Sending the request to the server is harmless, it can decide whether to accept or reject the request. If it's accepted you'll get a sleep action packet back.

Moondarker:
@extremeheat this plugin is supposed to actually put the bot to sleep, so these checks make sense - otherwise you can just activate bed block and handle everything else manually

Moondarker:
Also here are the results, couldn't let the time be wasted in vain :)
https://user-images.githubusercontent.com/4098364/111851687-489d5d80-8925-11eb-82b6-f2f35e689879.mp4

qrvd:
> @ extremeheat this plugin is supposed to actually put the bot to sleep, so these checks make sense - otherwise you can just activate bed block and handle everything else manually
> UPD: On the other hand, leaving the logic to server seems good too, the only thing that 100% has to be checked is dimension

@Moondarker I think dimension should be checked as well, but maybe there should be an override if some custom servers do support sleeping in other dimensions?

> Aight, I had to hop in Minecraft to test that once again, and yes, you can not sleep during rains in 1.16.4, however right-clicking on the bed sets the respawn point regardless of weather or day time (so it should be allowed even without the rain, but I'm not sure when was this added, definitely >1.12.2)

@Moondarker You're right, it was working for me because I was testing it when it was raining _at night_, oops (in the beta I had just been waiting for it to rain and didn't think to set the time). It doesn't work when it's only raining, even in beta 1.5.

qrvd:
Never mind I'm reverting that commit, I tried to make it recognize the type of weather on the client side, but it doesn't seem so simple (unless I'm missing something)

Moondarker:
When rain starts, server sends a bunch of reason 7 packets, gradually raising contained value (it is also called `gameMode` in node-minecraft-protocol from my testing, uhh, a bit weird), when rain comes to an end it goes in reverse, gradually going down.
Same for thunderstorm (with reason 8)
I think it would be easier and more informative to keep the last received float values in something `bot.rainState` and `bot.thunderState`, alongside `weatherUpdate` event instead of trying to cast every change in boolean, `bot.isRaining` should be kept for backwards compatibility sake

rom1504 (approved):

PR #1781, opened 2021-03-19
Add code snippets

u9g:


wvffle (commented):
Honestly I see no point in supporting snippets in the library. There are many  different snippet formats and many different editors besides vscode. In my opinion this should be an external plugin.

u9g:
> Honestly I see no point in supporting snippets in the library. There are many different snippet formats and many different editors besides vscode. In my opinion this should be an external plugin.

Why? VSCode is by far the most used IDE from contributors, it's just a nice productivity boost

DrMoraschi (changes requested):
On L8 it should be mineflayer-createbot* right?

u9g:
> On L8 it should be mineflayer-createbot* right?

fixed

wvffle:
> > Honestly I see no point in supporting snippets in the library. There are many different snippet formats and many different editors besides vscode. In my opinion this should be an external plugin.
> 
> Why? VSCode is by far the most used IDE from contributors, it's just a nice productivity boost

Are those snippets for contributors or for end users? Judging from the changes, for the latter.

Correct me if I am wrong but I just do not see the point of adding those type of files as a part of a library. How would it be distributed? With an npm module? I can see many in-production instances bloated by it. So adding it to the npmignore would be a must in my opinion. Then what? Shipping it with `git clone`? That doesn't make any sense as well. As I said before, those snippets you proposed are mostly for the end users who would have mineflayer installed as a dependency..
 
The best solution in my opinion (though I have no idea how it is done in VS Code) is to create a VS Code addon/plugin (if it even allows such things) or do it in external library that would add all of the snippets. 

Actually I'd also move the typescript typings to an external library as well.

rom1504:
There is no "bloating the npm package" issue. Going from 3.5kB to 3.8kB will cause no issue.

rom1504:
Status here ?

rom1504:
reopen if you want to work on it again

PR #1785, opened 2021-03-20
Fix farmer example

Karang:
* findBlock was used incorrectly
* added maxDistance check to avoid trying to break out of reach blocks
* node's once is leaking! minimal reproduceable example:

```js
const { once, EventEmitter } = require('events')

async function run() {
  const A = new EventEmitter()
  const B = new EventEmitter()

  A.on('removeListener', (name, listener) => {
    console.log('remove', name)
    B.off(name, listener)
  })

  A.on('newListener', (name, listener) => {
    console.log('add', name)
    B.on(name, listener)
  })

  for (let i=0 ; i < 3 ; i++) {
    process.nextTick(() => {
      B.emit('e', 42)
    });

    console.log('before once', A._events['e'])
    const [value] = await once(A, 'e')
    console.log('after once', A._events['e'])
  }
}

run()
```
This happen only if we forward the event to an other eventemitter

### DIFF
--- a/examples/farmer.js
+++ b/examples/farmer.js
@@ -1,4 +1,4 @@
-const Vec3 = require('vec3')
+const { Vec3 } = require('vec3')
 const mineflayer = require('mineflayer')
 
 if (process.argv.length < 4 || process.argv.length > 6) {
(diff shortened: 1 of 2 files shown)


rom1504 (approved):

rom1504:
Thanks

Do we use once incorrectly at other places?

rom1504:
https://github.com/nodejs/node/issues/37828 for reference

Karang:
> Thanks
> 
> Do we use once incorrectly at other places?

I did some git grep on mineflayer, it doesnt seems so, but in plugins maybe. This only concerns the world events that have the proxy on them.

Issue #1786, opened 2021-03-20
Add docs, examples, .github to .npmignore


wvffle:
## Is your feature request related to a problem? Please describe.
The npm package is bloated.

## Describe the solution you'd like
Add `docs`, `examples`, `.github` to `.npmignore`

Also any configuration dotfiles that are not needed in production.

rom1504:
I don't see any benefit to doing that. We would gain a few kB.
However it increase complexity and decrease a bit the usability of what's in node_modules

PR #1787, opened 2021-03-20
Add more exclusions to .npmignore

u9g:
closes https://github.com/PrismarineJS/mineflayer/issues/1786


rom1504:
I don't think we should do that, as commented in the issue

Issue #237, opened 2015-03-24
Support all 12 types of inventory window


thejoshwolfe:
http://wiki.vg/Inventory
- [x] Chest
- [x] Workbench
- [x] Furnace
- [x] Dispenser
- [x] Enchantment table
- [x] Npc trade
- [ ] Brewing Stand
- [ ] Beacon
- [ ] Anvil
- [ ] Hopper
- [ ] Dropper
- [ ] Horse

rom1504:
Related to #113

rom1504:
I think we need to refactor inventory.js (https://github.com/andrewrk/mineflayer/issues/223#issuecomment-86729991) before we try adding more to it.

rom1504:
refactoring is done, but I think we should solve https://github.com/andrewrk/mineflayer/issues/118 before adding more stuff.

rom1504:
Some information : 
* [Npc trade](http://minecraft.gamepedia.com/Trading)
* [Beacon](http://minecraft.gamepedia.com/Beacon#Powers)
* [Anvil](http://minecraft.gamepedia.com/Anvil#Repairing)
* [Hopper](http://minecraft.gamepedia.com/Hopper#Container)
* [Dropper](http://minecraft.gamepedia.com/Dropper)
* [Horse](http://minecraft.gamepedia.com/Horse#Equipment)

rom1504:
The windows code is currently in 3 separate places : in windows.js in name_of_window.js and in plugins/name_of_window.js. I think we can do better than that.
Are classes really needed for windows ?

rom1504:
These 3 places are at three different semantic level : 
1. the window definitions is windows.js is the more low-level one : it's used to create a window when receiving a open_window packet
2. the window classes in lib/ are used in openBlock(block,Class) which return an object of that class : these classes are event emitter which provide an event interface (open, close, updateSlot)
3. the window plugins define the higher level : they use openBlock to open a window and then define some additional function to make manipulating this window easier (deposit and withdraw for a chest for example)

Gjum:
some thoughts on 3.:
- chest, dispenser, hopper, dropper are essentially the same (a storage container), so they should share much code
- some inventories have special slots, so these should be adressable specifically (examples: horse saddle, crafting grid, -result, trade/repair ins and out)
- some inventories store items when the window is closed and reopened (chest, brewing), some don't (crafting, trade)
[...]

PR #1790, opened 2021-03-20
Advanced horse/donkey/llama support

Moondarker:
1. Makes horse window name follow the same pattern in <1.14 and >=1.14 (window.type = EntityHorse and ChatMessage-like title)
2. Makes activateEntity follow the same pattern in <1.16 and >=1.16 (sneaking state sent in packet depends on bot's sneaking state)
3. Extends horse-like entities with all neccessary attributes (I'm not sure if you want it to be implemented this way, should I maybe move it into a separate object in entity or maybe don't modify entity at all? Will update the doc once this is clear)
4. Adds `bot.openHorse(entity)` allowing to reliably open horse-like entity inventories without mounting them

### DIFF
--- a/lib/plugins/horse.js
+++ b/lib/plugins/horse.js
@@ -32,9 +32,9 @@ function inject (bot) {
   Object.assign(metadataMap.llama, metadataMap.base)
   metadataMap.mule = metadataMap.donkey
 
-  async function openHorse(horseEntity) {
+  async function openHorse (horseEntity) {
     if (!metadataMap[horseEntity.name]) return
-    const oldSneakState = bot.controlState['sneak']
+    const oldSneakState = bot.controlState.sneak
 
     bot.setControlState('sneak', true)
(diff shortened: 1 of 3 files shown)


Moondarker:
Also, it should make situation with #237 better, but if horses need any more work to be considered as "supported", please let me know :)

lib/plugins/horse.js
```diff
@@ -0,0 +1,80 @@
+const { callbackify } = require('../promise_utils')
+
+module.exports = inject
+
+function inject (bot) {
+  const metadataMap = {
+    base: {
```
rom1504, line 7: is this correct for all versions ?

Moondarker: Good call, looks like it is very likely not. Seems like another thing to be added in minecraft-data /._.

rom1504:
reopen if you want to work on it again

PR #1791, opened 2021-03-20
add example with chatregex plugin

u9g:


rom1504:
https://github.com/PrismarineJS/mineflayer/pull/1792#issuecomment-803471291

PR #1792, opened 2021-03-20
add some plugins to readme

u9g:


rom1504:
https://github.com/u9g/ChatRegex looks a lot like https://github.com/PrismarineJS/mineflayer/blob/master/examples/chatAddPattern.js

did you think about improving chat add pattern ?

u9g:
I submitted a pr for the chat one to add it directly to mineflayer

rom1504:
is this still relevant ? chat part is now in mineflayer
https://github.com/u9g/mineflayer-altlogin feels kind of overkill for such a small change, should it not be user code ?

PR #1794, opened 2021-03-22
Don't include npcs from player_info

U5B:
I guess you could probably find common NPC names and exclude them.


U5B:
fixed conflict, but this is pretty server specific

flowl:
It's nice you added a comment at all but please be more specific.
Do all npc's really have an "item.name" of `§` ?

U5B:
Yeah the placeholders are called that. You are right and I should probably make it a plugin

PR #1795, opened 2021-03-22
Slightly improve packet_info handling

Karang:
@U5B

### DIFF
--- a/lib/plugins/entities.js
+++ b/lib/plugins/entities.js
@@ -45,8 +45,7 @@ function inject (bot, { version }) {
       }
       return false
     }
-    const resultSet = Object.keys(bot.entities)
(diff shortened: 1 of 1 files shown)


lib/plugins/entities.js
```diff
@@ -446,6 +444,7 @@ function inject (bot, { version }) {
           player.displayName = new ChatMessage(JSON.parse(item.displayName))
         }
 
+        const playerEntity = Object.values(bot.entities).find(e => e.type === 'player' && e.username === item.name)
```
rom1504, line 447: Difference with previous code is this doesn't do .to lowercase()
But that's probably ok

rom1504 (approved):

PR #1796, opened 2021-03-24
Add a common errors md file

u9g:


DrMoraschi:
Should add the solution to the error that happens when you dont have node 14

rom1504:
What about using the faq file?

u9g:
> Should add the solution to the error that happens when you dont have node 14

done

u9g:
> What about using the faq file?

done

rom1504 (approved):

PR #1797, opened 2021-03-24
I'm getting an error while trying to use Mineflayer, this should help

javaveryhot:


index.js
```diff
@@ -60,12 +60,12 @@ module.exports = {
...
-  options.username = options.username ?? 'Player'
-  options.version = options.version ?? false
-  options.plugins = options.plugins ?? {}
-  options.hideErrors = options.hideErrors ?? true
-  options.logErrors = options.logErrors ?? true
-  options.loadInternalPlugins = options.loadInternalPlugins ?? true
+  options.username = options.username || 'Player'
+  options.version = options.version || false
+  options.plugins = options.plugins || {}
+  options.hideErrors = options.hideErrors || true
```
Moondarker, line 66: Please read [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator)
hideErrors, logErrors and loadInternalPlugins will be forever true

Moondarker (changes requested):

Moondarker:
Also, if you are getting an error, sharing it in such PR (or in issues first) would be a great idea

BlueBurgersTDD:
Having a error here is because your node is old, update it

Karang:
Update node, mineflayer support node 14+

PR #1823, opened 2021-03-30
await the block change in commandblock test

u9g:
as of now the current ci for mineflayer is failing so this pr should make the test not fail anymore


u9g:
ready for review

rom1504 (approved):

PR #1827, opened 2021-03-31
allow passing nmpclient as an object in createbot

u9g:
instead of always creating it in mf, allow for it to be passed


rom1504 (changes requested):
why ?

doc should be updated

u9g:
added, it is discussed here why https://discord.com/channels/413438066984747026/743199746876768318/826696291668721664

rom1504:
ok can you put a short explanation about that use case in the api.md line ? (maybe say it can be used for a proxy + mineflayer use case)

u9g:
added

rom1504 (approved):

Issue #1562, opened 2021-01-03
Add screenshot example using pviewer


rom1504:
Make a simple example for mineflayer using https://github.com/PrismarineJS/prismarine-viewer/blob/master/examples/exporter/screenshot.js

rom1504:
This is done by #1828 

See https://github.com/PrismarineJS/mineflayer/tree/master/examples/screenshot-with-node-canvas-webgl

PR #1828, opened 2021-03-31
Added pviewer screenshot.js example

IceTank:
A simple screenshot example that joins a server and takes 5 pictures.

Solves https://github.com/PrismarineJS/mineflayer/issues/1562#issue-777556792 

Works for 1.13 or later versions.
Should work for 1.10, 1.11 and 1.12 when ~https://github.com/PrismarineJS/minecraft-data/issues/386~ https://github.com/PrismarineJS/minecraft-data/pull/387#issue-603562205 is merged.

### DIFF
--- a/examples/viewer/package.json
+++ b/examples/viewer/package.json
@@ -3,7 +3,6 @@
   "version": "0.0.0",
   "private": true,
   "dependencies": {
-    "prismarine-viewer": "^1.0.3"
     "prismarine-viewer": "^1.0.3",
     "mineflayer": "file:../../"
   },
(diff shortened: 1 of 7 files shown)


Karang (approved):

rom1504 (changes requested):
that's nice but could you put it in a different folder ? node canvas webgl is not easy to install in all platforms so it's good to keep the main viewer example without it

IceTank:
Moved to new folder and added some installation info to README

examples/viewer/package.json
```diff
@@ -3,8 +3,9 @@
   "version": "0.0.0",
   "private": true,
   "dependencies": {
+    "prismarine-viewer": "^1.0.3"
     "prismarine-viewer": "^1.0.3",
```
rom1504, line 7: Duplicated

rom1504 (approved):

PR #1829, opened 2021-04-01
Add getExplosionDamages() function

Karang:
Sources:
https://minecraft.fandom.com/wiki/Explosion
https://minecraft.fandom.com/wiki/Armor#Damage_protection
https://minecraft.fandom.com/wiki/Attribute
https://github.com/extremeheat/extracted_minecraft_data/blob/client1.16.4/client/net/minecraft/world/level/Explosion.java#L198
https://github.com/extremeheat/extracted_minecraft_data/blob/client1.16.4/client/net/minecraft/world/damagesource/CombatRules.java
https://github.com/kami-blue/client/blob/master/src/main/kotlin/org/kamiblue/client/util/combat/CrystalUtils.kt#L87
https://github.com/kami-blue/client/blob/master/src/main/kotlin/org/kamiblue/client/util/combat/CombatUtils.kt#L47

### DIFF
--- a/examples/crystal.js
+++ b/examples/crystal.js
@@ -1,0 +1,32 @@
+const mineflayer = require('mineflayer')
+
+if (process.argv.length < 4 || process.argv.length > 6) {
+  console.log('Usage : node crystal.js <host> <port> [<name>] [<password>]')
+  process.exit(1)
+}
+
+const bot = mineflayer.createBot({
+  host: process.argv[2],
+  port: parseInt(process.argv[3]),
+  username: process.argv[4] ? process.argv[4] : 'crystal',
+  password: process.argv[5]
+})
+
+bot.on('chat', (username, message) => {
+  if (message === 'compute') {
+    const target = bot.players[username]?.entity
+    if (!target) {
+      bot.chat('I don\'t know where you are')
+      return
+    }
+    const crystal = bot.nearestEntity(entity => entity.name.includes('crystal'))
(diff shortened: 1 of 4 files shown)


rom1504:
that looks good but:
* is there any sub package where this could fit ?
* could we add some test?

LINKdiscordd:
Yes it would fit into mineflayer-autocrystal

u9g:
or prismarine-physics

LINKdiscordd:
or flying-squid

u9g:
no, this wouldnt fit in flying squid since we need it to be used by both mineflayer and flying squid

LINKdiscordd:
oh sorry my bad lol

LINKdiscordd:
also for the test karang can add one where the bot always will recieve a set amount of damage and if the calculation of the function doesn't match up with the damage that is expected it fails essentially

at least thats how I would do it

Karang:
prismarine-physics would make sense yes, but i think its better to move it once its feature complete. At least in mineflayer we have access to everything we need to develop it. To finish it properly we need to implement the whole attribute system and that's a big chunk (outside of the scope of this PR). I'll add the supportFeatures() and a small test then it'll be ready to be merged.

rom1504:
this at least needs doc before merging

KaffinPX (changes requested):
Adding events and functions to doc will useful

rom1504 (approved):

PR #1830, opened 2021-04-01
Implement a bot.placeEntity() function

u9g:


rom1504 (changes requested):
could you add a test?

u9g:
wip still

rom1504:
I like it but it's old and stale, reopen if you want to work on it

u9g:
awaiting https://github.com/PrismarineJS/prismarine-item/pull/33 to be merged, then needs to be replaced with having the function here

docs/api.md
```diff
@@ -1656,6 +1656,17 @@ This function also returns a `Promise`, with `void` as its argument upon complet
 
 The new block will be placed at `referenceBlock.position.plus(faceVector)`.
 
+#### bot.placeEntity(referenceBlock, faceVector, cb)
```
rom1504, line 1659: let's remove the cb (not adding it for new methods)

lib/plugins/place_entity.js
```diff
@@ -0,0 +1,175 @@
...
+    assert(['end_crystal', 'boat', 'spawn_egg', 'armor_stand'].includes(type), 'Unimplemented')
+
+    let name = bot.heldItem.name // used for finding entity after spawn
+      .replace(/.+_boat/, 'boat')
+
+    if (name.endsWith('spawn_egg')) {
+      name = bot.heldItem.spawnEggMobName
+    }
+    // Look at the center of the face
+    let dx = 0.5 + faceVector.x * 0.5
```
rom1504, line 24: is there any code from all this you could put in common with place block ? looks like yes

lib/plugins/place_entity.js
```diff
@@ -0,0 +1,175 @@
...
+
+      bot.on('entitySpawn', listener)
+    })
+  }
+
+  bot.placeEntity = callbackify(placeEntity)
+  bot._placeEntityWIthOptions = placeEntityWithOptions
+}
+
+function vectorToDirection (v) {
```
rom1504, line 160: don't we have this function elsewhere already ?

rom1504:
overall lgtm, let's just try to remove some code duplication

lib/plugins/place_entity.js
u9g: most of the place logic is the same, yeah

lib/plugins/place_entity.js
u9g: yeah, in placeBlock :p maybe should go in util?

rom1504: yeah something like that

lib/plugins/place_entity.js
rom1504: so try to do a function with the common stuff and call it here and in place block
will decrease the maintenance cost

docs/api.md
u9g: Done

lib/plugins/place_entity.js
u9g: Done

lib/plugins/place_entity.js
u9g: Done

rom1504 (approved):

PR #1844, opened 2021-04-09
Add a modular example of mineflayer

u9g:


u9g:
Tested on 1.16.5, works great :)

rom1504 (approved):

PR #1853, opened 2021-04-13
Reword inventory errors

u9g:
Old error: `Error: missing source item 621:undefined in (54,90)`
New error: `Error: Can't find bread in slots [54 - 90], (item id: 621)`

Also added a check to make sure a valid itemType was being passed

### DIFF
--- a/lib/plugins/inventory.js
+++ b/lib/plugins/inventory.js
@@ -213,7 +213,6 @@ function inject (bot, { version, hideErrors }) {
         (nbt != null && window.selectedItem.nbt !== nbt)) {
         // we are not holding the item we need. click it.
         const sourceItem = window.findItemRange(sourceStart, sourceEnd, itemType, metadata, false, nbt)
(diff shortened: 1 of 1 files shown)


lib/plugins/inventory.js
```diff
@@ -212,7 +213,8 @@ function inject (bot, { version, hideErrors }) {
         (nbt != null && window.selectedItem.nbt !== nbt)) {
         // we are not holding the item we need. click it.
         const sourceItem = window.findItemRange(sourceStart, sourceEnd, itemType, metadata, false, nbt)
-        if (!sourceItem) throw new Error(`missing source item ${itemType}:${metadata} in (${sourceStart},${sourceEnd})`)
+        assert(mcData.itemsArray.find(x => x.id === itemType), 'Invalid itemType')
```
rom1504, line 216: Put that in a variable to avoid duplicating

u9g: fixed

rom1504 (approved):

PR #1855, opened 2021-04-13
Respawn event now correctly parsed + added test

u9g:
Before now there was a bug where gamemode was being set to survival when the bot died in creative because the packet's `packet.gameMode` would be undefined (and added a test since this was tricky to fix), this fixes that + adds the newest supportfeature for dimension parsing as it was missing.

### DIFF
--- a/lib/plugins/game.js
+++ b/lib/plugins/game.js
@@ -63,9 +63,11 @@ function inject (bot) {
       bot.game.dimension = dimensionNames[packet.dimension]
     } else if (bot.supportFeature('dimensionIsAString')) {
       bot.game.dimension = packet.dimension
+    } else if (bot.supportFeature('dimensionIsAWorld')) {
+      bot.game.dimension = packet.worldName
     }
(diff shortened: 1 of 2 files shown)


U5B (approved):

extremeheat (approved):

lib/plugins/game.js
```diff
@@ -63,9 +63,11 @@ function inject (bot) {
       bot.game.dimension = dimensionNames[packet.dimension]
     } else if (bot.supportFeature('dimensionIsAString')) {
       bot.game.dimension = packet.dimension
+    } else if (bot.supportFeature('dimensionIsAWorld')) {
+      bot.game.dimension = packet.worldName
     }
     bot.game.difficulty = difficultyNames[packet.difficulty]
-    bot.game.gameMode = parseGameMode(packet.gameMode)
+    bot.game.gameMode = parseGameMode(packet.gameMode ?? packet.gamemode)
```
rom1504, line 70: Can we fix this in protocol.json instead ?

rom1504 (approved):

PR #1858, opened 2021-04-15
changed resource pack example

U5B:
added that hash is needed for 1.9 and below

added more examples


KaffinPX:
LGTM

u9g:
at this point should we just add it as a plugin?

KaffinPX:
I think its enough

rom1504 (changes requested):
I don't think we should have examples in mineflayer telling people to use packets directly

this should be added as mineflayer methods and events

rom1504:
now directly in mineflayer

PR #1864, opened 2021-04-19
Add autoreconnect

CyberPatrick:
The solution of task https://github.com/orgs/PrismarineJS/projects/11#card-44565850


DrMoraschi:
Would this bind events? For example if you have a bot.on "chat" event, would it work after reconnect?

CyberPatrick:
> Would this bind events? For example if you have a bot.on "chat" event, would it work after reconnect?

Hmm, no.

DrMoraschi:
So... What do we do?

aesthetic0001 (commented):
I don't see the benefit of this when you could just createBot again (especially since it doesn't bind events)

PR #1896, opened 2021-05-03
fix cannot read property id of undefined

TheAndrey777:
add undefined check

### DIFF
--- a/lib/plugins/inventory.js
+++ b/lib/plugins/inventory.js
@@ -344,9 +344,6 @@ function inject (bot, { version, hideErrors }) {
       onAccepted()
       click = windowClickQueue.shift()
        }
-    if (click !== undefined){
-        assert.ok(click)
-    }
 
     if (accepted) {
       onAccepted()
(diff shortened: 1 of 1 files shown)


lib/plugins/inventory.js
```diff
@@ -340,11 +340,13 @@ function inject (bot, { version, hideErrors }) {
       return
     }
     assert.ok(click.id <= actionId)
-    while (actionId > click.id) {
+    while (click !== undefined && actionId > click.id) {
       onAccepted()
       click = windowClickQueue.shift()
+       }
+    if (click !== undefined){
+        assert.ok(click)
```
extremeheat, line 348: What’s the purpose of this assert? It’s hidden behind an != undefined

TheAndrey777: yes i should remove it

extremeheat:
Why the close? No longer an issue?

extremeheat:
Ah, you opened #1901

PR #1898, opened 2021-05-04
Move problems to gh discussions

u9g:


extremeheat (approved):

KaffinPX (commented):
pls no

extremeheat:
This was discussed earlier, having it on Discussions means that the conversations are logged so people don't have to ask the same questions over and over again. If there needs to be extended discussions, the Discord is an option but I think this is the best way for everyone : search through issues and discussions, post a discussion if your problem isn't found, post an issue if you think you have a bug/suggestion, then go to discord if there needs to be bigger discussion. This way

* people can help themselves by looking through past issues/discussions with a search feature
* the conversations are organized, and we can reply to them later on. For example if a bug gets fixed or more info to add.
* we can add Github Actions or bots/AI/etc. to easily help guide people to duplicate issues, [example](https://github.com/microsoft/vscode/issues/122956#issuecomment-832213923)

rom1504 (approved):

Issue #1579, opened 2021-01-09
TypeError: Cannot read property 'id' of undefined


MoonVN571:
*\node_modules\mineflayer\lib\plugins\inventory.js:467
    while (actionId > click.id) {
                            ^

TypeError: Cannot read property 'id' of undefined
    at confirmTransaction (*\node_modules\mineflayer\lib\plugins\inventory.js:467:29)
    at Client.<anonymous> (*\node_modules\mineflayer\lib\plugins\inventory.js:568:5)
    at Client.emit (events.js:315:20)
    at FullPacketParser.<anonymous> (*\node_modules\minecraft-protocol\src\client.js:89:12)
    at FullPacketParser.emit (events.js:315:20)
    at addChunk (*\node_modules\readable-stream\lib\_stream_readable.js:298:12)
    at readableAddChunk (*\node_modules\readable-stream\lib\_stream_readable.js:280:11)
    at FullPacketParser.Readable.push (*\node_modules\readable-stream\lib\_stream_readable.js:241:10)
    at FullPacketParser.Transform.push (*\node_modules\readable-stream\lib\_stream_transform.js:139:32)
    at FullPacketParser._transform (*\node_modules\protodef\src\serializer.js:89:10)

rom1504:
https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/inventory.js#L460 that is not possible with that last code.
Is your mineflayer up to date ?

MoonVN571:
> https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/inventory.js#L460 that is not possible with that last code.
> Is your mineflayer up to date ?

Im use node install mineflayer for it, thanks you

Karang:
The test line 460 have been there for a while + the code at line 467 is matching the latest version. However as rom stated, this error shouldn't be possible. Can you go in your `\node_modules\mineflayer\lib\plugins\inventory.js` and copy here the code of the function confirmTransaction ? https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/inventory.js#L456

Karang:
@rom1504 ah I got what might be happening:
```js
while (actionId > click.id) {
      onAccepted()
      click = windowClickQueue.shift()
    }
```
click get redefined in the loop and not tested for undefined again.

MoonVN571:
> @rom1504 ah I got what might be happening:
> 
> ```js
> while (actionId > click.id) {
[...]

PR #1901, opened 2021-05-05
fix cannot read property id of undefined

TheAndrey777:
add undefined check


rom1504 (changes requested):
can you explain why this is not a workaround ?
in what case does this make sense ?

TheAndrey777:
for me when bot open menu this error appears. There is an issue about it #1579

PR #1905, opened 2021-05-13
Change scope of createActionNumber

Camezza:
Allows for low level client window packet transactions to be made by specifying a valid actionId

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -265,6 +265,7 @@
       - [bot.clickWindow(slot, mouseButton, mode, cb)](#botclickwindowslot-mousebutton-mode-cb)
       - [bot.putSelectedItemRange(start, end, window, slot, cb)](#botputselecteditemrangestart-end-window-slot-cb)
       - [bot.putAway(slot, cb)](#botputawayslot-cb)
+      - [bot.createActionNumber()](#botcreateactionnumber)
       - [bot.closeWindow(window)](#botclosewindowwindow)
       - [bot.transfer(options, cb)](#bottransferoptions-cb)
       - [bot.openBlock(block)](#botopenblockblock)
@@ -1805,6 +1806,10 @@ This function also returns a `Promise`, with `void` as its argument upon complet
 Can be useful in case the client is supposed to simulate without feedback from the server.
 
 Put the item at `slot` in the inventory.
+
+#### bot.createActionNumber()
+
+Create a valid actionId for lower level window operations
 
 #### bot.closeWindow(window)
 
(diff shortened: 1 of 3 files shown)


u9g:
Can you give an example of why something like this would need to be used?

Camezza:
For example, clicking a window slot:

```
                this.bot._client.write('window_click', {
                    windowId: window.id,
                    slot: slot,
                    mouseButton: mouseButton
                    action: this.bot.createActionNumber(),
                    mode: 0,
                    item: { blockId: -1 },
                });
```

u9g:
I suggest you take a look at [this](https://github.com/PrismarineJS/mineflayer/blob/master/docs/api.md#botclickwindowslot-mousebutton-mode-cb) because `bot.clickWindow(slot, mouseButton, 0)` does the same thing while having the benefit of abstraction from the raw packet.

Camezza:
Unfortunately, this solution didn't work for me due to inability to specify windows passed from the windowOpen event

u9g:
> Unfortunately, this solution didn't work for me due to inability to specify windows passed from the windowOpen event

this is a bug with bot.clickWindow because the opened window’s id should be entered itself

Camezza:
Is this able to be approved? Changes are complete, now it's a case of them being accepted or rejected

Karang:
> Is this able to be approved? Changes are complete, now it's a case of them being accepted or rejected

This PR only changes the doc, it doesn't modify the scope of the createActionNumber function. You need to modify https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/inventory.js to do so.

Karang:
> > Unfortunately, this solution didn't work for me due to inability to specify windows passed from the windowOpen event
> 
> this is a bug with bot.clickWindow because the opened window’s id should be entered itself

To rephrase a bit what U9G said: you don't need to give clickWindow the window from the windowOpen event because it keeps track internally of what window is open. There can always be only one window open at a time, so you shouldn't ever care about specifying the id yourself.

Camezza:
> > Is this able to be approved? Changes are complete, now it's a case of them being accepted or rejected
> 
> This PR only changes the doc, it doesn't modify the scope of the createActionNumber function. You need to modify https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/inventory.js to do so.

Ah sorry, my mistake. It has been updated.

> > > Unfortunately, this solution didn't work for me due to inability to specify windows passed from the windowOpen event
> > 
> > 
> > this is a bug with bot.clickWindow because the opened window’s id should be entered itself
> 
> To rephrase a bit what U9G said: you don't need to give clickWindow the window from the windowOpen event because it keeps track internally of what window is open. There can always be only one window open at a time, so you shouldn't ever care about specifying the id yourself.

I'll create another PR regarding this 👍

rom1504 (approved):

rom1504:
as explained above, this isn't needed

PR #1908, opened 2021-05-15
Fix crash

jkcoxson:
This adds a check to make sure the block actually exists, because this is crashing my bots.


KaffinPX (approved):
Looks legit

rom1504 (changes requested):
this is the wrong fix

in what case does this happen ? probably the digging should be canceled

rom1504:
please reopen if you want to finish it

PR #1919, opened 2021-05-21
Fix bot sometimes not being able to dig when raycasting

mat-1:


rom1504 (changes requested):
can you explain why that change make sense ? why 0 rather than 0.5 ?

mat-1:
> 
> 
> can you explain why that change make sense ? why 0 rather than 0.5 ?

Sometimes when the bot is too close to the block it can't dig it, although maybe I should've done some testing to find out exactly why it doesn't work.

rom1504:
if you have more information /tests ; please re-open

PR #1920, opened 2021-05-22
Add window parameter to clickWindow function

Camezza:
closeWindow has it, so why doesn't this?

By default, clickWindow would use the currently open window or player inventory for packet transactions. By specifying a parameter, windows can be passed down directly from the windowOpen event.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -262,7 +262,7 @@
       - [bot.supportFeature(name)](#botsupportfeaturename)
       - [bot.waitForTicks(ticks)](#botwaitforticksticks)
     - [Lower level inventory methods](#lower-level-inventory-methods)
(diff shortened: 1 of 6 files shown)


rom1504 (changes requested):
there can be only one window open
do you want to change that ? there are many more things to do to change this

Camezza:
No, I only wanted to implement a quick fix for specifying a window to click rather than defaulting to inventory or an unspecific one

u9g:
> No, I only wanted to implement a quick fix for specifying a window to click rather than defaulting to inventory or an unspecific one

theres no specifying a window. There is always an active window, there can’t be any other windows used at that point. Theres no reason then to worry about other windows if we store the opened window’s id internally.

Camezza:
Closed as it has been deemed unnecessary

PR #1927, opened 2021-05-23
Add pathfinder example to README.md

IceTank:


rom1504 (changes requested):
why do we want this in the readme ?

Hazuki-san:
mineflayer-pathfinder is not related to this.

PR #1850, opened 2021-04-12
Prevents exceeding Max Listeners when bot dies simultaneously digging

MoneyMakingTornado:
Prevents MaxListenersExceededWarning _by not exceeding max listeners_ such as from when the bot dies while simultaneously digging  
```js
(node:6948) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 diggingAborted listeners added to [EventEmitter]. Use emitter.setMaxListeners() to increase limit
    at _addListener (node:events:451:17)
    at EventEmitter.addListener (node:events:467:10)
    at resetPath ({ProjectPath}\node_modules\mineflayer-pathfinder\index.js:90:11)
    at EventEmitter.<anonymous> ({ProjectPath}\node_modules\mineflayer-pathfinder\index.js:294:5)
    at EventEmitter.emit (node:events:376:20)
    at addColumn ({ProjectPath}\node_modules\mineflayer\lib\plugins\blocks.js:111:9)
    at Client.<anonymous> ({ProjectPath}\node_modules\mineflayer\lib\plugins\blocks.js:327:5)
    at Client.emit (node:events:376:20)
    at FullPacketParser.<anonymous> ({ProjectPath}\node_modules\minecraft-protocol\src\client.js:91:12)
    at FullPacketParser.emit (node:events:376:20)
(node:6948) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 diggingCompleted listeners added to [EventEmitter]. Use emitter.setMaxListeners() to increase limit
    at _addListener (node:events:451:17)
    at EventEmitter.addListener (node:events:467:10)
    at resetPath ({ProjectPath}\node_modules\mineflayer-pathfinder\index.js:91:11)
    at EventEmitter.<anonymous> ({ProjectPath}\node_modules\mineflayer-pathfinder\index.js:294:5)
    at EventEmitter.emit (node:events:376:20)
    at addColumn ({ProjectPath}\node_modules\mineflayer\lib\plugins\blocks.js:111:9)
    at Client.<anonymous> ({ProjectPath}\node_modules\mineflayer\lib\plugins\blocks.js:327:5)
    at Client.emit (node:events:376:20)
    at FullPacketParser.<anonymous> ({ProjectPath}\node_modules\minecraft-protocol\src\client.js:91:12)
    at FullPacketParser.emit (node:events:376:20)
```
[...]

PR #1936, opened 2021-05-26
Adds initial resource pack support

KaffinPX:
adds resource pack support to mineflayer


u9g (changes requested):
1. Lint
2. `resource_pack` should be changed to `resourcepack` or `resourcePack` to match existing event naming standards
3. Has this been tested or maybe we should have tests for this?

lib/features.json
```diff
@@ -246,9 +246,14 @@
   },
   {
     "name": "selectingTradeMovesItems",
-    "description": "Selecting a trade automatically puts the required items into trading slots.",
+    "description": "selecting a trade automatically puts the required items into trading slots",
     "versions": ["1.14", "1.16.5"]
   },
+  {
+    "name": "resourcePackUsesHash",
```
extremeheat, line 253: do you use this feature anywhere since you added it?

lib/plugins/resource_pack.js
```diff
@@ -0,0 +1,43 @@
...
+      bot._client.write('resource_pack_receive', {
+        result: TEXTURE_PACK_RESULTS.ACCEPTED
+      })
+      bot._client.write('resource_pack_receive', {
+        result: TEXTURE_PACK_RESULTS.SUCCESSFULLY_LOADED
+      })
+    }
+  }
+
+  function denyResourcePack () {
```
extremeheat, line 38: same as above

lib/features.json
u9g: L19 of reaource_pack.js in this branch

KaffinPX:
@extremeheat its ready: https://github.com/PrismarineJS/mineflayer/pull/1850

docs/api.md
```diff
@@ -1016,6 +1019,10 @@ event instead.
 
 Emitted when the server changes any of the game properties.
 
+#### "resourcePack"
```
DrMoraschi, line 1022: Should indicate that it also emits data from the resource pack

DrMoraschi (changes requested):

docs/api.md
KaffinPX: right

docs/api.md
```diff
@@ -1016,6 +1019,10 @@ event instead.
 
 Emitted when the server changes any of the game properties.
 
+#### "resourcePack" (url, hash)
```
DrMoraschi, line 1022: No, it emits an object that contains the url and hash, it only has one parameter

DrMoraschi (changes requested):

docs/api.md
extremeheat: I liked the idea of an object here like before, its more future proof and can be expanded later on to add fields like pack size. What you can do is update the doc to show that it’s an object instead of parameters.

rom1504 (approved):

PR #1942, opened 2021-05-28
Add timeout to #consume()

SeanMcCord:
There are currently many ways the bot could be interrupted during consume. Examples include activating another item, being killed, etc. This change is meant as a robust fallback in case the server does not acknowledge the consumption.


KaffinPX (approved):
legit lets see it passes tests?

rom1504 (approved):

PR #1948, opened 2021-05-31
Optimize findblock distance filter

a1cd:
by not using `distanceTo()`, we dont use `Math.sqrt()` which is costly in processing. Instead, we can square the `maxDistance` once before anything else and use distance squared for the distance from the beginning search point. This works because we are only comparing distances, not performing arithmetic or using it for anything significant.

This also works when sorting the blocks by distance because we only need a positive value, negative value, or `0`. Getting the difference between their exact euclidian distances is overkill.

### DIFF
--- a/lib/plugins/blocks.js
+++ b/lib/plugins/blocks.js
@@ -176,11 +176,11 @@ function inject (bot, { version, storageBuilder }) {
     // the octahedron iterator can sometime go through the same section again
     // we use a set to keep track of visited sections
     const visitedSections = new Set()
+    const maxDistanceSquared = Math.pow(maxDistance, 2)
 
     let blocks = []
     let startedLayer = 0
     let next = start
-    let maxDistanceSquared = Math.pow(maxDistance, 2)
     while (next) {
       const column = bot.world.getColumn(next.x, next.z)
       if (next.y >= 0 && next.y < 16 && column && !visitedSections.has(next.toString())) {
--- a/lib/plugins/blocks.js
+++ b/lib/plugins/blocks.js
@@ -180,6 +180,7 @@ function inject (bot, { version, storageBuilder }) {
     let blocks = []
     let startedLayer = 0
     let next = start
+    let maxDistanceSquared = Math.pow(maxDistance, 2)
     while (next) {
       const column = bot.world.getColumn(next.x, next.z)
       if (next.y >= 0 && next.y < 16 && column && !visitedSections.has(next.toString())) {
@@ -189,7 +190,7 @@ function inject (bot, { version, storageBuilder }) {
           for (cursor.x = next.x * 16; cursor.x < next.x * 16 + 16; cursor.x++) {
             for (cursor.y = next.y * 16; cursor.y < next.y * 16 + 16; cursor.y++) {
               for (cursor.z = next.z * 16; cursor.z < next.z * 16 + 16; cursor.z++) {
-                if (fullMatcher(cursor) && cursor.distanceTo(point) <= maxDistance) blocks.push(cursor.clone())
+                if (fullMatcher(cursor) && cursor.distanceSquared(point) <= maxDistanceSquared) blocks.push(cursor.clone())
               }
             }
           }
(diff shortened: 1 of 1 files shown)


rom1504:
What performance gain did you observe with this change ?

a1cd:
I have not observed a performance increase but i have had dramatic speed increases in similar code. It was trying to compare the distance between thousands of points as fast as possible.

It also makes intuitive sense. For example, is it easier to multiply two numbers, square root it and compare the result to another number that you also had to square root? OR. Just compare two numbers that were multiplied together with another set of numbers that were multiplied? (it's easier not to square root) This is also true with computers. it takes so much more time to divide or root a number than to multiply or add it.

a1cd:
i can test to see if it performs better but this may take a while. (mineflayer is acting weird on my computer)

rom1504:
I'm curious to see the results of a benchmark.
I expect reading the blocks is the bottleneck rather than distance computation. Sqrt is highly optimised

a1cd:
wow, all failed, that's sad. Easy fix though. 

I do not believe it will be such a big improvement. like you said, getting the blocks from memory is likely a bottleneck but i have found that it is much faster to not use sqrt.

rom1504:
Ok anyway it should break anything to make that change, it's pretty simple.
Can you fix the lint ? (Run `npm run fix`)

Karang:
> I'm curious to see the results of a benchmark.
> I expect reading the blocks is the bottleneck rather than distance computation. Sqrt is highly optimised

I would be surprised if this make a difference at all, reading the blocks is indeed the bottleneck. That being said, the code changes proposed here are not harmful, so there is no reason to not accept it.

(But if you are willing to go to this level of micro-optimisations, don't use Math.pow() for squaring...)

To complete a bit the story on sqrt, and back up rom's claims: all modern cpus hardware includes some form of sqrt instructions that usually execute the operation in a few cycles (throughput of 3 cycles on modern x86 https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=sqrt_ss&expand=5389). So unless your language or library isn't taking advantage of those or if you do nothing else than sqrt, its unlikely to have a significant impact on your performances.

a1cd:
<meta name="ProgId" content="Excel.Sheet">
<meta name="Generator" content="Microsoft Excel 15">
<link id="Main-File" rel="Main-File" href="file:////Users/24wilber/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip.htm">
<link rel="File-List" href="file:////Users/24wilber/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip_filelist.xml">
<style>

</style>





&nbsp; | t1 time | t1 total | t2 time | t2 total
-- | -- | -- | -- | --
0 | 32.48 | 8158 | 31.48 | 8209
1 | 27.08 | 8257 | 18.31 | 8159
2 | 21.84 | 8159 | 17.56 | 8158
3 | 26.65 | 8107 | 15.26 | 8056
4 | 19.97 | 8106 | 22.91 | 8158
5 | 20.27 | 8058 | 21.46 | 8158
6 | 15.53 | 7740 | 27.31 | 7930
7 | 20.34 | 7948 | 21.38 | 8022
8 | 25.24 | 7935 | 22.14 | 8019
&nbsp; | 23.2666667 | 8052 | 21.9788889 | 8096.55556
&nbsp; | &nbsp; | &nbsp; | &nbsp; | &nbsp;
&nbsp; | &nbsp; | &nbsp; | difference | &nbsp;
&nbsp; | &nbsp; | &nbsp; | -1.2877778 | &nbsp;

it is 1.2ms faster on average.

a1cd:
yes this is fine, its possible that there is no difference, the tests are +/- 1.8 ms inaccurate but my math may be off. This means it could have just been slightly faster or up to 3ms faster which is still not a huge gain.

rom1504:
What code did you run to get these results?

a1cd:
i used a previous test that was committed and the test on this commit request

a1cd:
i just copied and pasted them into excel

PR #1953, opened 2021-06-01
👌 smaller format

MJEND7:
I simply changed all your old functions to arrow functions. smaller 👌

### DIFF
--- a/lib/conversions.js
+++ b/lib/conversions.js
@@ -1,4 +1,3 @@
-'use strict';
 
 const Vec3 = require('vec3').Vec3
 const math = require('./math')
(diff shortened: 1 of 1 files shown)


lib/conversions.js
```diff
@@ -1,3 +1,5 @@
+'use strict';
```
u9g, line 1: not needed, we lint with standardjs instead

Karang:
Tests are not passing, lint with `npm run fix`

lib/conversions.js
KaffinPX: yeah

MJEND7: ok

MJEND7:
O I see my problem

PR #1955, opened 2021-06-01
Bye blockInSight :( 

KaffinPX:
Bye bye our helper. RIP


rom1504:
Why

KaffinPX:
It was deprecated, now i removed it

rom1504:
Did you check that nothing in this repo, in plugins and in user repos is using it ?

Hazuki-san:
I think this function should be left inside the code, even though it is outdated... but some people might be using it.

PR #1961, opened 2021-06-02
Change workflow to seperate mocha tests from lint

u9g:
1. Renamed steps to what they are (ie. mocha, standard from just build)
2. Separated minecraft tests from linting test

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,6 @@
   "types": "index.d.ts",
   "scripts": {
     "test": "mocha --reporter spec --exit",
-    "pretest": "npm run lint",
     "lint": "standard",
     "fix": "standard --fix",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -6,7 +6,6 @@
   "types": "index.d.ts",
   "scripts": {
     "test": "mocha --reporter spec --exit",
-    "pretest": "npm run lint",
```
rom1504, line 9: problem of that is then users running `npm test` won't get the lint errors
could you instead keep that and have a "mocha_test" script that would get call from the ci ?

extremeheat:
Unrelated Tip: you can also run tests by doing “npx mocha” in any repo with jest/mocha and it’ll find the local bin script and run it.

rom1504 (approved):

PR #1965, opened 2021-06-02
add offhand support to placeBlockWithOptions

aesthetic0001:
Allows for ppl to use {offhand: true} as an option to place the block in the offhand slot when calling "placeBlockWithOptions"


lib/plugins/place_block.js
```diff
@@ -7,7 +7,15 @@ function inject (bot, { version }) {
   const Item = require('prismarine-item')(version)
 
   async function placeBlockWithOptions (referenceBlock, faceVector, options) {
-    if (!bot.heldItem) throw new Error('must be holding an item to place a block')
+    // 0 is placing with mainhand whilst 1 is off-hand
```
u9g, line 10: no need for the comments

u9g (changes requested):
would be good to add to typings/api

lib/plugins/place_block.js
Zn10plays: Do we usually handle errors like this? In the contributors' guide it's says to return the error insted of throwing it.

aesthetic0001:
> would be good to add to typings/api

I don't think there's docs for placeBlockWithOptions so should I add that along with the new option?

lib/plugins/place_block.js
aesthetic0001: @Zn10plays it does, but I just copied what was written to handle a scenario where bot.helditem is null, so should I change it?

KaffinPX (approved):
LGTM but adding types will helpful.

lib/plugins/place_block.js
Zn10plays: Up to you on this one because, .catch will work with either

aesthetic0001:
Is there anything I should change for this to be approved?

DrMoraschi:
Why no typings?

aesthetic0001:
> Why no typings?

Check the contribute channel

u9g:
I just started tests.

aesthetic0001:
yay tests passed

aesthetic0001:
why does it say "Merging can be performed automatically once the requested changes are addressed" even tho I made the changes a while back

rom1504:
can you please run git pull and push again ?

aesthetic0001:
yup should be done

DrMoraschi:
Why the .idea folder?

rom1504 (changes requested):
please remove .idea

aesthetic0001:
bruh i forgot to rm it mb

rom1504 (changes requested):
can you add the option in the doc ?

aesthetic0001:
which file should it be added to bcs its a _ method and there aren't docs for it yet afaik

rom1504:
true
I really don't like these non documented methods but not your fault

u9g:
This can be added to the _ functions docs, its in docs/

aesthetic0001:
> This can be added to the _ functions docs, its in docs/

https://github.com/PrismarineJS/mineflayer/blob/master/docs/unstable_api.md?

rom1504:
let's do it in another pr

PR #1968, opened 2021-06-04
run tests

u9g:


Hazuki-san:
update is real

u9g:
> update is real

What?

Hazuki-san:
nothing, just too many readme update lol

Issue #1714, opened 2021-02-24
No colours in `bot.tablist`


GoogleSites:
- [ ] The [FAQ](https://github.com/PrismarineJS/mineflayer/blob/master/docs/FAQ.md) doesn't contain a resolution to my issue 

## Versions
 - mineflayer: 3.2.0
 - server: Vanilla 1.8.9
 - node: 14.4.0

## Detailed description of a problem
A clear and concise description of what the problem is, with as much context as possible.
What are you building? What problem are you trying to solve?

I'm trying to get the list of players in the tab (with coloured names) that are located in Hypixel BedWars in order to get a list of all teams (each team has a different colour for their names).

## What did you try yet?

I've tried `bot.tablist` (it's empty), and `player.displayName`, but neither show the colours that I want.

Did you try any method from the API?
None of them do what I'm looking for.

Did you try any example? Any error from those?

## Your current code
```js
import mineflayer from 'mineflayer';

const bot = mineflayer.createBot({
	username: 'Bot',
	version: '1.8.9',
	host: 'localhost',
	port: 57573
});

bot.once('login', () => {
	setInterval(console.log, 5000, bot.tablist);
});
```

## Expected behavior

`bot.tablist` should include a list of players in the tablist, along with the correct colours (anywhere at all works, it doesn't need to be nicely formatted).

## Additional context

I have a yellow name that I set using scoreboard commands:
`/scoreboard teams add MyTeam Yellow`
`/scoreboard teams join MyTeam`

goncharovchik:
> - [ ] The [FAQ](https://github.com/PrismarineJS/mineflayer/blob/master/docs/FAQ.md) doesn't contain a resolution to my issue 
> 
> ## Versions
>  - mineflayer: 3.2.0
>  - server: Vanilla 1.8.9
>  - node: 14.4.0
> 
> ## Detailed description of a problem
> A clear and concise description of what the problem is, with as much context as possible.
> What are you building? What problem are you trying to solve?
> 
> I'm trying to get the list of players in the tab (with coloured names) that are located in Hypixel BedWars in order to get a list of all teams (each team has a different colour for their names).
> 
> ## What did you try yet?
> 
> I've tried `bot.tablist` (it's empty), and `player.displayName`, but neither show the colours that I want.
> 
> Did you try any method from the API?
> None of them do what I'm looking for.
> 
> Did you try any example? Any error from those?
> 
> ## Your current code
> ```js
[...]

Issue #1377, opened 2020-10-16
Scoreboard-data cant be read correctly


Schrubbii:
#866  Versions
 - mineflayer: 2.31.0
 - server: spigot 1.8.9
 - node: most recent

Im trying to read my scoreboard-data once im on the server. Displayed in this screenshot:
https://cdn.discordapp.com/attachments/413438150594265099/766467191645667328/Screenshot_10.png
Especially the value under Kontostand (here 0$).

## What did you try yet?
Tried to access that data through all listeners. I get them, but NOT with any value. Screenshot here:
https://image.prntscr.com/image/q1Iu_hhVT06P7DrbRB_0FQ.png

Code for that:
``
    bot.on('scoreUpdated', () => {  
        console.log('--------------------');  
        for (const scoreboardName in bot.scoreboard) {  
            const scoreboard = bot.scoreboard[scoreboardName];  
            if (typeof scoreboard !== 'undefined') {  
                for (const user in scoreboard.itemsMap) {  
                    const scoreboardEntry = scoreboard.itemsMap[user]  
                    console.log(scoreboardEntry.name, scoreboardEntry.value)  
                }  
            }  
        }  
        console.log('--------------------');  
    })  
``
(Sorry, but cant get the code displayed right. Heres it:
https://prnt.sc/v0n083

Tried in in every listener of scoreboard. 'updated' doesnt even fire once, only 'created'. Data doesnt get parsed after some delay aswell, tried a timeout of 20seconds before displaying the data.

Any idea if its a problem on my side, a server problem in parsing the data, or something else?

TheDudeFromCI:
Try listing the display names of all players in the `bot.players` map. It's possible the server is simply creating a fake player and just changing the display name of that player to update the scoreboard value.

Schrubbii:
Displayed all users and counted them. Exactly the same amount of players on the server as they are in the list.
Found both of my accounts aswell.
```javascript
-----------------------------
{
  username: 'Schrubber',
  ping: 16,
  uuid: '1d65baa4-c8e0-48be-aa83-7ed4c1aadafe',
  displayName: ChatMessage {
    json: { extra: [Array], text: '' },
    text: '',
    extra: [ [ChatMessage], [ChatMessage], [ChatMessage] ],
    bold: undefined,
    italic: undefined,
    underlined: undefined,
    strikethrough: undefined,
    obfuscated: undefined,
    color: undefined
  },
  entity: null
}
{
[...]

PR #1984, opened 2021-06-11
Add support for teams

builder-247:
In this PR I'll be adding support for teams.

### TODO
- [x] finish Team class implementation
- [x] pre-1.13 support
- [x] add a display method to scoreboards
- [x] tests
- [x] documentation

Related: https://github.com/PrismarineJS/mineflayer/issues/1377, https://github.com/PrismarineJS/mineflayer/issues/1714, https://github.com/PrismarineJS/mineflayer/pull/864


lib/team.js
```diff
@@ -0,0 +1,43 @@
...
+
+  remove (name) {
+    const removed = this.membersMap[name]
+    delete this.membersMap[name]
+    return removed
+  }
+
+  // Return a chat component with prefix + color + name + suffix
+  displayName (member) {
+    // TODO
```
builder-247, line 39: Is there an existing method of joining multiple `ChatMessage`s?

lib/team.js
```diff
@@ -0,0 +1,93 @@
...
+    return this.membersMap[name]
+  }
+
+  remove (name) {
+    const removed = this.membersMap[name]
+    delete this.membersMap[name]
+    return removed
+  }
+
+  update (packet) {
```
rom1504, line 73: could you do the packet handling in the team plugin ?
we try to keep such classes independent from the packets so they stay fairly stable
(that means taking as input directly {name, friendlyFire, ...} here and explicitely passing each of those fields in the team plugin
that will make it easier to adapt when mojang (inevitably) changes the fields

lib/team.js
```diff
@@ -0,0 +1,93 @@
...
+
+module.exports = loader
+
+function loader (mcVersion) {
+  ChatMessage = require('prismarine-chat')(mcVersion)
+  MessageBuilder = ChatMessage.MessageBuilder
+  return Team
+}
+
+function colorString (color) {
```
rom1504, line 12: could you check whether prismarine-chat could be used to do that ? I feel we already have that info there

rom1504 (changes requested):
Looks good overall!
Thanks for working on this.

I put 2 comments of things to improve.

lib/team.js
builder-247: while we have the data, it's not accessible currently

rom1504 (approved):

Issue #1981, opened 2021-06-10
reimplement removeChatPattern to remove race condition


u9g:
it's now possible to delete the regex before the chat parser itself looks whether to reset position and you can throw an undefined error, instead, maybe simply adding a disabled option would be better that way we don't have this race condition possible

PR #1986, opened 2021-06-13
Refactor chat.js (explained)

u9g:
What does this pr do:

1. internally the chat.js plugin no longer uses an array for storing chatpattern(set)s, they are stored in an object with a length variable that is incremented with each addition. the actual object is now called `_patterns` and the length is called `_length`.
2. internally, when the chat.js plugin gets a new message, it only processes the chatpattern(set) that the message in chat matches instead of processing every single stored pattern.
3. when registering a chat pattern with addChatPattern()/addChatPatternSet()/chatAddPattern() a number is now returned, this number is the index of that pattern in the internal _patterns object
4. removeChatPattern() now works based on what's passed to it for the name parameter: if name is a string, all patterns that have that name will be removed, else if name is a number, only that exact pattern will be removed

closes https://github.com/PrismarineJS/mineflayer/issues/1981

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1435,6 +1435,8 @@ the event will be called `"chat:name"`, with name being the name passed
   * `parse` - instead of returning the actual message that was matched, return the capture groups from the regex
   * `deprecated` - (**unstable**) used by bot.chatAddPattern to keep compatability, likely to be removed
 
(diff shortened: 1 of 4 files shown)


KaffinPX (approved):
LGTM, can u approve it @rom1504

PR #1993, opened 2021-06-16
Add support for 1.17

Archengius:


u9g:
needs lint

rom1504:
blocked by https://github.com/PrismarineJS/node-minecraft-protocol/pull/866

lib/plugins/entities.js
```diff
@@ -271,7 +270,7 @@ function inject (bot, { version }) {
 
   // TEMP: Misnamed packet in 1.17 protocol, entityIds is a single varint
```
rom1504, line 271: in 1.17 it's now a single varint named entityId

lib/plugins/game.js
```diff
@@ -1,78 +1,108 @@
...
+  } else if (bot.supportFeature('dimensionIsAString')) {
+    return packet.dimension
+  } else if (bot.supportFeature('dimensionIsAWorld')) {
+    return packet.worldName
+  }
+  throw new Error('Unsupported dimension type in login packet')
+}
+
+function handleRespawnPacketData (bot, packet) {
+  bot.game.levelType = determineLegacyLevelType(packet)
```
extremeheat, line 59: can you inline these if they only get called once? It makes the code easier to read IMO

lib/plugins/blocks.js
```diff
@@ -81,6 +81,32 @@ function inject (bot, { version, storageBuilder }) {
...
+      chunkInitializationData.minWorldHeight = dimensionData.min_y
+      chunkInitializationData.maxWorldHeight = dimensionData.min_y + dimensionData.height
+    }
+
+    return new Chunk(chunkInitializationData)
+  }
+
+  function flattenMaskArray (sectionBitMask) {
+    // We need to handle arrays specially because protocol will return
+    // array of BigInt on new versions, while we really need an array of 32-bit integers,
```
extremeheat, line 99: where does the protocol return BigInts?

Archengius: In the primaryBitMask field of packet_map_chunk of 1.17 protocol. All chunk related mask fields are now serialized as long arrays representing bit sets in fact, including lighting packet ones. That's needed to support extended height.

lib/plugins/game.js
Archengius: I could, but I literally moved them out into separate functions because not having version dependent logic in the common code actually makes code easier to understand.

lib/plugins/blocks.js
rom1504: I think the format of chunk packet in protocol 1.17.json is wrong, we should change it so things like this are not required.

lib/plugins/blocks.js
```diff
@@ -275,18 +301,28 @@ function inject (bot, { version, storageBuilder }) {
...
     }
 
-    column.loadLight(packet.data, packet.skyLightMask, packet.blockLightMask, packet.emptySkyLightMask, packet.emptyBlockLightMask)
+    // Before 1.13, light packet contained combined skylight and blocklight data
+    // Now they are stored separately in their exclusive fields, and their type are arrays
+    if (packet.data) {
+      column.loadLight(packet.data, packet.skyLightMask, packet.blockLightMask, packet.emptySkyLightMask, packet.emptyBlockLightMask)
+    } else {
+      // Use new method for handling 1.13+ lighting data
+      const flatSkyLightMask = flattenMaskArray(packet.skyLightMask)
```
rom1504, line 314: Flatten should not be needed. If it is needed, then the format of chunk packet is wrong.

Archengius: I don't really know what you mean by "wrong", because format of the packet is correct and matches what the game actually sends (and for that particular field, game sends varint length prefixed array of 64-bit integers)

Archengius: Same applies to section mask in map_chunk

lib/plugins/blocks.js
extremeheat: I looked through the Minecraft java source and it is writing it as a 64-bit big endian array, so I don't get what this flipping this is about.
```java
         var1.writeLongArray(this.storage.getRaw()); // PalettedContainer.java
```
```java
   public FriendlyByteBuf writeLongArray(long[] var1) { // FriendlyByteBuf.java
      this.writeVarInt(var1.length);
      long[] var2 = var1;
      int var3 = var1.length;

      for(int var4 = 0; var4 < var3; ++var4) {
         long var5 = var2[var4];
         this.writeLong(var5);
      }

      return this;
   }
   ...
   public ByteBuf writeLong(long var1) {
      return this.source.writeLong(var1);
   }

   public ByteBuf writeLongLE(long var1) {
      return this.source.writeLongLE(var1);
   }
```

lib/plugins/blocks.js
rom1504: yes my point is that if we need to unparse what protodef just parsed at every time we use it, then the format we used is useless, and we should for example return a buffer and parse it directly in pchunk
mineflayer should not do any chunk parsing

lib/plugins/blocks.js
rom1504: this should at worse be happening in pchunk, not here
at best be unneeded by improving the definition of the packet in protocol.json

Archengius: Flipping is needed because BitSet implementation in prismarine-chunk requires least significant bits to come first pretty much because bit addressing starts from zero and grows upwards.
Original implementation in Java doesn't need to do this because it actually operates on 64-bit integers and not 32-bit ones, but in JS we're kinda forced to use 32-bit numbers.

lib/plugins/entities.js
u9g: re this: before this is merged, we should just emit a RemoveEntities packet with the one entity from the RemoveEntity packet because in 1.17.1, we get the RemoveEntities packet back

rom1504:
pchunk now has 1.17, let's finish this

lib/plugins/blocks.js
rom1504: solved this

lib/plugins/blocks.js
rom1504: solved

rom1504:
from @nickelpro : 
> On login, you get min_y and height from the login packet. These need to be passed prismarine-world (which needs to be updated) somehow, which then needs o pass them as minY and worldHeight in the options object to the ChunkColumn constructor.
> The defaults will work for worlds that don't mess with min_y and height, but Mojang has already released official world generator data packs that modify these

lib/plugins/game.js
rom1504: this should be inlined indeed

malafk (approved):
loooks good

rom1504:
>      Uncaught TypeError: Cannot read property 'min_y' of undefined
in 1.8

extremeheat:
Should be fixed, what's failing in 1.17 currently:
* crafting
* digAndBuild
* enchanting

lib/features.json
```diff
@@ -298,10 +298,5 @@
     "name": "entityNameLowerCaseNoUnderscore",
```
rom1504: this commit broke the tests

why are you undoing that ?

u9g: theres no reason to make this change if as soon as 1.17.1 the packet is literally back to what it was

u9g: fixed

rom1504: can we assume no server will be running 1.17 and all 1.17.1 ?

u9g: no lol, why would we just leave a bug like that

rom1504: no as in we can't assume it ?
this revert introduces a bug for 1.17
we can choose not to support 1.17 and only 1.17.1 if nobody uses 1.17

u9g: 1. How does it introduce a bug? 
2. Why would people skip 1.17?

rom1504: 1. because 1.17 has the different entity destroy packet which you just removed support for
2. I don't know, that's what I'm asking you. If they don't skip 1.17, then this commit shouldn't be done

rom1504: ah looking at the code I see you actually kept support for 1.17 , didn't read in detail before
all good then

lib/plugins/entities.js
```diff
@@ -31,6 +31,11 @@ function inject (bot, { version }) {
   const entitiesArray = require('minecraft-data')(version).entitiesArray
   const Item = require('prismarine-item')(version)
   const ChatMessage = require('prismarine-chat')(version)
+  // ONLY 1.17 has this destroy_entity packet which is the same thing as entity_destroy packet except the entity is singular
+  // 1.17.1 reverted this change so this is just a simpler fix
+  bot._client.on('destroy_entity', (packet) => {
+    bot._client.emit('entity_destroy', { entityIds: [packet.entityId] })
```
rom1504, line 37: it's a hack to emit packet events like this, it will prevent things like mineflayer proxy from working well
let's use an event on bot instead ?

lib/features.json
u9g: Yeah I was very confused

jkcoxson:
The crafting, inventory and enchanting GitHub tests are likely failing from issue #2064

nickelpro:
We came to the same conclusion on the Discord. Working on a fix

rom1504:
we're continuing at #2106

PR #1998, opened 2021-06-18
Remove anvil tests from blocked list since their test works now

u9g:


u9g:
closes https://github.com/PrismarineJS/mineflayer/pull/1998

rom1504 (approved):

PR #2008, opened 2021-06-24
Made _genericPlace options argument optional

IceTank:


u9g:
Can you explain why this is necessary? this is only intended for internal use.

IceTank:
I used _placeBlockWithOptions for some special block placement cenarios. But you are right if the function is only used internally its not a big deal.

u9g:
> I used _placeBlockWithOptions for some special block placement cenarios. But you are right if the function is only used internally its not a big deal.

_genericPlace  isn't meant to be used alone, it doesn't await anything being placed or anything like that, it just sends the packet, it's meant to be a lower level utility used by _place(Block|Entity)withOptions()

PR #2019, opened 2021-07-03
Add inventory item click delay

IceTank:
Adds a delay to every clickWindow function call to lower the cance of inventory desync, rejected transactions or anti cheat detection.


u9g:
I think this should be implemented in user code/where the clicks are used, this is the wrong place to put a delay. there's no reason to hold up resolving the click event

rom1504:
agreed

PR #2021, opened 2021-07-03
Add socks decleration to socks code snippet

IceTank:


u9g (approved):
Thanks for the contribution!

PR #2022, opened 2021-07-04
Fix crash on blocks without registered blockId

Furry:
A simple addition to the already-existing if statement on line 59 ( !blocks[packet.blockId] ) to ignore handling packets coming from unregistered blocks that may have been introduced by protocol modifications (IE a Forge Protocol Extension, https://www.npmjs.com/package/minecraft-protocol-forge)

### DIFF
--- a/lib/plugins/block_actions.js
+++ b/lib/plugins/block_actions.js
@@ -56,7 +56,7 @@ function inject (bot, { version }) {
     const pt = new Vec3(packet.location.x, packet.location.y, packet.location.z)
     const block = bot.blockAt(pt)
 
-    if (block === null) { return }
+    if (block === null || !blocks[packet.blockId]) { return }
 
     const blockName = blocks[packet.blockId].name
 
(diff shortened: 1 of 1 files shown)


lib/plugins/block_actions.js
```diff
@@ -56,7 +56,7 @@ function inject (bot, { version }) {
     const pt = new Vec3(packet.location.x, packet.location.y, packet.location.z)
     const block = bot.blockAt(pt)
 
-    if (block === null) { return }
+    if (block === null || !blocks[packet.blockId]) { return }
```
rom1504, line 59: can you add a comment this is for ignore non vanilla blocks ?

rom1504 (changes requested):

lib/plugins/block_actions.js
Furry, line 59: Ignore non-vanilla blocks

Furry:
Added the comment, sorry if I screwed up the Github flow with it, was abit confusing.

rom1504 (approved):

rom1504 (approved):

PR #2023, opened 2021-07-05
Fix placeEntity test stability by awaiting entityGone

u9g:


u9g:
it failed, but 1/9 is better then 2/2 fails in master rn, so I merged anyways, a better fix would be to record ci and figure out the problem that is so rare

PR #2025, opened 2021-07-05
add pviewer recording to ci

u9g:
This does work, but I don't think it is helpful after watching it

to make this work:
1. run: 
```
sudo apt update
sudo apt-get install ffmpeg xvfb -y 
```
2. change https://github.com/PrismarineJS/prismarine-viewer/blob/691dd688ee56ea823caca6f3540475eb8f92e686/viewer/lib/worldrenderer.js#L22-L23 to 
```js
src += '/worker.js'
```
3. run mocha with: `xvfb-run -s "-ac -screen 0 1280x1024x24"  npm run mocha_test`

### DIFF
--- a/lib/plugins/kick.js
+++ b/lib/plugins/kick.js
@@ -9,6 +9,6 @@ function inject (bot) {
   })
   bot.quit = (reason) => {
     reason = reason ?? 'disconnect.quitting'
-    bot._client.end(reason)
+    bot.end(reason)
   }
 }
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "main": "index.js",
   "types": "index.d.ts",
   "scripts": {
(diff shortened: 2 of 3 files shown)


rom1504:
Why isn't it helpful ?

u9g:
> Why isn't it helpful ?

- you can't really see what the bot is intending to do, because there's no chat and no way to see what's in the bots hand <-- this is the most important reason
- to use it, we have to modify node_modules, but that can be fixed
- the output video is just too small to do anything with (in my opinion)
- there's a ton of textures missing still, which make any real debugging hard because I'm unsure what i'm looking at

rom1504:
sounds like it's not the concept that's wrong but the current implem
all the points we mentionned could be fixed in pviewer

PR #2034, opened 2021-07-13
Add openContainer typing

IceTank:


u9g:
openContainer is internal, can you describe what this solves?

IceTank:
Hm jeah I missed openChest if they do the same thing then I guess its irrelevant

Issue #1482, opened 2020-12-05
Promisify follow up


rom1504:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%240%20Funded-%2300A156.svg" />](https://issuehunt.io/r/PrismarineJS/mineflayer/issues/1482)

* [x] promisify the few remaining examples
* [x] promisify the tests
* [ ] promisify most used plugins

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/PrismarineJS/mineflayer/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

TheDudeFromCI:
Plugin-specific related issues:
https://github.com/PrismarineJS/mineflayer-statemachine/issues/81
https://github.com/PrismarineJS/mineflayer-cmd/issues/15
https://github.com/PrismarineJS/mineflayer-tool/issues/14
https://github.com/PrismarineJS/mineflayer-collectblock/issues/41
https://github.com/PrismarineJS/mineflayer-pvp/issues/24
https://github.com/PrismarineJS/MineflayerArmorManager/issues/19
https://github.com/PrismarineJS/mineflayer-pathfinder/issues/89

(I didn't add one for prismarine-viewer since I'm not sure of the current state/direction.)

rom1504:
collectbot and pvp seem to be the main missing ones

I want to have this finished and remove cb, they're annoying

Saiv46:
Now it's only PrismarineJS/mineflayer-collectblock#73 left

rom1504:
This is done

PR #2035, opened 2021-07-13
Add nbt option to withdraw and deposit

IceTank:
Exposes options for withdraw in containers and adds an extra matcher function. 
Allows for withdrawing items with specific nbt tags.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -20,8 +20,8 @@
     - [Biome](#biome)
     - [Item](#item)
     - [windows.Window (base class)](#windowswindow-base-class)
-      - [window.deposit(itemType, metadata, count)](#windowdeposititemtype-metadata-count)
-      - [window.withdraw(itemType, metadata, count)](#windowwithdrawitemtype-metadata-count)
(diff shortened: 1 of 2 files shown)


rom1504 (changes requested):
could you instead change deposit and withdraw ?
this pattern of duplicating methods is not good, we should remove it for place with options

IceTank:
I don't quit understand what you mean. Should I rename the function or replace withdraw and deposit with my functions? Just replacing them would make it a breaking change.

rom1504:
I think you should change the existing functions to add the parameters you want without making a breaking change

rom1504:
afraid this would be breaking due to callbacks
let's wait a little that we remove the callbacks

IceTank:
Is there a time frame for removing callbacks?

rom1504:
yes, whenever someone does #1482 ; I think it would be great to do ASAP

rom1504:
Can you now rebase this on master @IceTank ? Should be possible now that we removed callbacks

PR #2039, opened 2021-07-15
Add direction support to activateBlock, openBlock

IceTank:
Why? Because you cannot open containers if they are at y level 255 because you cannot activate blocks from the top at y level 255. So openContainer does not work at y 255.


u9g:
failing tests

IceTank:
> failing tests

Jeah and this would break anything still using callbacks. So I think I will put this on hold until callbacks are gone.

rom1504:
blocked by #1482

rom1504:
Can you rebase this on master @IceTank ?

lib/plugins/inventory.js
```diff
@@ -338,8 +338,8 @@ function inject (bot, { version, hideErrors }) {
     }
   }
 
-  async function openBlock (block) {
-    bot.activateBlock(block)
+  async function openBlock (block, direction, cursorPos) {
```
rom1504, line 341: looks like you forgot the default here

IceTank: Yeah I just noticed that there are also types missing

IceTank:
todo: Change direction from number to Vec3

u9g:
has conflicts

u9g (approved):
LGTM

PR #2049, opened 2021-07-19
remake pviewer in tests pr

u9g:


u9g:
it works... a bit? seems like some of the videos corrupted when downloading? odd...

rom1504:
that's cool but idk if we can afford to have 1GB of artifact for each commit
what do you think ? should we merge ?

rom1504:
cool but let's not do it

rom1504:
or maybe we can add an option somewhere to enable it only if needed ?

u9g:
> or maybe we can add an option somewhere to enable it only if needed ?

maybe lets wait for player models to land

rom1504:
old

Issue #2044, opened 2021-07-18
Use Typescript definitions as documentation


stzups:
## Is your feature request related to a problem? Please describe.
While using this project, I constantly found issues understanding and using the API. 

`api.md` and `index.d.ts` can be inconsistent. For example, some functions like bot.openContainer are documented in [api.md](https://github.com/PrismarineJS/mineflayer/blob/master/docs/api.md#botopencontainercontainerblock-or-containerentity) but missing in [index.d.ts](https://github.com/PrismarineJS/mineflayer/blob/master/index.d.ts). 

I have also found `api.md` hard to read. It can be hard to find input parameter types or return types. For example, I had trouble figuring out that the `jsonMsg` parameter from the [chat](https://github.com/PrismarineJS/mineflayer/blob/master/docs/api.md#chat-username-message-translate-jsonmsg-matches) event was actually a `ChatMessage` object. I then had to figure out what a `ChatMessage` object is, which is not documented in `api.md`. Once I saw the definition in [index.d.ts](https://github.com/PrismarineJS/mineflayer/blob/93f179557ba44097c090d75ee2fcfb62ca0b0476/index.d.ts#L450), I understood how to use the `chat` event.

## Describe the solution you'd like
I'd like to see `index.d.ts` being using for the api. Something like [tsdoc](https://tsdoc.org/), where documentation is combined with the definition. This prevents inconsistencies and is more helpful - parameters and return values all have types, and the layout is standardized. This can also be turned in to actual documentation with [typedoc](https://github.com/TypeStrong/typedoc). This format also allows for documentation in the IDE. Many IDEs will show information about a type or function that is hovered over, or provide autocomplete suggestions when given these definitions. I like the idea of being able to navigate the API in my IDE. Instead of using control+f to find something on `api.md`, I could just control+click to jump to the definition.

The only hurdle I can see is localization, which is discussed [here](https://github.com/TypeStrong/typedoc/issues/823).

u9g:
Open to contributions :P

stzups:
Great! I'll start to draft a PR to see how it looks with documentation in `index.d.ts`

PR #2050, opened 2021-07-19
Typescript documentation

stzups:
This is a very rough draft of moving `api.md` to `index.d.ts`, and generating documentation with TypeDoc. This was discussed in https://github.com/PrismarineJS/mineflayer/issues/2044

Still need to figure out:
- Localization
    - https://github.com/IgniteUI/typedoc-plugin-localization seems like a good candidate
    - Alternatively, since TypeDoc generates HTML, it may be easiest to direct users to use something like Google Translate.
- Deployment
    - The easy solution is to check the documentation generated by TypeDoc in to version control, but this generally isn't appropriate for reasons shared [here](https://stackoverflow.com/a/54672930/11975214)
    - It may be possible to automate TypeDoc generation through some kind of GitHub workflow whenever a new release is put out, then deploy to GitHub pages somehow. Otherwise, this may require a separate website.


rom1504:
do you intend to finish this ?

stzups:
I might pick this up later once I use mineflayer more.

PR #2056, opened 2021-07-20
Wait between clicks option

IceTank:


IceTank:
This pr adds the option to wait between clicks in the mineflayer api. This should help with servers that are lagging or have high ping.

u9g:
imo waiting between clicks should be opt-in, not opt-out

IceTank:
> imo waiting between clicks should be opt-in, not opt-out

Were is it not opt-in? Also I just realized that this will break any legacy code that uses callbacks. So this might have to wait until callbacks are removed.

rom1504:
I do not understand the purpose of this PR
why would you want to make your bot slow ?

PR #2059, opened 2021-07-24
Remove polling, recursive calling, rechecks

Eagle-Anne:
This code is more complex but requires only a single lookup per chunk, does not poll, returns as soon as the chunk is loaded instead of at 1/10th second granularity, and does not fill the queue with recursive calls to itself.
In Rom's original waitForChunksToLoad if at least one chunk were unloaded when running the check, the code would poll the loaded chunks by waiting 100ms and then recursively calling waitForChunksToLoad() until each chunk loaded. That would fill up the task queue with recursive calls as well as repeat every check for the 5x5 area up until the blocking chunk each time it's called. Once that missing chunk finally loaded, as the functions complete and return they must complete the rest of the checks after the blocking chunk in each parent stack frame; the end result being continually checking those same chunks over and over until it finally drains the task queue. The worst case scenario for that code's chunk lookups until returning from the original waitForChunksToLoad() is 24 * 10 * length in seconds the chunk was unloaded.

### DIFF
--- a/lib/plugins/blocks.js
+++ b/lib/plugins/blocks.js
@@ -126,7 +126,7 @@ function inject (bot, { version, storageBuilder }) {
           resolve()
         }
       }
-      
+
       if (chunkPosToCheck.length > 0) { // begin listening for remaining chunks to load
         bot.world.on('chunkColumnLoad', waitForLoadEvents)
       } else { // all chunks were already loaded, skip listening for events
--- a/lib/plugins/blocks.js
+++ b/lib/plugins/blocks.js
@@ -122,13 +122,13 @@ function inject (bot, { version, storageBuilder }) {
           chunkPosToCheck.splice(position, 1)
         }
         if (chunkPosToCheck.length === 0) { // no chunks left to find
-          this.bot.world.off('chunkColumnLoad', waitForLoadEvents) // remove this listener instance
+          bot.world.off('chunkColumnLoad', waitForLoadEvents) // remove this listener instance
           resolve()
         }
       }
       
       if (chunkPosToCheck.length > 0) { // begin listening for remaining chunks to load
-        this.bot.world.on('chunkColumnLoad', waitForLoadEvents)
+        bot.world.on('chunkColumnLoad', waitForLoadEvents)
       } else { // all chunks were already loaded, skip listening for events
         resolve()
       }
--- a/lib/plugins/blocks.js
+++ b/lib/plugins/blocks.js
@@ -1,7 +1,7 @@
 const Vec3 = require('vec3').Vec3
 const assert = require('assert')
 const Painting = require('../painting')
-const { callbackify, sleep } = require('../promise_utils')
+const { callbackify } = require('../promise_utils')
 
 const { OctahedronIterator } = require('prismarine-world').iterators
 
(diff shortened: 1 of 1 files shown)


Karang (commented):
The logic look good to me but it would be better to write it using async / await instead of the explicit Promise.

lib/plugins/blocks.js
```diff
@@ -104,16 +104,35 @@ function inject (bot, { version, storageBuilder }) {
...
+          this.bot.world.off('chunkColumnLoad', waitForLoadEvents) // remove this listener instance
+          resolve()
+        }
+      }
+      
+      if (chunkPosToCheck.length > 0) { // begin listening for remaining chunks to load
+        this.bot.world.on('chunkColumnLoad', waitForLoadEvents)
+      } else { // all chunks were already loaded, skip listening for events
+        resolve()
+      }
```
Karang, line 134: This could be simplified using the `once` helper and `Promise.all()`

Eagle-Anne:
CI failed - forgot to bind the "this" for use inside the Promise. Fixing.

extremeheat:
Yeah, we discussed this on the discord. The issue with doing once + promise.all is that the chunk coordinates aren’t included in the event name for chunk loading, so that would need extra checking.

lib/plugins/blocks.js
Eagle-Anne: I might be misunderstanding your comment, but there are multiple chunks being loaded. If the code is run as a once, to the best of my knowledge there's no easy way to keep listening for other chunks if this one chunk isn't the one we want. If there is, then just running a Promise.all() on the map of a filtered list of missing chunks would be nice.

Eagle-Anne:
> The logic look good to me but it would be better to write it using async / await instead of the explicit Promise.

Could you clarify what you mean by that? I'm not sure how I can send a "all chunks were loaded" signal out of the event hander without resolve() or something similar like another event emitter, or a poll/busy wait.

lib/plugins/blocks.js
Karang: Yes you are right. My comment was a bit quick. However adding an event that include the chunk coordinates could be useful: https://github.com/PrismarineJS/prismarine-world/blob/master/src/world.js#L109

rom1504: yeah agreed, this event is missing

rom1504: https://github.com/PrismarineJS/prismarine-world/blob/master/docs/API.md#blockupdatex-y-z-oldblock-newblock we already have one for blocks, having one for chunks would be coherent

Eagle-Anne:
Hi, I see that this was marked as requiring input from me. What information or actions are needed at this time? I'm not sure why the CI is failing for pre 1.15 versions, if that's the question I'll have to punt to someone with more experience with older versions.

rom1504:
rebase

rom1504:
so tests are stuck with this change, which seems to indicate this change does not work

that's what would be needed to fix

Eagle-Anne:
I'm now a bit confused because before it was working fine for 1.15.2 and 1.16.5 and broken for everything else, and after the rebase it's now broken for everything but 1.14.4, including breaking on 1.15.2 and 1.16.5. I don't know what could have changed between the two. Would it be possible to run the CI a few times in sequence to test if this is a probabilistic issue?

rom1504:
The CI simply shows that this new implementation does not always work.
So just write a simple example using this function and run it 10 times manually.
Very likely it will fail part of the times
Then you can investigate and fix it

extremeheat:
Should be fixed now, but the prismarine-world change should also be simple to make

rom1504 (approved):

PR #2063, opened 2021-07-26
Bypass anticheats that detect sensitivity

mat-1:
Anticheats can detect the player's sensitivity by comparing how much the player's look direction changes each tick and getting the GCD of that. In vanilla when using 100% sensitivity, the player's look direction always changes by a multiple .15 when they're manually moving the camera with the mouse. This change makes mineflayer do that too. I have a proof-of-concept anticheat that does this (which is on my unfinished server, matdoes.dev) but there's likely other anticheats that do this as well.

### DIFF
--- a/lib/plugins/physics.js
+++ b/lib/plugins/physics.js
@@ -227,9 +227,9 @@ function inject (bot, { physicsEnabled }) {
 
     // this is done to bypass certain anticheat checks that detect the player's sensitivity
     // by calculating the gcd of how much they move the mouse each tick
-    const sensitivity = conv.fromNotchianPitch(.15) // this is equal to 100% sensitivity in vanilla
-    let yawChange = Math.round((yaw - bot.entity.yaw) / sensitivity) * sensitivity
-    let pitchChange = Math.round((pitch - bot.entity.pitch) / sensitivity) * sensitivity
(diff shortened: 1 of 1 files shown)


u9g:
by having this, does this mean you can't get an exact yaw/pitch anymore, just in intervals of .15?

mat-1:
> 
> 
> by having this, does this mean you can't get an exact yaw/pitch anymore, just in intervals of .15?

You can't set an *exact* yaw/pitch anymore, but I don't think there's ever a situation where you need to do this (since you can't in vanilla anyways) so it shouldn't break anything and it's not noticeable either.

u9g:
wouldnt this stop some parkour jumps from being possible then?

mat-1:
> 
> 
> wouldnt this stop some parkour jumps from being possible then?

Probably not, also those parkour jumps wouldn't be possible in vanilla.

rom1504:
there is no example of this helping to bypass any known anticheat and it increases complexity

re-open if you have more information

U5B:
I can confirm that Vulcan does check for sensitivity and that will flag.

mat-1:
@rom1504 USB confirmed this PR bypasses Vulcan anticheat.

PR #2073, opened 2021-08-03
Add placing crystal example

u9g:


u9g:
eventually, I would like this example to also use the calculateExplosionDamage function to pick the ideal place for most damage dealt, but this is fine for now

RandomVersion:
lets go anarchy bots

rom1504 (approved):

PR #2077, opened 2021-08-04
Add entityAtCursor function

O-of:


u9g:
needs a lint

lib/plugins/ray_trace.js
```diff
@@ -23,4 +24,36 @@ module.exports = (bot) => {
 
     return bot.world.raycast(eyePosition, viewDirection, maxDistance, matcher)
   }
+
+  bot.playerAtCursor = (maxDistance = 3.5) => {
+    const players = Object.keys(bot.players).map((key) => bot.players[key])
+      .filter((player) => {
+        return player.username !== bot.username && player.entity &&
+          Math.abs(player.entity.position.x - bot.entity.position.x) <= maxDistance &&
```
extremeheat, line 32: Position is a Vec3. You can use .distanceTo() and distanceSquared() on another Vec3.

You can also use Object.values when you want to get the values of an object, or Object.entries when you want an array of [key, value].

extremeheat (changes requested):

u9g:
shouldnt this be entityAtCursor instead? since you will hit any entity first even if you have a player in range of your cursor?

O-of:
I used `bot.players` rather than `bot.entities` so it should only target players.

O-of:
> > I used `bot.players` rather than `bot.entities` so it should only target players.
> 
> why? defeats the purpose of this since if you have a mob infront of the player you wont notice

It's mostly because I haven't played around with bounding boxes of other entities (and the use case I had in mind didn't have other entities), but I can always make changes to it so it uses different entities as well.

O-of:
Seeing as there's no changes requested, is there anything I need to do?

lib/plugins/ray_trace.js
```diff
@@ -23,4 +24,29 @@ module.exports = (bot) => {
...
+    const entities = Object.values(bot.entities)
+      .filter(entity => entity.type !== 'object' && entity.username !== bot.username && entity.position.distanceSquared(bot.entity.position) <= maxDistance ** 2)
+      .sort((entity1, entity2) => entity1.position.distanceSquared(bot.entity.position) - entity2.position.distanceSquared(bot.entity.position))
+
+    const dir = new Vec3(-Math.sin(bot.entity.yaw) * Math.cos(bot.entity.pitch), Math.sin(bot.entity.pitch), -Math.cos(bot.entity.yaw) * Math.cos(bot.entity.pitch))
+    const iterator = new RaycastIterator(bot.entity.position.offset(0, bot.entity.height, 0), dir.normalize(), maxDistance)
+
+    let pos = iterator.next()
+    while (pos) {
+      for (let i = 0; i < entities.length; i++) {
```
Karang, line 38: This code is checking all entities in range at each step of the recast, this is a lot of redundant work.
A way to improve that is to do independent raycasts for blocks and entities and then simply compare the distance of the first entity and the fist block.

In pseudocode:
* Do a block raycast using `const b = bot.world.raycast(eyePosition, viewDirection, maxDistance)`
* Update maxDistance: `maxDistance = Math.min(maxDistance, eyePosition.distanceSquared(b.intersect))`
* Do your filtering on entities
* Test entity intersects

Note that the 2 last steps can be merged in 1 loop to avoid creating temporary arrays
Note2: sorting the entities is not enough to ensure the first entity is returned, because some entities can be larger than others. You have to test them all and return the nearest, which shouldn't be an issue if you do the other simplifications.

O-of:
There seems to be a weird bug in `RaycastIterator.intersect` where an intersection is detected at
```
{
  pos: Vec3 {
    x: -1.4554090302898173,
    y: 83.76438681030076,
    z: 6.829271060309972
  },
  face: 3
}
```
when pos is 
```
{ x: -2, y: 83, z: 3, face: 3 }
```
Am I just using the function wrong, or is it an actual bug?

Karang:
> There seems to be a weird bug in `RaycastIterator.intersect` where an intersection is detected at
> 
> ```
> {
>   pos: Vec3 {
>     x: -1.4554090302898173,
>     y: 83.76438681030076,
>     z: 6.829271060309972
>   },
>   face: 3
> }
> ```
> 
> when pos is
> 
> ```
> { x: -2, y: 83, z: 3, face: 3 }
> ```
> 
> Am I just using the function wrong, or is it an actual bug?

What do you mean ? It's unlikely that the position returned by intersect has only integer coordinates, so the first one seems more legit.
The intersect method returns the exact point of intersection of the bounding box with the ray, not the center of the entity.

O-of:
> > There seems to be a weird bug in `RaycastIterator.intersect` where an intersection is detected at
> > ```
> > {
> >   pos: Vec3 {
> >     x: -1.4554090302898173,
> >     y: 83.76438681030076,
> >     z: 6.829271060309972
> >   },
> >   face: 3
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > when pos is
> > ```
> > { x: -2, y: 83, z: 3, face: 3 }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Am I just using the function wrong, or is it an actual bug?
> 
> What do you mean ? It's unlikely that the position returned by intersect has only integer coordinates, so the first one seems more legit.
> The intersect method returns the exact point of intersection of the bounding box with the ray, not the center of the entity.

I'm more focused on the distance between the 2 points rather than the position being only integer coordinates. The first coordinate I posted was the intersection while the second one was the pos returned by RaycastIterator.next() (I probably should have explained that earlier oops). There is an approx 3 block distance between the 2 points, while at most it should be roughly 1 block.

lib/plugins/ray_trace.js
```diff
@@ -23,4 +24,41 @@ module.exports = (bot) => {
 
     return bot.world.raycast(eyePosition, viewDirection, maxDistance, matcher)
   }
+
+  bot.entityAtCursor = (maxDistance) => {
```
Karang, line 28: maxDistance=3.5 was better

lib/plugins/ray_trace.js
```diff
@@ -23,4 +24,41 @@ module.exports = (bot) => {
...
+      .filter(entity => entity.type !== 'object' && entity.username !== bot.username && entity.position.distanceSquared(bot.entity.position) <= maxDistance ** 2)
+
+    const dir = new Vec3(-Math.sin(bot.entity.yaw) * Math.cos(bot.entity.pitch), Math.sin(bot.entity.pitch), -Math.cos(bot.entity.yaw) * Math.cos(bot.entity.pitch))
+    const iterator = new RaycastIterator(bot.entity.position.offset(0, bot.entity.height, 0), dir.normalize(), maxDistance)
+
+    const intersectArr = []
+
+    let pos = iterator.next()
+    while (pos) {
+      for (let i = 0; i < entities.length; i++) {
```
Karang, line 43: you are still testing all entities for each block, this loop needs to be outside the while

lib/plugins/ray_trace.js
```diff
@@ -23,4 +24,41 @@ module.exports = (bot) => {
...
+        if (intersect) {
+          if (intersect.pos.distanceTo(new Vec3(pos.x, pos.y, pos.z)) <= 2) { // Hotfix to prevent attacking behind head
+            const dist = bot.entity.position.distanceTo(intersect.pos)
+            intersectArr.push({ entity: entity, dist: dist })
+          }
+        }
+      }
+
+      if (intersectArr.length) {
+        return intersectArr.reduce((prev, curr) => prev.dist < curr.dist ? prev : curr).entity // There might be a better way to do this
```
Karang, line 58: yes, don't use reduce, do it on the fly

Karang:
> > > There seems to be a weird bug in `RaycastIterator.intersect` where an intersection is detected at
> > > ```
> > > {
> > >   pos: Vec3 {
> > >     x: -1.4554090302898173,
> > >     y: 83.76438681030076,
> > >     z: 6.829271060309972
> > >   },
> > >   face: 3
> > > }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > when pos is
> > > ```
> > > { x: -2, y: 83, z: 3, face: 3 }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Am I just using the function wrong, or is it an actual bug?
> > 
> > 
> > What do you mean ? It's unlikely that the position returned by intersect has only integer coordinates, so the first one seems more legit.
> > The intersect method returns the exact point of intersection of the bounding box with the ray, not the center of the entity.
> 
> I'm more focused on the distance between the 2 points rather than the position being only integer coordinates. The first coordinate I posted was the intersection while the second one was the pos returned by RaycastIterator.next() (I probably should have explained that earlier oops). There is an approx 3 block distance between the 2 points, while at most it should be roughly 1-2 blocks, so it really should not be intersecting.
> 
> I forgot to mention, but this was primarily looked into cause I noticed the bot was attacking behind itself.

Now that i looked the code it makes sense. The 2 pos you are talking about are unrelated.

For the bot attacking behind itself you need to check the sign of the dot product of (pos-eyePosition) and viewDirection

rom1504:
is this ready @Karang @O-of ?

O-of:
Waiting for @Karang to verify, but it's all right on my part.

lib/plugins/ray_trace.js
```diff
@@ -23,4 +24,39 @@ module.exports = (bot) => {
...
+    const iterator = new RaycastIterator(bot.entity.position.offset(0, bot.entity.height, 0), dir.normalize(), maxDistance)
+
+    let targetEntity = null
+    let targetDist = maxDistance
+
+    for (let i = 0; i < entities.length; i++) {
+      const entity = entities[i]
+      const w = entity.width / 2
+
+      const shapes = [[-w, 0, -w, w, entity.height + (entity.type === 'player' ? 0.18 : 0), w]]
```
extremeheat, line 45: where does the 0.18 constant come from?

extremeheat (approved):
LGTM, but I think this would be better suited to go into prismarine-world like bot.blockAtCursor.

O-of:
> LGTM, but I think this would be better suited to go into prismarine-world like bot.blockAtCursor.

I thought that `bot.blockAtCursor` is also in [ray_trace.js](https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/ray_trace.js#L18) too?

lib/plugins/ray_trace.js
O-of: Adding the 0.18 constant comes is due to [this](https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/physics.js#L248) where the bot's height is set to its eye height, 1.62, rather than the bot's actual height 1.80.

rom1504:
this needs docs

Karang:
> > LGTM, but I think this would be better suited to go into prismarine-world like bot.blockAtCursor.
> 
> I thought that `bot.blockAtCursor` is also in [ray_trace.js](https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/ray_trace.js#L18) too?

pworld doesn't have entities yet, so it makes sense to be in mineflayer for now. can be moved later

Saiv46 (approved):
LGTM

amoraschi:
Can you also add it in `index.d.ts`?

Pix3lPirat3:
One issue I've ran into with this myself (which I just added a blacklist argument for my own purposes) is that it likes to target experience orbs, which may create unwanted results for some users of this function, as you can interact through experience orbs and such.

PR #2078, opened 2021-08-05
Make physicsEnabled true by default

u9g:
api.md says this should be true by default


u9g:
this should be done by https://github.com/PrismarineJS/mineflayer/blob/8dc89f9ef8fbf225215e8ff5d92c64dbd6986f98/lib/plugins/physics.js#L39

Issue #2074, opened 2021-08-04
Awaiting look 


O-of:
- [x] The [FAQ](https://github.com/PrismarineJS/mineflayer/blob/master/docs/FAQ.md) doesn't contain a resolution to my issue 

## Versions
 - mineflayer: 3.9.0
 - server: PaperMC 1.8.8
 - node: 14.16.0

## Detailed description of a problem
Bot movements snap directly to set pitch/yaw when using `bot.look` rather than smoothly transitioning. 

## Your current code
```js
// Modified code to increase conciseness 
bot.on("message") {
    bot.setControlState("forward", true);
    bot.setControlState("sprint", true);
}

bot.on("physicTick") {
    const x = bot.entity.position.x;
    const y = bot.entity.position.y;
    const z = bot.entity.position.z;

    if (getDistance2d(x, z, 0, 0) >= 9) {
        await bot.look(new Vec3(0, y, 0));
    }
}

```

## Expected behavior
Bot moves in the direction it is facing, rather than set pitch/yaw.

Zn10plays:
cool cool, bot.look args are wrong... 
It that on purpose for sudo code?

O-of:
> cool cool, bot.look args are wrong...
> It that on purpose for sudo code?

More or less (I kinda just whipped something up quickly). The main point to look at is that physics.js uses the set yaw rather than the yaw sent to the server, making the bot's movement weird

PR #2082, opened 2021-08-05
Add option to bot.look

O-of:
With the current codebase, when having a control state active and doing a bot.look, the physics simulation snaps the bot's yaw to what was set, rather than the direction the bot sent to the server. This change gives option to let the bot head in the direction it sent to the server.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -249,12 +249,13 @@ export interface Bot extends TypedEmitter<BotEvents> {
 
   clearControlStates: () => void
 
-  lookAt: (point: Vec3, force?: boolean, callback?: () => void) => Promise<void>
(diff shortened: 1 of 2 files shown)


OofAlt:
Addresses #2074

OofAlt:
there seems to be some bugs with set yaw, need to rethink implementation

OofAlt:
fixed the bugs, didn't have to change much fortunately. Bot should only turn when a lookingTask is running

rom1504 (changes requested):
either do a custom callbackify to make the param really optional or this is a breaking change that will need to wait for callback to be removed

rom1504:
blocked by  #1482

rom1504:
can you please fix conflicts ?

PR #2085, opened 2021-08-07
Default swingArm to left hand and make if work

u9g:


rom1504 (approved):

Gjum:
What caused this PR? Looking at the packet ids this is obviously wrong (0 should be right hand, not left). This also changes the previous behavior of all existing `swingArm()` calls (no parameters) in the Mineflayer code; they used to swing the right hand and now swing the left hand.

PR #2086, opened 2021-08-09
Add the world height from the login packet

jkcoxson:
This will allow the world height to be passed into prismarine-world for world heights that aren't 256.


u9g:
all the 1.17 mineflayer changes are going in one pr

PR #2108, opened 2021-08-15
Fix chest lid move

imharvol:
Apparently since 1.17, when a chest is opened/closed 'block_action'  is emitted every 200ms (instead of once every time the chest is opened/closed), which was triggering multiple 'chestLidMove' events. Since useChests.js checks that 'chestLidMove' is only emitted once (even if it's a double chest) it was failing.

I've updated it so that it keeps track of how many players are looking at a chest (`openCountByPos`). And 'chestLidMove' is only emitted once the amount of players looking at that chest changes.

useChests.js seem to pass now

### DIFF
--- a/lib/plugins/block_actions.js
+++ b/lib/plugins/block_actions.js
@@ -19,6 +19,9 @@ const FACING_MAP = {
 function inject (bot, { version }) {
   const { instruments, blocks } = require('minecraft-data')(version)
 
+  // Stores how many players have currently open a container at a certain position
+  const openCountByPos = {}
+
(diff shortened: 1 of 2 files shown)


rom1504:
can you please fix your commits ?
0. git branch -D fix-chestLidMove
1. git checkout 1.17
2. git checkout -b fix-chestLidMove
3. put your commits (you can use git cherry-pick X )
4. git push --force fix-chestLidMove

imharvol:
Already merged into nickelpro's 1.17 branch

Issue #2002, opened 2021-06-21
Fix random port in tests


rom1504:
https://github.com/PrismarineJS/mineflayer/blob/master/test/externalTest.js#L35 what's the seed here ?
Not likely to work
#667

DrMoraschi:
We could use a package like portfinder that finds a free port

rom1504:
Using port 0 is the right solution

PR #2109, opened 2021-08-15
automatic port discovery for tests

Kashalls:
Address #2002 

Using identical fix from https://github.com/PrismarineJS/node-minecraft-protocol/pull/870


rom1504 (approved):
this is a pretty bad solution as it doesn't solve the race condition but slightly better than before

PR #2111, opened 2021-08-16
Added containerItems() in class Chest

olzx:
Hi, in Chest's examples this containerItems () method was used, but ts file (in the tooltips) it is not there, so I decided to add

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -666,6 +666,8 @@ export class Chest extends (EventEmitter as new () => TypedEmitter<StorageEvents
   count (itemType: number, metadata: number | null): number;
 
   items (): Item[];
+
+  containerItems (): Item[];
 }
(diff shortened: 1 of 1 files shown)


u9g:
this isnt the right fix, the window object should extend prismarine-window

olzx:
> this isnt the right fix, the window object should extend prismarine-window

But how does this relate to my fix?

u9g:
the prismarine-window class already has the .containerItems funxtion, so this change would be unnecessary

olzx:
> the prismarine-window class already has the .containerItems function, so this change would be unnecessary

Definitely! 
I didn’t see this hint because I am writing a plugin (in another file, but they are not presented there) I did not notice, but I saw. Thanks

PR #2112, opened 2021-08-16
Add bot.usingHeldItem

mat-1:
This adds bot.usingHeldItem (also to the docs and tests) which is necessary for eventually fixing #91 since you can't sprint while using the item in your hand.

Also the API table of contents was outdated so a VSC plugin I have fixed that too.


docs/api.md
```diff
@@ -264,9 +265,72 @@
...
+    - [bot.creative](#botcreative)
+      - [bot.creative.setInventorySlot(slot, item, [callback])](#botcreativesetinventoryslotslot-item-callback)
+      - [bot.creative.flyTo(destination, [cb])](#botcreativeflytodestination-cb)
+      - [bot.creative.startFlying()](#botcreativestartflying)
+      - [bot.creative.stopFlying()](#botcreativestopflying)
+<<<<<<< Updated upstream
       - [bot.getControlState(control)](#botgetcontrolstatecontrol-state)
       - [bot.getExplosionDamages(entity, position, radius, [rawDamages])](#botgetexplosiondamages)
+=======
+>>>>>>> Stashed changes
```
u9g, line 331: is this intended?

mat-1: Yeah that was an accident because I'm bad at git, fixed now.

lib/plugins/inventory.js
```diff
@@ -30,11 +30,27 @@ function inject (bot, { version, hideErrors }) {
...
+    }
+    bot.usingHeldItem = false
+  })
+
+  bot._client.on('set_cooldown', () => {
+    if (!eatingTask.done) {
+      eatingTask.finish()
+    }
+    bot.usingHeldItem = false
   })
```
imharvol, line 54: I would check that the item id the set_cooldown packet is referring to is the same as the item id of the item the bot is holding as it appears to be used only with ender pearls and chorus https://wiki.vg/Protocol#Set_Cooldown

Something like this under line 49 would do:
```js
if (bot.heldItem && bot.heldItem.type !== packet.itemID) return
```

mat-1: Alright I changed that. Also the packet is also used when someone disables your shield with an axe.

rom1504 (approved):

Issue #1589, opened 2021-01-16
New place block function


rom1504:
Suggested by @Karang 

> I'm thinking about designing a new place block function. The one we have isn't really like vanilla and lack some functionalities > + I've seen in some issues that it's confusing people.
> The functionalities and api breaking changes i'd like to make are:
> * give the block position instead of the reference block and have the function deduce reference block + face vector
> * give a stateId to the function and have the function select the correct item + parse the properties to deduce the face it should > click (if there is an axis property) and the half it should click ( top or bottom)
> * this one is important for pathfinder: in vanilla client the block is always placed before the server answer, this make it possible to stand on the block even if the server haven't answered yet. (if the block was illegal, the server will respond by resetting the block to what it was anyway)
> For it to work, the function will be sync and all the checks would have to be performed before (reach, visibility, etc) the function > can return a boolean for success. Maybe this should be part of pworld ? Let me know if you have some other ideas for this. >@rom1504 @TheDudeFromCI (AI irl)
> there is also the problem of positioning the player to place blocks with a rotation based on player position (not the face clicked) > an example is standing signs I don't think this function should position the player but it should at least warn if the player >position doesnt match the stateId property

It would be a completely new function.

Details at  https://discord.com/channels/413438066984747026/671423547541028884/795380060496068628

Karang:
Some additional informations:

I had some code in flying squid to take care of that, its not complete but its a start:
some doc: https://github.com/PrismarineJS/flying-squid/blob/master/docs/block-updates.md
the implem: https://github.com/PrismarineJS/flying-squid/blob/master/src/lib/plugins/placeBlock.js#L124

the getProperties function of prismarine-block https://github.com/PrismarineJS/prismarine-block/blob/master/index.js#L114

prismarine-schematic have the opposite function if you need it https://github.com/PrismarineJS/prismarine-schematic/blob/master/lib/states.js#L32 (converting a list of properties + block type to a stateId)

sefirosweb:
Add value for for place block with off hand?

Can be useful use pickaxe main and off hand block for avoid changing equip and avoid reject server equipment

PR #2129, opened 2021-08-20
Don't place blocks in an existing block's position

u9g:


lib/plugins/place_block.js
```diff
@@ -4,6 +4,8 @@ module.exports = inject
 
 function inject (bot) {
   async function placeBlockWithOptions (referenceBlock, faceVector, options) {
+    const {x, y, z} = referenceBlock.position
```
nickelpro: Need to add the face vector to this, which is the ultimate position of the placed block. What we're trying to avoid is people passing blocks that are "buried" behind other blocks as the `referenceBlock`.

rom1504 (changes requested):
this should throw when reference block *doesn't* exist

nickelpro:
It should throw in both cases, reference block doesn't exist and when the reference block is buried

rom1504:
be aware of https://github.com/PrismarineJS/mineflayer/issues/1589 ; finding the position of the placed block is not completely trivial

lib/plugins/place_block.js
u9g: would that be:
```js
referenceBlock.position.plus(faceVector)
```
?

u9g:
> this should throw when reference block _doesn't_ exist

when would that happen?

u9g:
> be aware of #1589 ; finding the position of the placed block is not completely trivial

so I guess I would just have to account for whether we can place a second slab there?

rom1504:
> when would that happen?

for example if you use air as a reference block

rom1504:
re-open if you want to do it

PR #2136, opened 2021-08-22
remake afk example

u9g:
Old example is a lot more code and a lot more confusing


examples/plugins/rotator.js
```diff
@@ -0,0 +1,20 @@
+module.exports = bot => {
+  let rotater
+  let rotated = false
+  bot.afk = {}
```
rom1504, line 4: The name should still be afk imo
+ Can you add back the comments ?

u9g: done

rom1504 (approved):

PR #2140, opened 2021-08-24
remake crafting plugin

u9g:


nickelpro (approved):
LGTM with one reservation: I can't come up with a complete explanation for what connection these changes have on the `useChests` test. I think this code should get merged regardless, but I think don't think the `useChests` test is fixed by them. You can see similar random flailing with regards to useChest in the 1.17 PR.

u9g (commented):
stop using openContainer to open the crafting table

rom1504:
this is almost done, why not finish it ?

rom1504:
ci failing

rom1504:
Old

PR #2141, opened 2021-08-24
1.17 support

u9g:


.github/workflows/ci.yml
```diff
@@ -26,7 +26,7 @@ jobs:
         mcVersion: ['1.8.8', '1.9.4', '1.10.2', '1.11.2', '1.12.2', '1.13.2', '1.14.4', '1.15.2', '1.16.5']
         include:
          - javaVersion: 16
-           mcVersion: '1.17.1'
+           mcVersion: ['1.17', '1.17.1']
```
rom1504, line 29: no that doesn't work,
do 
 - javaVersion: 16
           mcVersion: ['1.17']

u9g: then how would we add 1.17.1 tests?

rom1504: that's how, you put both, like this:
```
         - javaVersion: 16
           mcVersion: '1.17'
         - javaVersion: 16
           mcVersion: '1.17.1'
```

u9g: I'll do this, but is there a cleaner way to do this for the future?

rom1504: maybe but I don't know it

test/internalTest.js
```diff
@@ -297,9 +297,15 @@ for (const supportedVersion of mineflayer.testedVersions) {
       it('sets players[player].entity to null upon despawn', (done) => {
         let serverClient = null
         bot.once('entitySpawn', (entity) => {
-          serverClient.write('entity_destroy', {
-            entityIds: [8]
-          })
+          if (mcData.protocol.play.toClient.types.packet[1][1].type[1].entity_destroy) {
```
rom1504, line 300: just depend on the version

u9g: done

test/internalTest.js
```diff
@@ -297,9 +297,15 @@ for (const supportedVersion of mineflayer.testedVersions) {
       it('sets players[player].entity to null upon despawn', (done) => {
         let serverClient = null
         bot.once('entitySpawn', (entity) => {
-          serverClient.write('entity_destroy', {
-            entityIds: [8]
-          })
+          if (bot.version !== '1.17') {
```
rom1504, line 300: Why not support feature ?

PR #1666, opened 2021-02-13
Add ability to use an anvil fully

u9g:

PR #2143, opened 2021-08-24
Scrub noWaiting from function signature and docs

nickelpro:
`noWaiting` never did what the docs said it did (I'm not entirely clear what it _did_ do, but not that). With the 1.17 PR merged it's gone entirely, so this change just reflects that.


rom1504:
Let's try and find the commit adding that to understand what was the intended purpose

nickelpro:
Well that helped me find the one place we were passing it, but it doesn't help me understand what it was supposed to do

https://github.com/PrismarineJS/mineflayer/pull/1666/commits/a2aca59c03cfcd410d5366fff4e131c063e17933

PR #2162, opened 2021-08-29
Added `getExplosionDamages` to index.d.ts

link-discord:


link-discord:
While I'm at it I also made it so if `rawDamages` is true in the `explosion.js` plugin it also checks if the armour props are present. Since  using the calculateExplosionDamages function without armour and `rawDamages` disabled it would cause an error

index.d.ts
```diff
@@ -257,6 +257,8 @@ export interface Bot extends TypedEmitter<BotEvents> {
 
   clearControlStates: () => void
 
+  getExplosionDamages: (entity: Entity, position: Vec3, raidus: number, rawDamages?: boolean) => number
```
u9g, line 260:
```suggestion
  getExplosionDamages: (targetEntity: Entity, position: Vec3, radius: number, rawDamages?: boolean) => number
```

lib/plugins/explosion.js
```diff
@@ -77,7 +77,7 @@ function inject (bot) {
     // The following modifiers are constant for the input targetEntity and doesnt depend
     // on the source position, so if the goal is to compare between positions they can be
     // ignored to save computations
-    if (!rawDamages) {
+    if (!rawDamages && targetEntity.attributes['generic.armor']) {
```
u9g, line 80: instead of returning undefined maybe we throw here? or return a null?

link-discord: The main reason I added this change is so that it doesn't throw. I would say we just return null instead

u9g:
can you add to typings that this returns number or null and to docs, and explain why in docs

link-discord:
> can you add to typings that this returns number or null and to docs, and explain why in docs

Ok I will do it now

u9g (approved):
LGTM

PR #2184, opened 2021-09-04
Default hideErrors to false

U5B:
Inspired by: https://discord.com/channels/413438066984747026/413438150594265099/883788843613057055


rom1504:
can we instead log to a file ?
if we merge this, tomorrow there will be 10 issues about "it doesn't work, I saw a warning"

U5B:
Where should I log the file?

u9g:
I disagree with logging to a file because what if you have 10 mineflayer bots all logging to a single file?

rom1504:
Indeed logging to a single file is not a good idea

PR #2188, opened 2021-09-10
Change entity.vehicle implementation

OofAlt:
`entity.vehicle` was only set when receiving the `set_passengers` packet, which only is sent in versions 1.9+, so an extra listener was added to the `attach_entity` packet

`entity.vehicle` was deleted after entity dismounts. Instead, I set it to null

`entity.passenger` was added to circumvent the need to iterate through all the loaded entities to find the passenger of an entity.

### DIFF
--- a/lib/plugins/entities.js
+++ b/lib/plugins/entities.js
@@ -333,19 +333,6 @@ function inject (bot, { version }) {
     if (eventName) bot.emit(eventName, entity)
   })
 
-  bot._client.on('attach_entity', (packet) => {
-    // attach entity
-    const entity = fetchEntity(packet.entityId)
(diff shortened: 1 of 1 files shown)


IceTank:
Would be cleaner if you used mcData.isNewerOrEqualTo or isOlderThan like this https://github.com/PrismarineJS/prismarine-windows/blob/f0452cb61e9065be1e263c5f844ddd5ddca43011/index.js#L7 for the different versions. Also can you update the docs with the changes?

rom1504:
I don't feel confident merging this without at least one person having tested it or unit tests

will close in a week if no feedback is provided

rom1504:
too old, please reopen if you want to check it works

PR #2213, opened 2021-09-18
Add pitch speed to look

IceTank:
Right now pitch movement in bot.look is always instant. With this pitch movement is the same as yaw movement were it slowly looks into a given direction using physics.pitchSpeed.

Depends on https://github.com/PrismarineJS/prismarine-physics/pull/49


IceTank:
This version is currently using a forked prismarine-physics version that has to be changed back to the primarine-physics once the change in prismarine-physics has been merged and released.

rom1504:
can you do a release pr in pphysics and then update with that version here ?

PR #2215, opened 2021-09-18
Small typings fixes

u9g:


rom1504:
Are you sure it's not a bug if it's undefined ?

u9g:
> Are you sure it's not a bug if it's undefined ?

No. If the player doesn't exist, the player is undefined, ie:
```js
const a = {}
console.log(a.b) // => undefined
```
And when the entity gets destroyed, we just set the entity to null: https://github.com/PrismarineJS/mineflayer/blob/u9g-patch-4/lib/plugins/entities.js#L270

rom1504:
> No. If the player doesn't exist, the player is undefined, ie:

how can the player not exist?

u9g:
> > No. If the player doesn't exist, the player is undefined, ie:
> 
> how can the player not exist?

if the player isn't on the server, they don't exist... What If Notch never logged onto your server, he doesn't exist as far as we know.

rom1504:
@u9g then why would it be in this.players ?

PR #2217, opened 2021-09-21
added rejection of invalid transaction packets (anticheat fix)

U5B:
For servers with anticheats that send actionIds like
```{ windowId: 0, action: 16707, accepted: false }```
or
```{ windowId: 0, action: -7967, accepted: false }```

Fixes https://github.com/PrismarineJS/mineflayer/issues/2189

### DIFF
--- a/lib/features.json
+++ b/lib/features.json
@@ -318,5 +318,10 @@
     "name": "setSlotAsTransaction",
     "description": "use setslot as transaction instead of just hoping it'll work",
     "versions": ["1.17", "1.17.1"]
+  },
+  {
+    "name": "pingPacket",
+    "description": "ping instead of transaction in 1.17",
+    "versions": ["1.17", "1.17.1"]
   }
 ]
--- a/lib/features.json
+++ b/lib/features.json
@@ -318,10 +318,5 @@
     "name": "setSlotAsTransaction",
     "description": "use setslot as transaction instead of just hoping it'll work",
     "versions": ["1.17", "1.17.1"]
-  },
-  {
-    "name": "pingPacket",
-    "description": "ping instead of transaction in 1.17",
-    "versions": ["1.17", "1.17.1"]
   }
 ]
--- a/lib/plugins/game.js
+++ b/lib/plugins/game.js
@@ -90,4 +90,12 @@ function inject (bot, options) {
   bot._client.on(brandChannel, (serverBrand) => {
     bot.game.serverBrand = serverBrand
   })
+  
+  if (bot.supportFeature('pingPacket') {
+    bot._client.on('ping', (packet) => {
+      bot._client.write('pong', {
+        id: packet.id
+      })
+    })    
+  }
 }
--- a/lib/plugins/game.js
+++ b/lib/plugins/game.js
@@ -91,11 +91,9 @@ function inject (bot, options) {
     bot.game.serverBrand = serverBrand
   })
 
-  if (bot.supportFeature('pingPacket')) {
-    bot._client.on('ping', (data) => {
-      bot._client.write('pong', {
-        id: data.id
-      })
+  bot._client.on('ping', (data) => {
+    bot._client.write('pong', {
+      id: data.id
     })
-  }
+  })
 }
--- a/lib/plugins/game.js
+++ b/lib/plugins/game.js
@@ -90,11 +90,7 @@ function inject (bot, options) {
   bot._client.on(brandChannel, (serverBrand) => {
     bot.game.serverBrand = serverBrand
   })
-  bot._client.on('ping', (data) => {
-    bot._client.write('pong', {
-      id: data.id
-    })
-  })
+
   if (bot.supportFeature('pingPacket')) {
     bot._client.on('ping', (data) => {
       bot._client.write('pong', {
--- a/lib/plugins/game.js
+++ b/lib/plugins/game.js
@@ -91,7 +91,7 @@ function inject (bot, options) {
     bot.game.serverBrand = serverBrand
   })
   
-  if (bot.supportFeature('pingPacket') {
+  if (bot.supportFeature('pingPacket')) {
     bot._client.on('ping', (packet) => {
       bot._client.write('pong', {
         id: packet.id
--- a/lib/plugins/game.js
+++ b/lib/plugins/game.js
@@ -90,12 +90,16 @@ function inject (bot, options) {
   bot._client.on(brandChannel, (serverBrand) => {
     bot.game.serverBrand = serverBrand
   })
-  
+  bot._client.on('ping', (data) => {
+    bot._client.write('pong', {
+      id: data.id
+    })
+  })
   if (bot.supportFeature('pingPacket')) {
-    bot._client.on('ping', (packet) => {
+    bot._client.on('ping', (data) => {
       bot._client.write('pong', {
-        id: packet.id
+        id: data.id
       })
-    })    
+    })
   }
 }
--- a/lib/plugins/game.js
+++ b/lib/plugins/game.js
@@ -91,6 +91,7 @@ function inject (bot, options) {
     bot.game.serverBrand = serverBrand
   })
 
+  // mimic the vanilla 1.17 client to prevent anticheat kicks
   bot._client.on('ping', (data) => {
     bot._client.write('pong', {
       id: data.id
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by U5B** [comment]
Test Code
```js
  host: "mc.thearchon.net",
  version: "1.8.9",
  username: "ign",
  password: "bestpassword"
})
// TO DO WHEN SPAWN
bot.on("spawn", function () {
  bot.chat("/outlands")
})

// LOAD resourcePack
bot.once('resourcePack', () => { // resource pack sent by server
  bot.acceptResourcePack()
  console.log('ResourcePack Done')
})

bot._client.on('transaction', (data, packetMeta) => {
  console.log(data)
  console.log(packetMeta)
})```

**Review by Greenlio — COMMENTED** [review]
On Vulcan 1.17.1, after using this fix, I receive the same kick reason and the same feedback is given to staff members. Using test code provided by you, here is what was logged:

kicked: {"text":"Internal Exception: io.netty.handler.timeout.ReadTimeoutException"}
Ofcourse, that's not a real error. That's just what Vulcan says whenever it's one of those kicks.
Refer to https://gist.github.com/freppp/6f669d666835011f3661025c2d9533e2 Line 151 for more information.

**Comment by U5B** [comment]
> On Vulcan 1.17.1, after using this fix, I receive the same kick reason and the same feedback is given to staff members. Using test code provided by you, here is what was logged:
> 
> kicked: {"text":"Internal Exception: io.netty.handler.timeout.ReadTimeoutException"}
> Ofcourse, that's not a real error. That's just what Vulcan says whenever it's one of those kicks.
> Refer to https://gist.github.com/freppp/6f669d666835011f3661025c2d9533e2 Line 151 for more information.

Ah, this was for a 1.8.9 Vulcan server, not a 1.17.1 Vulcan server. There are probably other checks. Could you send me on discord or here the 1.17.1 server ip?

**Review by Greenlio — APPROVED** [review]
Epic, works on my end, thank you again.

File: lib/features.json [hunk 1]
```diff
@@ -318,5 +318,10 @@
     "name": "setSlotAsTransaction",
     "description": "use setslot as transaction instead of just hoping it'll work",
     "versions": ["1.17", "1.17.1"]
+  },
+  {
+    "name": "pingPacket",
```
u9g, line 323: shouldn't be a thing, ping packet nisnt the same thing as transaction packet

u9g:
Are you sure this is the right fix? Are you sure that the vanilla client doesn't just accept any transaction id below the current one?

lib/features.json
U5B: so should I just remove it or say that ping packet was added in 1.17?

U5B:
> Are you sure this is the right fix? Are you sure that the vanilla client doesn't just accept any transaction id below the current one?

I believe this is the correct fix. I tested with a vanilla client and it rejects any id that is not a regular inventory id. So any negatives and high positives are counted.

u9g:
> > Are you sure this is the right fix? Are you sure that the vanilla client doesn't just accept any transaction id below the current one?
> 
> I believe this is the correct fix. I tested with a vanilla client and it rejects any id that doesn't show up in the click array. So any negatives and high positives are counted.

I don’t mean just test it, I mean is shifting the click queue the right option? I don’t think it is, i think just if the id is lower than current id, accept it

U5B:
> > > Are you sure this is the right fix? Are you sure that the vanilla client doesn't just accept any transaction id below the current one?
> > 
> > 
> > I believe this is the correct fix. I tested with a vanilla client and it rejects any id that doesn't show up in the click array. So any negatives and high positives are counted.
> 
> I don’t mean just test it, I mean is shifting the click queue the right option? I don’t think it is, i think just if the id is lower than current id, accept it

I didn't make a modification to shift the queue?

U5B:
I believe I fixed the issue of shifting and also found an issue with out of order transaction packets and fixed that.

lib/plugins/inventory.js
```diff
@@ -463,6 +469,15 @@ function inject (bot, { version, hideErrors }) {
       })
       bot.emit(`confirmTransaction${click.id}`, false)
     }
+
+    function onRejectedError () {
```
u9g, line 473: how come its called onRejectedError but you reply `accepted: true`, and you also comment out an emit, that's breaking for no reason, can you just write the packet in the if statement and revert the changes here?

U5B: I have done that

u9g:
Can you make sure it works still? Other than that, lgtm

U5B:
Works, might need to edit something mini because random test failed with no relation to inventory.

lib/plugins/game.js
```diff
@@ -90,4 +90,10 @@ function inject (bot, options) {
   bot._client.on(brandChannel, (serverBrand) => {
     bot.game.serverBrand = serverBrand
   })
+
+  bot._client.on('ping', (data) => {
```
rom1504, line 94: pretty sure this is already handled in nmp

rom1504:
did you test this on vanilla and some spigot servers having weird inventories ?

U5B:
> did you test this on vanilla and some spigot servers having weird inventories ?

Yes, I have tested on both.

lib/plugins/game.js
U5B: I cannot find where it is in node-minecraft-protocol.

U5B: https://wiki.vg/Protocol#Ping_2
https://wiki.vg/Protocol#Pong
This is a different ping packet.

rom1504: "Unknown what this packet does just yet, not used by the Notchian server or client. Most likely added as a replacement to the removed window confirmation packet.

"

rom1504: if you have more information, please edit wiki.vg
otherwise remove this code change

U5B: Edited wiki.vg also I gave the wrong link for Pong.
https://wiki.vg/Protocol#Pong_2
https://wiki.vg/Protocol#Ping_2

U5B:
Don't know what user input is needed for this PR. I believe this is done.

EndermanHackerVN (approved):

lib/plugins/inventory.js
```diff
@@ -427,13 +427,24 @@ function inject (bot, { version, hideErrors }) {
   function confirmTransaction (windowId, actionId, accepted) {
     // drop the queue entries for all the clicks that the server did not send
     // transaction packets for.
-    let click = windowClickQueue.shift()
-    if (click === undefined) {
+    // Also reject transactions that aren't sent from mineflayer
+    let click = windowClickQueue[0]
```
rom1504, line 431: Can you explain the logic?
This function is "confirmTransaction"
What your new code is doing is whenever we want to confirm a transaction
And there is no click in the queue we first reject the transaction
Why is that ?

U5B: We reject the transaction if the specific transaction id is not in windowClickQueue, or if there is no transaction ids in windowClickQueue.

If we don't, the server can send an out-of-order transaction packet, that will mess up how we respond.
For example, say we click a slot in the inventory. The server may send an different response to that click and then the valid response. 

Example for Mineflayer:
Mineflayer sends click window with action id of 1
Server sends transaction with action id of -1064
---- (Vanilla sends transaction with action id of -1064)
Server sends transaction with action id of 1
Mineflayer sends transaction with action id of 1

U5B: Code from Vanilla:
public void handleContainerAck(ClientboundContainerAckPacket clientboundContainerAckPacket) {
  PacketUtils.ensureRunningOnSameThread(clientboundContainerAckPacket, this, this.minecraft);
  AbstractContainerMenu abstractContainerMenu = null;
  LocalPlayer localPlayer = this.minecraft.player;
  if (clientboundContainerAckPacket.getContainerId() == 0) {
      abstractContainerMenu = localPlayer.inventoryMenu;
  } else if (clientboundContainerAckPacket.getContainerId() == localPlayer.containerMenu.containerId) {
      abstractContainerMenu = localPlayer.containerMenu;
  }
  if (abstractContainerMenu != null && !clientboundContainerAckPacket.isAccepted()) {
      this.send(new ServerboundContainerAckPacket(clientboundContainerAckPacket.getContainerId(), clientboundContainerAckPacket.getUid(), true));
  }
}

PR #2224, opened 2021-09-25
Remove unnecessary and buggy inventory check

Karang:
This check added in the last 1.17 commit breaks block placing in survival. It also breaks placing blocks with pathfinder. If the block cannot be placed (for instance if an entity is in the way) the slot is never updated and therefore the promise never resolve.
The correct way to check if a block has been placed is to check for block update event, which is done in https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/place_block.js#L13.


Karang:
@nickelpro @u9g any idea why the 1.17 bed test is failing ? It seems unrelated but this change is causing it to fail. And why would the 1.17 fail but the 1.17.1 pass ?

rom1504 (approved):

nickelpro:
This will cause a race in inventory handling. In 1.17+ you need to await the update from the server when you place a block from inventory.

nickelpro:
And the 1.17 tests are there because inventory interactions work differently in 1.17 than other versions. Without testing we're basically guaranteed to break 1.17 without noticing. Again I agree that the setup prior to this commit was broken, but just removing those tests is also wrong.

PR #2228, opened 2021-09-26
add a check for isAlive

U5B:
I believe this will prevent the bot from yeeting itself into the void on death (if the server doesn't lag him back)

### DIFF
--- a/lib/plugins/physics.js
+++ b/lib/plugins/physics.js
@@ -245,11 +245,6 @@ function inject (bot, { physicsEnabled }) {
 
   // player position and look (clientbound)
   bot._client.on('position', (packet) => {
-    if (doPhysicsTimer === null) {
-      lastPhysicsFrameTime = performance.now()
-      doPhysicsTimer = setInterval(doPhysics, PHYSICS_INTERVAL_MS)
-    }
-
     bot.entity.height = 1.62
     bot.entity.velocity.set(0, 0, 0)
 
@@ -277,6 +272,11 @@ function inject (bot, { physicsEnabled }) {
     shouldUsePhysics = true
     bot.entity.timeSinceOnGround = 0
     lastSentYaw = bot.entity.yaw
+ 
+    if (doPhysicsTimer === null) {
+      lastPhysicsFrameTime = performance.now()
+      doPhysicsTimer = setInterval(doPhysics, PHYSICS_INTERVAL_MS)
+    }
 
     bot.emit('forcedMove')
   })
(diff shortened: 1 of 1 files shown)


rom1504:
does it work? what's the difference before/after ?

lib/plugins/physics.js
```diff
@@ -1,4 +1,4 @@
-const { Vec3 } = require('vec3')
+const Vec3 = require('vec3').Vec3
```
Karang, line 1: Change not needed

lib/plugins/physics.js
```diff
@@ -142,7 +142,7 @@ function inject (bot, { physicsEnabled }) {
       sendPacketPosition(position, onGround)
     } else if (lookUpdated && bot.isAlive) {
       sendPacketLook(yaw, pitch, onGround)
-    } else if (performance.now() - lastSent.time >= 1000) {
+    } else if (performance.now() - lastSent.time >= 1000 && bot.isAlive) {
```
Karang, line 145: Did you check with a proxy that it is vanilla behavior ? I'm pretty sure this packet need to be sent no matter if the bot is alive.

lib/plugins/physics.js
```diff
@@ -245,6 +245,11 @@ function inject (bot, { physicsEnabled }) {
 
   // player position and look (clientbound)
   bot._client.on('position', (packet) => {
+    if (doPhysicsTimer === null) {
+      lastPhysicsFrameTime = performance.now()
+      doPhysicsTimer = setInterval(doPhysics, PHYSICS_INTERVAL_MS)
+    }
+
```
Karang, line 252: why move this code ?

U5B: Did not mean to move this.

lib/plugins/physics.js
U5B: I believe this is vanilla behavior. For the duration that the bot is dead, it does not send a position packet.

U5B:
> does it work? what's the difference before/after ?
https://ptb.discord.com/channels/413438066984747026/519952494768685086/891867964251447306

In laggy condtions (where the server is lagging) the bot will sometimes yeet itself into the void if the server does not correct the position.

U5B:
The other thing is that we could do is disable physics simulation until we are "alive", (by adding '&& bot.isAlive') to https://github.com/U5B/mineflayer/blob/8752e5fe9680efbdcb67daf62c1f2502e24ef71f/lib/plugins/physics.js#L62.

U5B:
Actually this probably isn't the correct fix, I will research this more later.

PR #2235, opened 2021-09-30
Update deps on some packages

u9g:
to cause less problems when people have older deps


rom1504 (approved):

rom1504:
doesn't hurt but also, can you think of one case where it could help ? (eg the user did run npm install and it didn't install the last version of the libs)

PR #2238, opened 2021-10-02
Fix removing many players at once from tab list

mat-1:
When a server sends a packet to remove players, Mineflayer only removes the first player, causing bot.players to be wrong. This can be seen for example when you join a Hypixel duels game and check bot.players, almost all the players from the lobby you were in are still in tab when they shouldn't be.


u9g (approved):
LGTM

PR #2245, opened 2021-10-06
Added folders for examples.

### LINKED ISSUE #1690: Add folders to examples
There's too many examples, would be better to categorize.

Add a folder structure, the structure is to be decided.

- make a readme for the examples folder

## Additional context
discussed in discord, but never reached a consensus

ClientCrash:
Closes #1690  

- [x] add the folders
- [ ] rewrite tests to new folder locations


extremeheat:
Should not be any spaces or casing inconsistencies in folder names. I’m also not sure this makes things much easier, it’s more clicks to dig through various folders with arbitrary categories. Some things make sense to combine (for example things that need a package.json), but things like “chat and network” doesn’t sound right. 

I think putting in a README into the example folder with the categorization there might make more sense.

rom1504:
please re-open if you want to work on it

PR #2247, opened 2021-10-07
add option to disable default chat patterns

U5B:


U5B:
Somehow `azure/msal-common@^5.0.1` cannot be found?

At least tests aren't failing to standard... yet.

U5B:
https://regex101.com/r/GQrK7F/1
this might work a bit better while still doing its job of matching the chat formats

U5B:
ugly regex can do the job, but it is pretty much a hotpatch to prevent crashing.

```js
/(?:[\(\[][^\(\)\[\}]+[\]\)]|[^\(\)\[\]\n])*?(\w+)/

File: lib/plugins/chat.js [hunk 1]
```diff
@@ -171,7 +172,9 @@ function inject (bot, options) {
   bot.tabComplete = callbackify(tabComplete)
 
   function addDefaultPatterns () {
-    const USERNAME_REGEX = '(?:\\(.+\\)|\\[.+\\]|.)*?(\\w+)'
+    if (!defaultChatPatterns) return
+    // original regex: (?:\\(.+\\)|\\[.+\\]|.)*?(\\w+)
```
u9g, line 176: why leave the original regex

U5B: In case someone wanted to see what the original regex looked like before I changed it

Through git blame exists

rom1504:
can you explain the objective of this new feature?

lib/plugins/chat.js
rom1504: please remove it

rom1504: however that seems unrelated with this PR, why did you make that change?

U5B:
The objective of this feature is to be able to disable the default chat patterns to prevent the bot from crashing when a series of '[]' or '()' is recieved.

However, I ended up patching the regex so it wouldn't crash without breaking what the original regex matched.

lib/plugins/chat.js
U5B: I made that change since I believed it would be a vulnerability if anyone could just chat a series of `[]` or `()` and crash any mineflayer bot.

U5B: I'll revert my regex change.

U5B:
I reverted the regex change because it wasn't matching
`[wvffle] meh`
which I honestly don't know how to fix that.

rom1504:
merged but this is a very poor fix

instead we should fix the regexes

PR #2262, opened 2021-10-13
Console spam fix

IceTank:
When mineflayer tries to move items in its inventory the server sometimes rejects the inventory clicks. 
Mineflayer then json stringifys the entire inventory content including item meta and nbt data. This can result in mineflayer printing out more then 5000 lines of json if the inventory is really full.


extremeheat:
Is it really necessary to print the entire window in the error at all here? This sounds like it should be debug information, not part of exception

PR #2272, opened 2021-10-22
Add action compatibility 

IceTank:
First draft for adding task holding and task compatibility checking. This is probably missing a lot of task that should still be added with identifiers. This includes just the task that used createTask.


IceTank:
Someone might have to check with a few third party plugins if they still work with these changes. And the list of compatible tasks may not be complete at the moment.

rom1504:
I'd appreciate if people could help review this (maybe @u9g ?)

extremeheat:
I think that some of these should throw an exception, they sound like errors in the user logic, for example not using await. 

Otherwise I think it's OK with some added tests to make sure catching incompatible tasks works and some doc for people to update the list in the future

rom1504:
yeah rereading this why not throw instead of await ?
awaiting is handling the task scheduling for the user, which I think we shouldn't do

lib/plugins/tasks.js
```diff
@@ -0,0 +1,75 @@
...
+    if (status === true) return
+    onceWithCleanup(bot.tasks.taskUpdate, 'update', {
+      checkCondition: () => {
+        const { status } = taskCompatibleWithRunningTask(taskIdentifier)
+        return status !== false
+      }
+    })
+  }
+
+  bot.taskCompatibleWithRunningTask = taskCompatibleWithRunningTask
```
rom1504, line 71: If you want to expose that they should be in the doc at API.md

u9g:
Will this effect usage by the end user? If so, please add a new example to the examples folder of how to use this.

IceTank:
> Will this effect usage by the end user? If so, please add a new example to the examples folder of how to use this.

Depends on how strict you want to make the task conflict configuration. Things like inventory actions should not be done at the same time as it can break both actions. But things like moving while fishing for instance can be more relax. I can add an example to how to deal with actions blocking you from doing certain things. 

This will probably be a difficult change for everyone that is not too familiar with how js or mineflayer works. As you can quickly mess up and have things try and run in parallel making functions suddenly throw errors when they worked fine before. So maybe only activate this when something like a strict option is set? What do you guys think @u9g  @rom1504 ?

IceTank:
I am not sure if the current implementation is working. I will add some tests and examples to test it.

rom1504:
failing and old

PR #2295, opened 2021-11-02
Fix bot not swinging arm on block place

IceTank:
placeEntity has its own armSwing call so placeBlock has to have its own arm swing function call too. The only issue is that I am pretty sure that arm swing animation packets are send before block placement packets. Some servers reject block placement if arm swing animations are not send. It seams like with my testing that the server should not be able to tell the difference between arm swing packets being send before or after block placement packets. But this may cause issues as it should be easily detectable by servers. A real fix should be giving _genericPlace an extra option to specify what arm to swing if block placement should be valid just before sending block placement packets.

### DIFF
--- a/lib/plugins/generic_place.js
+++ b/lib/plugins/generic_place.js
@@ -7,7 +7,7 @@ function inject (bot, { version }) {
    *
    * @param {import('prismarine-block').Block} referenceBlock
    * @param {import('vec3').Vec3} faceVector
-   * @param {{half?: 'top'|'bottom', delta?: import('vec3').Vec3, forceLook?: boolean | 'ignore', offhand?: boolean}} options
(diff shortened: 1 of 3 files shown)


rom1504:
>  A real fix should be giving _genericPlace an extra option to specify what arm to swing if block placement should be valid just before sending block placement packets.

let's do that?

IceTank:
I updated it so _genericPlace as an option for arm swinging. But bot.swingArm('right') currently swings the left arm and bot.swingArm('left') swings the right arm.

IceTank:
> I updated it so _genericPlace as an option for arm swinging. But bot.swingArm('right') currently swings the left arm and bot.swingArm('left') swings the right arm.

This has since been fixed with #2311

PR #2298, opened 2021-11-06
increase position packet accuracy

U5B:
1. Changed position packet time to be more accurate, as it uses the same time as the timer

2. lastSent.time resets whenever a position or position_look packet is sent (matches vanilla behavior)

3. vanilla sends position packets when "dead" but only for  it is moving / has velocity for a full second and then stops

4. on login, shouldUsePhysics is set to false until a position packet is received

5. Changed time (for sending a postion packet every second) from 1000 to 998, as in testing, position packets sent by the bot could be 50ms delayed if the number was 999.99 or 998.99

Probably fixes https://github.com/PrismarineJS/mineflayer/issues/1582

### DIFF
--- a/lib/plugins/physics.js
+++ b/lib/plugins/physics.js
@@ -121,18 +121,16 @@ function inject (bot, { physicsEnabled }) {
     bot.emit('move', oldPos)
   }
 
-  function lastSentPosition (position, onGround, now) {
+  function lastSentPosition (position, now) {
     lastSent.x = position.x
     lastSent.y = position.y
     lastSent.z = position.z
-    lastSent.onGround = onGround
     lastSent.time = now
   }
 
-  function lastSentLook (yaw, pitch, onGround) {
+  function lastSentLook (yaw, pitch) {
     lastSent.yaw = yaw
     lastSent.pitch = pitch
-    lastSent.onGround = onGround
   }
 
   function deltaYaw (yaw1, yaw2) {
@@ -146,7 +144,7 @@ function inject (bot, { physicsEnabled }) {
   function updatePosition (now) {
     // If you're dead, you're probably on the ground though ...
     // You can be dead in the air however
-    if (!isAlive) bot.entity.onGround = true
+    // if (!bot.isAlive) bot.entity.onGround = true
 
     // Increment the yaw in baby steps so that notchian clients (not the server) can keep up.
     const dYaw = deltaYaw(bot.entity.yaw, lastSentYaw)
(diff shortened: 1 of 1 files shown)


rom1504:
how did you test this?

U5B:
> how did you test this?

I checked by looking at vanilla client position packets in pakkit.
Vanilla client sends them every 50ms 
https://g.usbwire.net/i/🖼usb-204..png

However the interval is reset after every position and position look sent by the client
(Tested by jumping and looking around in the vanilla client)
https://g.usbwire.net/i/🎄usb-205..png

I also used pakkit to see how many position packets were sent after the vanilla client was dead if it was falling.
(Teleported the vanilla client to y 500 and /killed it while it was falling)
https://g.usbwire.net/i/✏️usb-206..png
After being killed, it was 20 ticks before it stopped sending any more position packets until it was respawned.

Changing the position timer check from 1000 to 998 was just a nitpick
However sometimes the timer would run faster and you would get 999.9 instead of 1000
Since updatePosition() is being run at 50ms, we should accept values that 1000 by a tiny bit
https://g.usbwire.net/i/🍌usb-207..png

U5B:
I decided to use rounding instead, which rounds to the nearest 50ms

Did some testing with different setInterval timings

For every 5 position packets, the time difference increments by +1ms with a setInterval loop of 50ms
For every 3 position packets, the time difference increments by +1ms with a setInterval loop of 25ms (how???)
For every 6 position packets, the time difference increments by -1ms and immediately after +1ms with a setInterval loop of 10ms
No incrementation with a setInterval loop of 5ms or below

However I don't know if you want to change the setInterval delay.

rom1504:
Can you say exactly what is the goal?

rom1504:
If the objective is to bypass some anti cheat test, did you test that this manages to bypass such tests and which one ?

U5B:
> Can you say exactly what is the goal?

The original goal was to fix Mineflayer skipping ticks by 50ms which it did.
https://g.usbwire.net/i/%F0%9F%98%A9usb-208..png

The anticheat I tested with was https://github.com/Leymooo/BungeeCord/

U5B:
Found a new botfilter to test https://github.com/Elytrium/LimboAPI

rom1504:
Can you describe how you test this ?

rom1504:
Is there any way to automate it ?

U5B:
I haven't completed testing it. I don't know exactly how you would test it unless there was an event that emitted every time a packet was written

U5B:
Going to remake this later. Finals are coming up and the commits are unreadable.

PR #2350, opened 2021-12-20
Fix regex dos

IceTank:
I think this should fix the regex dos. I am not sure if all types of whispers and chat messages still work. Might have to see if people report issues.


rom1504:
let's do that first https://github.com/PrismarineJS/prismarine-chat/issues/51

O-of:
Seems to fail on the cases where prefix is longer than 5 characters https://regex101.com/r/OhOJIM/5

PR #2357, opened 2021-12-23
Add reason for the end of a mineflayer bot

U5B:
This adds the parameter `reason` to bot.end() and the event `end`.

node-minecraft-protocol emits a default reason of `SocketClosed`.

If we start adding reasons for why the connection closed, then it will make future debugging easier.


U5B:
This also fixes bot.quit(reason) doing nothing, as the reason from bot.quit would not go anywhere.

TheDudeFromCI (approved):

u9g:
maybe say something about the default disconnect string?

PR #2367, opened 2021-12-27
fix bug with force lastSentPitch in bot.look

KadaverBrutalo10:
When doing:
```js
await bot.lookAt(chest, true)
await bot.openChest(chest)
```
it does not open the chest, it only looks at it.
it seems to have been some desync because in the if check it only sets 'lastSentYaw' but not 'lastSentPitch', my change fixes this.


amoraschi:
Isn't openChest deprecated?

PR #2378, opened 2022-01-02
Fix arm_animation and use_entity

amoraschi:
This should fix the "Sword doesn't wait for cooldown" error

From 1.8 to 1.8.9, Minecraft sends the arm_animation packet before the use_entity packet, but on 1.9+, Minecraft sends the use_entity packet before the arm_animation packet

### DIFF
--- a/lib/features.json
+++ b/lib/features.json
@@ -318,5 +318,10 @@
     "name": "setSlotAsTransaction",
     "description": "use setslot as transaction instead of just hoping it'll work",
     "versions": ["1.17", "1.17.1"]
+  },
+  {
+    "name": "armAnimationBefore",
+    "description": "arm animation packet sent before use entity packet",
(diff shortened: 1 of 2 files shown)


amoraschi:
Probably sould've used the "hasAttackCooldown" feature

lib/features.json
```diff
@@ -318,5 +318,10 @@
     "name": "setSlotAsTransaction",
     "description": "use setslot as transaction instead of just hoping it'll work",
     "versions": ["1.17", "1.17.1"]
+  },
+  {
+    "name": "armAnimationBefore",
```
IceTank, line 323: A bit off an nitpick but maybe change armAnimationBefore to maybe `armAnimationBeforeUse` to make it more descriptive.

amoraschi: Will do 👍

PR #2381, opened 2022-01-03
(feat): add option to `FindBlockOptions` to *not* sort all blocks

matteopolak:
Title says it all. The default `Array.prototype.sort` function is not very fast, and using it by default with no option to disable it is not ideal.

For example,
```javascript
const small = Array.from({ length: 1000 }, Math.random);
const medium = Array.from({ length: 100000 }, Math.random);
const large = Array.from({ length: 10000000 }, Math.random);

function sortAndTime(array, name) {
	console.time(name);
	array.sort((a, b) => b - a);
	console.timeEnd(name);
}

sortAndTime(small, 'small');
sortAndTime(medium, 'medium');
sortAndTime(large, 'large');
```

```console
small: 0.37ms
medium: 45.348ms
large: 8.632s
```

When searching for millions of blocks (e.g. digging only stone out of multiple chunks), the `findBlocks` function is very slow

### DIFF
--- a/lib/plugins/blocks.js
+++ b/lib/plugins/blocks.js
@@ -228,9 +228,9 @@ function inject (bot, { version, storageBuilder }) {
       next = it.next()
     }
     if (options.sort) {
-        blocks.sort((a, b) => {
-            return a.distanceTo(point) - b.distanceTo(point)
-        })
+      blocks.sort((a, b) => {
+        return a.distanceTo(point) - b.distanceTo(point)
+      })
     }
     // We found more blocks than needed, shorten the array to not confuse people
     if (blocks.length > count) {
--- a/lib/plugins/blocks.js
+++ b/lib/plugins/blocks.js
@@ -227,11 +227,11 @@ function inject (bot, { version, storageBuilder }) {
       startedLayer = it.apothem
       next = it.next()
     }
-		if (options.sort) {
-			blocks.sort((a, b) => {
-				return a.distanceTo(point) - b.distanceTo(point)
-			})
-		}
+    if (options.sort) {
+        blocks.sort((a, b) => {
+            return a.distanceTo(point) - b.distanceTo(point)
+        })
+    }
     // We found more blocks than needed, shorten the array to not confuse people
     if (blocks.length > count) {
       blocks = blocks.slice(0, count)
--- a/lib/plugins/blocks.js
+++ b/lib/plugins/blocks.js
@@ -227,9 +227,11 @@ function inject (bot, { version, storageBuilder }) {
       startedLayer = it.apothem
       next = it.next()
     }
-    blocks.sort((a, b) => {
-      return a.distanceTo(point) - b.distanceTo(point)
-    })
+		if (options.sort) {
+			blocks.sort((a, b) => {
+				return a.distanceTo(point) - b.distanceTo(point)
+			})
+		}
     // We found more blocks than needed, shorten the array to not confuse people
     if (blocks.length > count) {
       blocks = blocks.slice(0, count)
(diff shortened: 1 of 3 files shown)


Karang:
I have nothing against this PR. But if you want to search for more than 100 blocks, findBlocks isn't the best option anyway and you should instead use the block iterators directly.

matteopolak:
Ideally you could avoid sorting altogether by searching in a spherical pattern around the bot, making them already ordered by proximity. Might be too much of a hassle for the time saved for most applications though.

rom1504:
Can you clarify your use case ?
Is there a good use case to use find block for million of results instead of just filtering the whole chunk sequentially ?

Karang:
> Ideally you could avoid sorting altogether by searching in a spherical pattern around the bot, making them already ordered by proximity. Might be too much of a hassle for the time saved for most applications though.

No that would be slower. Findblock is optimized to check many blocks but return few. If you need the opposite you should make a custom function.

See https://github.com/PrismarineJS/mineflayer/pull/1035#issue-632536088 for the full writeup.

matteopolak:
I think the more relevant use-case for me is that I *don't* need returned blocks to be ordered by their distance from the player.

For example, I've been counting the # of cactus blocks in a cactus farm, which doesn't need sorted data.

The sorting of blocks just adds extra overhead in the worst case-scenario, and saves a single sorting step in the best case-scenario, assuming the user wants to sort only by the distance from the player. With this presumption, it doesn't allow someone to sort from another point (for example, the destination of the bot once they traverse some terrain), as you would need to manually sort the already-sorted-but-in-a-different-way data.

Karang:
> I think the more relevant use-case for me is that I _don't_ need returned blocks to be ordered by their distance from the player.
> 
> For example, I've been counting the # of cactus blocks in a cactus farm, which doesn't need sorted data.
> 
> The sorting of blocks just adds extra overhead in the worst case-scenario, and saves a single sorting step in the best case-scenario, assuming the user wants to sort only by the distance from the player. With this presumption, it doesn't allow someone to sort from another point (for example, the destination of the bot once they traverse some terrain), as you would need to manually sort the already-sorted-but-in-a-different-way data.

If your use case is counting, then you shouldn't use find block at all. The same argument as sorting overhead could be applied to the overhead of pushing blocks positions to a list (and allocating lots of temporary objects you'll never gonna use) when you only really need a single counter. A much more optimized version would be to loop on the sections in a cuboid, skip empty sections and sections which doesn't contain your block, count the blocks by looping in the sections that contain them. So you really don't need findBlocks for that, but we could accept a PR for a countBlocks function.

If you need sorting from an other position, findBlocks has a point parameter you can give https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/blocks.js#L193 which is exactly meant for that (because it's not only changing the comparator function but also the order sections are processed).

lib/plugins/blocks.js
```diff
@@ -227,9 +227,11 @@ function inject (bot, { version, storageBuilder }) {
...
-    blocks.sort((a, b) => {
-      return a.distanceTo(point) - b.distanceTo(point)
-    })
+    if (options.sort) {
+      blocks.sort((a, b) => {
+        return a.distanceTo(point) - b.distanceTo(point)
+      })
+    }
     // We found more blocks than needed, shorten the array to not confuse people
     if (blocks.length > count) {
```
rom1504, line 236: if options.sort is false this will break this slicing

rom1504:
As discussed above, there is no use case to add such a no sort option.
can you instead open a PR explaining not to use find block for your use cases and how to solve your use cases more efficiently? (explained how above)

PR #2418, opened 2022-01-22
Improved crafting algorithm

SaubereSache:
Simpified crafting code which improves crafting speed drastically

Video:
[Old Implementation](https://cdn.discordapp.com/attachments/413438150594265099/934446359392690246/javaw_iyBckBPPYl.mp4)
[New Implementation](https://cdn.discordapp.com/attachments/413438150594265099/934447741638160424/javaw_e5uAC0ntCy.mp4)


SaubereSache:
Currently the issue is a [rejected click transaction](https://github.com/PrismarineJS/mineflayer/blob/f3005df1256ee55d32d8081beb5df886a371c249/lib/plugins/craft.js#L133).
After crafting, the code attempts to cleanup any leftovers from the crafting area (basically only applicable for cake which leaves empty milk buckets), but mineflayer does not update the slots to being empty after crafting finished.
The old code implemented a workaround which updated the slots to the out shape manually, but I don't feel like that's something we should do if the protocol allows otherwise.

Any input is welcome :)

SaubereSache:
Server sends a window_items packet which seems to be ignored by mineflayer...

image

nickelpro:
Excellent work on this and congrats on test passing, I'm going to take a minute to review before giving my :heavy_check_mark:

lib/plugins/craft.js
```diff
@@ -8,184 +8,164 @@ function inject (bot, { version }) {
   const Item = require('prismarine-item')(version)
   const Recipe = require('prismarine-recipe')(version).Recipe
 
+  const craftingTableSlots = [1, 2, 3, 4, 5, 6, 7, 8, 9]
+  const inventoryCraftingSlots = [1, 2, 3, 4]
+  const resultSlot = 0
+
   async function craft (recipe, count, craftingTable) {
```
nickelpro, line 15: API.md says we accept a callback here, and clearly that isn't true and wasn't true before your commit. Can you update API.md to reflect that?

nickelpro (approved):
Approved for the actual code. I think we should add a cake test ~~and update the API.md for `bot.craft()`~~ but that's not blocking for me personally

cc @rom1504

lib/plugins/craft.js
SaubereSache: Actually, we do: ``bot.craft = callbackify(craft)``
But if it was my decision I'd just yeet callbacks altogether ^^

nickelpro: Oh derp, of course, we went over this already. Disregard my continued ignorance about how the hell this codebase works

SaubereSache:
@nickelpro As the cake test currently won't pass (missing outShape data in prismarine-recipe) I'll put it into a seperate PR ^^

nickelpro:
Give me two seconds, I'll update mcd and we can just do the right thing in one fell swoop

lib/plugins/craft.js
```diff
@@ -8,184 +8,166 @@ function inject (bot, { version }) {
...
     assert.ok(recipe)
-    count = parseInt(count ?? 1, 10)
```
rom1504, line 13: Can you add back this ?

index.d.ts
```diff
@@ -332,8 +332,8 @@ export interface Bot extends TypedEmitter<BotEvents> {
 
   craft: (
     recipe: Recipe,
-    count: number | null,
-    craftingTable: Block,
+    count?: number,
```
rom1504, line 335: Let's undo these changes

SaubereSache: Even craftingTable? It is indeed optional ^^

test/externalTests/crafting.js
```diff
@@ -54,14 +54,31 @@ module.exports = () => async (bot) => {
   bot.chat('/give @p stick 21')
   await once(bot.inventory, 'updateSlot')
   const craftingTable = bot.findBlock({ matching: blocksByName.crafting_table.id })
-  await bot.craft(bot.recipesFor(itemsByName.ladder.id, null, null, true)[0], 9, craftingTable)
-  if (bot.inventory.count(itemsByName.ladder.id, null) !== 9) { throw new Error('CraftLadder: count mismatch') }
 
-  bot.chat('/give @p milk_bucket 6')
+  const recipe = bot.recipesFor(itemsByName.ladder.id, null, null, true)[0]
+  await bot.craft(recipe, recipe.result.count * 3, craftingTable)
+  await bot.waitForTicks(5)
```
nickelpro, line 60: I don't like the implication of this, what `bot.craft` returns the inventory transactions should be complete

SaubereSache: Me neither, only happens in 1.17+. Needs more investigation :/

index.d.ts
rom1504: Ok yeah i guess that's fine to keep then

SaubereSache:
``npm run mocha_test -- -g "1.18.1.*craft"``:
4/4 of my debug runs of  were successful, 0/4 without debugger. 

Seems like this is a bigger Inventory related issue, how shall we proceed with this PR?

lib/plugins/craft.js
```diff
@@ -8,184 +8,167 @@ function inject (bot, { version }) {
   const Item = require('prismarine-item')(version)
   const Recipe = require('prismarine-recipe')(version).Recipe
 
+  const craftingTableSlots = [1, 2, 3, 4, 5, 6, 7, 8, 9]
+  const inventoryCraftingSlots = [1, 2, 3, 4]
+  const resultSlot = 0
+
   async function craft (recipe, count, craftingTable) {
```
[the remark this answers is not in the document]
rom1504: yes, please keep the same meaning as before

rom1504:
https://github.com/PrismarineJS/prismarine-recipe/issues/10

SaubereSache:
The issue with click window in versions >=1.17.1 is the following
https://github.com/PrismarineJS/mineflayer/blob/52650483a80412cf766c09256d007ab39d9ad50c/lib/plugins/inventory.js#L551
https://github.com/PrismarineJS/mineflayer/blob/52650483a80412cf766c09256d007ab39d9ad50c/lib/plugins/inventory.js#L410

rom1504:
I'm afraid this is still failing tests

PR #2435, opened 2022-01-29
Update typings - breaking change in dep 1.0 -> 2.0

Eagle-Anne:
Per https://github.com/andywer/typed-emitter/issues/28 the interfaces must be types or it will throw a type error.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -39,7 +39,7 @@ export interface PluginOptions {
 
 export type Plugin = (bot: Bot, options: BotOptions) => void
 
-interface BotEvents {
+type BotEvents = {
   chat: (
     username: string,
     message: string,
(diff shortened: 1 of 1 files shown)


extremeheat:
I think we might as well go back to v1 or switch the dependency here. If we merged this then we would need to either swap the linter or add a bunch of eslint-ignores in all of the updated places. I don't think this is a good idea.

My concerns with 2.0 of the package as discussed earlier otherwise are that their breaking change doesn't adequately explain the rationale for the change (aside from just "switch to constrained types")[1]. This update also adds an extra 4MB dependency to their package.

[1] The PR you mentioned earlier in this package, https://github.com/andywer/typed-emitter/issues/22 mentions that `rawListeners` and `listeners` are untyped. It seems that https://github.com/binier/tiny-typed-emitter/blob/master/lib/index.d.ts does not have this issue, and also works with the existing interfaces.

rom1504:
As a result of this PR and conversation let's:
* Add https://github.com/SamVerschueren/tsd to check types work in CI
* Revert to 1.0 of this dependency
* Open an issue in typed emitter to get this breaking change reverted or to have a rationale as to why it's worth it

Would be appreciated if you were to handle it @Eagle-Anne

PR #2438, opened 2022-01-30
Use prismarine-chunk for block entities 

extremeheat:
should be moved to prismarine-world soon, and prismarine-block should instead have the abstractions related to signs


lib/plugins/blocks.js
```diff
@@ -583,7 +593,14 @@ function inject (bot, { version, storageBuilder }) {
   bot.canSeeBlock = canSeeBlock
   bot.blockAt = blockAt
   bot._updateBlockState = updateBlockState
-  bot._blockEntities = blockEntities
+  Object.defineProperty(bot, '_blockEntities', {
```
rom1504, line 596: you can just remove that

lib/plugins/blocks.js
```diff
@@ -451,19 +447,33 @@ function inject (bot, { version, storageBuilder }) {
...
-      prepareString(4)
-    ].join('\n')
+    const blockAt = column.getBlock(pos)
+    if (!blockAt.entity) {
+      blockAt.entity = nbt.comp({
+        id: nbt.string('minecraft:sign')
+      })
+    }
+
+    blockAt.entity.value.Text1 = nbt.string(prepareString(1))
```
rom1504, line 457: can we put that in pchunk/pworld maybe ?

lib/plugins/blocks.js
```diff
@@ -293,9 +283,6 @@ function inject (bot, { version, storageBuilder }) {
     }
     if (oldBlock.type !== newBlock.type) {
       const pos = point.floored()
-      delete blockEntities[pos]
-      delete signs[pos]
-
       const painting = paintingsByPos[pos]
```
rom1504, line 286: we can probably make paintings not be in blockAt

lib/plugins/blocks.js
```diff
@@ -254,9 +238,15 @@ function inject (bot, { version, storageBuilder }) {
...
 
     if (extraInfos) {
-      block.signText = signs[block.position]
       block.painting = paintingsByPos[block.position]
-      block.blockEntity = blockEntities[block.position]
+      if (block.entity) {
+        const id = block.entity.value.id.value
+        // Mege in abstractions related to signs
+        Object.assign(block.blockEntity, addExtraBlockEntityData(id, block.entity))
+        if (id === 'minecraft:sign' || id === 'Sign') {
```
rom1504, line 246: this should be in pchunk/pworld

this is basically a more high level way to represent sign and there could be something like that for other block entities in the future

rom1504:
https://github.com/PrismarineJS/mineflayer/pull/2441 this may help

extremeheat:
Yeah, the changes to blocks.js in #2441 would fix the issue here with quotation marks. Not sure what's going on with the examples part of the PR

rom1504:
can you maybe just pick the changes you like from that PR and put them here so we can merge this one?

lib/plugins/blocks.js
```diff
@@ -335,11 +288,13 @@ function inject (bot, { version, storageBuilder }) {
     })
 
     if (typeof packet.blockEntities !== 'undefined') {
+      const column = bot.world.getColumn(packet.x, packet.z)
```
rom1504, line 291: why not use chunk.loadBlockEntities ?

PR #2441, opened 2022-01-31
Sign example and implementation fix

IceTank:
There are some things that did not work with the sign example and the sign implementation.
The example was broken and did not work for all versions. It also had references to painting blocks (?) that do not seam to exist anymore.
The updateSign function had an issue where it would stringify the sign text in versions 1.9 and above. But sign text is send as a string after 1.8 so empty lines showed up as `""` on a sign. updateSign also had an issue where it would brick node-minecraft-protocol by trying to send sign text lines that where undefined.


IceTank:
Someone could also remake the sign tests. But I don't have a lot of time right now.

examples/graffiti.js
```diff
@@ -32,37 +38,85 @@ bot.on('chat', (username, message) => {
...
+  }))
+  if (!freeBlock) {
+    bot.chat('No free block to place a sign on')
+    return
+  }
+
+  const signEditPromise = (async () => {
+    // Wait for the right event to fire that indicates that we have opened our sign gui
+    // If we update the sign text before this event fires the server might reject the sign text
+    for await (const event of on(bot._client, 'packet')) {
```
extremeheat, line 92: this is not good

rom1504:
can you please rebase on master ?

examples/graffiti.js
IceTank: Can you be more specific? What is not good about it?

extremeheat: User code shouldn't be using ._client or listening to packets

IceTank: Can you be more specific about what is not good about it?

rom1504: Indeed

IceTank: I changed it in the last commit

rom1504:
So this removes painting completely from all examples
I think we should have an example with paintings, maybe another one than this one

IceTank:
todo: Make example simpler

lib/plugins/blocks.js
```diff
@@ -419,6 +423,11 @@ function inject (bot, { version, storageBuilder }) {
     }
   })
 
+  // Emit open_sign event
+  bot._client.on('open_sign_entity', (data) => {
+    bot.emit('signOpen', data)
```
u9g, line 428: this isn't good, we should abstract over the packet

test/externalTests/sign.js
```diff
@@ -1,48 +1,104 @@
...
+          assert.strictEqual(sign.signText, '1\n2\n3\n')
 
-        if (sign.blockEntity && sign.blockEntity.Text1) {
-          assert.strictEqual(sign.blockEntity.Text1.toString(), '1')
-          assert.strictEqual(sign.blockEntity.Text2.toString(), '2')
-          assert.strictEqual(sign.blockEntity.Text3.toString(), '3')
-          assert.strictEqual(sign.blockEntity.Text4.toString(), '')
-        }
+          if (sign.blockEntity && sign.blockEntity.Text1) {
+            assert.strictEqual(sign.blockEntity.Text1.toString(), '1')
```
u9g, line 40: is there any way we can abstract over this? this is way too direct to have to use the sign's blockentity nbt.

lib/plugins/blocks.js
IceTank: I don't know what you mean. How should this be abstracted more?

u9g: I have no idea what's in this packet, and neither will people updating mineflayer for future updates, can you atleast destructure the packet in the arg, then put it back into an object when being emitted so people reading this code will know what to expect here

IceTank: Ok abstracted the location away to a position argument

test/externalTests/sign.js
IceTank: This test should be unnecessary anyway as we are check the sign.signText property

extremeheat (changes requested):

lib/plugins/blocks.js
```diff
@@ -454,6 +463,54 @@ function inject (bot, { version, storageBuilder }) {
...
+  /**
+   * Places a sign and writes to it
+   * @throws {Error} Throws error when no sign is equipped or the position to place the sign is is already occupied
+   * @param {import('vec3').Vec3} pos
+   * @param {string} text
+   * @param { { sneak?: boolean, writeDelay?: number } } options
+   * @returns {Promise<void>}
+   */
+  bot.placeSign = async (pos, text, options = {}) => {
+    if (!(pos instanceof Vec3)) throw new Error('Invalid argument for pos. Must be a instance of Vec3 got ' + typeof pos)
```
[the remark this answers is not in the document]
IceTank: I think more checks are never wrong if they do not interfere with performance or other things. I do not expect people to run placeSign hundreds off times a second so performance should not be an issue. But I can also remove that check again.

lib/plugins/blocks.js
```diff
@@ -454,6 +463,54 @@ function inject (bot, { version, storageBuilder }) {
...
+    if (!bot.heldItem || (bot.heldItem && !bot.heldItem.name.includes('sign'))) throw new Error('Need a sign equipped to place sign')
+    const signBlock = bot.blockAt(pos)
+    if (!signBlock || !signBlock?.position || (signBlock && !signBlock.name.includes('air'))) throw new Error('Invalid sign position')
```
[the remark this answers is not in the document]
IceTank: Yeah I don't think that is possible. Ill remove it.

lib/plugins/blocks.js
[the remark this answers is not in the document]
extremeheat: Yes, I understand that. I meant from a consistency standpoint to have clear code. If it's a common problem it may make sense, otherwise it sounds like noise to me if some have these checks and some don't

examples/graffiti.js
```diff
@@ -32,37 +37,90 @@ bot.on('chat', (username, message) => {
...
     bot.chat('There are no signs or paintings near me')
   }
 }
 
+async function placeSign (message) {
+  message = message.split(' ').slice(1).join(' ') // Remove the first word. ie place
+  // Split the message along its 15 char line limit
+  const messageArr = []
+  for (let i = 0; i < 4; i++) {
+    if (message.length > 15) {
```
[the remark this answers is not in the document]
IceTank: I think minecraft splits along chars and sometimes width. But from what I can tell most servers do not check the text real font width to block text that is to long. I can still implement this in the updateSign code.

lib/plugins/blocks.js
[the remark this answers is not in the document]
IceTank: I feel like a lot off problems for people that are new to mineflayer come from sometimes really obscure error messages. Catching the really obvious stuff should probably be implemented for every api facing function where it dose not hinder performance.

IceTank:
Should I make updateSign and place sign take in an array off strings for the sign text lines? I feel like this would a lot more sense then having to put newline characters into the sign.

lib/plugins/blocks.js
```diff
@@ -454,6 +463,55 @@ function inject (bot, { version, storageBuilder }) {
...
+    const directions = [new Vec3(1, 0, 0), new Vec3(0, 0, 1), new Vec3(-1, 0, 0), new Vec3(0, 0, -1)]
+    for (const dir of directions) {
+      const tmp = bot.blockAt(pos.minus(dir))
+      if (tmp?.boundingBox === 'block') {
+        wallPlacement = dir
+        break
+      }
+    }
+
+    if (!groundPlacable && wallPlacement === null) throw new Error('Cannot place block in position ' + pos.toString())
```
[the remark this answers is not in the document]
IceTank: This is suppose the check for potential wall placement spots. If it dose not find a ground placement block or any off the wall placement blocks it will throw an error. There is not way to specify where you want to place the sign. The function looks for a spot for you. I can maybe add a allowWallPlacement and allowFloorPlacement option.

lib/plugins/blocks.js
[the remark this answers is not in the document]
extremeheat: Perhaps a discussion for another time, there are type checkers for this purpose

lib/plugins/blocks.js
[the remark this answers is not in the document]
extremeheat: I think it should be explicit and well defined here. Maybe seperate functions to place a wall or standing sign would make things clear, which can prefill options to placeSign. A bit like currently placing blocks

IceTank:
I am not sure if the current implementation is working. I will add some tests and examples to test it.

MrAwesome:
For what it's worth, as an end user of the library, an array of strings seems more user-friendly and less brittle.

IceTank:
This should be done. Waiting for someone to merge or suggest changes.

rom1504:
can you rebase and push to retrigger ci ?

extremeheat:
Did you delete the commits ? What happened to the bot.placeSign function?

rom1504:
closed with label "was good"

Maybe someone will pick it up someday

PR #2443, opened 2022-02-02
Update physics.js

IsaacJuracich:
Sneak function


extremeheat:
Why not set the control state?

PR #2453, opened 2022-02-04
use bot.world.getBlock instead of bot.blockAt

rom1504:


rom1504:
#334

rom1504:
blocked my moving painting to pchunk

rom1504:
old

PR #2475, opened 2022-02-09
fix titles and actionbar for new versions

KalmeMarq:


KalmeMarq:
but what if a subtitle isn't emitted? And they are different packets now so how do I put them together anyway?

KalmeMarq:
but how does the person know which is which?

KalmeMarq:
> Do they care that there is a difference? But you do have a point. Correct me if I am wrong, but in vanilla, there isn't a difference when the server sends you a title and a subtitle, so that is probably why it was included together
> 
> You are welcome to add support for the other packet actions for titles.

but if it didn't have a difference how would the server know what to send to the other players? it must have a difference somewhere cuz it even has diff packets.

extremeheat:
The minecraft protocols they use are not efficient (or stable), so making abstractions that maximize future compatibility is important. See how boss bars are currently handled for an example

rom1504:
@KalmeMarq want to finish this?

rom1504:
too old, please reopen if you want to finish

Issue #2481, opened 2022-02-11
Bots should be able to deny server listings


haykam821:
## Is your feature request related to a problem? Please describe.

Minecraft 1.18 adds the online option 'Allow Server Listings'. Much like skin settings, this setting is sent to the server to determine whether the player should show up in server listings. However, it cannot be controlled from a bot in the same way as skin settings.

## Describe the solution you'd like

The client should support the `allowServerListings` setting.

U5B:
Guess we do need to update the settings packet for that.
The two new setting options are two booleans
`enableServerListing` for 1.18+
`enableTextFiltering` for 1.17+

I might do it later if no one else does.

PR #2482, opened 2022-02-11
settings.js imporvements

U5B:
add support for the two new options
`enableTextFiltering` boolean added in 1.17: Unused, but is set to false by default in vanilla
`enableServerListing` boolean added in 1.18: Added because of some griefing incidents, allows you to stop showing yourself in the ping for server lists.

Fixes https://github.com/PrismarineJS/mineflayer/issues/2481


U5B:
Also adds support for using a positive number to set the view distance.

rom1504:
Please update doc

PR #2491, opened 2022-02-23
Add warnings for calling properties on bot too early

u9g:
Example code:
```js

  host: '95.111.249.143',
  port: 10000,
  username: 'we_win_these'
})

console.log(bot.entities)
```
Output: 
image


extremeheat:
Proxies are slow, this will add a lot of unnecessary overhead to every call to bot functions. Instead adding them for specific properties and throwing an exception on improper access would be a better approach.

Issue #2402, opened 2022-01-12
Bypass bot check


Plyrs1:
Mineflayer Version : 3.14.1
Minecraft Server Version : 1.16.5
Node.js Version : 16.10.0

## Is your feature request related to a problem? Please describe.

Bot passes the first stage of the bot check, which is a falling test (well, this is only my assumption because when I use Minecraft Console Client, the bot kicked immediately), but I don't know why the bot always got kicked after that.
I compared position packets between Vanilla Minecraft 1.16.5, Mineflayer, and MCC, and here are the results :
<details>
<summary>Vanilla Minecraft 1.16.5</summary>
<p>

A fall test, which puts ground on weird Y position, in this case, it's Y 4.25
![image](https://user-images.githubusercontent.com/97564087/149038656-3bbdd490-c0e0-4b58-8f3d-3b2cee671617.png)
![image](https://user-images.githubusercontent.com/97564087/149038775-37451974-621f-4561-bc89-78e8e75e63bf.png)
and that respawn packet at the bottom indicates that the client completed bot check and respawned into the hub
(actually, the 1st chat packet in the 2nd image basically tells the user that they passed the bot checks, and the 2nd chat packet is a message that the user have been connected to the hub)
<details>
<summary>1st chat packet</summary>
<p>

```json
{
  "message": "{\"extra\":[{\"bold\":true,\"color\":\"yellow\",\"text\":\"Blockcraft\"},{\"color\":\"light_purple\",\"text\":\" / \"},{\"color\":\"green\",\"text\":\"Beep!\"}],\"text\":\"\"}",
  "position": 0,
  "sender": "00000000-0000-0000-0000-000000000000"
}
```
</p>
</details>
<details>
<summary>2nd chat packet</summary>
<p>

```json
{
  "message": "{\"color\":\"green\",\"text\":\"Connected to HUB-2\"}",
  "position": 1,
  "sender": "00000000-0000-0000-0000-000000000000"
}
```
</p>
</details>

</p>
</details>

<details>
<summary>Mineflayer</summary>
<p>

here we can see when onGround is true, bots correctly sent the weird position just like vanilla minecraft, which is Y 4.25
(the fall starts at different Y in every connection, but when onGround = true, the Y is always 4.25)
![image](https://user-images.githubusercontent.com/97564087/149040127-33b015e1-0c95-4a58-b288-56b955b741f9.png)
[...]

PR #2492, opened 2022-02-23
Mineflayer physics refactor

U5B:
This PR will have some physics refactors to be more like vanilla client.

I will try to make each commit readable and add a comment explaining what each commit does.

Issues that PR does not fix:
lastSentYaw and lastSentPitch not be accounted for in prismarine-physics


U5B:
Description of each test:

`position_look should be sent 1000ms if we are looking around` - looks every tick, should send a position_look instead of a position packet if we are sending a look update

`absolute position & relative position (velocity)` reproducing vanilla client gimmicks
absolute position sets the velocity to 0
relative position keeps the current velocity
 
 `position is sent every 1000ms` - send a position packet every second if idle, no skipping ticks
 
 `falling check (position is sent every 50ms)` - self explanatory,  no skipping ticks

U5B:
https://github.com/PrismarineJS/mineflayer/pull/2492/commits/4e5fb70e2f9fe659e4841a473c139f0f222ce052
This is a big one.

The variable `positionUpdated` now checks if 1000ms has passed since the last position packet was sent
It rounds `lastSent.time` to the nearest 50ms when checking to ensure that it does not skip a tick even if lastSent.time is `999` or `998`

This changes the functions of `sendPacketPositionAndLook`, `sendPacketPosition` and `sendPacketLook` to not set `lastSent`.
Instead, `lastSent` is updated in `lastSentPosition` (if `positionUpdated` is true) and `lastSentLook` (if `lookUpdated` is true)
`lastSent.onGround` is always set (since it is always sent)

U5B:
Still need to figure out a reliable way for checking if a chunk is currently loaded where the bot is standing and a more reliable way to determine that the bot is "alive".

U5B:
https://github.com/PrismarineJS/mineflayer/issues/2402
Vanilla client only requires a login packet > map_chunk packet > position packet in order to start sending packets.

lib/plugins/physics.js
```diff
@@ -51,25 +53,36 @@ function inject (bot, { physicsEnabled }) {
...
+      tickPhysics(now)
       timeAccumulator -= PHYSICS_TIMESTEP
     }
   }
 
+  function tickPhysics (now) {
+    if (bot.physicsEnabled && shouldUsePhysics) {
+      physics.simulatePlayer(new PlayerState(bot, controlState), world).apply(bot)
+      bot.emit('physicsTick')
+      bot.emit('physicTick') // Deprecated, only exists to support old plugins. May be removed in the future
```
IceTank, line 79: You could add a event deprecation warning here. Check if there are listeners on bot with the name physicTick with [this](https://nodejs.org/api/events.html#emitterlistenercounteventname) and if there are print a deprecation warning once.

lib/plugins/physics.js
```diff
@@ -51,25 +53,36 @@ function inject (bot, { physicsEnabled }) {
...
+  // use WSL or switch to Linux
+  // see: https://discord.com/channels/413438066984747026/519952494768685086/901948718255833158
   let timeAccumulator = 0
   function doPhysics () {
     const now = performance.now()
-    const deltaSeconds = (now - lastPhysicsFrameTime) / 1000
+    let deltaSeconds = (now - lastPhysicsFrameTime) / 1000
     lastPhysicsFrameTime = now
 
+    if (deltaSeconds >= 0.2) deltaSeconds = 0.2 // limiter (limit catchups to 4 ticks behind)
```
IceTank, line 66: Maybe make this a variable instead of hard coding it.

lib/plugins/physics.js
```diff
@@ -78,35 +91,50 @@ function inject (bot, { physicsEnabled }) {
   function sendPacketPosition (position, onGround) {
     // sends data, no logic
     const oldPos = new Vec3(lastSent.x, lastSent.y, lastSent.z)
-    lastSent.x = position.x
-    lastSent.y = position.y
-    lastSent.z = position.z
-    lastSent.onGround = onGround
-    bot._client.write('position', lastSent)
+    const packet = {}
```
IceTank, line 94: Why change the properties on the packet object when you can just do 
```js
const packet = {
  x: position.x,
  y: position.y
  ...
}
```

lib/plugins/physics.js
```diff
@@ -117,17 +145,24 @@ function inject (bot, { physicsEnabled }) {
...
+    if (isDead()) return
 
     // Increment the yaw in baby steps so that notchian clients (not the server) can keep up.
     const dYaw = deltaYaw(bot.entity.yaw, lastSentYaw)
     const dPitch = bot.entity.pitch - (lastSentPitch || 0)
 
     // Vanilla doesn't clamp yaw, so we don't want to do it either
-    const maxDeltaYaw = dt * physics.yawSpeed
-    const maxDeltaPitch = dt * physics.pitchSpeed
+    const maxDeltaYaw = PHYSICS_TIMESTEP * physics.yawSpeed
```
IceTank, line 164: Why change dt for a constant? What if you want to run mineflayers physics faster? Is this still possible with this change?

lib/plugins/physics.js
```diff
@@ -276,19 +322,15 @@ function inject (bot, { physicsEnabled }) {
 
     if (bot.supportFeature('teleportUsesOwnPacket')) {
       bot._client.write('teleport_confirm', { teleportId: packet.teleportId })
-      // Force send an extra packet to be like vanilla client
-      sendPacketPositionAndLook(pos, newYaw, newPitch, bot.entity.onGround)
```
IceTank, line 280: Was this wrong or why was it removed?

lib/plugins/physics.js
u9g: or you can do 
```js
bot._client.write('position', 
  x: position.x,
  y: position.y
  ...
}
```

lib/plugins/physics.js
U5B: This was wrong.

lib/plugins/physics.js
U5B: dt was originally a constant.
This does not change behavior, you can simulate it faster if you decrease the values of PHYSICS_INTERVAL_MS and PHYSICS_SETINTERVAL_MS

lib/plugins/physics.js
U5B: > Why change the properties on the packet object when you can just do
> 
> ```js
> const packet = {
>   x: position.x,
>   y: position.y
>   ...
> }
> ```

I'll change it to this, because I believe it makes it more readable.

lib/plugins/physics.js
U5B: 'physicTick event is deprecated! Please use physicsTick instead'
Is this a fine message?

IceTank: yes

rom1504:
what's the status here ? is this ready ?

U5B:
> what's the status here ? is this ready ?

Waiting for more comments, since this is a pretty big change.
I also want to figure out a reliable way to not do physics until the bot's chunk is loaded.

U5B:
@Karang @nickelpro I would like for you to take a look at this.

IceTank:
> I also want to figure out a reliable way to for physics to wait until the bot's chunk is loaded like vanilla.

What wrong with `bot.waitForChunksToLoad()`?

U5B:
> > I also want to figure out a reliable way to for physics to wait until the bot's chunk is loaded like vanilla.
> 
> What's wrong with `bot.waitForChunksToLoad()`?

I just want to check if the chunk that the bot is in is loaded. `bot.waitForChunksToLoad()` has issues if the server doesn't send the other chunks.

IceTank:
> I just want to check if the chunk that the bot is in is loaded. bot.waitForChunksToLoad() has issues if the server doesn't send the other chunks.

Someone could refactor `waitForChunksToLoad()` to take a radius off chunks to wait for. Setting this to something like 0 or 1 would then make it possible to wait until the chunk the bot is standing in is loaded.

lib/plugins/physics.js
```diff
@@ -51,25 +54,42 @@ function inject (bot, { physicsEnabled }) {
...
+  function tickPhysics (now) {
+    if (bot.blockAt(bot.entity.position) == null) return // check if chunk is unloaded
+    if (bot.physicsEnabled && shouldUsePhysics) {
+      physics.simulatePlayer(new PlayerState(bot, controlState), world).apply(bot)
+      bot.emit('physicsTick')
+      bot.emit('physicTick') // Deprecated, only exists to support old plugins. May be removed in the future
+      if (bot.listenerCount('physicTick') > 0 && deprecatedEvent === false) { // remove when physicTick is removed
+        console.log('physicTick event is deprecated! Please use physicsTick instead')
+        deprecatedEvent = true
+      }
```
Karang, line 86: This is a hot loop, it's probably not the right place for such a check, a better place would be https://nodejs.org/api/events.html#event-newlistener

That being said, it's maybe time to remove the deprecated event...

lib/plugins/physics.js
```diff
@@ -12,7 +12,9 @@ module.exports = inject
 const PI = Math.PI
 const PI_2 = Math.PI * 2
 const PHYSICS_INTERVAL_MS = 50
-const PHYSICS_TIMESTEP = PHYSICS_INTERVAL_MS / 1000
+const PHYSICS_TIMESTEP = PHYSICS_INTERVAL_MS / 1000 // 0.05
+const PHYSICS_SETINTERVAL_MS = 10 // this runs the doPhysics() loop faster, not tickPhysics()
```
Karang, line 16: That really sounds like a horrible hack, you are running this loop 5x time faster than it needs to be. Is it still because of the inaccurate timers on windows ? If so it would be better to fix the problem at its source instead of trying to hack around it.

lib/plugins/physics.js
```diff
@@ -51,25 +54,42 @@ function inject (bot, { physicsEnabled }) {
...
+  // use WSL or switch to Linux
+  // see: https://discord.com/channels/413438066984747026/519952494768685086/901948718255833158
   let timeAccumulator = 0
   function doPhysics () {
     const now = performance.now()
-    const deltaSeconds = (now - lastPhysicsFrameTime) / 1000
+    let deltaSeconds = (now - lastPhysicsFrameTime) / 1000
     lastPhysicsFrameTime = now
 
+    if (deltaSeconds >= PHYSICS_CATCHUP_TICKS) deltaSeconds = PHYSICS_CATCHUP_TICKS // limiter (limit catchups)
```
Karang, line 67:
```js
deltaSeconds = Math.min(deltaSeconds, PHYSICS_CATCHUP_TICKS)
```

Karang: That is not the right place to limit the number of catchup ticks, you should add a counter to the while loop bellow and have the PHYSICS_CATCHUP_TICKS be in tick units. Otherwise you'll make the timeAccumulator have a wrong value and desync.

lib/plugins/physics.js
```diff
@@ -78,35 +98,54 @@ function inject (bot, { physicsEnabled }) {
...
-    lastSent.z = position.z
-    lastSent.onGround = onGround
-    bot._client.write('position', lastSent)
+    const packet = {
+      x: position.x,
+      y: position.y,
+      z: position.z,
+      onGround: onGround
+    }
+    packet.onGround = onGround
```
Karang, line 107: You are assigning onGround twice. And you are removing an optimization: lastSent was used to avoid creating too many temporary objects.

lib/plugins/physics.js
U5B: I could remove the deprecated event if you would like.

lib/plugins/physics.js
U5B: That was a my bad on the packet.onGround part.
I have changed it to use lastSent.
However lastSent.onGround is always updated in vanilla and lastSent.time is always reset when a position or a position_look packet is sent.

lib/plugins/physics.js
U5B: I am confused on where to put the counter and how to limit it properly.
```js
  let timeAccumulator = 0
  let catchupTicks = 0
  function doPhysics () {
    const now = performance.now()
    let deltaSeconds = (now - lastPhysicsFrameTime) / 1000
    lastPhysicsFrameTime = now

    timeAccumulator += deltaSeconds
    catchupTicks = 0
    while (timeAccumulator >= PHYSICS_TIMESTEP) {
      tickPhysics(now)
      timeAccumulator -= PHYSICS_TIMESTEP
      catchupTicks++
      if (catchupTicks >= PHYSICS_CATCHUP_TICKS) timeAccumulator -= PHYSICS_TIMESTEP * PHYSICS_CATCHUP_TICKS
    }
  }
  ```
  Please clarify for me.

File: lib/plugins/physics.js (hunk 7 above)
Comment (U5B) [reply lib/plugins/physics.js:16]: It isn't, I decided to read through that article, and they do run the loop faster (way too fast because they used a while loop instead of a setInterval).

The faster loop is used to get a more accurate doPhysics(), otherwise why aren't we just running the contents of tickPhysics() in a setInterval loop?

File: lib/plugins/physics.js [hunk 10]
```diff
@@ -309,5 +342,12 @@ function inject (bot, { physicsEnabled }) {
 
   bot.on('mount', () => { shouldUsePhysics = false })
   bot.on('respawn', () => { shouldUsePhysics = false })
+  bot.on('login', () => {
+    shouldUsePhysics = false
+    if (doPhysicsTimer === null) {
+      lastPhysicsFrameTime = performance.now()
+      doPhysicsTimer = setInterval(doPhysics, PHYSICS_SETINTERVAL_MS)
```
IceTank, line 349: This feels wrong. When the setInterval schedules the next physics tick to far in the future because of some de synch issue you could end up with the time accumulator having to catch up every time doPhysics is called. This would cause physics to take up way more cpu time then it should. Maybe a solution would be to set a new setTimeout whenever doPhysics ran after the time accumulator had a chance to adjust. And then give the setTimeout an adjusted time when the time accumulator ran late.

IceTank: You basically have two separate clocks running along side each other. But when on clock de syncs and runs late it has to block the event loop to catch up. But there is nothing stopping it from having to do the same thing in the next tick again.

IceTank: This could also explain why sometimes mineflayer is using a lot of cpu time when the bot is just sitting still and doing nothing.

U5B: Isn't that what doPhysics() is doing?
doPhysics() is a self adjusting loop that is called every 10ms. Every 50ms, it is supposed to call tickPhysics().
It catches up by itself if the setInterval loop falls behind. The problem is that Mineflayer runs the setInterval loop every 50ms, which means doPhysics() isn't doing it's job of playing catchup when the setInterval loop goes over 50ms.

U5B: > You basically have two separate clocks running along side each other. But when on clock de syncs and runs late it has to block the event loop to catch up. But there is nothing stopping it from having to do the same thing in the next tick again.

This is still a problem with the old design. The old design always had to catchup because setInterval would usually run late.
This is why in the new design, I am trying to incorporate a limiter so that it doesn't run too far behind.

lib/plugins/physics.js
Karang: > The faster loop is used to get a more accurate doPhysics(), otherwise why aren't we just running the contents of tickPhysics() in a setInterval loop?

The accumulator is here to compensate late intervals, running it faster do no improve accuracy, it just takes more event loop time...

Checking once every tick is the optimal interval time, if you go slower you'll accumulate delay, if you go faster you waste cpu.

U5B: How does the accumulator compensate late intervals if it is only run once every tick? It just keeps adding delay until deltaSeconds reaches 0.1 in which it fires twice and repeats the cycle. Unless that is the intended way?

Karang: yes that is exactly the intended way

lib/plugins/physics.js
Karang: Almost good, but you shouldn't remove time in the last if, just break and let the next tick continue catching up.

If you remove time you'll skip ticks, causing desync.

U5B: Thanks for clarification

lib/plugins/physics.js
U5B: Ah, because the double-fire behavior seems to happen so often that anticheats pick up on it, and I wanted to reduce the amount of double firing.

lib/plugins/physics.js
IceTank: But this dose not solve the issue when the two de sync. You should not be using setInterval at all when setInterval can fall behind. Or have a mechanism to adjust either setInterval or the catch up time in doPhysics. 

I agree that this is something that is already present in the current version. But this is a good opportunity to fix it.

Karang:
> Completely forgot about this PR.
> 
> Tests are failing because the delay between position packets is sometimes 100ms. @Karang any suggestions on fixing this delay? @IceTank says I shouldn't remove the offending test.

Testing that something should happen within 50ms in the test environment is not reasonable. Don't add this test.

U5B:
Alright, removed the test. The reason that it was testing for within 50ms, is because Minecraft sends a position packet every 50ms  or tick if it is falling.

This will not pass some falling checks if their threshold is extremely strict.

U5B:
I'll look into why tests are failing later. It probably is related to the setInterval change.

rom1504:
what's the state of this ?

U5B:
> what's the state of this ?

This is done. I believe it just needs review.

lib/plugins/physics.js
```diff
@@ -51,25 +53,44 @@ function inject (bot, { physicsEnabled }) {
...
+      bot.emit('physicTick') // Deprecated, only exists to support old plugins. May be removed in the future
+    }
+    if (shouldUsePhysics) {
+      updatePosition(now)
     }
   }
 
+  // remove this when 'physicTick' is removed
+  bot.on('newListener', (name) => {
+    if (name === 'physicTick') console.log('Mineflayer detected that you are using a deprecated event (physicTick)! Please use this event (physicsTick) instead.')
```
andriycraft, line 91: use console.warn

rom1504:
could you check this @IceTank please ?

lib/plugins/physics.js
```diff
@@ -12,7 +12,8 @@ module.exports = inject
 const PI = Math.PI
 const PI_2 = Math.PI * 2
 const PHYSICS_INTERVAL_MS = 50
-const PHYSICS_TIMESTEP = PHYSICS_INTERVAL_MS / 1000
+const PHYSICS_TIMESTEP = PHYSICS_INTERVAL_MS / 1000 // 0.05
+const PHYSICS_CATCHUP_TICKS = 4 // limit number of ticks to catchup if loop gets behind
```
IceTank, line 16: This seams like a really small number and no way to increase it?

lib/plugins/physics.js
IceTank, line 91: use console.warn

lib/plugins/physics.js
```diff
@@ -117,17 +138,24 @@ function inject (bot, { physicsEnabled }) {
     return dYaw
   }
 
-  function updatePosition (dt) {
-    // If you're dead, you're probably on the ground though ...
-    if (!bot.isAlive) bot.entity.onGround = true
+  function isDead () {
```
IceTank, line 141: This function should be named something else. It does not return the alive status of the bot but a helper variable to send packets after death.

lib/plugins/physics.js
```diff
@@ -280,19 +319,15 @@ function inject (bot, { physicsEnabled }) {
 
     if (bot.supportFeature('teleportUsesOwnPacket')) {
       bot._client.write('teleport_confirm', { teleportId: packet.teleportId })
-      // Force send an extra packet to be like vanilla client
-      sendPacketPositionAndLook(pos, newYaw, newPitch, bot.entity.onGround)
     }
     sendPacketPositionAndLook(pos, newYaw, newPitch, bot.entity.onGround)
 
     shouldUsePhysics = true
-    bot.entity.timeSinceOnGround = 0
```
IceTank, line 289: Why was this removed? Was it moved somewhere else?

lib/plugins/physics.js
```diff
@@ -280,19 +319,15 @@ function inject (bot, { physicsEnabled }) {
...
+    // bot.isAlive = true
```
IceTank, line 326: Remove debug comment

IceTank (commented):
The rest looks good

lib/plugins/physics.js
U5B: This was the number used before, I just added a comment saying what that number is. It is 0.05 seconds, which is 50ms.

U5B: Oh misread the review. Should I add a configuration option for this?

lib/plugins/physics.js
U5B: At the time, it was unused and wasn't specified in the API.

lib/plugins/physics.js
U5B: This function returns if the bot is considered "dead" or not to send physics packets. Vanilla client sends movement packets for 20 ticks after dying.

lib/plugins/physics.js
IceTank: fixed

lib/plugins/physics.js
IceTank: Maybe `isNotSendingPosition`

lib/plugins/physics.js
IceTank: ok

IceTank:
LGTM

IceTank (approved):

lib/plugins/physics.js
```diff
@@ -117,17 +138,25 @@ function inject (bot, { physicsEnabled }) {
     return dYaw
   }
 
-  function updatePosition (dt) {
-    // If you're dead, you're probably on the ground though ...
-    if (!bot.isAlive) bot.entity.onGround = true
+  // returns false if packet should be sent, true if not
+  function sendPositionPacketInDeath () {
```
frej4189, line 142: This seems like quite ambigious naming

IceTank: I think its a good name. Better then it was before. And it does not return where there the bot is dead but if the bot is dead and has been dead for a while. Do you have a better suggestion?

frej4189: The function does the opposite of it's name.

My best suggestion is either call it shouldNotSendPosition or invert its return values

PR #2494, opened 2022-02-23
bot.craft: typing

safazi:
allow count and craftingTable to be undefined


safazi:
there are probably many more types like this where an optional parameter isn't marked as such
also lots of nulls when we could just use the `param?` sugar

u9g (approved):
LGTM

PR #2497, opened 2022-02-27
Update missing/outdated typescript definitions

dumbasPL:
There is a lot of outdated typescript definitions that no longer match what's in the real code. I will try to add as much as I notice while working with mineflayer in TS.

PS: dont't merge this right away as there is a good chance I will find a lot more in the next few days

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -197,7 +197,7 @@ export interface Bot extends TypedEmitter<BotEvents> {
 
   end: (reason?: string) => void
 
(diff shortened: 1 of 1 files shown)


safazi:
Here's another: https://github.com/PrismarineJS/mineflayer/pull/2494

u9g (approved):
Looks good for now :)

safazi:
Also `bot.equip` - the destination slot should be optional

PR #2498, opened 2022-02-27
Adds how to use thealtening in mineflayer

Zn10plays:


Zn10plays:
I can't figure out what is wrong

PondWader:
I'm not sure PrismarineJS should be directly promoting the use of stolen accounts.

Issue #2513, opened 2022-03-06
Bots are flagging "bad packets" upon join with Vulcan anti-cheat


Cubxity:
- [x] The [FAQ](https://github.com/PrismarineJS/mineflayer/blob/master/docs/FAQ.md) doesn't contain a resolution to my issue 

## Versions
 - mineflayer: `4.0.0`
 - server: `git-Pufferfish-20 (MC: 1.17.1) (Implementing API version 1.17.1-R0.1-SNAPSHOT) (Git: b9556fd on HEAD)`
 - node: `16`

## Detailed description of a problem

Mineflayer bots flag Vulcan anti-cheat upon joining, due to the bots sending item held slot packet regardless of what the current held slot is.

![image](https://user-images.githubusercontent.com/27609129/156936220-e2be649b-6966-4aef-94dd-839fff2243ba.png)

## Expected behavior

It is expected that the bots don't send held item packet unless the held item slot changes.

## Additional context

Running Vulcan v2.6.2

u9g:
this packet is sent in two places: https://github.com/PrismarineJS/mineflayer/blob/694bee68e90b890b6ac58e920122bf2472b3ebd5/lib/plugins/game.js#L70 and https://github.com/PrismarineJS/mineflayer/blob/4fe54ad05a8c68892534f8f7c78e77cbff8a67b5/lib/plugins/simple_inventory.js#L56

Is the bot flagging after a little bit or right after join?

amoraschi:
I don't think bots are made to bypass anticheats, you can always fork the package and modify it

Cubxity:
> this packet is sent in two places:
> 
> https://github.com/PrismarineJS/mineflayer/blob/694bee68e90b890b6ac58e920122bf2472b3ebd5/lib/plugins/game.js#L70
> and
> 
> https://github.com/PrismarineJS/mineflayer/blob/4fe54ad05a8c68892534f8f7c78e77cbff8a67b5/lib/plugins/simple_inventory.js#L56
> 
> Is the bot flagging after a little bit or right after join?

It is flagged right after joining.

u9g:
> > this packet is sent in two places:
> > https://github.com/PrismarineJS/mineflayer/blob/694bee68e90b890b6ac58e920122bf2472b3ebd5/lib/plugins/game.js#L70
> > 
> > and
[...]

PR #2514, opened 2022-03-06
tiny anticheat fixes

U5B:
Fixes https://github.com/PrismarineJS/mineflayer/issues/2513

Removes held item slot being sent on login

Makes setting packet always sent on login

### DIFF
--- a/lib/plugins/game.js
+++ b/lib/plugins/game.js
@@ -67,7 +67,6 @@ function inject (bot, options) {
 
     bot.emit('login')
     bot.emit('game')
-    bot._client.write('held_item_slot', { slotId: 0 })
 
     // varint length-prefixed string as data
     bot._client.writeChannel(brandChannel, options.brand)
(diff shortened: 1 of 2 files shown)


lib/plugins/game.js
```diff
@@ -67,7 +67,6 @@ function inject (bot, options) {
 
     bot.emit('login')
     bot.emit('game')
-    bot._client.write('held_item_slot', { slotId: 0 })
```
rom1504, line 70: why

lib/plugins/settings.js
```diff
@@ -90,7 +90,7 @@ function inject (bot, options) {
     enableServerListing: options.enableServerListing || true
   }
 
-  bot._client.once('login', () => {
+  bot._client.on('login', () => {
```
rom1504, line 93: why

lib/plugins/game.js
U5B: Vanilla client only sends the server bound held item packet when recieving a client bound held item packet

PR #2515, opened 2022-03-07
Update tallWorld supportsFeature description

u9g:


lib/features.json
```diff
@@ -326,7 +326,7 @@
   },
   {
     "name": "tallWorld",
-    "description": "world starts at -64 and ends at 384",
+    "description": "world has infinite height in both directions",
```
rom1504, line 329: no that's not the meaning of this features
it's "in these versions, the default is -64 to 384"

infinite height started in 1.17

u9g: fixed

PR #2516, opened 2022-03-07
Move supportFeature to mcdata

u9g:


u9g:
requires https://github.com/PrismarineJS/node-minecraft-data/pull/162

rom1504:
can you also remove features.json from here ?

u9g:
> can you also remove features.json from here ?

Done

lib/loader.js
```diff
@@ -56,7 +55,7 @@ module.exports = {
   BossBar: require('./bossbar'),
   supportedVersions,
   testedVersions,
-  supportFeature
+  supportFeature: (version, feature) => require('minecraft-data')(version).supportFeature(feature)
```
rom1504, line 58: can't do that here, version is not known at the top level

rom1504: ah you're taking it as arg ok

u9g: fixed

PR #2523, opened 2022-03-10
Make resetState much faster for newer versions

u9g:
Later versions don't want to say anything in chat for gamemode changes unless we actually changes gamemode, not just creative->creative.

### DIFF
--- a/lib/plugins/creative.js
+++ b/lib/plugins/creative.js
@@ -1,6 +1,6 @@
 const assert = require('assert')
 const { Vec3 } = require('vec3')
-const { sleep } = require('../promise_utils')
+const { sleep, onceWithCleanup } = require('../promise_utils')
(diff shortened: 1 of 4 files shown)


rom1504:
Is it ready?

lib/plugins/creative.js
```diff
@@ -17,11 +17,6 @@ function inject (bot, { version }) {
   }
 
   const creativeSlotsUpdates = []
-  bot._client.on('set_slot', ({ windowId, slot, item }) => {
```
rom1504, line 20: Why

u9g: we don’t need it, below I replaced the check that used this with just waiting for the inventory event, which is probably better anyway since we can be sure that the inventory received the item

PR #2547, opened 2022-03-26
consistent badges in readme

autowert66:


extremeheat:
The current badges (with the big discord button) are intentional, it was supposed to give more attention to the important information

rom1504:
thanks @autowert66 
if you feel like it, you could do the same PR for https://github.com/PrismarineJS/prismarine-template and then for all prismarine projects

PR #2559, opened 2022-04-07
Cache 'positionUpdateSentEveryTick' feature lookup

IceTank:
This peace off code is called 20 times a second. Running `bot.supportFeature('positionUpdateSentEveryTick')` on a raspberrypi 4 takes up to 9ms. Not good.

We should look into caching supportFeature in more places that are called frequently or look into improving the performance off the implementation. Having to call this function multiple times on something like `physicsTick` can slow down the entire process significantly.

### DIFF
--- a/lib/plugins/physics.js
+++ b/lib/plugins/physics.js
@@ -18,6 +18,8 @@ function inject (bot, { physicsEnabled }) {
   const world = { getBlock: (pos) => { return bot.blockAt(pos, false) } }
   const physics = Physics(bot.registry, world)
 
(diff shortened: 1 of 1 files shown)


u9g (approved):

u9g:
Thanks!

extremeheat:
I agree the implementation could be made faster, iterating the features and setting attributes on the necessary data

rom1504:
Yeah i think we should simply pregenerate all the features values for the version we need.

PR #2562, opened 2022-04-09
Remove old teams from bot.teamMap

u9g:


rom1504:
Some tests wouldn't hurt

PR #2563, opened 2022-04-09
Add nameOnTablist() on player object

u9g:
allows for getting a chatmessage object of the player's name like it shows on the tablist (if it shows on the tablist)


IceTank:
Maybe name the property something like getNameOnTablist to indicate that it is a function and not a variable.

u9g:
> Maybe name the property something like getNameOnTablist to indicate that it is a function and not a variable.

It's now a readonly getter

IceTank:
Can you run the doctoc command to update the automated api.md formating? The command is also listed at the end of CONTRIBUTING.md

IceTank:
Looks like the failing test is due to some other issue and not related to the implemented feature. Having a test for this feature would be nice too.

rom1504:
conflicts

rom1504:
old and not quite right, closing

PR #2567, opened 2022-04-11
Add manual crafting test

IceTank:
This test only works for 1.12.2+ using this change: https://github.com/PrismarineJS/prismarine-windows/pull/60
I am not sure what is causing the test to fails for older versions tho.


frej4189:
This seems unnecessary

Issue #2580, opened 2022-04-15
Example program produces deprecation warnings


FM-96:
- [x] The [FAQ](https://github.com/PrismarineJS/mineflayer/blob/master/docs/FAQ.md) doesn't contain a resolution to my issue 

## Versions
 - mineflayer: 4.2.0
 - server: paper 1.18.2-287
 - node: 16.6.1

## Detailed description of a problem
Running a slightly modified version of the `reconnector.js` example file (see below for current code) leads to repeated deprecation warnings filling up the console:

```
{mcData}.findItemOrBlockById() is deprecated, scheduled to be removed on or after April 1 2022
{mcData}.findItemOrBlockById() is deprecated, scheduled to be removed on or after April 1 2022
{mcData}.findItemOrBlockById() is deprecated, scheduled to be removed on or after April 1 2022
{mcData}.findItemOrBlockById() is deprecated, scheduled to be removed on or after April 1 2022
{mcData}.findItemOrBlockById() is deprecated, scheduled to be removed on or after April 1 2022
{mcData}.findItemOrBlockById() is deprecated, scheduled to be removed on or after April 1 2022
{mcData}.findItemOrBlockById() is deprecated, scheduled to be removed on or after April 1 2022
{mcData}.findItemOrBlockById() is deprecated, scheduled to be removed on or after April 1 2022
{mcData}.findItemOrBlockById() is deprecated, scheduled to be removed on or after April 1 2022
{mcData}.findItemOrBlockById() is deprecated, scheduled to be removed on or after April 1 2022
```

## Your current code
```js
const mineflayer = require('mineflayer')

const loginData = {
	host: 'localhost',
	username: '[REDACTED]',
	password: '[REDACTED]',
	auth: 'microsoft',

}

function createBot() {
	const bot = mineflayer.createBot(loginData)

	bot.on('chat', (username, message) => {
		if (username === bot.username) return
		bot.chat(message)
	})

	bot.on('end', createBot)
}

createBot()
```

## Expected behavior
The `.findItemOrBlockById()` method referenced in the deprecation warning isn't used in the example code. Presumably it is called somewhere in the background during regular operation.

Ideally, a deprecated method shouldn't be used internally at all, but even if it is, it should not print out deprecation warnings when the end user can't do anything about it.

u9g:
can you do `npm install https://github.com/PrismarineJS/mineflayer/tree/u9g-patch-2` and tell me if you still see the error spam?

[...]

PR #2581, opened 2022-04-15
Use newer prismarine-item

u9g:


u9g:
should get rid of deprecation spam

IceTank:
The latest prismarine-item version is 1.11.5 not 1.15.0

u9g:
closes https://github.com/PrismarineJS/mineflayer/issues/2580

PR #2587, opened 2022-04-17
Adds new title events that work for later versions

u9g:
The "title" event doesn't work on later versions at all, plus it now sends chat messages, so the msg suffixed events are better overall. I'm considering setting a timer on deprecating the existing title event seeing as it doesn't work on later versions anyway


SMEDjs:
**it works**  👍

SMEDjs (approved):
works and seems good 👍

rom1504:
@u9g state of this ?

rom1504:
I don't like the API, please re-open if you want to finish it

PR #2590, opened 2022-04-17
adding new feature activateEntityWithOptions

Katzengott:
v1

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -304,7 +304,7 @@ export interface Bot extends TypedEmitter<BotEvents> {
 
   activateEntityAt: (block: Entity, position: Vec3) => Promise<void>
   
-  activateEntityWithOptions: (block: Entity, position: Vec3, mouse: 0 or 1) => Promise<void>
+  activateEntityWithOptions: (block: Entity, position: Vec3, mouse: 0 | 1) => Promise<void>
 
   consume: () => Promise<void>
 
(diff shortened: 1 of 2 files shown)


amoraschi:
Lint is failing

You also should add it to api.md and index.d.ts

IceTank:
Maybe also add the option to control the sneaking and hand parameter? Also that is not a valid index.d.ts entry. It should be something like `activateEntityWithOptions: (block: Entity, position: Vec3, mouse: 0 | 1) => Promise<void>`

Katzengott:
ok Where should I run npm run fix

IceTank:
> ok Where should I run npm run fix

In the project root directory where you also run `npm install` to install mineflayer dependencies

Katzengott:
Somehow that doesn't work for me. I'm just too stupid to do it right

IceTank:
> Somehow that doesn't work for me. I'm just too stupid to do it right

go to #help in the discord

Katzengott:
It looks good

rom1504:
Why not instead add options to the existing function?

Katzengott:
Would also go. I can also change it again. I thought it looks better

IceTank:
It is better to have one function that is as flexible as possible then to have multiple different functions doing almost the same stuff. You can add something like this
```js
activateEntityWithOptions(entity: Entity, options: { position?: Vec3, mouse?: 0 | 1, sneak?: boolean,  hand?: 'right' | 'left' })
```
and then make those options optional. Once you have this you can also refactor activateEntity to use activateEntityWithOptions. If you refactor it you make sure that there is only one place in mineflayer where it has to write the raw packet information making it easier in the future to update the packet.

rom1504:
old

PR #2595, opened 2022-04-20
Add bot.clickWindow mode disclaimer

IceTank:
If we are throwing with mode values other then 0 inside the clickWindow function we should also say that in the documentation.


u9g:
maybe say to refer to the simpleClick methods

amoraschi:
Maybe it's also a good idea to print a warning or something when using mode other than 0

u9g:
> Maybe it's also a good idea to print a warning or something when using mode other than 0

you can put that in another pr

PR #2609, opened 2022-05-01
Add bot.entityInSight() (recreation of #1371)

au5ton:
This is a re-opening of https://github.com/PrismarineJS/mineflayer/pull/1371 after making the changes requested in: https://github.com/PrismarineJS/mineflayer/pull/1371#issuecomment-712040319

I would've used the original pull request, but the author's GitHub account was deleted, so I wasn't sure how to go about doing that.

This worked well for me in my limited testing.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -201,6 +201,8 @@ export interface Bot extends TypedEmitter<BotEvents> {
 
   blockInSight: (maxSteps: number, vectorLength: number) => Block | null
 
+  entityInSight: (maxSteps?: number, vectorLength?: number) => Entity | null
+
   blockAtCursor: (maxDistance?: number, matcher?: Function) => Block | null
   blockAtEntityCursor: (entity?: Entity, maxDistance?: number, matcher?: Function) => Block | null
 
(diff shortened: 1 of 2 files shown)


extremeheat:
Dupe of https://github.com/PrismarineJS/mineflayer/pull/2077 ?

extremeheat:
This implementation seems cleaner with a width feature though. Making the name match the current block methods sounds like a good idea, like `entityAtCursor()` in the other PR

au5ton:
> This implementation seems cleaner with a width feature though. Making the name match the current block methods sounds like a good idea, like `entityAtCursor()` in the other PR

Are there any changes you'd like me to make, maybe specifically with the naming? It tried to follow the convention I saw in the other methods.

After looking at the other PR (#2077 ) it looks like I forgot to update the API docs, too.

u9g:
If you can make a test that would be great!

extremeheat:
This does not seem to test for obstructions. entityInSight() may return something even if a wall is in front of the bot.

rom1504:
add test or example

rom1504:

PR #2610, opened 2022-05-02
added bot.canSeeEntity(entity)

au5ton:
This pull request adds the function `bot.canSeeEntity: (entity: Entity, vectorLength?: number) => boolean`.

Basically, all 8 vertices of an entity's bounding box are computed and a ray trace is done from the bot's eyes to each vertex. For each trace, if non-transparent blocks are found along the way, that trace returns false. If any of the traces return true, the bounding box of the entity is partially visible from the bot.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -206,6 +206,8 @@ export interface Bot extends TypedEmitter<BotEvents> {
 
   canSeeBlock: (block: Block) => boolean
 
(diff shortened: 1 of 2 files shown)


Slezzzy-lab (approved):
Looks good

rom1504:
can you add an example or a test ?

otherwise how do you know it works ?

rom1504:

PR #2617, opened 2022-05-07
Added method to wait until sleep function is in reality sleeping

sefirosweb:

## Versions
 - mineflayer: 4.3.0
 - server: vanilla 1.18.1
 - node: Dockerized:  node:16.13-bullseye

If the bot click on bed at same time when other bot/player click on bed and steal him they don't return any error,
The function no return an error

I attach video you can see a bit better: https://youtu.be/S5PrYqO4KRk

## Expected behavior
Throw an error

### DIFF
--- a/lib/plugins/bed.js
+++ b/lib/plugins/bed.js
@@ -147,7 +147,7 @@ function inject (bot) {
       }
 
       await bot.activateBlock(bedBlock)
-      
+
       await timeout(3000)
       if (!bot.isSleeping) {
         throw new Error('bot is not sleeping')
--- a/lib/plugins/bed.js
+++ b/lib/plugins/bed.js
@@ -148,23 +148,25 @@ function inject (bot) {
 
       await bot.activateBlock(bedBlock)
 
-      await timeout(3000)
-      if (!bot.isSleeping) {
-        throw new Error('bot is not sleeping')
+      const err = await waitUntilSleep()
+      if (err) {
+        throw new Error(err)
       }
     }
   }
 
-  function timeout (ms) {
-    return new Promise(resolve => setTimeout(resolve, ms))
-  }
+  async function waitUntilSleep () {
+    return new Promise((resolve, reject) => {
+      const timeoutForSleep = setTimeout(() => {
+        resolve('bot is not sleeping')
+      }, 3000)
 
-  bot._client.on('game_state_change', (packet) => {
-    if (packet.reason === 0) {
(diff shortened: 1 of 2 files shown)


sefirosweb:
How i can rerun tests?

amoraschi:
Push some changes, but doesn't the sleeping duration vary with the server's TPS?

sefirosweb:
> Push some changes, but doesn't the sleeping duration vary with the server's TPS?

Yes, but the idea is not to get the sleeping duration, only they detect if bot are sleeping or not,
Because they don't return if the bot can be clicked correctly to bed

sefirosweb:
Seems this PR break the test of sleep, checking..

sefirosweb:
@amoraschi done!

I think now is much better, (See code)

the sleep function waits until the sleep event is triggered,

If there is no error and the event is not triggered then it returns an error indicating that it has not been able to sleep
So in this way we will avoid errors that the .catch event of sleep() is never fired

I removed in the test "await once(bot, 'sleep')" because sleep now wait to this trigger

lib/plugins/bed.js
```diff
@@ -147,15 +147,26 @@ function inject (bot) {
...
-  bot._client.on('game_state_change', (packet) => {
-    if (packet.reason === 0) {
-      // occurs when you can't spawn in your bed and your spawn point gets reset
-      bot.emit('spawnReset')
-    }
-  })
+  async function waitUntilSleep () {
+    return new Promise((resolve, reject) => {
+      const timeoutForSleep = setTimeout(() => {
+        resolve('bot is not sleeping')
```
amoraschi, line 161:
```suggestion
        reject('bot is not sleeping')
```
Shouldn't this be `reject`?

amoraschi (changes requested):
Promise should reject instead of resolve inside the `setTimeout`

lib/plugins/bed.js
sefirosweb: I tried, but with reject don't received data,
Also they don't throw an error correctly

sefirosweb: With reject
image

U5B: try `reject(new Error('bot is not sleeping'))`

sefirosweb: @U5B you are great i learned a new thing :D

u9g (approved):

u9g:
Thanks

PR #2642, opened 2022-05-29
Add ender chests as a chest type

RoseChilds:
Not really much to explain here, just adds ender chests so that a bot can open one without throwing an error.


u9g (approved):

u9g:
Thanks

PR #2647, opened 2022-06-02
Add player object to blockBreakProgressEnd & observed.

JackCrispy:
I needed to see which players were interacting with blocks, shoutout to U9G#0670 on discord for solution.


amoraschi:
Should the variable be renamed to entity? Since I think that returns an entity

PR #2684, opened 2022-07-03
Fix type error on chest open

IceTank:
There is a missing null check when trying to access a block. If that block is outside the render distance blockAt can return null causing a type error. See PrismarineJS discord message for an example https://discord.com/channels/413438066984747026/413438150594265099/993210395990949988

### DIFF
--- a/lib/plugins/block_actions.js
+++ b/lib/plugins/block_actions.js
@@ -44,7 +44,7 @@ function inject (bot) {
       const perpendicularCardinals = Object.keys(FACING_MAP[facing])
       for (const cardinal of perpendicularCardinals) {
         const cardinalOffset = CARDINALS[cardinal]
-        if (bot.blockAt(chestBlock.position.plus(cardinalOffset)).type === chestBlock.type) {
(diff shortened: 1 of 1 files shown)


IceTank:
So I think a chest can either be a single chest or a double chest. If the chest block does not store what chest type it is mineflayer tries to guess what chest type the chest it want's to open is. On 1.8 if a chunk is empty it is not send at all and mineflayer thinks that chunk is unloaded. So when you open a single chest at a chunk corner mineflayer will crash as blockAt will return null. There can be double chests at chunk borders that can still work if both chest blocks are within loaded chunks. I don't see how 1.8 would tell the client that a single chest at a chunk corner can be a double chest with the other chest block being within the unloaded chunk. For newer versions you can have chest blocks that say if they are single or double chest blocks. And mineflayer uses that to know what type off chest it opens. It looks like this function is only used to emit the chest lid move event. And that requires knowledge off if a chest is a double chest or a single chest. I have not tested it by my guess is that on 1.8 you would only see a single chest opening when you observer a double chest opening that is half in a loaded chunk.

PR #2718, opened 2022-07-28
Updated how awaitMessage works.

QimieGames:
Just for my conveniences, not even going to lie about it.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -262,7 +262,7 @@
       - [bot.addChatPattern(name, pattern, chatPatternOptions)](#botaddchatpatternname-pattern-chatpatternoptions)
       - [bot.addChatPatternSet(name, patterns, chatPatternOptions)](#botaddchatpatternsetname-patterns-chatpatternoptions)
       - [bot.removeChatPattern(name)](#botremovechatpatternname)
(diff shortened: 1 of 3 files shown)


amoraschi:
In this case I think timeout should be the first argument since rest parameters must be last

QimieGames:
> In this case I think timeout should be the first argument since rest parameters must be last

Done. Check the latest commit. Felt like im missing something still. Ill try to look more. Ill tell ya when im done. https://github.com/PrismarineJS/mineflayer/pull/2718/commits

QimieGames:
Yeah im done with everything. Pretty sure anyways (except other languages cause I can only understand English and my native lang). Have fun reviewing and thanks! <3

QimieGames:
Making awaitMessage async also allows user to run commands in chat after awaitMessage has been executed. Instead of running the command before or using setTimeout to run the command for example /stats or any commands that needs the player to run a command before giving output in chat. (My main intention of this whole PR xD)

rom1504:
this is breaking, can you instead make a new function ?

QimieGames:
> this is breaking, can you instead make a new function ?

Im new to all this stuff but I will try my best.

PR #2727, opened 2022-08-07
digging.js timing fix

Vakore:
This fixes inconsistent timings when compared to vanilla when the mineflayer client performs and action that causes the time it takes to dig a block in the middle of digging a block, such as starting to dig a block in the air, followed by landing and continuing to dig. The old version would not finish digging until the air time finished, while this fix allows for it to be more in line with vanilla. This has not been extensively tested, especially against anticheats.

This fix adds a new variable called "bot.digPercentage" which is on a scale from 0.0(just started) to 1.0(finished). This variable will reset to 0.0 when the bot's held item is switched, just like the vanilla client. Again, not tested extensively. These fixes appear to work to me, but I would test them a little more thoroughly against the older version.

THINGS THAT STILL NEED FIXING:
* swingInterval will not clear sometimes, and as a result the bot will swing its arm indefinitely, which is not only visually unappealing but affects the bot's ability to PvP in 1.9+ due to the cooldown resetting whenever the arm swing animation plays in either hand. My personal fix has been to disable it and make the bot swing its arm on my own terms, but I'm not sure how it would be fixed here. May be easier to replicate on weaker machines, as my laptop is super low end and I get it very quickly.
* Inconsistent timings in certain aspects of mining. I.e. bot's head above water while standing on a dirt block but the dig time is not those of a vanilla client. Easiest to test in a swamp.
* I got a memory leak about waiting for events when testing this new version. I have no clue if it is related with any of the changes I made or if it is just a result of the project I was working on. Either way, I would test this more before implementing.

### DIFF
--- a/lib/plugins/digging.js
+++ b/lib/plugins/digging.js
@@ -7,12 +7,17 @@ module.exports = inject
 
 function inject (bot) {
   let swingInterval = null
-  let waitTimeout = null
+  let waitInterval = null
 
   let diggingTask = createDoneTask()
 
   bot.targetDigBlock = null
   bot.lastDigTime = null
+
+  bot.digPercentage = 0;
+  bot.on('heldItemChanged', () => {
+    bot.digPercentage = 0;
+  });
 
   async function dig (block, forceLook, digFace) {
     if (block === null || block === undefined) {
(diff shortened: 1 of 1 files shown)


rom1504:
can you please fix the lint

PR #2729, opened 2022-08-07
Added bot.onMessage() function for more automation.

QimieGames:
Idk how to sync the old PR to the latest sorry. But here I tried making a new function.


QimieGames:
@rom1504 Done. Will this do the trick?

u9g:
How is this different from doing promise.race with a settimeout promise and bot.awaitMessage ? It’s two lines of code in the users program to do this.

QimieGames:
> How is this different from doing promise.race with a settimeout promise and bot.awaitMessage ? It’s two lines of code in the users program to do this.

Well it would be easier to have it as a function rather than making a function in main file then requiring it in every single file that uses it. These are examples of the usage. [My old PR](https://github.com/PrismarineJS/mineflayer/pull/2718)
image
image

QimieGames:
and I just realised I can combined those leaderboard commands into one file. ffs

QimieGames:
> How is this different from doing promise.race with a settimeout promise and bot.awaitMessage ? It’s two lines of code in the users program to do this.

Also awaitMessage promise wont be ever resolve if message is not found right? Unless u reject or resolve them urself but tha would need more code. Which ngl im lazy. So I make this new function just to give user a bit more automation.

rom1504:
that seems useful

can we call it bot.onMessage instead ?

QimieGames:
Yep sure will do when I get home

QimieGames:
> that seems useful
> 
> can we call it bot.onMessage instead ?

Indeed. Ive been using it for my projects. Just check any of my repo all of them uses it i believe. Owh and its done. Ive changedd the name to onMessage, fixed the tests, definitions thing, and quick tutorial in api md

4lve (approved):
Looks good

lib/plugins/chat.js
```diff
@@ -243,4 +243,30 @@ function inject (bot, options) {
...
+        }
+      }
+      const timeoutOnMessage = setTimeout(async () => {
+        resolve(false)
+        bot.off('messagestr', messageListener)
+      }, onMessageTimeout)
+      bot.on('messagestr', messageListener)
+    })
+  }
+  bot.onMessage = onMessage
```
IceTank, line 271: Instead of making a new promise that resolver or rejects at some point you should use mineflayers [onceWithCleanup](https://github.com/PrismarineJS/mineflayer/blob/e5c7f7bad35a4778ade759297d4a63a8524c38a8/lib/promise_utils.js#L47) function. It also has a check condition to see if the promise should resolve.

test/externalTests/chat.js
```diff
@@ -88,6 +88,25 @@ module.exports = () => {
...
+    await p1
+    const p2 = bot.onMessage(3000, ['<flatbot> hello', '<flatbot> world'])
+    bot.chat('world')
+    await p2
+    const p3 = bot.onMessage(3000, /<.+> hello/)
+    bot.chat('hello')
+    await p3
+    const p4 = bot.onMessage(3000, [/<.+> hello/, /<.+> world/])
+    bot.chat('world')
+    await p4
```
IceTank, line 103: Just do `await bot.onMessage(3000, [/<.+> hello/, /<.+> world/])`

IceTank:
So as I understand it you can combine different resolve criteria in a `or` or `and` way. But why? If you want a `or` behavior you could use something like `Promise.any([onMessage(2000, 'hallo'), onMessage(2000, 'hi')])` and if you want `and` `Promise.all(...)`

frej4189:
> So as I understand it you can combine different resolve criteria in a `or` or `and` way. But why? If you want a `or` behavior you could use something like `Promise.any([onMessage(2000, 'hallo'), onMessage(2000, 'hi')])` and if you want `and` you can use `Promise.all(...)`

I guess that makes sense for convience though

PR #2740, opened 2022-08-15
Fixed openContainer and added missing functions extended from windows

sefirosweb:
The function of containerItems is missing from chest / dispenser / and furnance
This function is extended from windows

Also "openContainer" no works with furnance


sefirosweb:
Close for a now still missing components of windows,

I will try to made another PR fixing all types

PR #2751, opened 2022-08-21
Add options to control block validation

CrawLeyYou:
setting for a problem I've had and think others may have as well.
The problem is that the bot teleports to a certain place and placing a command block there by commands (like /setblock) but it cannot detect it and cannot write anything in it but it can write something if we add this options.

### DIFF
--- a/lib/plugins/command_block.js
+++ b/lib/plugins/command_block.js
@@ -9,7 +9,7 @@ function inject (bot) {
     assert.notStrictEqual(pos, null)
     assert.notStrictEqual(command, null)
     if (options.checkblock !== false) {
-     assert.strictEqual(bot.blockAt(pos).name.includes('command_block'), true, new Error("The block isn't a command block"))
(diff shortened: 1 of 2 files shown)


lib/plugins/command_block.js
```diff
@@ -8,9 +8,11 @@ function inject (bot) {
     assert.strictEqual(bot.player.gamemode, 1, new Error('The bot has to be in creative mode to open the command block window'))
     assert.notStrictEqual(pos, null)
     assert.notStrictEqual(command, null)
-    assert.strictEqual(bot.blockAt(pos).name.includes('command_block'), true, new Error("The block isn't a command block"))
-
+    if (options.checkblock !== false) {
```
rom1504, line 11: that is the wrong fix

the check is correct, the fix is to call that function at a later time when the bot got the information (using the block update event for example)

rom1504 (changes requested):
incorrect change, looks like a bug on your side

PR #2768, opened 2022-09-09
Extend every window that is opened

IceTank:
Not all windows that are emitted with windowOpen include withdraw, deposit and close. 
Especially close is really important when having windows open. I am not 100% sure if withdraw and deposit work with all types off windows that can be opened. Maybe someone can test it?


rom1504:
do you want to finish this ?

IceTank:
I think this should work

rom1504:
That method name is way too vague

PR #2785, opened 2022-10-04
add example for GUI

firejoust:


examples/gui.js
```diff
@@ -0,0 +1,39 @@
...
+
+  // log all the teams
+  for (const team of teams) {
+    const display = bot.gui.item.getName(team)
+    console.log(display.toString())
+  }
+}
+
+const bot = mineflayer.createBot()
+bot.loadPlugin(gui)
```
DEVTomatoCake, line 38:
```suggestion
bot.loadPlugin(gui.plugin)
```
Doesn't work otherwise.
(Also this plugin is fairly difficult to use, esp the Options and ItemMatch declarations need comments on how to use and expand them, e. g. for items with item translations)

firejoust: yep fair enough, I may scrap the ItemMatch althogether and use strings instead (using Options to specify what the string is matching)

firejoust: I will continue to work on making the plugin simpler to use in the meantime, I'll re-open once this has been completed

Issue #2450, opened 2022-02-04
Use prismarine-registry for everything


rom1504:
* [x] pblock
* [x] pchunk
* [x] pbiome
* [x] pchat
* [x] pentity
* [x] pitem
* [x] precipe
* [x] mineflayer 
   * [x] internal mineflayer
   * [x] examples (should be done once other p* are done)

rom1504:
Only pentity and precipe left

rom1504:
Last step here is to remove all usages of minecraft-data here and instead use bot.registry
In particular everything in examples folder

rom1504:
Then this will unlock using more of the dynamic features provided by packets (eg recipes)

rom1504:
Almost all examples done
Only mineflayer pathfinder ones left, require pathfinder support first

rom1504:
Maybe https://github.com/PrismarineJS/prismarine-viewer too

rom1504:
Maybe it would make sense in https://github.com/PrismarineJS/flying-squid/issues/613 but low priority

rom1504:
Probably in https://github.com/PrismarineJS/prismarine-block/search?q=minecraft-data&type=

rom1504:
It seems that's it, nothing else. So summary
* [ ] Pathfinder and usage
* [ ] prismarine-block
* [ ] prismarine-windows
* [ ] prismarine-viewer


Follow up will include flying-squid and prismarine-web-viewer but can be tracked in their dedicated repos and not here.

rom1504:
Once those are done, next step here is https://github.com/PrismarineJS/mineflayer/issues/2794

rom1504:
https://github.com/PrismarineJS/mineflayer/search?q=minecraft-data&type= quite a few tests to adapt here

rom1504:
main thing left is prismarine-block

rom1504:
ah it's done already

rom1504:
I'm going to call this done. Next step https://github.com/PrismarineJS/mineflayer/issues/2794

PR #2793, opened 2022-10-12
refactoring examples to use bot.registry (*except for some that use mineflayer-pathfinder, which still doesn't support it)

Epirito:
#2450


rom1504:
Looks good thanks
Want to do the same for pathfinder too ?

Epirito:
ok

PR #2795, opened 2022-10-12
mcData to registry refactoring (final step, at least for mineflayer)

Epirito:
#2450

### DIFF
--- a/test/externalTests/consume.js
+++ b/test/externalTests/consume.js
@@ -1,7 +1,7 @@
 const assert = require('assert')
 
 module.exports = () => async (bot) => {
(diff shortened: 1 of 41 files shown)


test/externalTests/consume.js
```diff
@@ -2,9 +2,8 @@ const assert = require('assert')
 
 module.exports = () => async (bot) => {
   const Item = require('prismarine-item')(bot.version)
```
rom1504, line 4: This should take the registry instead of the version

rom1504:
CI failing

Epirito:
Fixed

PR #2801, opened 2022-10-15
block face position fix + lint

WhoTho:


WhoTho:
Previous calculations for block faces were doing the distance to the corner of the block and then adding 0.5 instead of just doing the bots distance to the center of the block. That error caused the bot to think it was supposed to be looking at different sides of the block and skipping past open faces.

PR #2804, opened 2022-10-17
world.entities api consumption draft

Epirito:
it's incomplete. when you see a removed bot._client.on(event, listener) it's probably because it really should be there but I just didn't reimplement it yet. But most other stuff is really supposed to move to pworld.

### DIFF
--- a/lib/plugins/entities.js
+++ b/lib/plugins/entities.js
@@ -1,8 +1,4 @@
-const { Vec3 } = require('vec3')
 const conv = require('../conversions')
-const NAMED_ENTITY_HEIGHT = 1.62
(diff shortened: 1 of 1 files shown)


rom1504:
Why did you close? Is it invalid?

Epirito:
Well I was waiting for a confirmation that this draft was on the right direction before fully implementing the changes but perhaps closing it was not the correct way of signalling that? Idk the right protocol for this.
Anyway, what do you think about the draft? (It's supposed to interact with this: https://github.com/Epirito/prismarine-world )

rom1504:
old

PR #2813, opened 2022-10-21
Add a native mineflayer event for particles

### LINKED ISSUE #1887: Add a native mineflayer event for particles
Shouldn't have to rely on node-minecraft-protocol for particle packets

`particles` event

using `bot._client.once('world_particles', onParticles)`

## Additional context
N/A

NyxaYu:
Closes #1887

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -81,7 +81,7 @@
       - [Particle.forceRender](#particleforcerender)
       - [Particle.position](#particleposition)
       - [Particle.offset](#particleoffset)
-      - [Particle.particleData](#particleparticledata)
+      - [Particle.speed](#particlespeed)
       - [Particle.particles](#particleparticles)
       - [Particle.data](#particledata)
   - [Bot](#bot)
@@ -757,9 +757,9 @@ Vec3 instance of where the particle was created
 
 Vec3 instance of the particle's offset
 
-#### Particle.particleData
+#### Particle.speed
 
-Extra particle data, normally speed
+Particle speed in a random direction
 
 #### Particle.particles
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -75,6 +75,15 @@
       - [BossBar.isDragonBar](#bossbarisdragonbar)
       - [BossBar.createFog](#bossbarcreatefog)
       - [BossBar.color](#bossbarcolor)
+    - [mineflayer.Particle](#mineflayerparticle)
+      - [Particle.id](#particleid)
+      - [Particle.name](#particlename)
+      - [Particle.longDistance](#particlelongdistance)
+      - [Particle.position](#particleposition)
+      - [Particle.offset](#particleoffset)
+      - [Particle.particleData](#particleparticledata)
+      - [Particle.particles](#particleparticles)
+      - [Particle.data](#particledata)
   - [Bot](#bot)
     - [mineflayer.createBot(options)](#mineflayercreatebotoptions)
     - [Properties](#properties)
@@ -239,6 +248,7 @@
       - ["heldItemChanged" (heldItem)](#helditemchanged-helditem)
       - ["physicsTick" ()](#physicstick-)
       - ["chat:name" (matches)](#chatname-matches)
+      - ["particle"](#particle)
     - [Functions](#functions)
       - [bot.blockAt(point, extraInfos=true)](#botblockatpoint-extrainfostrue)
       - [bot.waitForChunksToLoad()](#botwaitforchunkstoload)
@@ -724,6 +734,40 @@ Determines whether or not boss bar creates fog
 #### BossBar.color
 
 Determines what color the boss bar color is, one of `pink`, `blue`, `red`, `green`, `yellow`, `purple`, `white`
+
+### mineflayer.Particle
+
+#### Particle.id
+
+Particle ID, as defined in the [protocol](https://wiki.vg/Protocol#Particle)
+
+#### Particle.name
+
+Particle Name, as defined in the [protocol](https://wiki.vg/Protocol#Particle)
+
+#### Particle.longDistance
+
+Determines whether or not to increase the particle distance
+
+#### Particle.position
+
+Vec3 instance of where the particle was created
+
+#### Particle.offset
+
+Vec3 instance of the particle's offset
+
+#### Particle.particleData
+
+Extra particle data, normally speed
+
+#### Particle.particles
+
+Amount of particles created
+
+#### Particle.data
+
+An object of variable particle data if required otherwise null, as defined in the [protocol](https://wiki.vg/Protocol#Particle)
 
 ## Bot
 
@@ -1464,6 +1508,10 @@ Fires every tick if bot.physicsEnabled is set to true.
 #### "chat:name" (matches)
 
 Fires when the all of a chat pattern's regexs have matches
+
+#### "particle"
+
+Fires when a particle is created
 
 ### Functions
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -78,7 +78,7 @@
     - [mineflayer.Particle](#mineflayerparticle)
       - [Particle.id](#particleid)
       - [Particle.name](#particlename)
-      - [Particle.forceRender](#particleforcerender)
+      - [Particle.longDistanceRender](#particlelongdistancerender)
       - [Particle.position](#particleposition)
       - [Particle.offset](#particleoffset)
       - [Particle.speed](#particlespeed)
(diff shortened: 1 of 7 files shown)


lib/particle.js
```diff
@@ -0,0 +1,84 @@
...
+
+module.exports = loader
+
+function loader (registry) {
+  mcdata = registry
+  return Particle
+}
+
+class Particle {
+  constructor (id, longDistance, position, offset, particleData, particles, data) {
```
extremeheat, line 11: Do we need this class with all the getters and setters?

Sounds like a simple Particle.fromNetwork(packet) => Particle would suffice. Using constructors like this is not future proof if fields change.

lib/particle.js
```diff
@@ -0,0 +1,84 @@
+let mcdata
+
+module.exports = loader
+
+function loader (registry) {
+  mcdata = registry
```
extremeheat, line 6: Global registry variable doesn't make sense. One instance will overwrite the other. You need to put the class inside the loader function.

docs/api.md
```diff
@@ -725,6 +735,40 @@ Determines whether or not boss bar creates fog
...
+
+Particle ID, as defined in the [protocol](https://wiki.vg/Protocol#Particle)
+
+#### Particle.name
+
+Particle Name, as defined in the [protocol](https://wiki.vg/Protocol#Particle)
+
+#### Particle.longDistance
+
+Determines whether or not to increase the particle distance
```
extremeheat, line 750: Can you explain this a bit more and where it applies ? Not familiar with this concept in game.

docs/api.md
```diff
@@ -725,6 +735,40 @@ Determines whether or not boss bar creates fog
...
+
+#### Particle.position
+
+Vec3 instance of where the particle was created
+
+#### Particle.offset
+
+Vec3 instance of the particle's offset
+
+#### Particle.particleData
```
extremeheat, line 760: This type of data should be abstracted away, not exposed directly. It's likely to break between versions and between different editions of Minecraft (eg bedrock). Try to only expose data/fields that would make sense to an ingame player.

extremeheat, line 760: This type of data should be abstracted away, not exposed directly. It's likely to break between versions and between different editions of Minecraft (eg bedrock). Try to only expose data/fields that would make sense to an ingame player (high level).

extremeheat (changes requested):
Left some comments about abstraction and future compatibility.

docs/api.md
NyxaYu: It's only use is for rendering purposes. In game if you use the `/particles` command and set the `force` option it will force any player to render the particle(even with particles set to minimum) and increase the distance it can be seen from 256 blocks to 65536 blocks.

docs/api.md
NyxaYu: This could be changed to `maxSpeed` as defined in the latest protocol however I wasn't sure if it had any other use in previous versions as it was originally called `particleData`.

docs/api.md
extremeheat: ok, you should rename it to something better (longDistanceRender) and put that explanation into the doc

docs/api.md
extremeheat: yeah, would be a good idea to check this then. Users shouldn't have to do it on their side

NyxaYu: Checked a few Minecraft version's source code, from what I can tell it has no other use other than determining the speed to move a particle in a random direction, changed accordingly.

docs/api.md
NyxaYu: Renamed to `forceRender` and detailed in docs

lib/particle.js
NyxaYu: Changed to a function

lib/particle.js
```diff
@@ -0,0 +1,20 @@
+const { Vec3 } = require("vec3")
+
+module.exports = loader
+
+function loader (registry) {
+  function fromNetwork(packet){
```
extremeheat, line 6: I didn't mean to get rid of the class. I meant to make fromNetwork a static method inside the class.

docs/api.md
```diff
@@ -725,6 +735,40 @@ Determines whether or not boss bar creates fog
...
+
+Particle ID, as defined in the [protocol](https://wiki.vg/Protocol#Particle)
+
+#### Particle.name
+
+Particle Name, as defined in the [protocol](https://wiki.vg/Protocol#Particle)
+
+#### Particle.forceRender
+
+Determines whether or not to force the rendering of a particle despite client particle settings and increases maximum view distance from 256 to 65536
```
extremeheat, line 750: `forceRender` is not a great name, because it has an extended but not unlimited rage

lib/particle.js
```diff
@@ -0,0 +1,20 @@
...
+    return {
+      id: packet.particleId,
+      name: registry.particles[packet.particleId].name,
+      forceRender: packet.longDistance,
+      position: new Vec3(packet.x, packet.y, packet.z),
+      offset: new Vec3(packet.offsetX, packet.offsetY, packet.offsetZ),
+      speed: packet.particleData,
```
extremeheat, line 13: Did you check the accuracy on supported versions?

docs/api.md
```diff
@@ -725,6 +735,40 @@ Determines whether or not boss bar creates fog
...
+
+Particle speed in a random direction
+
+#### Particle.count
+
+Amount of particles created
+
+#### Particle.data
+
+An object of variable particle data if required otherwise null, as defined in the [protocol](https://wiki.vg/Protocol#Particle)
```
extremeheat, line 770: This also needs to be abstracted

extremeheat (changes requested):

lib/particle.js
NyxaYu: Checked against MC 1.8, MC 1.8.8, MC 1.13, MC 1.16.3 and MC 1.19, can check more version if required

docs/api.md
NyxaYu: Thought that would be better suited given it was the primary purpose and the tab complete called the option for it `force`, changed to `longDistanceRender` as suggested

lib/particle.js
NyxaYu: Apologies, reverted to class with static function

lib/particle.js
```diff
@@ -0,0 +1,87 @@
+const { Vec3 } = require("vec3")
+
+module.exports = loader
+
+function loader (registry) {
+  class Particle {
```
extremeheat, line 6: In my earlier comment I meant you can remove the unnecessary `get` and `set` functions. Something like this will take alot less code :

```js
class Particle {
  constructor(id, position, count = 1, longDistanceRender, extraData) {
    Object.assign(this, registry.particlesById[id]) // Add in static data about particle 
    // Add in general instance fields for this particle
    // https://minecraft-data.prismarine.js.org/?v=1.8&d=protocol#toClient_world_particles
    // and https://minecraft-data.prismarine.js.org/?v=bedrock_1.16.220&d=protocol#packet_spawn_particle_effect
    this.id = id
    this.position = position
    this.count = count
    this.longDistanceRender = longDistanceRender
    Object.assign(this, extraData) // Particle-specific data that can vary by version
  }

  // This doesn't need to be here, it can also be handled in packet event handler
  static fromNetwork(packet) {
     return new Particle(packet.particleId, new Vec(particle.x, ...), packet.particles, packet.longDistance, {
       movementSpeed: packet.particleData,
     })
  }
}
```

Pretty much everything on particle should be well defined and not copied 1:1 from the network without understanding what it is. Otherwise in the future when the packet structures fields change or different editions get supported it's likely something will break.

lib/particle.js
```diff
@@ -0,0 +1,32 @@
...
+
+    static fromNetwork (packet) {
+      return new Particle(
+        packet.particleId,
+        new Vec3(packet.x, packet.y, packet.z),
+        new Vec3(packet.offsetX, packet.offsetY, packet.offsetZ),
+        packet.particles,
+        packet.particleData,
+        packet.longDistance,
+        packet.data
```
extremeheat, line 26: Don't expose this data if it's not abstracted

docs/api.md
```diff
@@ -725,6 +735,40 @@ Determines whether or not boss bar creates fog
...
+
+#### Particle.count
+
+Amount of particles created
+
+#### Particle.movementSpeed
+
+Particle speed in a random direction
+
+### Particle-specific properties
```
extremeheat, line 768: remove (see other comment)

extremeheat:
Needs a test

test/internalTest.js
```diff
@@ -721,5 +721,38 @@ for (const supportedVersion of mineflayer.testedVersions) {
         })
       })
     })
+
+    it('particles', (done) => {
```
extremeheat, line 725: Can you add an externalTest ? The internal test for this is not very meaningful

lib/particle.js
NyxaYu: Checked with newly written external test, working against all versions

extremeheat (approved):

Issue #2819, opened 2022-10-28
entityAttach and entityDetach events not triggering post 1.8


drakemj:
- [x] The [FAQ](https://github.com/PrismarineJS/mineflayer/blob/master/docs/FAQ.md) doesn't contain a resolution to my issue 

## Versions
 - mineflayer: 4.5.1
 - server: vanilla/spigot/paper 1.9.0 -> 1.19
 - node: v19.0.0

## Detailed description of a problem
This is a followup to #2805, where @DatArnoGuy found that these two events did not trigger at all in 1.17.1, and `set_passengers` replaced `attach_entity` for all vehicles. `attach_entities` is triggered when leads are used, not sure if they apply anywhere else yet. As a result, `entityAttach` and `entityDetach` do not work as described in the API. I wrote in that issue:
> According to the [docs](https://minecraft-data.prismarine.js.org/?d=protocol&v=1.9), it looks like set_passengers was added in 1.9. After testing, attach_entity works as expected in version 1.8, but fails to work after 1.9 when set_passengers took over.

It turns out that in 1.8, `attach_entity` works even when the boat is destroyed. Meanwhile, `set_passengers`  is _not_ triggered when a boat is destroyed.  Switching in between boats/vehicles will update `entity.vehicle` until the vehicle's id is -1, in which case `entity.vehicle` is deleted. 

## What did you try yet?
I haven't tried anything, but I have several ideas that I want some feedback on. 

## Your current code
```js
bot.on('entityAttach', (passenger, vehicle) => {
  console.log('ATTACH')
  console.log(passenger.name)
  console.log(vehicle.name)
})

bot.on('entityDetach', (p, v) => {
  console.log('DETACH')
  console.log(p.name)
  console.log(v.name)
})
```

## Expected behavior
'DETACH' and 'ATTACH' to appear in console after entities change vehicles near the bot.

## Additional context
This is quite the annoying issue since `attach_entity` works well with this functionality quite easily, but `set_passengers` does not. `attach_entity` passes a -1 in the case there is no longer a vehicle, whereas post 1.8, `set_passengers` is not called at all when a vehicle is destroyed. Additionally, post 1.8, set_passengers does not have a way by itself that I can see where its previous state of passengers is known, since it only holds the id of the vehicle in addition to the list of passengers as of the time of the packet being sent.

Therefore, there are a few ways of dealing with this that I can think of:

[...]

PR #2820, opened 2022-10-28
Added attachEntity and detatchEntity functionality to versions past 1.8 #2819

drakemj:
As mentioned in #2819, just added an event for `set_passengers` in entities.js. Attaching and detaching now works as expected as far as triggering the events; however, the return values only include the vehicle. This is because `set_passengers` itself has no way of tracking the previous passenger state of the vehicle unlike `attach_entity`, and thus there is no way with just that function to be able to tell which entity has been attached/detached in certain cases.

### DIFF
--- a/lib/plugins/entities.js
+++ b/lib/plugins/entities.js
@@ -354,6 +354,27 @@ function inject (bot) {
     }
   })
 
+  bot._client.on('set_passengers', (packet) => {
+    //set_passengers for versions past 1.8
+    const entity = fetchEntity(packet.entityId)
+    if (!entity.passengerCount){
+      entity.passengerCount = packet.passengers.length
+      bot.emit('entityAttach', entity)
+    }
+    else if (packet.passengers.length === 1){
+      entity.passengerCount = 1
+      bot.emit('entityDetach', entity)
+    }
+    else if (!packet.passengers.length){
+      delete entity.passengerCount
+      bot.emit('entityDetach', entity)
+    }
+    else {
+      entity.passengerCount = 2
+      bot.emit('entityAttach', entity)
+    }
+  })
+
   bot._client.on('entity_metadata', (packet) => {
     // entity metadata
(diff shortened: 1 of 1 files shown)


lib/plugins/entities.js
```diff
@@ -354,6 +354,24 @@ function inject (bot) {
     }
   })
 
+  bot._client.on('set_passengers', (packet) => {
+    // set_passengers for versions past 1.8
+    const entity = fetchEntity(packet.entityId)
+    if (!entity.passengerCount) {
```
rom1504, line 360: could you add a comment explaining these branches ? it's highly non intuitive

lib/plugins/entities.js
```diff
@@ -354,6 +354,24 @@ function inject (bot) {
...
+    // set_passengers for versions past 1.8
+    const entity = fetchEntity(packet.entityId) // entity of the vehicle
+    if (!entity.passengerCount) { // if passengerCount DNE, create it. entity must be attaching
+      entity.passengerCount = packet.passengers.length
+      bot.emit('entityAttach', entity)
+    } else if (packet.passengers.length === 1) { // if pasengerCount exists + passengers array is len 1, must be detaching
+      entity.passengerCount = 1
+      bot.emit('entityDetach', entity)
+    } else if (!packet.passengers.length) { // if passengers array is empty, detaching
+      delete entity.passengerCount
```
extremeheat, line 367: Why the need to delete and not leave this value at 0?

lib/plugins/entities.js
```diff
@@ -354,6 +354,24 @@ function inject (bot) {
...
+      bot.emit('entityDetach', entity)
+    } else { // else passengers array must have two members and must be attaching
+      entity.passengerCount = 2
```
extremeheat, line 370: Why not leave this as passengers.length?

extremeheat (changes requested):

rom1504:
too old

PR #2821, opened 2022-10-30
Update to support validating 1.19.2 minecraft-data protocol.json

jtsiskin:


lib/plugins/chat.js
```diff
@@ -131,10 +131,19 @@ function inject (bot, options) {
       const { profileKeys } = bot.players[sender]
       if (profileKeys) verified = bot._client.verifyMessage(profileKeys.publicKey, packet)
     }
-    const parameters = {
-      sender: JSON.parse(packet.senderName),
-      content: JSON.parse(message)
+    let parameters = {};
```
Kashalls, line 134: This code is messy.

Consider doing something like:
```js
const parameters = {
    sender: JSON.parse(packet.networkName ? packet.networkName: packet.senderName),
    content: JSON.parse(packet.networkName ? { "text": packet.plainMessage } : message)
  }

rom1504: I think the code in the PR is better actually. Makes more sense to branch on this once.
Could maybe be improved by using support feature though

jtsiskin: This commit in general is just "here's the minimum diff to let us verify the minecraft-data changes and make it easier to continue development"

liuk3 (approved):

lib/plugins/chat.js
Shoro123456789: how long would it take  for this bot to update to 1.19.2

rtm516:
Fixed /tell in https://github.com/jtsiskin/mineflayer/pull/1

rom1504:
this is not working

rom1504:
another pr doing it

PR #2833, opened 2022-11-15
Fix unhandled promise rejection in onceWithCleanup

IceTank:
From MDN:

> Returns an equivalent Promise with its finally handler set to the specified function. If the handler throws an error or returns a rejected promise, the promise returned by finally() will be rejected with that value instead

So finally will throw an unhandled promise rejection if the task promise rejects. Causing a un-catchable promise rejection on the onceWithCleanup function

### DIFF
--- a/lib/promise_utils.js
+++ b/lib/promise_utils.js
@@ -67,7 +67,7 @@ function onceWithCleanup (emitter, event, { timeout = 0, checkCondition = undefi
     })
   }
 
-  task.promise.finally(() => emitter.removeListener(event, onEvent))
+  task.promise.finally(() => emitter.removeListener(event, onEvent)).catch((err) => {})
 
   return task.promise
 }
--- a/lib/promise_utils.js
+++ b/lib/promise_utils.js
@@ -67,7 +67,7 @@ function onceWithCleanup (emitter, event, { timeout = 0, checkCondition = undefi
     })
   }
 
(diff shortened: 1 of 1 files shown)


lib/promise_utils.js
```diff
@@ -67,7 +67,7 @@ function onceWithCleanup (emitter, event, { timeout = 0, checkCondition = undefi
     })
   }
 
-  task.promise.finally(() => emitter.removeListener(event, onEvent))
+  task.promise.finally(() => emitter.removeListener(event, onEvent)).catch((err) => {})
```
extremeheat, line 70: This does not seem right. You're silently letting this fail.

IceTank: The promise returned by finally is a copy off the promise the task itself will return. This is why there is a unhandled promise rejection in the first place. No one is handling the promise returned by finally because it is also returned by the promise in task.

IceTank: Altho this does not work with the linter. Maybe `promise.catch(() => {}).finally(() => ...)` would be better?

extremeheat: You’re right, I missed that. It’s a chained promise, so it should be catching error removing the event (if I’m not mistaken)

IceTank:
@extremeheat Changed it

extremeheat:
Why did you change it ? I said you were right.

IceTank:
I changed it because the old version caused a lint error. This version works.

PR #2894, opened 2023-01-15
1.19.2 support

### LINKED ISSUE #2797: Support 1.19.1/1.19.2

Would it be possible to upgrade Mineflayer from Protocol Version 759 to Protocol Version 760? (1.19 to 1.19.1/1.19.2)

Probably the main change would be secure chat here as 1.19 changes are already implemented

## Additional context

n/a

frej4189:
Depends on ~~https://github.com/PrismarineJS/minecraft-data/pull/670~~

Depends on ~~PrismarineJS/node-minecraft-protocol#1055~~ and ~~PrismarineJS/node-minecraft-protocol#1054~~ for full 1.19.2 support.

Mineflayer part is done

- Chat parsing using NMP convenience events `playerChat` and `systemChat`
- Chat sending using NMP convenience method `.chat`
- Add previousMessages field when sending chat_command packet for 1.19.2
- Update node-minecraft-protocol version in package.json
- No breaking changes to current API

Not sure if anything else needs to be done

Closes #2797


rom1504:
* add 1.19.2 to tested versions
* fix lint
* fix tests

frej4189:
> * [ ]  add 1.19.2 to tested versions
> * [ ]  fix lint
> * [ ]  fix tests

Lint and tests should be okay. Just needs to re-run CI

rom1504:
ok, need to add 1.19.2 in tested versions there 

https://github.com/PrismarineJS/mineflayer/blob/master/lib/version.js#L3 there
https://github.com/PrismarineJS/mineflayer/blob/master/.github/workflows/ci.yml#L33 and there

lib/plugins/chat.js
```diff
@@ -108,43 +108,52 @@ function inject (bot, options) {
...
-    if (sender) {
-      const { profileKeys } = bot.players[sender]
-      if (profileKeys) verified = bot._client.verifyMessage(profileKeys.publicKey, packet)
-    }
-    const parameters = {
-      sender: JSON.parse(packet.senderName),
-      content: JSON.parse(message)
+      if (message) content = JSON.parse(message)
+      else if (options.allowInsecureChat && data.unsignedContent) content = JSON.parse(data.unsignedContent)
+      else content = { text: data.plainMessage }
```
extremeheat, line 120: Instead of this, you can emit both the message and the unsigned content for users that want to get access to the data. When we implemented this initially the idea was that only one would be relevant -- but if both are valid then the users can handle which message to look at. To make it not breaking you can add a .unsignedContent to the emitted ChatMessage

extremeheat (changes requested):

lib/plugins/chat.js
frej4189: If we're just attempting to mirror the vanilla client then only one will be valid.

extremeheat: Yes, one will be the default but the other still available from the data. Only should be unexposed if it's one or the other (ie this or that)

rom1504:
CI failed for 1.19 and 1.19.2

rom1504:
Chat related failures. Before each of external and chat test of internal

frej4189:
Bot gets kicked because server can't validate chat. Reason seems to be "unknown message". Gonna have to look further into what that might mean

lib/plugins/chat.js
frej4189: I have (more or less) done what you requested, but I honestly have to disagree. For NMP, sure, we should expose everything, but for Mineflayer we should really only expose what the Vanilla client would expose.

frej4189:
Depends on https://github.com/PrismarineJS/node-minecraft-protocol/pull/1060

lib/plugins/chat.js
extremeheat: mineflayer is mainly to serve as a higher level abstraction from the lower level networking things, yes. But that doesn't mean valuable information shouldn't be exposed or put behind flags. If there's an stable API to get at the data, then in this case think it would make sense as an [underscore method](https://github.com/PrismarineJS/mineflayer/blob/master/docs/unstable_api.md). For features like this, it can be left to the user libraries to do what they need. 

For example, in prismarine-web-client we might want to show the unsigned content when a user clicks on a message for additional context. Shouldn't be hard to do or require special config on the bot / listening to raw packets. The previous `message: signed || unsigned` was put under the assumption one would be empty and one not, so user code didn't need to worry about it. As you mention in the nmp PR, both these fields are set messages, but one may be server formatted and the other the players' original message. 

For simplicity/compatibility we want to put out the message that would appear to a vanilla client. But we can also send out the original message in a documented field in the chat object (or emit it as a new parameter).

extremeheat (changes requested):

lib/plugins/chat.js
frej4189: I guess if anyone can find a usecase for it as you have here, then let's implement it. I would just personally refer to things that aren't accessible in the Vanilla client as "low-level" and Mineflayer is not a low-level library.

Gonna push a new commit to implement the changes but I do not think it makes sense.

extremeheat: Sure, we can just not expose this. My main point of objection is there should not be a special flag that changes the presence/output of a event. There should just be a new field to the object/event. 

The flags we added to nmp are required because they impact login procedure and can't be changed later on. For example, disableChatSigning on the client skips a call to pauth for dealing with certs. enforceSecureProfile on the server affects the login procedure for clients entering.

frej4189: By doing it that way we leave chat formatting to higher-level libraries which I think is not optimal. I can get behind the way it is implemented currently, but if we decide to just emit multiple messages and let libraries choose which to select adds unnecessary complication.

I vote we keep it this way. It's the best of both worlds

extremeheat: Not a new event, in the above reply I meant a new field or parameter to the ChatMessage object we're emitting.

frej4189: Alright. Then we're on the same page. That is how it is now

rom1504:
CI failing

lib/plugins/chat.js
```diff
@@ -108,43 +108,54 @@ function inject (bot, options) {
...
-    const sender = bot.uuidToUsername[packet.senderUuid]
-    if (sender) {
-      const { profileKeys } = bot.players[sender]
-      if (profileKeys) verified = bot._client.verifyMessage(profileKeys.publicKey, packet)
-    }
-    const parameters = {
-      sender: JSON.parse(packet.senderName),
-      content: JSON.parse(message)
+      if (message) content = JSON.parse(message)
+      else if (options.allowInsecureChat && data.unsignedContent) content = JSON.parse(data.unsignedContent)
```
extremeheat, line 119: > options.allowInsecureChat

Remove per prior comment

lib/plugins/chat.js
```diff
@@ -108,43 +108,54 @@ function inject (bot, options) {
...
+      }
+
+      const parameters = {
+        sender,
+        content
+      }
+      msg = ChatMessage.fromNetwork(data.type, parameters)
+      Object.assign(msg, parameters)
+      if (data.unsignedContent) msg.unsignedContent = JSON.parse(data.unsignedContent)
+      msg.plainMessage = data.plainMessage
```
extremeheat, line 142: `msg.plainMessage` shouldn't be necessary, that should be the primary message

extremeheat (changes requested):
Since there are changes, these should be documented (like addition of .unsigned field)

rom1504:
1.19.2 is surprisingly passing

Was there no change besides for chat ?

lib/plugins/chat.js
frej4189: msg.plainMessage will not be the primary message when formattedMessage is defined. Regardless of whether we make the changes you requested or not

frej4189:
> 1.19.2 is surprisingly passing
> 
> Was there no change besides for chat ?

Not anything that is not already accounted for in NMP

lib/plugins/chat.js
frej4189: Without this check, the behaviour of chat patterns (due to relying on the messagestr event) will be unpredictable as you can't compare to Vanilla.

lib/plugins/chat.js
extremeheat: Yeah, we can set the plainMessage to always be the default

frej4189: What do you mean?

lib/plugins/chat.js
frej4189: Removing for now. I still think we should keep it, but let's see what happens.

lib/plugins/chat.js
frej4189: It makes more sense to expose it.

lib/plugins/chat.js
```diff
@@ -134,12 +128,13 @@ function inject (bot, options) {
 
       const parameters = {
         sender,
-        content
+        content: message ? JSON.parse(message) : { text: data.plainMessage }
```
frej4189, line 131: Handling it this way breaks chat patterns

extremeheat: Which chat pattern? Do you mean passing to ChatMessage.fromNetwork ? Shouldn't be a problem if it's already formatted

extremeheat: Or are you saying we can skip the call to fromNetwork?

frej4189: If we don't consider the unsigned content then we might miss server formatting

frej4189: Without having tested it, I can say pretty confidently that most non-Vanilla server implementations will send the message as unsigned content.

extremeheat: There is another prismarine chat message in the .unsigned field. Can you explain the formatting issue here?

frej4189: msg.toString() will not be the expected message in a lot of cases so `messagestr` event will emit an "unpredictable" message. That means chat patterns added with `addChatPattern` will not necessarily match messages as they appear in the Vanilla Client. And a lot of messages can't possibly be matched because we're only accounting for the raw message

frej4189: If we don't want to parse unsigned contents internally in Mineflayer then we shouldn't parse formattedMessage either

extremeheat: Under the previous commit these unsigned messages would not show either unless you set a special option in the bot, so not an optimal fix there.

frej4189: I did remove that check in my latest commit

frej4189: See https://github.com/PrismarineJS/mineflayer/pull/2894/commits/0229bd73ab230d2b6de77058007c94d7f9460f95

frej4189: I still believe the check should be there though, as that would make it easier to simulate the Vanilla client (so you don't have to disable secure chat only in vanilla MC to mimic mineflayer behaviour)

frej4189: But we can always add that check later so let's not take that discussion in this PR.

extremeheat:
According to some user testing it works on vanilla/paper servers and Hypixel but is failing on servers with velocity. Either kicking on join or when sending a chat message.  Possibly an issue with login procedure and/or the chat signing, as it happens on the first sent message

rom1504:
https://github.com/PrismarineJS/minecraft-data/pull/651#issuecomment-1306240713 packets that changed in 1.19.2

It's really unlikely none of these changes need adaptation in mineflayer

frej4189:
> https://github.com/PrismarineJS/minecraft-data/pull/651#issuecomment-1306240713 packets that changed in 1.19.2
> 
> It's really unlikely none of these changes need adaptation in mineflayer 

The chat packets should all be covered by NMP (except for chat suggestions, I will check whether we will have to account for that here). The explosion packet is covered by NMP/protodef. The respawn packet did not actually change as far as I can see. I am on my phone though, so might be wrong.

Multi block change just changed the meaning of the last Boolean. I don't think we actually use that as the Update Light packet has all the information needed.

The changes to entity metadata should not affect mineflayer as the parsing is already very sparse.

rom1504:
Sounds good.

Last thing to do here is do try booting a few of the examples to make sure
things still work. I usually look at jumper and inventory examples

frej4189:
Depends on https://github.com/PrismarineJS/node-minecraft-protocol/pull/1064

rom1504:
just tried to boot examples/inventory.js and ask for `list` command. Got this:
```
media/documents/Documents/programmation/interlangage/minecraft/mineflayer/node_modules/prismarine-chat/index.js:390
      const fstr = format.formatString
                          ^

TypeError: Cannot read properties of undefined (reading 'formatString')
    at Function.fromNetwork (/media/documents/Documents/programmation/interlangage/minecraft/mineflayer/node_modules/prismarine-chat/index.js:390:27)
    at Client.<anonymous> (/media/documents/Documents/programmation/interlangage/minecraft/mineflayer/lib/plugins/chat.js:133:25)
    at Client.emit (node:events:527:28)
    at Client.<anonymous> (/media/documents/Documents/programmation/interlangage/minecraft/mineflayer/node_modules/minecraft-protocol/src/client/chat.js:140:12)
    at Client.emit (node:events:527:28)
    at FullPacketParser.<anonymous> (/media/documents/Documents/programmation/interlangage/minecraft/mineflayer/node_modules/minecraft-protocol/src/client.js:91:12)
    at FullPacketParser.emit (node:events:527:28)
    at addChunk (/media/documents/Documents/programmation/interlangage/minecraft/mineflayer/node_modules/protodef/node_modules/readable-stream/lib/_stream_readable.js:298:12)
    at readableAddChunk (/media/documents/Documents/programmation/interlangage/minecraft/mineflayer/node_modules/protodef/node_modules/readable-stream/lib/_stream_readable.js:280:11)
    at FullPacketParser.Readable.push (/media/documents/Documents/programmation/interlangage/minecraft/mineflayer/node_modules/protodef/node_modules/readable-stream/lib/_stream_readable.js:241:10)
```

rom1504:
seems bot.chat not working

rom1504:
nevermind, fixed by updating deps

rom1504:
getting some `[01:41:06] [Server thread/INFO]: inventory lost connection: Invalid move player packet received`

rom1504:
fixed by merging and releasing pphysics

rom1504:
jumper, inventory and digger work
good enough

PR #2907, opened 2023-01-25
add example for using the node:readline module

Jovan-04:


Jovan-04:
double checked the bot on vanilla servers for both 1.19 & 1.19.2. works as intended with no issues.

rom1504:
we have too many examples ,let's start doing things right

can you create a subfolder like examples/cli and put your example there ?

Jovan-04:
can do. I think there's a repl example that can go in there too. 

also, i bet there's a better way to organize the examples we have now, along with making sure they're all up-to-date. i'll think on that. 

i'd like to contribute, but i don't think i'm experienced with mineflayer or js to do any serious work on the source code.

rom1504:
Moving the examples would definitely be appreciated

Someone tried once but didn't finish

Also having a doc page pointing to relevant example for relevant features would help

But first let's finish this PR

Jovan-04:
i think that should do it

rom1504:
Yes feel free to do in another PR

PR #2908, opened 2023-01-25
Added missing colon in example code

MightyPiggie:


MightyPiggie:
Whoops. It should be a colon that's commented

PR #2913, opened 2023-01-28
Added missing type value

sefirosweb:
Adding missing type value,

Is used on transfer but not are in types, this value is optinal:

https://github.com/PrismarineJS/mineflayer/blob/4734f0e808b52c762bc22380ecd78aa9135a6e37/lib/plugins/inventory.js#L228-L243


extremeheat (approved):

extremeheat:
Thanks!

PR #2915, opened 2023-01-29
Improve crafting stacks

sefirosweb:
Improve a bit the speed to craft items, (only for stacks)

One of the issue when the bot is craftin every item is crafter they open and close the windows, this is very slow,

Well when you use a ```bot.craft(plankRecipe, 64, craftingTable)``` they open and close the windows 64 times,
I fixed than only open 1 time for craft

If you call 64 times ```bot.craft(plankRecipe, 1, craftingTable)``` the bot still slow, i hope fix them in next pr

I attach video of the improved speed: https://youtu.be/J5FAd-IuCGc


sefirosweb:
I have another PR ready to prevent windows from opening on each craft item when the items are different, but I need to merge this PR first

I think this will improve performance a bit and it should also be more "clean" to avoid creating 1000 windows when you can do everything in 1 shot

Ah-Zheng:
Is this branch merged to master sucessful ?

Ah-Zheng:
I got a problem.  When I use for loop to craft only one can done, but too slow.
```
for (let i = 0, i < times; i += 1) {
  await this.mcBot.craft(this.recipe!, 1, this.craftingTable!);
}
```
But when I just send the times to the count param, it will not realy working, it may craft something different from my target
```
await this.mcBot.craft(this.recipe!, times, this.craftingTable!);
```

for example: I want to composite quartz_block, but when I put 64 to count param, then it have the opportunity to get  
24 quartz_bricks or 2 quartz_pillar and some quartz_block

I need more advice Thanks！

sefirosweb:
I did a several test and you are right,

its not working propely:
```js
      await bot.craft(recipe, 64, craftingTable);
```
It is a bit curous, if i use debug mode it works fine, but when I use a is withouth debuging (production) they don't craft the correct stack

I investigate and i tell you something

Issue #2574, opened 2022-04-13
Craft API 


luizhrios:
- [x] The [FAQ](https://github.com/PrismarineJS/mineflayer/blob/master/docs/FAQ.md) doesn't contain a resolution to my issue 

## Versions
 - mineflayer: 4.2.0
 - server: Idk
 - node: v14

## Detailed description of a problem
Not sure if it's a bug or a feature but it's about the craft API.

The bot.craft does exactly like it's on the recipe but what if I want to put more ingredients and get more results on one single "craft"?
I know the bot.craft has a count parameter but it just loops the "craftOnce" and craft 1 by 1, which is a problem in my case because if I do like count = 64, the bot just hangs at some point and crashes, probably sending too much packets and after it restarts, I can see the bot crafted 10~20 times, its slower than doing manually.

Imagine 1 Bone to 3 Bone Meal and I want to craft the whole stack of Bone

## What did you try yet?
Tried using bot.craft with a high count and as it does 1 by 1, it hangs after sending too much packets to the server.
Tried changing the recipe ingredients count, gets ignored.
Tried changing the clickWindow method to put the whole stack instead 1 single item, no effect.

## Your current code
```js
let boneMealRecipe = bot.recipesFor(mcData.itemsByName["bone_meal"].id);
if(boneMealRecipe.length)
	bot.craft(boneMealRecipe[0], 64);
```

## Expected behavior
Be able to craft more result items in one craft instead of using the exact ingredients on the recipe and loop 1 by 1

Zangoi1:
Unfortunately I can't provide you with a perfect answer because I barely know what I'm doing but I got just got this exact scenario to work with my bot. 

```
bot.moveSlotItem(36, 1) // Move hotbar slot 1 (slot 36 in inventory) to slot 1 of crafting bar slot
await new Promise(resolve => setTimeout(resolve, 1000)) // hacky wait
bot.clickWindow(0, 0, 1) // Click crafting result slot WITH mode 1, click id 0. This results in a shift click [reference](https://wiki.vg/index.php?title=Protocol&oldid=7368#Click_Window)
```

Now this doesn't work by default, because in the prismarineJS project, different modes for clicking in windows is not implemented. See issues: [1](https://github.com/PrismarineJS/mineflayer/pull/1812) [2](https://github.com/PrismarineJS/mineflayer/issues/1937) and [3](https://github.com/PrismarineJS/mineflayer/pull/1812)

[...]

Issue #257, opened 2015-04-06
Crafting is slow


demipixel:
I've had experience and @roblabla suggested that I make a post about since he believes we're doing a lot of unnecessary work.

In some cases I can craft faster than the bot.

roblabla:
I believe we shouldn't be waiting after each click for the transaction to complete. Instead, we could send all clicks at once (except the final one), check if we got any errors, and if we didn't send the final click.

rom1504:
Yeah I guess we could optimize it. It might even not be very hard since nobody tried yet afaik (craft was broken not long ago).

rom1504:
related to #129

rom1504:
244fc7b76eb2a08d54411b5bd6e02d10cdb4aa61 improves bot.transfer by left clicking when possible instead of always right clicking, bot.craft could probably be improved in a similar way (using different kind of clicks)

FeldrinH:
In minecraft 1.12 and above, fast crafting could be achieved by using [craft recipe request packet](https://wiki.vg/Protocol#Craft_Recipe_Request) to move items into the crafting grid (multiple consecutive packets can be sent to increase the number of items in the crafting grid), and a shift click to withdraw all items at once.

sefirosweb:
Checking this issue, the slow performance comes for waiting to open window,
I measured all steps to craft items and this step are the 98% of lost time

https://github.com/PrismarineJS/mineflayer/blob/20862737c733fecee92b78fa12ea360003126290/lib/plugins/craft.js#L24

What I don't quite understand is that the FIRST execution takes only 22ms,

But if another crafting is requested shortly (example x 64 items) the second request takes 900ms~ why? I think it is due to a windowOpen failure

I attach the perform test:

![image](https://user-images.githubusercontent.com/20754836/215291448-e38dce4d-3966-4f9d-b7f6-56f8fc51a4a6.png)

Of course it could be optimized to not close the window and finish crafting, but I think it would be important to see why it takes so long to open the window

rom1504:
there are at least 2 PRs solving this issue but they didn't get finished

you could pick one and finish it

sefirosweb:
@rom1504 yep i'm on it, only are notifying on people if know why happens to save time =P

sefirosweb:
fixed bug! now flies!

PR #2916, opened 2023-01-29
Improve look and fix bug slow craft

sefirosweb:
Hi guys i found the issue why crafting are too slow,

The issue comes because is executing a lot of time ```bot.activateBlock```
This command execute bot.look, 

But when are crafting 1 times is fine, but when you craft a alot of items they are executed one time for craft,
The issue comes because this function is added on task, and need to wait to execute task to finish craft, this delays 1 second for craft,

I fixed easly, if the look are already looking to point then no eed to create new task, and return true,

This also improve the speed for other functions when no need to look the point

Issues related:
https://github.com/PrismarineJS/mineflayer/issues/2574
https://github.com/PrismarineJS/mineflayer/issues/257

Note: one of the bug "still" are opening and closing the windows each time to craft, i think i will make another pr to fix them, but is a different issue

### DIFF
--- a/lib/plugins/physics.js
+++ b/lib/plugins/physics.js
@@ -235,6 +235,10 @@ function inject (bot, { physicsEnabled }) {
     const yawChange = Math.round((yaw - bot.entity.yaw) / sensitivity) * sensitivity
     const pitchChange = Math.round((pitch - bot.entity.pitch) / sensitivity) * sensitivity
 
+    if (yawChange === 0 && pitchChange === 0) {
+      return true
+    }
+
     bot.entity.yaw += yawChange
     bot.entity.pitch += pitchChange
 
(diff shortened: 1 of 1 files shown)


lib/plugins/physics.js
```diff
@@ -235,6 +235,10 @@ function inject (bot, { physicsEnabled }) {
     const yawChange = Math.round((yaw - bot.entity.yaw) / sensitivity) * sensitivity
     const pitchChange = Math.round((pitch - bot.entity.pitch) / sensitivity) * sensitivity
 
+    if (yawChange === 0 && pitchChange === 0) {
+      return true
```
rom1504, line 239: Make it return, not return true

sefirosweb: Done

PR #2921, opened 2023-01-31
*: update dep and make tweaks

VoltrexKeyva:
- Update the `typed-emitter` dependency to it's latest version.
- The `BotEvents` interface is supposed to be a type instead of an interface, required by the latest versions of the `typed-emitter` dependency.
- Set the `esModuleInterop` compiler option to `true` required to import type definitions of some of the updated dependencies.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -40,7 +40,7 @@ export interface PluginOptions {
 
 export type Plugin = (bot: Bot, options: BotOptions) => void
 
-interface BotEvents {
+type BotEvents = {
   chat: (
     username: string,
     message: string,
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "prismarine-windows": "^2.5.0",
     "prismarine-world": "^3.6.0",
     "protodef": "^1.14.0",
-    "typed-emitter": "^1.0.0",
+    "typed-emitter": "^2.1.0",
     "vec3": "^0.1.7"
   },
   "devDependencies": {
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,7 +1,8 @@
 {
     "compilerOptions": {
       "module": "commonjs",
-      "strictNullChecks": true
+      "strictNullChecks": true,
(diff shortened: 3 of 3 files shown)


extremeheat:
Bumping the versions here will have no effect on installs as npm uses semantic versioning and we use the `^` designator.  Read https://docs.npmjs.com/about-semantic-versioning

The event emitter change is inappropriate as you lose validity checking, such as typos in the event name.

As for the type config changes, the `tsconfig.json` is used for the linter (ts-standard), lint CI seems passing so can you explain the need for the change?

VoltrexKeyva:
> Bumping the versions here will have no effect on installs as npm uses semantic versioning and we use the `^` designator.  Read https://docs.npmjs.com/about-semantic-versioning

I'm well aware, I only updated them along with the other ones that do have an effect just for convenience.

> The event emitter change is inappropriate as you lose validity checking, such as typos in the event name.

Thank you for pointing that out, I misunderstood the change the new versions of `typed-emitter` requires, pushing a fix for that.

> As for the type config changes, the `tsconfig.json` is used for the linter (ts-standard), lint CI seems passing so can you explain the need for the change?

The TypeScript compiler chokes when trying to import the type definitions of some of the other modules because they're exported with `export =`, the lint CI passes because I made that change.

extremeheat (approved):

rom1504:
> I'm well aware, I only updated them along with the other ones that do have an effect just for convenience.

So we don't know what change is useful since you changed too many

VoltrexKeyva:
> > I'm well aware, I only updated them along with the other ones that do have an effect just for convenience.
> 
> So we don't know what change is useful since you changed too many

I can update the commit message to clarify what dependency updates are actually useful/does have an effect if that's better.

rom1504:
Simply undo the unneeded updates

VoltrexKeyva:
> Simply undo the unneeded updates

Done.

rom1504:
there are conflicts

PR #2922, opened 2023-01-31
meta: update GitHub workflow actions

VoltrexKeyva:
- Update the GitHub workflow actions to their latest versions.
- Switch to the `softprops/action-gh-release` action as the `actions/create-release` action is no longer maintained.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,21 +11,22 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
     - name: Use Node.js 14.x
-      uses: actions/setup-node@v1.4.4
+      uses: actions/setup-node@v3
       with:
-        node-version: 14.x
+        node-version: 14
+        check-latest: true
     - run: npm i && npm run lint
 
   MinecraftServer:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        javaVersion: [1.8]
+        javaVersion: [8]
         mcVersion: ['1.8.8', '1.9.4', '1.10.2', '1.11.2', '1.12.2', '1.13.2', '1.14.4', '1.15.2', '1.16.5']
         include:
-         - javaVersion: 16
+         - javaVersion: 17
            mcVersion: '1.17.1'
          - javaVersion: 17
            mcVersion: '1.18.2'
@@ -38,16 +39,19 @@ jobs:
       fail-fast: false
 
     steps:
-      - uses: actions/checkout@v2
-      - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v1.4.4
+      - uses: actions/checkout@v3
+      - name: Use Node.js 14
+        uses: actions/setup-node@v3
         with:
-          node-version: 14.x
+          node-version: 14
+          check-latest: true
       - name: Setup Java JDK
-        uses: actions/setup-java@v1.4.3
+        uses: actions/setup-java@v3
         with:
+          distribution: temurin
           java-version: ${{ matrix.javaVersion }}
           java-package: jre
+          check-latest: true
       - name: Install Dependencies
         run: npm install
       - name: Start Tests
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -9,11 +9,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Checkout repository
-      uses: actions/checkout@master
+      uses: actions/checkout@v3
     - name: Set up Node.js
-      uses: actions/setup-node@master
+      uses: actions/setup-node@v3
       with:
-        node-version: 14.0.0
+        node-version: 14
+        check-latest: true
     - id: publish
       uses: JS-DevTools/npm-publish@v1
       with:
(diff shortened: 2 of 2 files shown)


extremeheat:
Why?

VoltrexKeyva:
For some reason the actual CI isn't running, I don't see why.

VoltrexKeyva:
> Why?

To updating the actions?

1. The older versions use outdated commands and requests that GitHub has deprecated (such as the `set-output` command being used by the `actions/*` actions) and will soon be disabled, rendering the actions unusable.
2. Updating the actions can also improve the performance of the CI overall.

I don't see a reason to not update.

Or is that question forwarded to something else?

VoltrexKeyva:
The actual CI still isn't running, any ideas?

extremeheat:
Ah, I think that makes sense then. We'd probably want to update the other repos also to account for that.

As for the node.js version bump in the CI, like Rom in mentioned https://github.com/PrismarineJS/bedrock-protocol/pull/338 we don't intend to support for older node versions for the time being. Maybe we can move to node 16 at some point but node 18 seems to me too recent as Node 14 is still in maintenance support.

The CI is failing because an error in the workflow syntax. We specifically use different Java versions because different versions of Minecraft require different versions of Java, so this change is not correct.

VoltrexKeyva:
> As for the node.js version bump in the CI, like Rom in mentioned https://github.com/PrismarineJS/bedrock-protocol/pull/338 we don't intend to drop support for older node versions for the time being. Maybe we can move to node 16 at some point but node 18 seems to me too recent as Node 14 is still in maintenance support.

Understandable, I will revert the change to the Node.js version.

> The CI is failing because an error in the workflow syntax.

I don't see any syntax errors, all YAML validators I used also says it's valid...

> We specifically use different Java versions because different versions of Minecraft require different versions of Java, so this change is not correct.

I see, I'll revert the change to the Java version.

VoltrexKeyva:
@extremeheat I got the CI to work although the Java version that was already specified isn't available, the ones that are available is shown in the CI's logs, can you tell me which one is the appropriate one for this case?

.github/workflows/ci.yml
```diff
@@ -38,16 +39,19 @@ jobs:
...
+      - uses: actions/checkout@v3
+      - name: Use Node.js 14
+        uses: actions/setup-node@v3
         with:
-          node-version: 14.x
+          node-version: 14
+          check-latest: true
       - name: Setup Java JDK
-        uses: actions/setup-java@v1.4.3
+        uses: actions/setup-java@v3
```
extremeheat, line 49: This change is likely breaking CI

VoltrexKeyva: In what way? Due to the already specified Java version not being available or?

extremeheat: The v3 implies a breaking change. Reverting this change may make the CI pass if the breakage here[1] is due to the change in API upstream. 

[1] https://github.com/PrismarineJS/mineflayer/actions/runs/4051469496/jobs/6969827556#step:4:1

VoltrexKeyva: The latest versions of the `actions/setup-java` action rely on different distributions mentioned [here](https://github.com/actions/setup-java#supported-distributions).

And I already mentioned that the already specified Java version isn't available in this comment: https://github.com/PrismarineJS/mineflayer/pull/2922#issuecomment-1409849423

So which Java version do you think is appropriate here? You can also check all the other distributions if the one needed isn't available.

extremeheat: What I meant in my comment above was to revert this change. The CI on master is currently passing. It seems likely this change is breaking  CI on this PR.

VoltrexKeyva: I get that, I mean like reverting that change isn't ideal because that the older version of `actions/setup-java` is also subject to this point I made earlier:

> The older versions use outdated commands and requests that GitHub has deprecated (such as the set-output command being used by the actions/* actions) and will soon be disabled, rendering the actions unusable.

That's why I want to update it 😅.

extremeheat: Java 8 should be synonymous with 1.8, you can try replacing 1.8 here with 8

VoltrexKeyva: Got it, though some of the Minecraft server tests are using Java v16 which is generally unavailable since it's not LTS.

Is there a reason Java v17 isn't being used for all of them? I'm pretty sure those Minecraft versions work fine with Java v17.

VoltrexKeyva: The CI passed with Java v17 for that Minecraft version.

rom1504:
If you want to work on this

1. Explain why you want to make this change
2. Update in prismarine-template
3. Update in all PrismarineJs repos

Updating only in mineflayer will only cause confusion

VoltrexKeyva:
> If you want to work on this
> 
> 1. Explain why you want to make this change
> 2. Update in prismarine-template
> 3. Update in all PrismarineJs repos
> 
> Updating only in mineflayer will only cause confusion 

1. I've already explained why I want to make this change. Ref: https://github.com/PrismarineJS/mineflayer/pull/2922#issuecomment-1409766947
2 & 3. Sure that can be done although I'm not sure if I have the time to update all of them, this won't cause confusion and can be done at anytime, at least before the aforementioned deprecations turn into removals/features get disabled.

rom1504:
> Sure that can be done although I'm not sure if I have the time to update all of them, this won't cause confusion and can be done at anytime, at least before the aforementioned deprecations turn into removals/features get disabled.

Makes sense. So let's not do this change here.

PR #2927, opened 2023-02-03
Craft batch

sefirosweb:
Hi guys, to improve craft when you need to craft multiple different items in one shot, to avoid open and close the windows every time, I made a new function:

image
To:
image

Basically craft batch does same as craft but avoiding close the windows while is crafting,
This new function receive an array of recipes and the count of how many items want to craft,

Usage example:

```js
const plankRecipe = bot.recipesFor(mcData.itemsByName.oak_planks.id)[0]

    const recipes = [
      {
        recipe: plankRecipe,
        count: 2
      },
      {
        recipe: plankRecipe,
        count: 2
      },
      {
        recipe: plankRecipe,
        count: 2
      },
      {
        recipe: plankRecipe,
        count: 2
      },
      {
        recipe: plankRecipe,
        count: 2
      },
      {
        recipe: plankRecipe,
        count: 2
      },
      {
        recipe: plankRecipe,
        count: 2
      },
      {
        recipe: plankRecipe,
        count: 2
      },
      {
        recipe: plankRecipe,
        count: 2
      },
      {
        recipe: plankRecipe,
        count: 2
      },
      {
        recipe: plankRecipe,
        count: 2
      },
      {
        recipe: plankRecipe,
        count: 2
      },
      {
        recipe: plankRecipe,
        count: 2
      },
      {
        recipe: plankRecipe,
        count: 2
      }
    ]

    bot.craftBatch(recipes, craftingTable)
      .then(() => {
        console.log('Finish to craft')
      })
      .catch((e) => {
        console.log(e)
      })
```

I really think craftBatch should replace the old "craft", but this can cause compatibility problems for users,
It should be done, since having 2 functions that do similar things should not exist (it's my opinion)

I hope this simplify a bit the code of users

If you would like me to create a test for this function let me know.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1983,6 +1983,20 @@ This function returns a `Promise`, with `void` as its argument when the crafting
    use. If the recipe does not require a crafting table, you may use
    `null` for this argument.
 
+#### bot.craftBatch([{recipe, count}], craftingTable)
+
+This function returns a `Promise`, with `void` as its argument when the crafting is complete and your inventory is updated.
+
+Used to create multiple items in 1 request, to use send an array of items with recipe and count
+
+ * `recipe` - A `Recipe` instance. See `bot.recipesFor`.
+ * `count` - How many times you wish to perform the operation.
+   If you want to craft planks into `8` sticks, you would set
+   `count` to `2`.
+ * `craftingTable` - A `Block` instance, the crafting table you wish to
+   use. If the recipe does not require a crafting table, you may use
+   `null` for this argument.
+
 #### bot.writeBook(slot, pages)
 
 This function returns a `Promise`, with `void` as its argument when the writing was successfully or an error occurred.
--- a/docs/es/api_es.md
+++ b/docs/es/api_es.md
@@ -1841,16 +1841,27 @@ Todas las direcciones son relativas a donde está mirando el bot
 
  * `slot` - puede ser de 0 a 8, la casilla de la barra de acceso rápido
 
-#### bot.craft(recipe, count, craftingTable, [callback])
+#### bot.craft(recipe, count, craftingTable)
 
-Esta función también devueve un `Promise`, con `void` como argumento al finalizar.
+Esta función devuelve una `Promise`, tipo `void`.
 
  * `recipe` - Una instancia `Recipe`. Mira `bot.recipesFor`.
  * `count` - Cuantas veces quieres repetir la acción.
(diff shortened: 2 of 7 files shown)


rom1504:
can you measure how much faster it is ?

sefirosweb:
Hi Rom this is not much faster now with the fixes have i been did, we are talking only about of milisseconds

I request this changes to simplify the user code when want to craft multiple different recipes, 

Also I think emit windows open & close for each recipe is not correct

Best

rom1504:
can you please add tests?

rom1504:
@kaduvert are you interested in reviewing/finishing this ? looks similar to your interests

kaduvert:
so i think the problem is with the craft function, which opens and closes a window from a crafting bench block

to me `bot.craft` should take a window as an argument, craft the stuff with it and not do anything more than that.

i think the user should'nt have to create an array of crafting recipes, he should just be able to call `bot.craft` with one recipe and manage the crafting bench window himself

kaduvert:
to conclude:
i think this pr has a valid point, which is that `bot.craft` is opening and closing windows, which it shouldn't.

but i think the path to fixing this is not to implement better management of the window for the user, but giving management of the window completely to the user.

so instead of
```js
const plankRecipe = bot.recipesFor(mcData.itemsByName.oak_planks.id)[0]
const craftingTable = bot.blockAt(...)

bot.craft(plankRecipe, 2, craftingTable)
```
we should let the user obtain the window before letting him craft:
```js
const plankRecipe = bot.recipesFor(mcData.itemsByName.oak_planks.id)[0]
const craftingTable = bot.blockAt(...)

const craftingWindow = bot.activateBlock(craftingTable)

bot.craft(plankRecipe, 1, craftingWindow)

craftingWindow.close()
```

this way he can do multiple recipes and decide himself when to open or close the window

also some servers have commands like `/craft` which open a crafting window directly.
with the current implementation of `bot.craft` or the requested `bot.craftBatch` the user can not make use of this.
but he could if `bot.craft` took a window as argument

rom1504:
makes sense to me

do you want to implement that @kaduvert ?

kaduvert:
maybe.. ^^
let's hear from @sefirosweb on what he thinks, as he's probably most interested.

sefirosweb:
Yes for me have more sense too, 

I would like to help to implement them but right now I'm with other work projects :(

PR #2941, opened 2023-02-12
Made placeBlock easier

SMEDjs:
added option to use face `top`, `bottom`, `east`, `west`, `south` or `north`  or vector

### DIFF
--- a/lib/plugins/place_block.js
+++ b/lib/plugins/place_block.js
@@ -1,9 +1,24 @@
 const { onceWithCleanup } = require('../promise_utils')
+const { Vec3 } = require("vec3")
 
 module.exports = inject
 
 function inject (bot) {
-  async function placeBlockWithOptions (referenceBlock, faceVector, options) {
+  async function placeBlockWithOptions (referenceBlock, face, options) {
+    var faceVector = face;
+    if(typeof face === "string") {
+      const directions = {
+        'top': new Vec3(0, 1, 0), 
+        'bottom': new Vec3(0, -1, 0), 
+        'east': new Vec3(1, 0, 0), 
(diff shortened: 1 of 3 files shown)


lib/plugins/place_block.js
```diff
@@ -1,9 +1,24 @@
...
-  async function placeBlockWithOptions (referenceBlock, faceVector, options) {
+  async function placeBlockWithOptions (referenceBlock, face, options) {
+    let faceVector = face
+    if (typeof face === 'string') {
+      const directions = {
+        top: new Vec3(0, 1, 0),
+        bottom: new Vec3(0, -1, 0),
+        east: new Vec3(1, 0, 0),
+        west: new Vec3(-1, 0, 0),
+        south: new Vec3(0, 0, 1),
```
rom1504, line 15: Isn't this reversed

SMEDjs: it's the direction that the block face is facing at

sefirosweb:
@SMEDjs with your PR can't be use vec3 directions directly? need to be used facevector?  
I think there should be backwards compatibility, many of us use some formula to detect the "best" position to place the block that is not based on its coordinate orientation, if there is simply an adjacent block that helps to place the block

SMEDjs:
> @SMEDjs with your PR can't be use vec3 directions directly? need to be used facevector? I think there should be backwards compatibility, many of us use some formula to detect the "best" position to place the block that is not based on its coordinate orientation, if there is simply an adjacent block that helps to place the block

you can, it checks if it's a string but you can still use vector + if you give me the formula i can add it to the PR

rom1504:
can you use this in a test or example ?

rom1504:

PR #2944, opened 2023-02-14
add brewing stand methods

Jovan-04:
similar to the furnace methods, but for brewing stands  

not ready to merge, but figured i had a good start and would make a pr.  

there are still some problems with different mc versions. pre-1.9 doesn't need blaze powder as fuel for brewing, so the `fuel` property returns the wrong item, and the `takeFuel()` and `putFuel()` methods error out in an unhelpful way.  

post-1.9 also has different issues with the fuel gui. the windowUpdate event doesn't pass through any information about the gui besides the slots, so the current fuel level stays undefined until it gets updated.


extremeheat (commented):
Missing:
* tests
* doc
* example

Jovan-04:
yeah, i still have some more work to do before it'll be fully functional. i'll write up a quick doc today, and i'll make an example once i get everything working. 

what do you mean by tests?

rom1504:
See tests/ folder

Jovan-04:
from what I've been able to tell, the issues are arising from the test I wrote. I'm not sure if that's an issue with the tests themselves, or an issue with the underlying plugin that the tests are just exposing. I've been pretty busy with classes and midterms lately, which is why there hasn't been much progress. as of right now, my plan is to update the test so that it uses `bot.supportFeature` instead of strict version checking (which i think will involve an update to mc-data), then add some more checks in the test.

extremeheat:
You can try commenting out some of the code in the test and see what lines cause the errors, that will help debug the issues

rom1504:
looks like this is not far from merging

@Jovan-04 can you fix 1.8 tests ?

rom1504:
@Jovan-04 any time to finish it now by any chance :) ?

rom1504:
Old, closing

PR #2948, opened 2023-02-16
Persist properties received in player_info packet

Paulomart:
This changes persists the 'properties' property in the player object when received
via the player_info packet. Now its possible to access properties on the player object
like the textures property which contains skin information about the player.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -459,10 +459,16 @@ export interface Player {
   gamemode: number
   ping: number
   entity: Entity
+  skinData: SkinData | undefined
   profileKeys?: {
     publicKey: Buffer
     signature: Buffer
   }
+}
+
+export interface SkinData {
+  url: string
+  model: string | null
 }
 
 export interface ChatPattern {
--- a/lib/plugins/entities.js
+++ b/lib/plugins/entities.js
@@ -462,6 +462,7 @@ function inject (bot) {
         if (packet.action & 1) {
           obj.username = item.player.name
           obj.displayName = player.displayName || new ChatMessage({ text: '', extra: [{ text: item.player.name }] })
+          obj.skinData = extractSkinInformation(item.player.properties)
         }
 
         if (packet.action & 4) {
@@ -510,6 +511,7 @@ function inject (bot) {
               ping: item.ping,
               uuid: item.UUID,
               displayName: new ChatMessage({ text: '', extra: [{ text: item.name }] }),
+              skinData: extractSkinInformation(item.properties),
               profileKeys: item.crypto
                 ? {
                     publicKey: item.crypto.publicKey, // DER-encoded public key
@@ -525,6 +527,7 @@ function inject (bot) {
             // Just an Update
             player.gamemode = item.gamemode
             player.ping = item.ping
+            player.skinData = extractSkinInformation(item.properties)
             if (item.crypto) {
               player.profileKeys = {
                 publicKey: item.crypto.publicKey,
@@ -705,4 +708,26 @@ function parseMetadata (metadata, entityMetadata = {}) {
   }
 
   return entityMetadata
+}
+
+function extractSkinInformation (properties) {
+  if (!properties) {
+    return undefined
+  }
+
+  const props = Object.fromEntries(properties.map((e) => [e.name, e]))
+  if (!props.textures || !props.textures.value) {
+    return undefined
+  }
+
+  const skinTexture = JSON.parse(Buffer.from(props.textures.value, 'base64').toString('utf8'))
+
+  const skinTextureUrl = skinTexture?.textures?.SKIN?.url ?? null
+  const skinTextureModel = skinTexture?.textures?.SKIN?.metadata?.model ?? null
+
+  if (!skinTextureUrl) {
+    return undefined
+  }
+
+  return { url: skinTextureUrl, model: skinTextureModel }
 }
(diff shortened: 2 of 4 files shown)


index.d.ts
```diff
@@ -459,6 +459,11 @@ export interface Player {
   gamemode: number
   ping: number
   entity: Entity
+  properties: Array<{
```
extremeheat, line 462: Can you add to the doc what kind of information is here?

lib/plugins/entities.js
```diff
@@ -510,6 +511,7 @@ function inject (bot) {
               ping: item.ping,
               uuid: item.UUID,
               displayName: new ChatMessage({ text: '', extra: [{ text: item.name }] }),
+              properties: item.properties,
```
extremeheat, line 514: Can we abstract this?

index.d.ts
Paulomart: You mean this? https://wiki.vg/Mojang_API#UUID_to_Profile_and_Skin.2FCape

extremeheat: Yes, mineflayer should not be returning arbitrary server that can change and break user code. Should be abstracted as much as possible

Paulomart: @extremeheat maybe you didnt see my message here?

extremeheat: What would example usage of this PR look like in its current form? What information can you extract from this? If it's skin data, you add a `skin` object to the player object, for example.

Paulomart: I currently use the data in the following form:

```typescript
function extractSkinInformation(player: mineflayer.Player & any): {url: string; model: string} | null {
  if (!player.properties) {
    return null;
  }

  const skinProperty = player.properties.find((property: any) => property.name === 'textures');
  if (!skinProperty) {
    return null;
  }

  const skinTexture = JSON.parse(Buffer.from(skinProperty.value, 'base64').toString('utf8'));

  const skinTextureHash = skinTexture?.textures?.SKIN?.url ?? null;
  const skinTextureModel = skinTexture?.textures?.SKIN?.metadata?.model ?? null;

  if (!skinTextureHash) {
    return null;
  }

  return { url: skinTextureHash, model: skinTextureModel };
}
```

But I find it hard to anticipate any future usage patterns. Even more because Mojang designed the `properties` object in a way that allows further expansion by them. I dont really think that implementing a "skin parser" would be very future proof, it would just add maintenance burden for this project.

extremeheat: > find it hard to anticipate any future usage patterns

This is exactly why we do abstraction (the point of mineflayer over nmp). Likewise, we don't just emit plain packet structures back through mineflayer because they can break functionality from one version to the next. 

If there is a specific field field you're looking to use, that should be exposed, put in the doc and ideally tested (if possible). If it breaks in a future version, our tests should capture it and it can be handled internally in mineflayer to avoid a breaking change for the user.

extremeheat: That function you provided, and its output, seems like what should be in and returned by mineflayer, for example:

```ts
player.skinData = { textureURL: '', model: '' } as { textureURL: string, model: string }
```

Paulomart:
@extremeheat I have updated the implementation to extract the skin data and store in the player object.

Paulomart:
@extremeheat I dont think the test failures are related.

lib/plugins/entities.js
```diff
@@ -462,6 +462,7 @@ function inject (bot) {
         if (packet.action & 1) {
           obj.username = item.player.name
           obj.displayName = player.displayName || new ChatMessage({ text: '', extra: [{ text: item.player.name }] })
+          obj.properties = item.properties
```
extremeheat, line 465: remove

lib/plugins/entities.js
```diff
@@ -510,6 +511,7 @@ function inject (bot) {
               ping: item.ping,
               uuid: item.UUID,
               displayName: new ChatMessage({ text: '', extra: [{ text: item.name }] }),
+              properties: extractSkinInformation(item.properties),
```
extremeheat, line 514: remove

lib/plugins/entities.js
```diff
@@ -706,3 +709,25 @@ function parseMetadata (metadata, entityMetadata = {}) {
 
   return entityMetadata
 }
+
+function extractSkinInformation (properties) {
+  if (!properties) {
+    return undefined
+  }
+
+  const skinProperty = properties.find((property) => property.name === 'textures')
```
extremeheat, line 718: ok, but instead of searching you can also use a map
```js
const props = Object.fromEntries(properties.map(e) => [e.name, e])
props.textures.value ...
```

lib/plugins/entities.js
```diff
@@ -706,3 +709,25 @@ function parseMetadata (metadata, entityMetadata = {}) {
...
+  const skinTexture = JSON.parse(Buffer.from(skinProperty.value, 'base64').toString('utf8'))
+
+  const skinTextureUrl = skinTexture?.textures?.SKIN?.url ?? null
+  const skinTextureModel = skinTexture?.textures?.SKIN?.metadata?.model ?? null
+
+  if (!skinTextureUrl) {
+    return undefined
+  }
+
+  return { url: skinTextureUrl, model: skinTextureModel }
```
extremeheat, line 732: mismatch between doc and types

extremeheat (commented):
Missing doc, left some comments

extremeheat:
Can you add this to an example? Would be great if someone could test this to work.

Paulomart:
@extremeheat done

extremeheat (approved):

PR #2962, opened 2023-03-04
Fix reference error with block updates

IceTank:
Both oldBlock and newBlock in block updates can be null. This was causing some reference errors in mineflayer.

### DIFF
--- a/lib/plugins/place_block.js
+++ b/lib/plugins/place_block.js
@@ -17,7 +17,7 @@ function inject (bot) {
     if (!oldBlock && !newBlock) {
       return
     }
-    if (oldBlock?.type === newBlock?.type) {
+    if (oldBlock?.type === newBlock.type) {
       throw new Error(`No block has been placed : the block is still ${oldBlock?.name}`)
     } else {
       bot.emit('blockPlaced', oldBlock, newBlock)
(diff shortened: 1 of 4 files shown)


lib/plugins/place_block.js
```diff
@@ -13,8 +13,12 @@ function inject (bot) {
       [oldBlock, newBlock] = await onceWithCleanup(bot, `blockUpdate:${dest}`, { timeout: 5000 })
     }
 
-    if (oldBlock.type === newBlock.type) {
-      throw new Error(`No block has been placed : the block is still ${oldBlock.name}`)
+    // blockUpdate emits (null, null) when the world unloads
+    if (!oldBlock && !newBlock) {
+      return
+    }
+    if (oldBlock?.type === newBlock?.type) {
```
extremeheat, line 20: Can new block actually be undefined?

IceTank: Yes. When you register a listener for block updates and the world unloads you receive a block update even with both new block and old block being null. https://github.com/PrismarineJS/mineflayer/blob/8f87e055d9dc6ab58e2efff036be589b3086b970/lib/plugins/blocks.js#L481

IceTank: Actually this looks like it only applies to block update events tied to specific coordinates. The generic blockUpdate event can indeed not have null for newBlock. Ill fix that.

extremeheat (approved):

extremeheat:
Thanks

PR #2968, opened 2023-03-07
Standardize dimensions for all versions

sefirosweb:
Hello guys, i checked when game is lower than 1.19- the game.dimension is returning minecraft:XXX but in 1.19+ is the dimension directly,

I made changes to unify them

### DIFF
--- a/lib/plugins/blocks.js
+++ b/lib/plugins/blocks.js
@@ -275,7 +275,7 @@ function inject (bot, { version, storageBuilder, hideErrors }) {
       bitMap: packet.bitMap,
       heightmaps: packet.heightmaps,
       biomes: packet.biomes,
-      skyLightSent: bot.game.dimension === 'minecraft:overworld',
+      skyLightSent: bot.game.dimension === 'overworld',
       groundUp: packet.groundUp,
       data: packet.chunkData,
       trustEdges: packet.trustEdges,
--- a/docs/api.md
+++ b/docs/api.md
@@ -866,7 +866,7 @@ Whether the bot is using the item that it's holding, for example eating food or 
 
 #### bot.game.dimension
 
-The bot's current dimension, such as `overworld, `the_end` or `the_nether`.
+The bot's current dimension, such as `overworld`, `the_end` or `the_nether`.
 
 #### bot.game.difficulty
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -866,6 +866,8 @@ Whether the bot is using the item that it's holding, for example eating food or 
 
 #### bot.game.dimension
 
+The bot's current dimension, such as `overworld, `the_end` or `the_nether`.
+
 #### bot.game.difficulty
 
 #### bot.game.gameMode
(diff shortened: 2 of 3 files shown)


frej4189:
The namespaces are there for a reason. Don't remove them

extremeheat:
We don't use minecraft namespaces in the PrismarineJS projects (it's implied), so it should probably be removed for consistency.

Also, since the network packet structure and format is unstable, we shouldn't be directly exposing network things 1:1 but expose things that are stable across all versions. That way when new versions come out or when we support bedrock for example we don't need to make needless breaking API changes. So the exact possible outputs of `bot.game.dimension` should be put into the doc and types. This is the expected output according to the types:

```ts
export type Dimension = 'the_nether' | 'overworld' | 'the_end'
```

extremeheat (approved):

sefirosweb:
I'm not clear why the tests are failing, I hardly changed anything, i will check

frej4189:
> We don't use minecraft namespaces in the PrismarineJS projects (it's implied), so it should probably be removed for consistency.

If server sends the names with a namespace identifier attached, then it probably has a reason to do so. Most notably consequent packets (like respawn) will use the same format, so we'd have to mirror it when handling those aswell. Codebase will quickly become a mess if we have to do `replace('minecraft:', '')` each time we are sent a string from the server (obviously exaggerating here; hopefully point comes across). It's best to just save resource tags in the format the server sends them in for parsing purposes.

If you really care for the consistency (which is another discussion) then it'd make more sense to add in an internal dimension storage that is left unmodified and is used for parsing respawn packets and similar.

Also, in the implementation of this PR, namespaces like something_minecraft would be broken as the call to str.replace does not account for minecraft: being in the start of the string. This is of course not a truly relevant concern, but might aswell get those edge cases handled, just in case.

extremeheat:
You changed it in the wrong place (you meant to directly update the `dimensionNames` object used <=1.18). 
It was failing on 1.19+ because prismarine-registry's `dimensionsByName` doesn't use `minecraft:` prefix, and you only replaced the naming after the call to `bot.registry.dimensionsByName[bot.game.dimension]` which would return undefined and threw on destructuring. I've fixed the PR.

lib/plugins/blocks.js
```diff
@@ -275,7 +275,7 @@ function inject (bot, { version, storageBuilder, hideErrors }) {
       bitMap: packet.bitMap,
       heightmaps: packet.heightmaps,
       biomes: packet.biomes,
-      skyLightSent: bot.game.dimension === 'minecraft:overworld',
+      skyLightSent: bot.game.dimension === 'overworld',
```
frej4189, line 278: This should be inferred from the dimension codec

extremeheat: Yeah, that can be addressed in a different PR, let's just normalize the output values for all versions first

sefirosweb:
@extremeheat i need to do something from my part?

extremeheat:
LGTM, it matches the rest of mcdata naming.

As discussed on the discord, dimensions.json still needs to be added to minecraft-data but for now pregistry creates a new `dimensionsByName` field from the login codec that's not present in the static data. But this static data can be added later on.

PR #2985, opened 2023-03-22
Emit inject_allowed after a timeout of 0

IceTank:
The current implementation can in some cases emit the inject_allowed event synchronously within createBot. This makes it impossible to listen to that event on the bot object after createBot. Using setTimeout will schedule the event emitting in the next cycle giving other code outside of createBot enough time to register event listeners for inject_allowed.


IceTank:
Test failure seams unrelated to the change.

PR #2989, opened 2023-03-23
feat(next): migrate to nextjs

AugustinMauroy:
# Feature added

* Reactjs
* Search
* Dark mode
* Edit this page on GitHub
* Button to open an issue

# Think had been remove 

* I18n (temporary)


extremeheat:
What improvement does this add? These is adding lots of complexity to the project over the current repo.

AugustinMauroy:
@extremeheat Take a look to pr description I had an explaining

extremeheat:
I don't see anything in the description minus UI and a search bar. If that's the only improvement I don't think it justifies the added complexity here. It seems to me that should be an improvement made in the markdown to HTML renderer (Jekyll) than adding the build complexity inside this repo.

AugustinMauroy:
https://augustinmauroy.github.io/beta-mineflayer-docs
Take a look to this preview.
It's not the same. But it's can help you to visualize the upgrading.
And also now the markdown are parse to html on client here it's parse at the build. So it's more optimize for the client

williamd5:
The UI looks fantastic (+1 for tailwind). I realise this is probably a theme, however, there are some design issues which IMO might be worth considering:

---

### 1. Main body and nav not aligned

The main body has max-width 6xl, but the navbar doesn't align with that because it uses 90rem

Currently:
image
Could be [(TW play)](https://play.tailwindcss.com/omboY7Mrx5?layout=horizontal):
image

I think `max-w-prose` might be even better precisely for this use case.


### 2. Search focus has white outline on focus
Could fix with `focus:outline-none`
image
### 3. Accent colour low contrast
(e.g. on links) on dark theme at least is too dark and has really low contrast ratio
### 4. The sidebar x scroll
The sidebar on the right has x-axis scroll. Nested items cause scrolling on this axis when you activate them, which looks terrible IMO. You can see here it crops from the text at the bottom:
image

### 5. Inconsistencies
There are some inconsistencies in the UI, e.g. the sidebar items on the left and the theme select menu items. In one case primary colours are used to indicate the active element and in the other instance they are used to just indicate hover state.
### 6. Copyright footer
has huge padding on `y` and would probably look better in the same container as the theme picker

AugustinMauroy:
@williamd5  Thanks for our point of vue. But yes it's an theme so if you want to custom you can create an pr on my fork.

rom1504:
I'm comparing with https://prismarinejs.github.io/mineflayer/#/ and lot of content was removed

PR #2993, opened 2023-03-25
fix type definition in regards to documentation

Moosbee:
As written in the API Docs(https://github.com/PrismarineJS/mineflayer/blob/master/docs/api.md#botfindblocksoptions) `useExtraInfo` can be a boolean or a function, the typings do not reflect that.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -565,7 +565,7 @@ export interface FindBlockOptions {
   matching: number | number[] | ((block: Block) => boolean)
   maxDistance?: number
   count?: number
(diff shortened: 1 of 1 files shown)


IceTank:
useExtraInfo is almost completely useless. All it does is add paintings to a block if one exists in that space. A better approach would be to remove useExtraInfo entirely or deprecate it.

extremeheat:
Maybe that can be discussed in a different PR. But interim let's fix the inconsistency to match the implementation.

extremeheat:
Thanks

PR #3003, opened 2023-04-02
Add window transaction timeout

firejoust:
Fixes a memory leak if the server doesn't respond to a window transaction, causing multiple listeners to be created without the promise resolving (if `bot.clickWindow` is called multiple times)

### DIFF
--- a/lib/plugins/inventory.js
+++ b/lib/plugins/inventory.js
@@ -547,6 +547,9 @@ function inject (bot, { hideErrors }) {
         confirmTransaction(window.id, actionId, true)
       }
       const [success] = await withTimeout(response, WINDOW_TIMEOUT)
+      .catch(() => {
+        throw new Error(`Server didn't respond to transaction for clicking on slot ${slot} on window with id ${window?.id}.`)
+      })
       if (!success) {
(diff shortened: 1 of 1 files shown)


WhoTho:
how does a vanilla client act when not receiving a confirm transaction packet?  if it just assumes its correct then i think adding the timeout would be fine.

add a different error message though, as the server never rejected the click it simply didnt respond. otherwise lgtm

IceTank:
Would not surprise me if vanilla has a memory leak too

IceTank:
Test look like they failed due to unstable tests and not due to the change made.

extremeheat:
There are a lot of potential leaks like this in mineflayer, pretty much everywhere where there's an naked await without a timeout. I'll look at addressing this when we refactor things for bedrock. But this PR LGTM.

extremeheat (approved):

lx30011:
But this doesn't play well with laggy servers

IceTank:
> But this doesn't play well with laggy servers

Make an issue when 5 seconds is not long enough. We van probably make an option to extend it.

Issue #2971, opened 2023-03-13
Disable bot respawn


DaryeDev:
I wanted to leave the bot connected but dead, to use commands but not interfere, but when I kill it, it auto-respawns.

Is there any way to disable auto-respawning?

extremeheat:
I don't believe so, you can open a PR to add this option

MrFiend179:
try this
~~~js
let isBotDead = false;

bot.on('death', () => {
    if (!isBotDead) {
      isBotDead = true;
      console.log("Bot has died.");

      // Prevent respawn
      bot._client.write('client_command', { actionId: 0 });
    }
  });
  
~~~

amoraschi:
> try this
> ~~~js
> let isBotDead = false;
> 
> bot.on('death', () => {
>     if (!isBotDead) {
>       isBotDead = true;
>       console.log("Bot has died.");
> 
>       // Prevent respawn
>       bot._client.write('client_command', { actionId: 0 });
>     }
>   });
>   
> ~~~

There's an option in createBot, this discussion should be closed

rom1504:
Is it documented?

MrFiend179:
> > try this
> > ```js
> > let isBotDead = false;
> > 
> > bot.on('death', () => {
> >     if (!isBotDead) {
> >       isBotDead = true;
> >       console.log("Bot has died.");
> > 
> >       // Prevent respawn
> >       bot._client.write('client_command', { actionId: 0 });
> >     }
> >   });
> >   
> > ```
> 
> There's an option in createBot, this discussion should be closed

keepAlive? other than that i dont see one

amoraschi:
![Screenshot_2023-08-27-22-00-37-127_com.github.android-edit.jpg](https://github.com/PrismarineJS/mineflayer/assets/68395930/19bd32c4-eb4e-4770-a2bc-46edd6833dbf)

MrFiend179:
Yes, this issue is resolved now.

PR #3010, opened 2023-04-05
Made bot auto respawning togglable

Averagess:
Implements #2971

### DIFF
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -69,6 +69,7 @@ function createBot (options = {}) {
   options.loadInternalPlugins = options.loadInternalPlugins ?? true
   options.client = options.client ?? null
   options.brand = options.brand ?? 'vanilla'
+  options.respawn = options.respawn ?? true
   const bot = new EventEmitter()
   bot._client = options.client
   bot.end = (reason) => bot._client.end(reason)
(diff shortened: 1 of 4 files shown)


lib/loader.js
```diff
@@ -118,6 +119,7 @@ function createBot (options = {}) {
       throw new Error(`Version ${version.minecraftVersion} is not supported. Latest supported version is ${latestTestedVersion}.`)
     }
 
+    bot.respawn = options.respawn
```
extremeheat, line 122: Why?

lib/plugins/health.js
```diff
@@ -24,6 +24,7 @@ function inject (bot) {
         bot.isAlive = false
         bot.emit('death')
       }
+      if (!bot.respawn) return
```
extremeheat, line 27: What are the side effects of this?

extremeheat (changes requested):

lib/loader.js
Averagess: Removed this

lib/plugins/health.js
Averagess: None that i know, please tell me if there are some

Jovan-04: > What are the side effects of this?
> None that i know, please tell me if there are some

it appears that there's no way to manually trigger the bot respawning aside from calling `bot._client.write('client_command', { payload: 0 })`

perhaps a `bot.respawn()` command would be useful?

PR #3012, opened 2023-04-07
Add `spectator` to gameModes array

williamd5:
This PR is a fix to the following bug: When in spectator mode, `bot.game.gameMode` is `undefined`.

The affected lines were first implemented in https://github.com/PrismarineJS/mineflayer/commit/ad52fcaee0fc0529796f3ed37ea11346b1efdc55.

### DIFF
--- a/lib/plugins/game.js
+++ b/lib/plugins/game.js
@@ -2,7 +2,7 @@ const nbt = require('prismarine-nbt')
 module.exports = inject
 
 const difficultyNames = ['peaceful', 'easy', 'normal', 'hard']
-const gameModes = ['survival', 'creative', 'adventure']
+const gameModes = ['survival', 'creative', 'adventure', 'spectator']
 
 const dimensionNames = {
   '-1': 'minecraft:nether',


williamd5:
> Lgtm, I guess. But this won't make spectator mode work in general. Mineflayers physics currently can't simulate spectator mode.

Isn't spectator mode literally no physics?

extremeheat (approved):

extremeheat:
Thanks for the contribution

IceTank:
> > Lgtm, I guess. But this won't make spectator mode work in general. Mineflayers physics currently can't simulate spectator mode.
> 
> Isn't spectator mode literally no physics? In any case, this PR was to fix that `bot.game.gameMode` is undefined when in spectator mode

If you want a real 'correct' implementation you would implement the flying physics. Things like drag when flying, correct speed limits and overshooting when letting go of control states.
Creative fly is also not implemented. Right now it just disables physics calculations and teleports the bot in a certain direction. It does not send the proper start flying packets and does not update the bots position to the server over time as physics calculations are turned off.

IceTank:
> @IceTank implementing spectator mode physics is beyond the scope of this PR. The scope of this PR is to fix a bug. With the bug, when in spectator mode, you have no way to know that you are. `bot.game.gameMode` used to be `undefined`. What this PR fixes is that. Now you can correctly check what your game mode is.

Yes. That is correct. I just want to clarify in case anyone thinks that creative flying or spectator flying is implemented. Which it is not.

PR #3016, opened 2023-04-08
Move to node18

rom1504:
node18 is now the only supported node, see https://endoflife.date/nodejs

let's switch to it


rom1504:
looks like something changed about the order of definition of things

AugustinMauroy (approved):
Nice to use modern version of node

rom1504:
ci failing

rom1504:
fails

PR #3031, opened 2023-04-23
Better code in `gps.js`

andriycraft:


andriycraft:
this will prob never get merged, so im closing this

PR #3056, opened 2023-05-22
Emit heldItemChanged when held item slot is changed

extremeheat:
As opposed to only emitting when the hotbar slot item contents has changed, the held item also changes when using scroll wheel

### DIFF
--- a/lib/plugins/inventory.js
+++ b/lib/plugins/inventory.js
@@ -587,6 +587,8 @@ function inject (bot, { hideErrors }) {
   bot._client.on('held_item_slot', (packet) => {
     // held item change
     bot.setQuickBarSlot(packet.slot)
+    // if slot is updated, held item is also updated
+    updateHeldItem()
   })
 
   function prepareWindow (window) {


u9g (approved):
LGTM

u9g:
Okay, did a little more digging, really this is already being hit, so not sure why we need this...

1. we call `bot.setQuickBarSlot` => [here](https://github.com/PrismarineJS/mineflayer/pull/3056/files#diff-43f2845439ef016d592378adeb4b9085fddb239031f857b1d76a51d631fa1d02R589)
2. we execute `setQuickBarSlot => [here](https://github.com/PrismarineJS/mineflayer/blob/4415e73bdb65edae8976fe0f84d7b608789f0194/lib/plugins/simple_inventory.js#L150) which points at [here](https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/simple_inventory.js#L51)
3. we call `bot.updateHeldItem` => [here](https://github.com/PrismarineJS/mineflayer/blob/4415e73bdb65edae8976fe0f84d7b608789f0194/lib/plugins/simple_inventory.js#L59)
4. we execute `updateHeldItem` => [here](https://github.com/PrismarineJS/mineflayer/blob/4415e73bdb65edae8976fe0f84d7b608789f0194/lib/plugins/inventory.js#L692) which points at [here](https://github.com/PrismarineJS/mineflayer/blob/4415e73bdb65edae8976fe0f84d7b608789f0194/lib/plugins/inventory.js#L366)

This pr adds an additional call to `updateHeldItem` but why? We already hit in from setQuickBarSlot.

extremeheat:
Yeah, you're right. It seems pwc was just missing the handler for this.

PR #2743, opened 2022-08-15
1.19 support 

rom1504:

PR #3088, opened 2023-06-25
1.20

PondWader:
After: https://github.com/PrismarineJS/minecraft-data/pull/723 https://github.com/PrismarineJS/prismarine-block/pull/87 https://github.com/PrismarineJS/prismarine-physics/pull/99 https://github.com/PrismarineJS/prismarine-chunk/pull/226
And probably after: https://github.com/PrismarineJS/mineflayer/pull/3090

Also includes a slight change to how checking if a place was successful works. I don't know if this is because I'm on a laptop, so stuff can be slower, but place kept thinking it'd failed when it'd succeeded around 80% of the time because it was getting a block update before the block update showing the block had changed. So I made it check all block updates within a time frame of 5 seconds (before timing out with an error) and seeing if the block changes within that time, instead of assessing it based on the first block update it receives.


rom1504:
your out of order error sounds like the one I had in #3016 PR

I advise you make that a separate PR that does this fix + update node 18, and we can merge that before merging any change for 1.20

PondWader:
A bunch of these tests seem to be failing because they can't connect to the server 🤔 And it's weird that 1.20.1 failed as it worked locally

rom1504:
yeah I'd really advise you do the "update to node18" and "add support for 1.20" in 2 independent PRs
otherwise it'll be confusing

PondWader:
> yeah I'd really advise you do the "update to node18" and "add support for 1.20" in 2 independent PRs otherwise it'll be confusing

Yeah agreed, just moved it in to a separate PR (the same one as place block because that's how I interpreted your message to do those 2 changes in the same PR)

lib/plugins/blocks.js
```diff
@@ -423,7 +423,7 @@ function inject (bot, { version, storageBuilder, hideErrors }) {
     }
   })
 
-  bot.updateSign = (block, text) => {
+  bot.updateSign = (block, text, isFrontText = true) => {
```
extremeheat, line 426: Instead of returning a boolean here you can return an integer for side (0 = main, 1 for alternate, and any future side)

test/externalTests/sign.js
```diff
@@ -22,19 +22,13 @@ module.exports = () => async (bot) => {
...
-          assert.strictEqual(sign.blockEntity.Text4.toString(), '')
-        }
-
         if (sign.blockEntity) {
           // Check block update
           bot.activateBlock(sign)
           assert.notStrictEqual(sign.blockEntity, undefined)
         }
 
+        bot._client.removeAllListeners('open_sign_entity')
```
extremeheat, line 31: Why? This seems like it could conflict with mineflayer's internal event handlers.

PondWader: It was causing issues when the world was unloaded, for some reason `open_sign_entity` was being triggered if I remember correctly, so the next tests `beforeAll` was failing, seems like some 1.20 thing. I checked when I did it and Mineflayer doesn't use `open_sign_entity` internally.

rom1504:
ah got it, it's failing because nmp is saying 18 is required

rom1504:
https://github.com/PrismarineJS/node-minecraft-protocol/pull/1235 will help

PondWader:
I think all the tests using java 1.8 will fail as they did on the place block one for some reason

rom1504:
they're not failing on master so I don't see why they would start failing here

PondWader:
Ok it seems to work now, for some reason previously all the servers using Java 1.8 failed to start

rom1504:
All deps released except pblock.
Can you finish the pblock PR @PondWader ?

Then I think this is almost ready (doesn't need to block on the place block improvement imo)

lib/plugins/blocks.js
extremeheat: Let's just flip the condition for now, that way it can be converted to numbers later on if needed

test/externalTests/sign.js
extremeheat: This should be using once then, I've updated the PR

PondWader:
By the way @rom1504 I think everything is ready

.github/workflows/ci.yml
```diff
@@ -37,6 +37,10 @@ jobs:
            mcVersion: '1.19.3'
          - javaVersion: 17
            mcVersion: '1.19.4'
+         - javaVersion: 17
+           mcVersion: '1.20'
+         - javaVersion: 17
```
extremeheat, line 42: Since there is no protocol change in 1.20.1 and 1.20, just testing 1.20.1 should save time than doing the same tests twice, to match how previous versions are tested. But I guess that can be handled later.

PondWader: Yup, it was helpful to have that originally just to be sure everything worked on both in case Mojang did something weird that wasn't in the changelog, but definitely makes sense to only test one once this is merged

rom1504:
Pblock released. can you update package.json here?

PondWader:
> Pblock released. can you update package.json here?

Done

PondWader:
Oh tests failing due to a change with pblock, one moment

rom1504:
last thing here: can you check these examples are working https://github.com/PrismarineJS/mineflayer/pull/2743#issuecomment-1276790778

* [ ] inventory.js
* [ ] chest.js
* [ ]  digger.js
* [ ] jumper.js

not everything is covered by tests

test/externalTests/commandBlock.js
```diff
@@ -13,7 +13,19 @@ module.exports = () => async (bot) => {
   await p
   bot.setCommandBlock(commandBlockPos, command, false)
 
-  const [message] = await once(bot, 'message')
-  assert(message.json.translate === 'advMode.setCommand.success')
-  assert(message.json.with[0] === command)
+  return new Promise((resolve, reject) => {
```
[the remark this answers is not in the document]
PondWader: I did this before I knew that existed, will change it

PondWader:
> last thing here: can you check these examples are working [#2743 (comment)](https://github.com/PrismarineJS/mineflayer/pull/2743#issuecomment-1276790778)
> 
> * [ ]  inventory.js
> * [ ]  chest.js
> * [ ]   digger.js
> * [ ]  jumper.js
> 
> not everything is covered by tests

Seem to be working

Also the tests seem to have failed just by chance, some invalid JSON response

rom1504 (approved):

rom1504:
thanks!

PR #3089, opened 2023-06-25
Fixed entityAttach and entityDetach events.

LetsChill:
Since all non-player entities can be ridden, this pull request makes it easy to access what rides entities and when does an entity ride another entity or a vehicle.

if this feature is good enough, [prismarine-entity](https://github.com/PrismarineJS/prismarine-entity) should also be updated to add the .passengers attribute.

it has been tested and the events are working as intended.

Fix for #2819


rom1504:
tests are failing

LetsChill:
Nothing related to my changes. It appears that it happens not only for me, but for multiple people.

LetsChill:
There is an issue that I discovered. On 1.8.9, the event gets repeated endlessly, I think set_passengers is fired even though passengers has already been assigned.

rom1504:
can you rebase on master?

rom1504:

PR #3090, opened 2023-06-25
Make the place block success check ignore block updates received with no block type changes

PondWader:
Originally from https://github.com/PrismarineJS/mineflayer/pull/3088
I don't know if it was due to lag (since I was on a laptop that performs poorly with newer versions of Minecraft) or what, but when testing for the 1.20 update, I kept getting block updates of the same block type before the block update showing the block placement had been successful, meaning I'd get an error even though the block had been placed. This fixed that
~~Includes changing node version to 18 for running tests and required engine~~


PondWader:
It seems like the server is failing to start for all the tests using Java 1.8

PondWader:
I think this will work, when I just tested locally even the 1.8 server can start on newer java versions

.github/workflows/ci.yml
```diff
@@ -12,17 +12,17 @@ jobs:
...
-        node-version: 14.x
+        node-version: 18.x
     - run: npm i && npm run lint
 
   MinecraftServer:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        javaVersion: [1.8]
+        javaVersion: [16]
```
rom1504, line 25: Why this change ?

lib/plugins/place_block.js
```diff
@@ -1,16 +1,33 @@
-const { onceWithCleanup } = require('../promise_utils')
-
 module.exports = inject
 
 function inject (bot) {
+  function waitForBlockChange (location, timeout) {
+    return new Promise((resolve, reject) => {
+      const blockUpdateListener = (oldBlock, newBlock) => {
+        if ((oldBlock || newBlock) && oldBlock.type === newBlock.type) return
```
rom1504, line 7: Why || ?

package.json
```diff
@@ -17,7 +17,7 @@
     "url": "git://github.com/PrismarineJS/mineflayer.git"
   },
   "engines": {
-    "node": ">=14"
```
rom1504, line 20: Why not make that 18 ?

lib/plugins/place_block.js
```diff
@@ -1,16 +1,33 @@
...
   async function placeBlockWithOptions (referenceBlock, faceVector, options) {
     const dest = referenceBlock.position.plus(faceVector)
-    let oldBlock = bot.blockAt(dest)
+    const oldBlock = bot.blockAt(dest)
     await bot._genericPlace(referenceBlock, faceVector, options)
 
-    let newBlock = bot.blockAt(dest)
+    const newBlock = bot.blockAt(dest)
     if (oldBlock.type === newBlock.type) {
-      [oldBlock, newBlock] = await onceWithCleanup(bot, `blockUpdate:${dest}`, { timeout: 5000 })
```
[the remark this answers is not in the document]
PondWader: No, the issue was I was receiving block updates before the block update triggered by placing the block. So I changed it to register an `on` listener which would check all block updates with a timeout of 5 seconds to determine whether the place had been successful or not.

It felt like it happened more on 1.20, but it was sending a block update with no changes before sending the block update where the block changed a lot of the time, so I changed it to check for multiple block updates.

Using `once` made it only check the first block update it receives for that block after it sends the packet. This caused the `exampleDigger` test to fail about 80% of the time.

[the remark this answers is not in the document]
extremeheat: If you need a once with timeout function, there's onceWithTimeout. Both onceWithTimeout and onceWithCleanup have checkCondition arguments you can use to check for the condition before resolving, so there is no need for a special function here.

PondWader:
Ok so for some reason in this PR, using Java version 16 for old versions makes the tests pass fine, but using Java 1.8 makes them fail. I don't see how any of the changes made could have affected whether or not the server runs on certain java versions

rom1504:
test failure look unrelated with java version

lib/plugins/place_block.js
```diff
@@ -10,7 +10,10 @@ function inject (bot) {
 
     let newBlock = bot.blockAt(dest)
     if (oldBlock.type === newBlock.type) {
-      [oldBlock, newBlock] = await onceWithCleanup(bot, `blockUpdate:${dest}`, { timeout: 5000 })
+      [oldBlock, newBlock] = await onceWithCleanup(bot, `blockUpdate:${dest}`, {
+        timeout: 5000,
+        checkCondition: (oldBlock, newBlock) => !oldBlock || !newBlock || oldBlock.type !== newBlock.type
```
rom1504, line 15: can you add a comment to explain this condition? it's not obvious to me why we want this

PondWader:
> test failure look unrelated with java version

When I switched the tests to Java 16 they all worked. I'm wondering if it's something to do with node-minecraft-wrap starting the server using Java 1.8 whilst on Node 18

lib/plugins/place_block.js
PondWader: When testing 1.20, I kept getting block updates of the same block type before the block update saying the block had changed from the server making the function think the place had failed.

rom1504:
we still want to support node 18 though

PondWader:
> we still want to support node 18 though

If I do that in the future I'll do it in a seperate PR as it seems more changes are needed than just bumping the version for the tests to pass on Java 1.8.

lib/plugins/place_block.js
rom1504: Rereading this and I still don't get it
The condition is :
* When there is no old block
* or when there no new block
* or when the new block is different from the old

Can you explain why the first 2 are wanted ?

For example why is a change from null to null ok ?

rom1504:
can you answer my comment here please

lib/plugins/place_block.js
PondWader: Because of this
image

rom1504: no that's different

this if is triggering only when both oldBlock and newBlock are null

PondWader: I think I just did that because if only one of oldBlock and newBlock is null the third check will be true anyway so easier to just write them all as or.

PR #3091, opened 2023-06-27
Fixing NBT and Name settings in modifyBook function

adambrangenberg:
In prismarine-item the property name isn't part of NBT but the Item itself. 

https://github.com/PrismarineJS/prismarine-item/blob/master/index.js#L29


u9g:
Can you add a failing test that is fixed by this pr?

TigerbyteDev:
There isn't a way to test the function directly since it's not being exported. We could test bot.writeBook() or bot.signBook() though?

u9g:
> 

Just test anything that supposed to be fixed by this pr

adambrangenberg:
Hey, book is currently excluded tests. Is there any specific reason for that since I'm also getting this: 
image

adambrangenberg:
@u9g ?

rom1504:
@adambrangenberg can you do what U9G is asking ? is anything new working now ?

adambrangenberg:
@rom1504 as I said before, the Test is in excluded tests and also timing out or failing. Is there any specific reason for that?

rom1504:
Probably because either the test or the implementation doesn't work. You can check what commit last changed that line with git blame

adambrangenberg:
It's a commit from you saying "temporarily disable flaky book test", but I don't know why that would make it fail

adambrangenberg:
I debugged it a bit more, inventory.js#L549 is causing it to timeout and loose connection (=> fail test)

rom1504:
I can't tell whether this is correct and there is no test for it, please re open if you want to finish this

PR #3097, opened 2023-07-06
Handle entity hand swapping (swapping off-hand and main hand)

### LINKED ISSUE #3095: `entity.equipment` and `entity.heldItem` not accurate when switching with offhand

## Versions
 - mineflayer: 4.10.0
 - server: paper 1.19.4-545
 - node: 18.14.2
 - typescript: 5.0.4

When you get an item from your inventory and put it in your main slot, the bot gets the item name correct. But if you put a different item in your offhand and swap the items (using F key), the bot will still assume you are holding the first item.

For example:
```
you get stone from creative menu

<You> .inhand
<Bot123>In hand: stone stone

put oak_planks in your offhand and swap with your offhand (using F key)

<You> .inhand
<Bot123>In hand: stone stone
```

## What did you try yet?

There was no method in the API that I could find that gets the offhand item or fixes the mainhand issue. I also could not find an example that showed getting the mainhand item.

I did try using entityEquip but the event did not fire when switching with offhand items.

## Your current code
```ts
import mineflayer from 'mineflayer'

const bot: mineflayer.Bot = mineflayer.createBot({
    username: 'Bot123',
    host: 'localhost',
    port: 25565,
    version: '1.19.4'
})

bot.on('chat', (username, message) => {
    if (message.startsWith('.')) {
        if (message.endsWith('inhand')) {
            bot.chat(`In hand: ${bot.players[username].entity.equipment[0]?.name} ${bot.players[username].entity?.heldItem?.name}`)
        }
    }
})
```

## Expected behavior
I expected there to be a way to access offhand items, and for the mainhand item to be correct when swapping offhand items. I also expected the entityEquip event to fire when swapping.
I also expected there to be a way to check what items are in a player's offhand.

## Additional context
I used typescript in the project, however I don't think that was the cause of the issue.

PondWader:
Fixes #3095


lib/plugins/entities.js
```diff
@@ -338,6 +339,12 @@ function inject (bot) {
     // entity status
     const entity = fetchEntity(packet.entityId)
     const eventName = entityStatusEvents[packet.entityStatus]
+
+    if (eventName === 'entityHandSwap' && entity.equipment) {
+      [entity.equipment[0], entity.equipment[1]] = [entity.equipment[1], entity.equipment[0]]
```
extremeheat, line 344: is this valid syntax?

MalkistCoder: Just asking, is it possible to check if an item in the offhand like a shield is being used?

PondWader: This isn't really the place to ask for help. I'd recommend you join the Discord: https://discord.com/invite/GsEFRM8
Or if that's not possible for you you can use the discussions: https://github.com/PrismarineJS/mineflayer/discussions

Anyway since you've already asked, you have to use the entity metadata like so:
```js
function getActiveHand (entity) {
    switch (entity.metadata[8]) {
        case undefined: 
        case 0:
        case 2: return 'none'
        case 1: return 'hand'
        case 3: return 'offhand'
    }
}
```

PR #3102, opened 2023-07-10
Test prismarine-windows changes

kaduvert:
Open a Pull Request with https://github.com/PrismarineJS/prismarine-windows/pull/74 to see if any tests fail,
as suggested by @rom1504 in https://github.com/PrismarineJS/prismarine-windows/pull/74#issuecomment-1605703610


IceTank:
There are mineflayer tests for using containers. For instance this one 
https://github.com/PrismarineJS/mineflayer/blob/master/test/externalTests/useChests.js
Can you extend those tests?

You can keep the git reference in the package.json. And then change it back once prismarine-window is released.
When prismarine-window is released we can merge the mineflayer pull request.

kaduvert:
this pull request shouldn't be merged, as it doesn't change anything.
it's purpose is to test the changes of https://github.com/PrismarineJS/prismarine-windows/pull/74 with current mineflayer as suggested.

kaduvert:
i might extend the tests in the future, but i don't intend to do so for now.

PR #3110, opened 2023-07-17
Use node-minecraft-protocol for chat

lkwilson:
- This pr leverages node-minecraft-protocol for sending chat messages and commands
- This also prevents chat commands without a header from being split up


lib/plugins/chat.js
```diff
@@ -161,7 +161,10 @@ function inject (bot, options) {
           messageCount: 0,
           acknowledged: Buffer.alloc(3),
           // 1.19.2 Chat Command packet also includes an array of last seen messages
-          previousMessages: []
+          previousMessages: bot._client._lastSeenMessages.map((e) => ({
```
extremeheat, line 164: This whole if block from L 151-170 can be removed as command parsing is handled in node-minecraft-protocol's chat(). Seems mineflayer wasn't updated for the change, but this can just become 
```js
if (bot.supportFeature('signedChat') && message.startsWith('/') {
  bot._client.chat(message)
}
```

extremeheat (changes requested):

frej4189:
Mineflayer does not sign chat commands so sending the lastSeenMessages will probably break things. 

If you want to do something about it, you should implement command signing in node-minecraft-protocol instead

extremeheat:
Yeah, the signing is to some degree implemented inside nmp. It just seems there was no update to mineflayer after nmp command handling.

lib/plugins/chat.js
lkwilson: Cool, that also fixed the issue I was having. I did however leave the return statement in since I assume that would cause it to send commands twice

lkwilson:
I made the change you suggested and tests are now failing.

extremeheat:
> I made the change you suggested and tests are now failing.

Looks like opening a PR to node-minecraft-protocol with added 
```js
          messageCount: 0,
          acknowledged: Buffer.alloc(3),
``` 
fields to chat_command packet in https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/src/client/chat.js#L343 may fix the issues

lkwilson:
What happens if I remove the block altogether? The below code calls _client.chat also, but it does length correcting. It seems that in 1.19.2, if I send a whisper command longer than the max msg length, it kicks me.

extremeheat:
Commands should definitely not be split like that. You don't want a bot.chat(/tell user really long message) command to be sent like

/tell player really
long message

Where long message was split into a normal message and leaks the contents to the server.

extremeheat:
To preserve splitting when there is a `header` you can do this instead:
```js
   if (!header && message.startsWith('/')) {
      // Do not try and split a command without a header
      bot._client.chat(message)
    }
```

rom1504:
is this ready?

frej4189:
> is this ready?

NMP dependency currently points to a fork instead of the release on NPM.

Apart from that it looks good.

lkwilson:
This is ready pending the node-minecraft-protocol pull request. I just rebased it onto the lastest master

lkwilson:
changed minecraft protocol version to official release. This should be good to go now

PR #3120, opened 2023-07-26
Fixed the enchantment table "ready" event, which has been broken for years. 

zisis912:
What used to happen was that the "ready" event would get emitted when all the level values were above 0, however 0 means that a packet hasnt been sent yet

image
image


pokahs:
very cool yeah getting enchantment name by string is just way more convenient also nice fix :p

u9g:
Please seperate this pr into two seperate prs, since it has two distinctly different changes.

zisis912:
One thing I am worried about is that the packets for `level` could be sent before the packets for `expected.enchant`, resulting in something like

```js
[
  { level: 5, expected: { enchant: 'unbreaking', level: -1 } },
  { level: 11, expected: { enchant: null, level: -1 } },
  { level: 30, expected: { enchant: null, level: -1 } }
]
```

Does anyone know the order with which the packets get sent? Maybe it would be smarter to make the "ready" event reliant on `expected.level`, depending on the order of the packets

zisis912:
> Please seperate this pr into two seperate prs, since it has two distinctly different changes.

i dont know how to do that 😭 I made a commit to my fork and it got auto-added to the pull request

zisis912:
After some stress testing I saw that its possible for the level to update without the expected.level updating, so I made sure to change the condition to check for the latter.

image

u9g:
> > Please seperate this pr into two seperate prs, since it has two distinctly different changes.
> 
> i dont know how to do that 😭 I made a commit to my fork and it got auto-added to the pull request
> 
> I think it's okay if we keep this as a 2 in 1, just a general enchanting library fix and overhaul to modernise it.

That's exactly the problem, the "overhaul" is a breaking change to anyone who uses this now.

zisis912:
> > > Please seperate this pr into two seperate prs, since it has two distinctly different changes.
> > 
> > i dont know how to do that 😭 I made a commit to my fork and it got auto-added to the pull request
> > 
> > I think it's okay if we keep this as a 2 in 1, just a general enchanting library fix and overhaul to modernise it.
> 
> That's exactly the problem, the "overhaul" is a breaking change to anyone who uses this now.

JavaScript is a language that requires one to actively update their code to the newest documentation, atleast as far as web development goes. I don't think this would go against that principle.

Also I would like to mention that there is not a single person that used the enchanting library, since it was broken this whole time. Think of this as a new feature getting added, instead of an old one being modified. Enchanting did not really exist in a usable form up to this point.

rom1504:
1.8.8 enchanting test is failing
can you please fix it ?

zisis912:
can you please merge this and ignore 1.8.8, the enchanting module does not work on ANY major version right now. While we are worrying that 1.8.8 might be broken, every major release in the last few years is broken

I cant manually do this tweak in each of my projects, pls just merge it

rom1504:
We can't merge PR which break the CI or it breaks merging any further PR

If you care about this change then take a few minutes to finish it

zisis912:
@rom1504 should be complete? the failed CI check is for the windowOpen event, it has nothing to do with my PR

lib/plugins/enchantment_table.js
```diff
@@ -75,6 +108,10 @@ function inject (bot) {
       if (!ready) await once(enchantmentTable, 'ready')
       choice = parseInt(choice, 10) // allow string argument
       assert.notStrictEqual(enchantmentTable.enchantments[choice].level, -1)
+
+      // if (enchantmentTable.enchantments[choice].level>bot.experience.level) throw new Error('insufficient xp to enchant')
```
rom1504, line 112: Remove commented code

lib/plugins/enchantment_table.js
```diff
@@ -27,32 +27,65 @@ function inject (bot) {
...
+          }
       } else if (packet.property < 10) {
         const slot = slots[packet.property - 7]
         slot.expected.level = packet.value
       }
 
-      if (slots[0].level >= 0 && slots[1].level >= 0 && slots[2].level >= 0) {
+      let readycheck;
+
+      if (bot.majorVersion === "1.8") {
```
rom1504, line 72: Use support feature

lib/plugins/enchantment_table.js
```diff
@@ -27,32 +27,65 @@ function inject (bot) {
...
+          // console.log('setting appropriate slots')
+          if (slots[i].level === 0) {
+            slots[i].level = 0
+            slots[i].expected.enchant = null
+            slots[i].expected.level = 0
+          }
+        }
+      }}
+
+      // console.log(packet.property)
```
rom1504, line 49: Remove commented code

rom1504:
Tests fail and the code is clearly not ready

Re open if you want to implement this

PR #3124, opened 2023-07-27
Add function to get the total xp points the bot has.

zisis912:
Unlike its title, the `bot.experience.points` property does NOT provide one with the xp points the bot has, but rather its score, displayed only when one dies.

image

This PR adds a function that does precisely that, correctly get the bot's xp points.

PR #3125, opened 2023-07-27
Change bot.experience.points to bot.score, the packed has been mislabeled

zisis912:
The packet has been incorrectly labeled as "xp points" for a while on both mineflayer and wiki.vg. In reality, it only represents a player's score.

One's score is all the xp they have gathered in one life, and is displayed on death. I would consider it a simplistic stat and a general attribute of a player which is why I am moving it to the bot object.

image

(https://discord.com/channels/413438066984747026/413438150594265099/1134240353738825790)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -177,6 +177,7 @@ export interface Bot extends TypedEmitter<BotEvents> {
   food: number
   foodSaturation: number
   oxygenLevel: number
+  score: number
   physics: PhysicsOptions
   physicsEnabled: boolean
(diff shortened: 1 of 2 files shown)


rom1504:
I don't see the value of this breaking change

zisis912:
> I don't see the value of this breaking change

sir this is clearly misleading, especially considering the fact that we have an actual xp points function in https://github.com/PrismarineJS/mineflayer/pull/3124

I think it should be looked into

PR #3131, opened 2023-07-29
Run CI dependent on p-item#106

frej4189:


frej4189:
Would recommend rerunning CI again here before merging https://github.com/PrismarineJS/prismarine-item/pull/105 due to changes

rom1504:
Merged now

PR #3132, opened 2023-07-30
Extended useChests.js tests

kaduvert:
Added randomized testing.
for now only for normal mouseClicks, but extending the tests to other click modes is very easy


kaduvert:
depends on https://github.com/PrismarineJS/mineflayer/pull/3133

kaduvert:
doesn't depend on #3133 anymore but uses less efficient click now

PR #3133, opened 2023-07-30
Allow more click modes

kaduvert:


kaduvert:
this may be incomplete because i don't know how the information sent in the packets differs between versions
in 1.8 it sends null item with mode 2 or 4

lib/plugins/inventory.js
```diff
@@ -528,7 +527,7 @@ function inject (bot, { hideErrors }) {
         mouseButton,
         action: actionId,
         mode,
-        item: Item.toNotch(click.item)
```
rom1504, line 531: Could you add a comment explaining this line ?

kaduvert: in 1.8 it sends null as item with mode 2 or 4
i assumed this is up to 1.16.5 but that may not be correct, haven't tested

rom1504: Ok can you add a comment for this ?

kaduvert: yes, did it

at first i thought you meant that i should explain it to you, my bad

rom1504:
would be good to add tests but lgtm

PR #3134, opened 2023-07-30
Import changedSlots computation from prismarine-windows

kaduvert:
moves the slot-comparing calculation of changedSlots here because it's of no use in prismarine-windows
the clicks can still return changedSlots directly to save computation. some of them do

depends on https://github.com/PrismarineJS/prismarine-windows/pull/102 for ci

### DIFF
--- a/lib/plugins/inventory.js
+++ b/lib/plugins/inventory.js
@@ -479,6 +479,23 @@ function inject (bot, { hideErrors }) {
     }
   }
 
+  function getChangedSlots (oldSlots, newSlots) {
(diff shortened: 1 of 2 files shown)


rom1504:
CI fails

kaduvert:
> CI fails

yes, because
> depends on https://github.com/PrismarineJS/prismarine-windows/pull/102 for ci

kaduvert:
added the primsarine-windows pr as dependency and fixed some minor bugs.
ci passes @rom1504

rom1504:
ok thanks

PR #3136, opened 2023-07-31
Update digging.js for state changes

Vakore:
Made digging.js compatible with state changes, i.e. when the bot switches from being on ground to not on ground and in water, the digging progress is saved and accurately updated.

### DIFF
--- a/lib/plugins/digging.js
+++ b/lib/plugins/digging.js
@@ -7,12 +7,25 @@ module.exports = inject
 
 function inject (bot) {
   let swingInterval = null
-  let waitTimeout = null
+  let waitInterval = null
 
   let diggingTask = createDoneTask()
 
   bot.targetDigBlock = null
   bot.lastDigTime = null
+  bot.digPercentage = 0;
+  bot.lastHeldItemName = "";
+  if (bot.heldItem) {bot.lastHeldItemName = bot.heldItem.name;}
+  bot.on('heldItemChanged', () => {
+    //if (bot.heldItem) {console.log("this why " + bot.heldItem.name);}
+    //if (!bot.heldItem) {console.log("this why " + bot.heldItem);}
+    var nameToCompare = "";
+    if (bot.heldItem) {nameToCompare = bot.heldItem.name;}
+    if (bot.lastHeldItemName != nameToCompare) {
+        bot.lastHeldItemName = nameToCompare;
+        bot.digPercentage = 0;
+    }
+  });
 
   async function dig (block, forceLook, digFace) {
     if (block === null || block === undefined) {
@@ -36,17 +49,15 @@ function inject (bot) {
         } else if (digFace.z) {
           diggingFace = digFace.z > 0 ? BlockFaces.SOUTH : BlockFaces.NORTH
         }
-        await bot.lookAt(
-          block.position.offset(0.5, 0.5, 0.5).offset(digFace.x * 0.5, digFace.y * 0.5, digFace.z * 0.5),
-          forceLook
-        )
+        await bot.lookAt(block.position.offset(0.5, 0.5, 0.5).offset(digFace.x * 0.5, digFace.y * 0.5, digFace.z * 0.5), forceLook)
       } else if (digFace === 'raycast') {
         // Check faces that could be seen from the current position. If the delta is smaller then 0.5 that means the
         // bot cam most likely not see the face as the block is 1 block thick
         // this could be false for blocks that have a smaller bounding box then 1x1x1
-        const dx = bot.entity.position.x - (block.position.x + 0.5)
-        const dy = bot.entity.position.y + bot.entity.height - (block.position.y + 0.5)
-        const dz = bot.entity.position.z - (block.position.z + 0.5)
+        const dx = bot.entity.position.x - block.position.x + 0.5
+        const dy = bot.entity.position.y - block.position.y - 0.5 + bot.entity.height // -0.5 because the bot position
+        // is calculated from the block position that is inside its feet so 0.5 - 1 = -0.5
+        const dz = bot.entity.position.z - block.position.z + 0.5
         // Check y first then x and z
         const visibleFaces = {
           y: Math.sign(Math.abs(dy) > 0.5 ? dy : 0),
@@ -57,20 +68,12 @@ function inject (bot) {
         for (const i in visibleFaces) {
           if (!visibleFaces[i]) continue // skip as this face is not visible
           // target position on the target block face. -> 0.5 + (current face) * 0.5
-          const targetPos = block.position.offset(
-            0.5 + (i === 'x' ? visibleFaces[i] * 0.5 : 0),
-            0.5 + (i === 'y' ? visibleFaces[i] * 0.5 : 0),
-            0.5 + (i === 'z' ? visibleFaces[i] * 0.5 : 0)
-          )
+          const targetPos = block.position.offset(0.5 + (i === 'x' ? visibleFaces[i] * 0.5 : 0), 0.5 + (i === 'y' ? visibleFaces[i] * 0.5 : 0), 0.5 + (i === 'z' ? visibleFaces[i] * 0.5 : 0))
           const startPos = bot.entity.position.offset(0, bot.entity.height, 0)
           const rayBlock = bot.world.raycast(startPos, targetPos.clone().subtract(startPos).normalize(), 5)
           if (rayBlock) {
             const rayPos = rayBlock.position
-            if (
-              rayPos.x === block.position.x &&
-                            rayPos.y === block.position.y &&
-                            rayPos.z === block.position.z
-            ) {
+            if (rayPos.x === block.position.x && rayPos.y === block.position.y && rayPos.z === block.position.z) {
               // console.info(rayBlock)
               validFaces.push({
                 face: rayBlock.face,
@@ -85,9 +88,7 @@ function inject (bot) {
           let distSqrt = 999
           for (const i in validFaces) {
             const tPos = validFaces[i].targetPos
-            const cDist = new Vec3(tPos.x, tPos.y, tPos.z).distanceSquared(
-              bot.entity.position.offset(0, bot.entity.height, 0)
-            )
+            const cDist = new Vec3(tPos.x, tPos.y, tPos.z).distanceSquared(bot.entity.position.offset(0, bot.entity.height, 0))
             if (distSqrt > cDist) {
               closest = validFaces[i]
               distSqrt = cDist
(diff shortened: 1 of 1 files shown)


Vakore:
I have no idea if this would break other people's projects due to variable name changes/complete functionality differences, but the reason for suggesting this fix would be because currently if the bot starts digging a block in one state, i.e. in the water, and then exits the water, or vice versa, the time needed till completion is not adjusted accordingly.

rom1504:
please remove commented code

Vakore:
I believe I have made the appropriate changes, though I missed a piece of commented debugging code on line 75.

lib/plugins/digging.js
```diff
@@ -111,19 +112,37 @@ function inject (bot) {
...
     })
     const waitTime = bot.digTime(block)
-    waitTimeout = setTimeout(finishDigging, waitTime)
+    // waitInterval = setTimeout(finishDigging, waitTime)
+    if (waitInterval) { clearInterval(waitInterval) }
+    // bot.digPercentage = (50 / bot.digTime(block));
+    // bot.referenceTime = window.performance.now();
+    waitInterval = setInterval(() => { //! !!!
+      bot.digPercentage += (50 / bot.digTime(block))
+      // bot.digPercentage += (Math.abs(bot.referenceTime - window.performance.now()) / bot.digTime(block));
```
rom1504, line 121: Remove commented code

rom1504:
Clean it up please

lib/plugins/digging.js
```diff
@@ -57,20 +66,12 @@ function inject (bot) {
         for (const i in visibleFaces) {
           if (!visibleFaces[i]) continue // skip as this face is not visible
           // target position on the target block face. -> 0.5 + (current face) * 0.5
-          const targetPos = block.position.offset(
-            0.5 + (i === 'x' ? visibleFaces[i] * 0.5 : 0),
-            0.5 + (i === 'y' ? visibleFaces[i] * 0.5 : 0),
-            0.5 + (i === 'z' ? visibleFaces[i] * 0.5 : 0)
-          )
+          const targetPos = block.position.offset(0.5 + (i === 'x' ? visibleFaces[i] * 0.5 : 0), 0.5 + (i === 'y' ? visibleFaces[i] * 0.5 : 0), 0.5 + (i === 'z' ? visibleFaces[i] * 0.5 : 0))
```
rom1504, line 69: this is only a formatting change, undo

lib/plugins/digging.js
```diff
@@ -36,17 +47,15 @@ function inject (bot) {
...
-        )
+        await bot.lookAt(block.position.offset(0.5, 0.5, 0.5).offset(digFace.x * 0.5, digFace.y * 0.5, digFace.z * 0.5), forceLook)
       } else if (digFace === 'raycast') {
         // Check faces that could be seen from the current position. If the delta is smaller then 0.5 that means the
         // bot cam most likely not see the face as the block is 1 block thick
         // this could be false for blocks that have a smaller bounding box then 1x1x1
-        const dx = bot.entity.position.x - (block.position.x + 0.5)
-        const dy = bot.entity.position.y + bot.entity.height - (block.position.y + 0.5)
-        const dz = bot.entity.position.z - (block.position.z + 0.5)
+        const dx = bot.entity.position.x - block.position.x + 0.5
```
rom1504, line 55: this is only a formatting change, undo

lib/plugins/digging.js
```diff
@@ -36,17 +47,15 @@ function inject (bot) {
...
+        const dy = bot.entity.position.y - block.position.y - 0.5 + bot.entity.height // -0.5 because the bot position
+        // is calculated from the block position that is inside its feet so 0.5 - 1 = -0.5
+        const dz = bot.entity.position.z - block.position.z + 0.5
```
rom1504, line 58: this is incorrect

lib/plugins/digging.js
```diff
@@ -85,9 +86,7 @@ function inject (bot) {
           let distSqrt = 999
           for (const i in validFaces) {
             const tPos = validFaces[i].targetPos
-            const cDist = new Vec3(tPos.x, tPos.y, tPos.z).distanceSquared(
-              bot.entity.position.offset(0, bot.entity.height, 0)
-            )
+            const cDist = new Vec3(tPos.x, tPos.y, tPos.z).distanceSquared(bot.entity.position.offset(0, bot.entity.height, 0))
```
rom1504, line 89: this is only a formatting change, undo

lib/plugins/digging.js
```diff
@@ -57,20 +66,12 @@ function inject (bot) {
...
           const startPos = bot.entity.position.offset(0, bot.entity.height, 0)
           const rayBlock = bot.world.raycast(startPos, targetPos.clone().subtract(startPos).normalize(), 5)
           if (rayBlock) {
             const rayPos = rayBlock.position
-            if (
-              rayPos.x === block.position.x &&
-                            rayPos.y === block.position.y &&
-                            rayPos.z === block.position.z
-            ) {
+            if (rayPos.x === block.position.x && rayPos.y === block.position.y && rayPos.z === block.position.z) {
```
rom1504, line 74: this is only a formatting change, undo

lib/plugins/digging.js
```diff
@@ -212,14 +228,7 @@ function inject (bot) {
...
     const creative = bot.game.gameMode === 'creative'
-    return block.digTime(
-      type,
-      creative,
-      bot.entity.isInWater,
-      !bot.entity.onGround,
-      enchantments,
-      bot.entity.effects
-    )
+    return block.digTime(type, creative, bot.entity.isInWater, !bot.entity.onGround, enchantments, bot.entity.effects)
```
rom1504, line 222: this is only a formatting change, undo

lib/plugins/digging.js
```diff
@@ -185,11 +205,7 @@ function inject (bot) {
   })
 
   function canDigBlock (block) {
-    return (
-      block &&
-            block.diggable &&
-            block.position.offset(0.5, 0.5, 0.5).distanceTo(bot.entity.position.offset(0, 1.65, 0)) <= 5.1
-    )
+    return block && block.diggable && block.position.offset(0.5, 0.5, 0.5).distanceTo(bot.entity.position.offset(0, 1.65, 0)) <= 5.1
```
rom1504, line 199: this is only a formatting change, undo

lib/plugins/digging.js
```diff
@@ -36,17 +47,15 @@ function inject (bot) {
         } else if (digFace.z) {
           diggingFace = digFace.z > 0 ? BlockFaces.SOUTH : BlockFaces.NORTH
         }
-        await bot.lookAt(
-          block.position.offset(0.5, 0.5, 0.5).offset(digFace.x * 0.5, digFace.y * 0.5, digFace.z * 0.5),
-          forceLook
-        )
+        await bot.lookAt(block.position.offset(0.5, 0.5, 0.5).offset(digFace.x * 0.5, digFace.y * 0.5, digFace.z * 0.5), forceLook)
```
rom1504, line 50: this is only a formatting change, undo

lib/plugins/digging.js
```diff
@@ -162,13 +184,14 @@ function inject (bot) {
     function onBlockUpdate (oldBlock, newBlock) {
       // vanilla server never actually interrupt digging, but some server send block update when you start digging
       // so ignore block update if not air
-      // All block update listeners receive (null, null) when the world is unloaded. So newBlock can be null.
-      if (newBlock?.type !== 0) return
+      if (!newBlock) return
```
rom1504, line 187: this doesn't change anything, see the ? below
you can remove this change

lib/plugins/digging.js
```diff
@@ -110,20 +121,30 @@ function inject (bot) {
...
 
     swingInterval = setInterval(() => {
-      bot.swingArm()
-    }, 350)
+      if (bot.targetDigBlock) {
+        bot.swingArm()
+      } else {
+        clearInterval(swingInterval)
+      }
+    }, 250)
```
rom1504, line 140: why the change from 350 to 250 ?

lib/plugins/digging.js
```diff
@@ -110,20 +121,30 @@ function inject (bot) {
       location: block.position,
       face: diggingFace // default face is 1 (top)
     })
-    const waitTime = bot.digTime(block)
-    waitTimeout = setTimeout(finishDigging, waitTime)
+    if (waitInterval) { clearInterval(waitInterval) }
+    waitInterval = setInterval(() => {
+      bot.digPercentage += (50 / bot.digTime(block))
```
rom1504, line 126: does the vanilla client follow this algorithm when state change while digging ? do you have a reference ?

Vakore:
image
What I am attempting to change can be mostly summarized by these two changes.

The first change(upper) is meant to fix an issue with the digging function. In Minecraft, if a player starts digging a block while on the ground, then there is a definitive time it will take to break the block if the player continues digging and stays on the ground. Likewise, with if the player mines a block while in the air, or underwater. This is what the original code assumes.

The problem with this is the time to finish breaking something will vary if the player switches between these states(on the ground, in the air, underwater), so if it were to take 1 second to mine a block while grounded the entire time, but 2 seconds when in the air, and the player spends the first half of the digging progress on the ground and the second half in the air, it'll break in 1.5 seconds. Mineflayer currently does not account for this. If the bot begins mining while in the air, then lands on the ground, it will not finish breaking the block until the first computed time, that is the 'in the air' time, has elapsed.

The changes relating to 'digPercentage' allow for the progress to be properly stored and calculated as the mineflayer bot breaks blocks while constantly switching these states, and more accurate to a vanilla client's behavior to my knowledge.

The second change is relating to a common 'arm swinging' bug when breaking blocks. Sometimes, the arm keeps swinging and will never stop. The changes I made add a safety check to the interval to make sure not to swing the arm if the bot isn't breaking a block, and if it isn't, clear the swing interval. The change from '350' to '250' is simply to make the arm swing speed faster and closer to vanilla, though I don't think I made it fast enough to match the vanilla client's arm swinging speed.

lib/plugins/digging.js
Vakore: Probably not this exact algorithm but it mimics what I've seen in game, and I've yet to see something abnormal from it.

lib/plugins/digging.js
```diff
@@ -162,13 +184,13 @@ function inject (bot) {
     function onBlockUpdate (oldBlock, newBlock) {
       // vanilla server never actually interrupt digging, but some server send block update when you start digging
       // so ignore block update if not air
-      // All block update listeners receive (null, null) when the world is unloaded. So newBlock can be null.
```
IceTank, line 165: Why did you remove this comment?

lib/plugins/digging.js
```diff
@@ -7,12 +7,23 @@ module.exports = inject
...
 
   bot.targetDigBlock = null
   bot.lastDigTime = null
+  bot.digPercentage = 0
+  bot.lastHeldItemName = ''
+  if (bot.heldItem) { bot.lastHeldItemName = bot.heldItem.name }
+  bot.on('heldItemChanged', () => {
+    let nameToCompare = ''
+    if (bot.heldItem) { nameToCompare = bot.heldItem.name }
+    if (bot.lastHeldItemName !== nameToCompare) {
```
IceTank, line 22: I think any held item change cancels the dig process? Even increasing the item amount in a slot.

IceTank:
Btw you don't have to use an external site for linting. Mineflayer comes with its own linter. Just do `npm run fix`.

lib/plugins/digging.js
Vakore: Surprisingly no. If you have two pickaxes with equal durability and/or nbt, and switch between them, or seitch between different fist slots while mining, it won't cancel the mining timer. Mining a tree with fists and picking up a log while mining will reset it, though constantly mining with logs while the stack size increases doesn't last time I checked. I'll double check later today but I'm pretty sure this is correct.

Though now that I'm looking at this the system here won't take into account durability/enchant differences, though that's just nbt and can likely be changed easily. This system will still work in most cases though and is better than nothing, but it should be fixed.

IceTank: A better way to check if items are equal is the stringify the item and compare that. You could make a string of the item you started mining with and reset the progress if the string of the held item changes.

rom1504:
@IceTank do you want to check this one and tell me if it looks good?

lib/plugins/digging.js
IceTank: Can you add the comment back?

lib/plugins/digging.js
IceTank: @Vakore can you change this back or give w reason to why you changed the interval?

Vakore: I changed the interval to fix the arm swing bug

lib/plugins/digging.js
Vakore: I have a more updated version of digging.js on the discord. Search for attachments from 'Vakore', unable to do stuff with npm rn.

rom1504:
Old closing

PR #3157, opened 2023-08-17
Use node 18 on CI

extremeheat:


rom1504:
Seems to run

Why delete the gitpod file ?

extremeheat:
It was loading node 14 and didn't seem necessary since the default container now uses node 18 + java 20, but I can add it back

index.js
```diff
@@ -1,6 +1,6 @@
-if (typeof process !== 'undefined' && parseInt(process.versions.node.split('.')[0]) < 14) {
+if (typeof process !== 'undefined' && parseInt(process.versions.node.split('.')[0]) < 16) {
```
rom1504, line 1: why not 18 ?

PR #3159, opened 2023-08-18
add gamemode on player init packet

zardoy:
In my testing singeplayer in the web client works well. I'm using flying-squid (which seems to work only on 1.16) where all events are processed in both clients synchronously. So right after initial player data set, spawn event fires where I should synchronously use gamemode of the current player


zardoy:
okay, nevermind, I see there is `bot.game.gameMode` that is more reliable than `bot.player.gamemode` eg it gets correctly updated on gamemode change. I think `bot.player.gamemode` should be deprecated

PR #3161, opened 2023-08-18
Fix ambigious function naming

frej4189:
As noted here: https://github.com/PrismarineJS/mineflayer/pull/2492#discussion_r1297309964, the function to determine whether an entity is dead for physics purposes has very ambigious naming which could potentially lead to confusion (and worsened maintainability).

Vanilla client handles this by actually removing the entity, so my naming reflects that behaviour (but note the PR does not actually implement such behaviour).


extremeheat (approved):
Yeah, I agree the current naming is missing a "should" prefix to make clear it's not executing anything

PR #3163, opened 2023-08-19
Add elytra flying support and rocket support

lkwilson:


lkwilson:
Need tests

amoraschi:
Doesn't this need some additional physics implementations? Maybe it should be moved to prismarine-physics

lkwilson:
> Doesn't this need some additional physics implementations? Maybe it should be moved to prismarine-physics

Yeah, see https://github.com/PrismarineJS/prismarine-physics/pull/106

I think the main feedback I need is whether I put things in the right spot

SilkePilon:
does this work?

lkwilson:
> does this work?

It does but not super well tested

SilkePilon:
will test it with https://github.com/SilkePilon/OpenDeliveryBot

lib/plugins/inventory.js
```diff
@@ -108,6 +110,35 @@ function inject (bot, { hideErrors }) {
         hand: offHand ? 1 : 0
       })
     }
+
+    if (bot.elytraFlying) {
```
lkwilson, line 114: Should this section be moved to `updateHeldItem`?

lkwilson:
It seems that entity_metadata packets can report whether the player is flying or not. The code as it is will sometimes get kicked when the server rejects its elytra request

lkwilson:
Turns out I had to pull the elytra is flying state and the firework attached to entity state from entity update msgs. I changed it to only set elytra flying / rockets when it detects them via that (similar to crouch). This fixes the occational kicked for flying issue I was seeing (now that server actually confirms flying and fireworks). There's also a test for using elytra in all the versions, and it seems to pass (but requires a minecraft-data bump, which I wasn't sure how to run in github)

lkwilson:
This is ready pending the following prs:
- minecraft-data: https://github.com/PrismarineJS/minecraft-data/pull/761 (plus bump to https://github.com/PrismarineJS/node-minecraft-data and npm publish)
- prismarine-entity: https://github.com/PrismarineJS/prismarine-entity/pull/69
- prismarine-physics: https://github.com/PrismarineJS/prismarine-physics/pull/106

rom1504:
merged mcdata and p-entity because they were trivial

for prismarine-physics and here we need proper reviews

docs/api.md
```diff
@@ -1417,6 +1429,10 @@ This occurs whether the process was completed or aborted.
 
  * `block` - the block that still exists
 
+#### "usedfirework"
+
+Fires when the bot uses a firework while elytra flying
```
AugustinMauroy, line 1434:
```suggestion
Fires when the bot uses a firework while elytra flying.
```

index.d.ts
```diff
@@ -164,6 +164,7 @@ export interface Bot extends TypedEmitter<BotEvents> {
   version: string
```
AugustinMauroy, line 164: you missed to add "Tanner Buchanan" in event same for "entityElytraFlew"

package.json
```diff
@@ -27,10 +27,10 @@
     "prismarine-block": "^1.17.0",
     "prismarine-chat": "^1.7.1",
     "prismarine-chunk": "^1.34.0",
-    "prismarine-entity": "^2.2.0",
+    "prismarine-entity": "lkwilson/prismarine-entity#elytra-flying",
```
AugustinMauroy, line 30: should be change when pr is publish

lib/plugins/physics.js
```diff
@@ -192,6 +192,43 @@ function inject (bot, { physicsEnabled, maxCatchupTicks }) {
...
+    } else if (bot.entity.isInWater) {
+      throw new Error('Unable to elytra fly while in water')
+    }
+
+    const mcData = require('minecraft-data')(bot.version)
+    if (getEffectLevel(mcData, 'Levitation', bot.entity.effects) > 0) {
+      throw new Error('Unable to elytra fly with levitation effect')
+    }
+
+    const item = bot.inventory.slots[6]
```
frej4189, line 221: This value should probably not be hardcoded

frej4189 (changes requested):

frej4189:
I see no breaking changes so should be relatively safe to merge. Elytra test itself looks fine so given that's passing I believe the PR is good to go.

lkwilson:
> I see no breaking changes so should be relatively safe to merge. Elytra test itself looks fine so given that's passing I believe the PR is good to go.

One thing I should point out is that the elytra test is passing since it doesn’t have the feature flags from minecraft data, so when we bump to that version we have to reevaluate where I got the entity metadata parsing correct for all versions

lib/plugins/physics.js
lkwilson: Changed to getEquipmentDestSlot

docs/api.md
lkwilson: Changed, thanks!

package.json
```diff
@@ -27,10 +27,10 @@
     "prismarine-block": "^1.17.0",
     "prismarine-chat": "^1.7.1",
     "prismarine-chunk": "^1.34.0",
-    "prismarine-entity": "^2.2.0",
+    "prismarine-entity": "^2.3.0",
```
lkwilson, line 30: Oh, and this bump adds a bunch of deprecation warnings, so we might not want to bump prismarine entity without addressing that

lkwilson: https://github.com/PrismarineJS/mineflayer/pull/3168

docs/api.md
```diff
@@ -1301,6 +1309,10 @@ Fires when an attribute of an entity changes.
 #### "entityEquip" (entity)
 #### "entitySleep" (entity)
 #### "entitySpawn" (entity)
+#### "entityElytraFlew" (entity)
```
AugustinMauroy, line 1312: we should explain what is entity or redirect to prismarine-entity

lkwilson: It's already documented in the [entity section](https://github.com/PrismarineJS/mineflayer/blob/master/docs/api.md#entity) of this document, so I'm not sure if I should do that

rom1504: no indeed don't

rom1504:
thanks for the PR!

Out of curiosity, why did you need elytra support / how do you intend to use it ?

lkwilson:
> thanks for the PR!
> 
> Out of curiosity, why did you need elytra support / how do you intend to use it ?

Thanks for merging! It started with wanting an afk bot that I could run on my raspberry pi instead of leaving my laptop on, but without letting it mine or build, it couldn't reach afk spots high in the sky or really far away. tp commands work but felt a bit cheaty, so implementing elytra seemed like a fun project. Plus it's really cool watching the bot elytra fly in ways that I can't think fast enough to do myself

PR #3168, opened 2023-08-26
Switch to entity.displayName

lkwilson:
- Bump to prismarine-entity 2.3.0, which deprecates mobType and objectType
- Update mineflayer to not use newly deprecated fields


fructozer (commented):
its good

PR #3178, opened 2023-08-31
Updating RU docs

FurnyGo:


FurnyGo:
i think sidebar works wrong... (it didn't adding "/ru/" in link)

PR #3191, opened 2023-09-09
Handle unknown scoreboard objectives

frej4189:
Currently mineflayer will throw if given a scoreboard objective packet with the update action if the objective does not exist.

This "fixes" that by simply emitting an error event instead, allowing the user to do their own error handling.

Vanilla client will also throw when this happens, but will not disconnect the client, so I think simply emitting the error event is the right thing to do, so the user can handle it themselves and decide whether or not to disconnect. Cross-checked only with Vanilla 1.8 client, might have different behaviour in higher versions, but regardless I don't think we can handle it any different than this.


lib/plugins/scoreboard.js
```diff
@@ -19,6 +19,10 @@ function inject (bot) {
     }
 
     if (packet.action === 2) {
+      if (!(packet.name in scoreboards)) {
```
rom1504, line 22: in operator has unexpected behavior in JavaScript, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in
In short it operates on Object which is not only a dictionary

One way that works is scoreboards[packet.name] !== undefined

frej4189: We know that scoreboards is a dictionary so that shouldn't really be an issue as far as I am concerned?

extremeheat: scoreboards is not analogous to a dictionary, it's an Object. Which has special properties inside of it, relating to the prototype. Even if a property inside scoreboards is undefined, `in` will return true because the property is not considered "empty" inside JS.

Eg.
```js
const x = { y: 2 }
x.y = undefined
'y' in x == true
```

Also, this should probably be using `Object.hasOwn(scoreboards, packet.name)` to not match on prototype data inside the object.

frej4189: Makes sense.

frej4189: Fixed with latest commit.

PR #3206, opened 2023-09-27
feat: bot.respawn, fix respawn with flying squid

zardoy:
respawn wasnt working with flying squid. It worked with vanilla client as I understand vanilla handle both payload and actionId. Maybe it would be better to remove acknowledgment of respawnIsPayload feature?


zardoy:
imo `respawn` option should be renamed to `autoRespawn`

zardoy:
so confused why it was closed

lib/plugins/health.js
```diff
@@ -25,10 +25,17 @@ function inject (bot, options) {
...
       bot.emit('spawn')
     }
   })
+
+  const respawn = () => {
+    if (bot.isAlive) return
+    bot._client.write('client_command', bot.supportFeature('respawnIsPayload') ? { payload: 0 } : { actionId: 0 })
+  }
+
+  bot.respawn = respawn
```
rom1504, line 40: this should be added to API.md

zardoy: done, thanks

Issue #3216, opened 2023-10-29
Access chunk infomation from worker threads


szdytom:
## Is your feature request related to a problem? Please describe.

I working on my own path finding of the bot, which includes heavy computation so I want to extract those into a worker thread. 

## Describe the solution you'd like
Use `SharedArrayBuffer` in [prismarine-chunk](https://github.com/PrismarineJS/prismarine-chunk/) so that those information can be send across threads. We can also restrict the worker threads read-only to those data, so there won't be many data-races problems. 

## Describe alternatives you've considered
I am currently listening to "chunk load" event and sending chunks to a shared memory buffer.

PR #3218, opened 2023-10-29
Change findblocks into promise have a sense?

sefirosweb:
Hi guys, I have been noticed that my bot sometimes freezes, I have detected that if I run findblock to search in a large radius, this causes that the bot freeze until the search is finished,

I think it would be a good idea change the method to a promise, and give it a short wait for the bot to exchange the action processes with the search and this not freeze the bot,

The change is simple, which will cause users to have to adapt their code to a promise

Obviously the performance is worse; the same test for "normal"/"promise" is 3-4 times slower,
What this slowness in my opinion compensates for the bot being much "fresher" in these actions without blocking,

image

The important code are only in block.js
Change findblocks into promise have a sense?

### DIFF
--- a/examples/anvil.js
+++ b/examples/anvil.js
@@ -110,7 +110,7 @@ function getAnvilIds () {
 }
 
 async function rename (bot, itemName, name) {
-  const anvilBlock = bot.findBlock({
+  const anvilBlock = await bot.findBlock({
     matching: getAnvilIds()
   })
   const anvil = await bot.openAnvil(anvilBlock)
@@ -124,7 +124,7 @@ async function rename (bot, itemName, name) {
 }
 
 async function combine (bot, itemName1, itemName2, name) {
-  const anvilBlock = bot.findBlock({
+  const anvilBlock = await bot.findBlock({
     matching: getAnvilIds()
   })
   const anvil = await bot.openAnvil(anvilBlock)
--- a/examples/block_entity.js
+++ b/examples/block_entity.js
@@ -22,14 +22,14 @@ bot.on('message', (cm) => {
   }
 })
 
-function spawner () {
+async function spawner () {
   let blockName
   if (bot.supportFeature('mobSpawner')) {
     blockName = bot.registry.blocksByName.mob_spawner.id
   } else if (bot.supportFeature('spawner')) {
     blockName = bot.registry.blocksByName.spawner.id
   }
-  const block = bot.findBlock({
+  const block = await bot.findBlock({
     matching: blockName,
     point: bot.entity.position
   })
--- a/examples/blockfinder.js
+++ b/examples/blockfinder.js
@@ -35,7 +35,7 @@ bot.on('chat', async (username, message) => {
     const ids = [bot.registry.blocksByName[name].id]
 
     const startTime = performance.now()
-    const blocks = bot.findBlocks({ matching: ids, maxDistance: 128, count: 10 })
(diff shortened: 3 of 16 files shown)


rom1504:
I think we should rather introduce a new method (potentially using a
partially common implementation with the current method) rather than make
this breaking change.

extremeheat:
Also, yielding to the event loop inside of any hot/iterative code is really bad as that prevents any JIT optimizations, on top of the event loop overhead (copying memory, waiting, etc). If it's really necessary to split up work, you would need to find a better algorithm to segment the data and yield as little as possible (maybe a few times depending on search area) to make it optimal. 

I think a better solution would be to profile the existing code and see if synchronous optimizations could be made (reducing copies, type conversions, checks, etc). As of right now, it seems findBlock() is running through every block in the box, when it could be exiting early on the first match. That's an easy optimization.

sefirosweb:
I don't think it can help much in optimizing the code, I don't understand it very well, and when I want to see the variable types I see that many of them are not complete.

image

If you want I can make a new method "searchBlock" & "searchBlocks" that use a common code or most of it

extremeheat:
> I don't think it can help much in optimizing the code

It will, pushing things to do later will make things slower, not faster. It will still block your program, including timers. The only way to not block aside a worker is to do less operations or more efficient ones.

> I don't understand it very well, and when I want to see the variable types I see that many of them are not complete.

Use control + f to search the codebase, it's not very big or complicated code, everything is defined in a few lib/plugins files.

> If you want I can make a new method "searchBlock" & "searchBlocks" that use a common code or most of it

Echo my previous comment: As of right now, it seems findBlock() is running through every block in the box, when it could be exiting early on the first match. That's an easy optimization.

sefirosweb:
I'm on it =)

about findBlock (one) already have fast exit on find one,
```ts
if (startedLayer !== it.apothem && blocks.length >= count) {
```

> It will, pushing things to do later will make things slower, not faster. It will still block your program, including timers. The only way to not block aside a worker is to do less operations or more efficient ones.

Yes and not, yes is more slower, but the bot dosent freeze on to do this operation, (of course this affect to the rest of bot while are executing )

anyways i try to see if it can also be improved without promises

szdytom:
Possibly related to #3216.

extremeheat:
> anyways i try to see if it can also be improved without promises

One of the best ways to make things faster is by minimizing memory allocations/copies, especially in hot code. Simplifying the code helps, but be cautious here since less code doesn't mean faster, if that "less code" is just an abstraction for really complex code under the hood. https://github.com/PrismarineJS/prismarine-physics/pull/111/commits/d29774ae4ca426d87a5209a9dc5e92aa45458b7c is a good example for that, as lots of garbage from temporary objects would be created in hot code.

Also, moving the world data and these findBlock methods to prismarine-world per #334 would make it easier to remove where the bottlenecks are in the current code.

rom1504:
ci is failing

sefirosweb:
I tried to improve the optimization but without results..

Finally are you interested to merge this pr?

rom1504:
yeah actually reading the previous discussion it seems this PR is not really helping right? maybe we should just close it?

sefirosweb:
There are contrary opinions, I think so, but I do not have the knowledge to improve performance, if perhaps it is not interesting we can close the pr

rom1504:
ok

PR #3222, opened 2023-11-01
Fixes `fireworkRocketDuration` calculation

### LINKED ISSUE #3221: `fireworkRocketDuration` is calculated wrong, results in elytra to fly shorter

## Versions
 - mineflayer: 4.14.0
 - server: Vanilla 1.20.1
 - node: _Not Related_

`fireworkRocketDuration` is set to wrong values that cause elytra to fly shorter. The correct formula is: 10 × (number of gunpowder + 1) + random value from 0 to 5 + random value from 0 to 6

example picture

The above figure illustrates the difference: the plank post on the left stands for mineflayer's fly height, the log post on the right stands for notchian client fly height (both are done with 1 rocket with flight of 1).

## What did you try yet?

I'm using `.activateItem()` to uses the firework rocket.

## Your current code

The problem is not with my code but mineflayer's code, please revise the following part:

https://github.com/PrismarineJS/mineflayer/blob/08208e2f110af2c6de41fac9a389597aac916412/lib/plugins/entities.js#L384-L399

There are two bug here:

- Line 385: the nbt may does not exists, but that means the rocket has a flight of 1 instead of 0.
- Line 391: the type should be `byte` instead of `int` (at least for 1.20.1)

## Expected behavior
Fly as long as a notchian client player.

## Additional context
I'll open a pull request to fix this soon.

szdytom:
## The problem

`fireworkRocketDuration` is set to wrong values that cause elytra to fly shorter (#3221). The correct formula is: 10 × (number of gunpowder + 1) + random value from 0 to 5 + random value from 0 to 6

example picture

The above figure illustrates the difference: the plank post on the left stands for mineflayer's fly height, the log post on the right stands for notchian client fly height (both are done with 1 rocket with flight of 1).

The problem is about the following part:

https://github.com/PrismarineJS/mineflayer/blob/08208e2f110af2c6de41fac9a389597aac916412/lib/plugins/entities.js#L384-L399

There are two bug in this code:

- Line 385: the nbt may does not exists, but that means the rocket has a flight of 1 instead of 0.
- Line 391: the type should be `byte` instead of `int` (at least for 1.20.1)

## Changes

In this pull request:

- Modified function `handleBotUsedFireworkRocket` to fix `fireworkRocketDuration` calculation.
- Made variable `knownFireworks` a [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set), this in more efficient than using Array.
- Fixed a typo about event `usedFirework` in the API document (`usedfirework` to `usedFirework`).
- Added an argument to event `usedFirework`:
  - `fireworkEntityId` - the entity id of the firework.
- Updated the document of Simplified Chinese to add the `usedfirework` event. I am a native speaker of that language.

Fixes #3221

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1429,9 +1429,11 @@ This occurs whether the process was completed or aborted.
 
  * `block` - the block that still exists
 
-#### "usedfirework"
+#### "usedFirework" (fireworkEntityId)
 
 Fires when the bot uses a firework while elytra flying.
+
+ * `fireworkEntityId` - the entity id of the firework.
 
 #### "move"
 
--- a/docs/zh/README_ZH_CN.md
+++ b/docs/zh/README_ZH_CN.md
@@ -240,6 +240,7 @@ Mineflayer 支持插件；任何人都可以创建一个插件，在 Mineflayer 
 ### 示例
 
 `npm run mocha_test -- -g "1.18.1.*BlockFinder"` 进行1.18.1寻路测试
+
 ## 许可证
 
 [MIT](../LICENSE)
--- a/docs/zh/api.md
+++ b/docs/zh/api.md
@@ -1339,6 +1339,12 @@ This occurs whether the process was completed or aborted.
 
  * `block` - 方块仍然存在
 
+#### "usedFirework" (fireworkEntityId)
+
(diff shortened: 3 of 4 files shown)


rom1504:
seems right

would be good to add a test for this

PR #3225, opened 2023-11-02
Fix `bot.heldItem` and `bot.entity.equipment`

### LINKED ISSUE #397: bot.entity.equipment is always empty array
As I understand, `equipment` property should display currently equipped items(such as helmet, leggins) but it remains empty even after bot equips full set so for now here is only workaround to get equipped items is to read from 5th inventory's index where equipped armor goes.

### LINKED ISSUE #3224: `bot.heldItem` sometimes desync

## Versions
 - mineflayer: 4.14.0
 - server: vanilla 1.20.1
 - node: 20.8.0

`bot.heldItem` sometimes is wrong and desync with `bot.inventory`.

## What did you try yet?

To reproduce:
- Place an hopper on the groud.
- Teleport the bot onto that hopper.
- Let the bot look at its foot: `await bot.look(0, -Math.PI/2, true)`.
- Give the bot a stack of items on its main hand.
- Let the bot drop its heldItem by calling: `bot.tossStack(bot.heldItem)`.

After those operations, the bot no longer holds the item, but `bot.heldItem` isn't set to `null`.

## Your current code
```js
bot.tossStack(bot.heldItem);
```

## Expected behavior
`bot.heldItem` yields `null`.

## Additional context

I wonder what is the reason for the `heldItem` isn't implemented by [Getter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get#defining_a_getter_on_existing_objects_using_defineproperty).

szdytom:
Made `bot.heldItem` and `bot.entity.equipment` two getter functions. Note `bot.entity.heldItem` may still be incorrect, and that was fixed in https://github.com/PrismarineJS/prismarine-entity/pull/73

Fixes #3224
Fixes #397

### DIFF
--- a/docs/zh/README_ZH_CN.md
+++ b/docs/zh/README_ZH_CN.md
@@ -240,6 +240,7 @@ Mineflayer 支持插件；任何人都可以创建一个插件，在 Mineflayer 
 ### 示例
 
 `npm run mocha_test -- -g "1.18.1.*BlockFinder"` 进行1.18.1寻路测试
+
 ## 许可证
 
 [MIT](../LICENSE)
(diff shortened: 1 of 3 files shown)


rom1504:
can you add a test ?

szdytom:
I've added a external test.

Issue #3209, opened 2023-10-03
villager.trade(index, count) fails for enchanted books


EvModder:
As titles states.

I wonder if Mineflayer is having trouble due to the fact there are two input items for the trade?

Simple example to recreate issue:
```js
const gui = await bot.openVillager(villager)
console.log('no errors')
await gui.trade(index, 1).then(success)
console.log('errors before getting here')
gui.close()
```
I have verified all of the below:
* The bot has more than sufficient emeralds
* The bot has more than sufficient plain books
* The bot has more than sufficient available empty inv slots

Perhaps related - I've noticed Mineflayer will hang indefinitely even for single-input trades. My workaround up to this point for *that* issue is to have a timeout, close & reopen the gui, and try again.

Issue #2681, opened 2022-07-01
`bot.trade()` hangs forever


wmantly:
- [X ] The [FAQ](https://github.com/PrismarineJS/mineflayer/blob/master/docs/FAQ.md) doesn't contain a resolution to my issue 

## Versions
 - mineflayer: mineflayer@4.3.0
 - server: vanilla/spigot/paper paper 1.18.12
 - node: 18.4.01

## Detailed description of a problem
When calling `bot.trade` sometimes it will just hang. I have tracked the issue to this line: https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/villager.js#L165 
The last time `updateSlot` is called, `countOfItemOneLeftToTake` is not zero. `upateSlots` never gets called again and the whole app hangs.

## What did you try yet?
I have edited the function in question to find, and  midagte the issue. It still hangs.
```js
let count_calls = 0;
let last_countOfItemOneLeftToTake = 9999;
const listener = (slot, oldItem, newItem) => {
  console.log('in listener count', ++count_calls)    
        // console.log('bot inventory', bot.inventory.slots.filter(e=>e).map(e=>`${e.slot}: ${e.count} ${e.displayName}`))

  if (!(slot >= villager.inventoryStart && slot <= villager.inventoryEnd)) return
  if (newItem === null) {
    if (oldItem.type === Trade.inputItem1.type) countOfItemOneLeftToTake -= oldItem.count
    else if (Trade.hasItem2 && oldItem.type === Trade.inputItem2.type) countOfItemTwoLeftToTake -= oldItem.count
  }
    console.log('in window listener', countOfItemOneLeftToTake,countOfItemTwoLeftToTake)
  if (countOfItemOneLeftToTake === 0 && countOfItemTwoLeftToTake === 0) {
    villager.off('updateSlot', listener)
    resolve()
  }

  console.log('villager', villager.slots.filter(e=>e).map(e=>`${e.slot}: ${e.count} ${e.displayName}`))
  if(last_countOfItemOneLeftToTake === countOfItemOneLeftToTake){
    console.log('work around worked!');
    villager.off('updateSlot', listener)
    resolve()
  }else{
    last_countOfItemOneLeftToTake = countOfItemOneLeftToTake;
  }


}
```
returns the following out, and just hangs forever

```
in listener count 1
in listener count 2
in listener count 3
in window listener 21 0
villager [
  '0: 64 Paper',
  '2: 1 Emerald',
[...]

PR #3230, opened 2023-11-12
Fix several bugs in villager trading

evan-goode:
These changes should fix issues with trades that have more than one item in the output stack, trades that cost more than one emerald, and trades with two input items, such as trades for enchanted books and tipped arrows.

Resolves https://github.com/PrismarineJS/mineflayer/issues/3209 and probably https://github.com/PrismarineJS/mineflayer/issues/2681.

I've been developing a bot to automate void trading, and while the current villager plugin works fine for trades of 1 emerald to 1 item, I encountered many issues with trades of 1 emerald to multiple items, multiple emeralds to one item, and trades with a second input item. After these changes, all trades I've tested work correctly.

I play on a vanilla 1.20.1 server.

I'll add comments on this PR detailing the reason for each change.


lib/plugins/inventory.js
```diff
@@ -391,15 +391,20 @@ function inject (bot, { hideErrors }) {
   }
 
   function tradeMatch (limitItem, targetItem) {
-    return targetItem.type === limitItem.type && targetItem.count >= limitItem.count
+    return (
+      targetItem !== null &&
+      limitItem !== null &&
+      targetItem.type === limitItem.type &&
+      targetItem.count >= limitItem.count
+    )
```
evan-goode, line 399: Sometimes, `tradeMatch` is called with a null `targetItem`. If there is nothing in the `targetItem` slot, then the trade requirements cannot be satisfied, so return false in that case.

lib/plugins/inventory.js
```diff
@@ -391,15 +391,20 @@ function inject (bot, { hideErrors }) {
...
   }
 
   function expectTradeUpdate (window) {
     const trade = window.selectedTrade
     const hasItem = !!window.slots[2]
 
     if (hasItem !== tradeMatch(trade.inputItem1, window.slots[0])) {
-      if (trade.hasItems2) {
```
evan-goode, line 402: This was a typo, the field is called `hasItem2`.

lib/plugins/inventory.js
```diff
@@ -418,7 +423,7 @@ function inject (bot, { hideErrors }) {
       }
     } else if (window.type === 'minecraft:merchant') {
       const toUpdate = []
-      if (slot <= 2 && !window.selectedTrade.tradeDisabled && expectTradeUpdate(window)) {
```
evan-goode, line 421: The `updateSlot:2` event is already awaited by `promisePutAway` in `putAway()`, so if we await it here too, we'll never get it. I think `waitForWindowUpdate` should only wait for updates on slots other than the slot it was called on.

lib/plugins/inventory.js
```diff
@@ -427,6 +432,12 @@ function inject (bot, { hideErrors }) {
         }
       }
       await Promise.all(toUpdate)
+
+      if (slot === 2 && !window.selectedTrade.tradeDisabled && expectTradeUpdate(window)) {
+        // After the trade goes through, if the inputs are still satisfied,
+        // expect another update in slot 2
+        await once(bot.currentWindow, 'updateSlot:2')
+      }
```
evan-goode, line 440: In the trade loop, if the requirements of a trade are still satisfied after `putAway` but the window doesn't have an item in slot 2, then `putRequirements` will not deposit anything, an update in slot 2 will not be awaited, and `putAway` will be immediately called again without anything in slot 2. So here, we should only consider the trading window "fully updated" if the state of slot 2 correctly reflects whether the trade inputs are satisfied.

lib/plugins/villager.js
```diff
@@ -141,8 +141,11 @@ function inject (bot, { version }) {
       itemCount2 = villager.count(Trade.inputItem2.type, Trade.inputItem2.metadata)
       hasEnoughItem2 = itemCount2 >= Trade.inputItem2.count * count
     }
-    if (!hasEnoughItem1 || !hasEnoughItem2) {
-      throw new Error('Not enough items to trade')
+    if (!hasEnoughItem1) {
+      throw new Error('Not enough item 1 to trade')
+    }
+    if (!hasEnoughItem2) {
+      throw new Error('Not enough item 2 to trade')
```
evan-goode, line 148: Log a bit more information here. I'm happy to leave this out if it amounts to an unnecessary breaking change.

lib/plugins/villager.js
```diff
@@ -152,31 +155,12 @@ function inject (bot, { version }) {
...
-              if (oldItem.type === Trade.inputItem1.type) countOfItemOneLeftToTake -= oldItem.count
-              else if (Trade.hasItem2 && oldItem.type === Trade.inputItem2.type) countOfItemTwoLeftToTake -= oldItem.count
-            }
-            if (countOfItemOneLeftToTake === 0 && countOfItemTwoLeftToTake === 0) {
-              villager.off('updateSlot', listener)
-              resolve()
-            }
-          }
-          villager.on('updateSlot', listener)
-        })
```
evan-goode, line 173: The bot consistently gets hung up here, and I honestly can't figure out what the purpose of this is. It's possible that it was necessary before some of the other changes I made. @nickelpro do you remember by chance?

lib/plugins/villager.js
```diff
@@ -215,24 +199,25 @@ function inject (bot, { version }) {
...
-
-    const input1 = !slot1
-      ? neededCount1
-      : (slot1.count < tradeCount1 ? neededCount1 - slot1.count : 0)
-    await deposit(window, type1, metadata1, input1, 0)
+    const { type: type1, metadata: metadata1 } = Trade.inputItem1
+
+    const input1 = slot1
+      ? Math.max(0, Trade.realPrice - slot1.count)
+      : Trade.realPrice
```
evan-goode, line 208: Only put the requirements for a single trade at a time. This seems to greatly increase reliability, especially for trades with multiple inputs.

lib/plugins/villager.js
```diff
@@ -215,24 +199,25 @@ function inject (bot, { version }) {
...
+    if (input1) {
+      await deposit(window, type1, metadata1, input1, 0)
```
evan-goode, line 210: Only call `deposit` if we actually need to move items

evan-goode:
Could someone with permissions perhaps re-run https://github.com/PrismarineJS/mineflayer/actions/runs/6843109062/job/18605343146?pr=3230? The tests pass locally for me, this seems like a transient error unrelated to the changes.

rom1504:
can you add a test to make sure what you fixed will keep working in the future ?

evan-goode:
> can you add a test to make sure what you fixed will keep working in the future ?

Sure, I added a couple tests in `test/externalTests/trade.js` for the types of trade that weren't working.

PR #3254, opened 2023-12-27
Fix for a possible error in lib/plugins/entities.js file

Mykola1453:
I'm not sure if this error happens with vanilla Minecraft, but when I tried to use mineflayer with node-minecraft-protocol-forge, I receive the following error shortly after my bot joins a server with TeamAbnormals mods:

```
    entity.type = entityData.type || 'object'
                             ^

TypeError: Cannot read properties of undefined (reading 'type')
```

The pull request fixes this error by adding an if statement around `entity.type = entityData.type || 'object'` line.


lib/plugins/entities.js
```diff
@@ -197,7 +197,9 @@ function inject (bot) {
     const entity = fetchEntity(packet.entityId)
     const entityData = bot.registry.entities[packet.type]
 
-    entity.type = entityData.type || 'object'
```
rom1504, line 200: can you rather move this line to setEntityData ?

PR #3256, opened 2023-12-27
Improve CI per version setup

rom1504:
- use +v to select the version properly (eg don't run all 1.19.x for 1.19)
- run on all tested versions specified in version.js


rom1504:
cleaner and decrease overall time from 12min to 4min

PR #3259, opened 2023-12-30
Fix version check

extremeheat:
Now throws
```
Error: Version '1.20.2' is not supported. Latest supported version is '1.20.1'.
    at Client.next (/workspaces/mineflayer/lib/loader.js:121:13)
    at Object.onceWrapper (node:events:628:28)
    at Client.emit (node:events:526:35)
    at /workspaces/mineflayer/node_modules/minecraft-protocol/src/client/autoVersion.js:53:12
    at /workspaces/mineflayer/node_modules/minecraft-protocol/src/ping.js:13:7
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```
on mineflayer attempting to connect to a 1.20.2 server.

A test is possible, but turns out it would require some monkey patching to minecraft-data and nmp as the node-minecraft-protocol version checks are strict.


rom1504:
That fix may work for this particular case

But I think it's not the semantic we want.

I think it makes sense that we explicitly put in version.js the range of versions that are supported. And then this check should make sure the actual server version is in that range.

The reason why I think it should be like this is that presence of the data in Minecraft data or support by nmp is not enough. What's needed is
* The supported versions should have a test of a version later than it
* someone should manually check things are working before including a version in the range of tested versions

I think it's unlikely we will support a non contiguous range in the future so what I would propose is to have something like 
supportedVersionRange: [1.7, 1.20.1] in version.js

And then check this thanks to the >= and <= operators that are defined for versions in node-minecraft-data

rom1504:
Ah actually rereading your code I think it's actually almost implementing what I want

What's missing is
* Checking the version is old not too old
* deciding if we need the supported version range or if tested versions is good enough. Maybe tested versions is indeed better and more explicit

lib/loader.js
```diff
@@ -107,22 +109,21 @@ function createBot (options = {}) {
...
+    const serverPingVersion = bot._client.version
+    bot.registry = require('prismarine-registry')(serverPingVersion)
+    if (!bot.registry?.version) throw new Error(`Server version '${serverPingVersion}' is not supported, no data for version`)
 
-    const latestTestedVersion = testedVersions[testedVersions.length - 1]
-    const latestProtocolVersion = require('prismarine-registry')(latestTestedVersion).protocolVersion
-    if (version.protocolVersion > latestProtocolVersion) {
-      throw new Error(`Version ${version.minecraftVersion} is not supported. Latest supported version is ${latestTestedVersion}.`)
+    const versionData = bot.registry.version
+    if (versionData.version > latestSupportedVersionData.version) {
```
rom1504, line 117: how is that operator working?

lib/version.js
```diff
@@ -1,4 +1,8 @@
+const supportedVersions = ['1.8', '1.9', '1.10', '1.11', '1.12', '1.13', '1.14', '1.15', '1.16', '1.17', '1.18', '1.19', '1.20']
+const testedVersions = ['1.8.8', '1.9.4', '1.10.2', '1.11.2', '1.12.2', '1.13.2', '1.14.4', '1.15.2', '1.16.5', '1.17.1', '1.18.2', '1.19', '1.19.2', '1.19.3', '1.19.4', '1.20.1']
 module.exports = {
-  supportedVersions: ['1.8', '1.9', '1.10', '1.11', '1.12', '1.13', '1.14', '1.15', '1.16', '1.17', '1.18', '1.19', '1.20'],
-  testedVersions: ['1.8.8', '1.9.4', '1.10.2', '1.11.2', '1.12.2', '1.13.2', '1.14.4', '1.15.2', '1.16.5', '1.17.1', '1.18.2', '1.19', '1.19.2', '1.19.3', '1.19.4', '1.20.1']
+  supportedVersions,
+  testedVersions,
+  latestSupportedVersion: testedVersions[testedVersions.length - 1],
+  oldestSupportedVersion: testedVersions[0]
```
extremeheat, line 7: Added these two variables to the versions.js. So if there is a change and tested versions is not the latest supported, it can be set here. Otherwise if needed we can also manually define these instead of being first/last tested versions.

lib/version.js
```diff
@@ -1,4 +1,8 @@
+const supportedVersions = ['1.8', '1.9', '1.10', '1.11', '1.12', '1.13', '1.14', '1.15', '1.16', '1.17', '1.18', '1.19', '1.20']
```
rom1504, line 1: I think we do not use this anymore, so I would remove it

lib/loader.js
extremeheat: versionData is a version object in minecraft-data, https://github.com/PrismarineJS/node-minecraft-data/blob/master/doc/api.md#mcdataversionversion. The `.version` is the protocol version.

lib/version.js
rom1504: that makes sense yeah; but I think we'll keep testing supported versions

rom1504: in fact I suggest we expose these 2 variables in https://github.com/PrismarineJS/mineflayer/blob/master/lib/loader.js#L58 (there and in the doc) and stop exposing supportedVersions

lib/loader.js
rom1504: is that using https://github.com/PrismarineJS/node-minecraft-data/blob/master/index.js#L44 ?

lib/loader.js
```diff
@@ -57,7 +57,7 @@ module.exports = {
   ScoreBoard: require('./scoreboard'),
   BossBar: require('./bossbar'),
   Particle: require('./particle'),
-  supportedVersions,
+  supportedVersions: testedVersions,
```
rom1504, line 60: I think that is not right

we support many more versions than the tested versions

let's expose oldestSupportedVersion and oldestSupportedVersion and stop exposing supportedVersions

it's slightly breaking but I think it should be ok

lib/loader.js
extremeheat: No, those functions check against dataVersion, this check was supposed to check protocol version but the previous logic was wrong

rom1504: I think protocol version is not enough, we should check against the version ordering in minecraft data

sometimes mojang does not increase the protocol version but makes mineflayer-level breaking changes

rom1504: why not use the operators defined in nmd ?

extremeheat: ok, updated to use dataVersion comparison functions in node-mcdata

rom1504:
thanks

rom1504:
maybe we should also adapt the readme to mention the begin and end of the range

PR #3243, opened 2023-12-12
Initial 1.20.2 support

crux153:
This PR adds preliminary support for version 1.20.2.

Before merging this PR, we need PrismarineJS/prismarine-nbt#78 and PrismarineJS/node-minecraft-protocol#1264 to be merged and released.

I have tested that a bot can successfully log in to an offline server and interact with chunks using this patch. But I believe additional testing is necessary.

PR #3262, opened 2023-12-30
Minecraft 1.20.2 support

rom1504:
migrated from https://github.com/PrismarineJS/mineflayer/pull/3243


lib/plugins/entities.js
```diff
@@ -212,6 +213,14 @@ function inject (bot) {
     entity.yaw = conv.fromNotchianYawByte(packet.yaw)
     entity.pitch = conv.fromNotchianPitchByte(packet.pitch)
     entity.objectData = packet.objectData
+
+    if (entity.type === 'player') {
```
rom1504, line 217: that looks pretty suspicious, not sure if it's right

wgaylord: I suspect it is correct based on existing code using the opposite condition. https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/entities.js#L42

Not sure if what its being used for is correct tho.

wgaylord:
I think the only thing left is to fix what ever is breaking the External Bed and Elytra tests. Then fix the Internal tests since they are all totally failing.

wgaylord:
Found https://github.com/PrismarineJS/minecraft-data/issues/817 to be why both the External Bed and Elytra tests are failing

lib/plugins/entities.js
extremeheat: named_entity_spawn was removed, so players spawn with the generic spawn_entity packet. However there are alot of unhandled fields that are not currently accounted for - https://github.com/PrismarineJS/mineflayer/blob/a0a577763d08f44494669793f8f4a2f8b2d58e0f/lib/plugins/entities.js#L132

wgaylord:
okay, so the major thing stopping this is changes to the entity spawning packet in both Mineflayer and in the internal tests
https://github.com/PrismarineJS/mineflayer/blob/mc-1.20.2/lib/plugins/entities.js#L132-L157 is how we used to handle player spawns. For 1.20.2+ players spawn using the same as other mobs, partially implemented here https://github.com/PrismarineJS/mineflayer/blob/mc-1.20.2/lib/plugins/entities.js#L196-L224 We need to fill out the rest of the fields for the player.

This same sort of change needs to happen on in the integral tests based on version.  Where https://github.com/PrismarineJS/mineflayer/blob/mc-1.20.2/test/internalTest.js#L635-L647 and https://github.com/PrismarineJS/mineflayer/blob/mc-1.20.2/test/internalTest.js#L700-L712 need to be setup so if we are not 1.20.2+ they are used and if we are 1.20.2+ something like https://github.com/PrismarineJS/mineflayer/blob/mc-1.20.2/test/internalTest.js#L786-L800 is used (of course with the correct values.)

rom1504:
@wgaylord anything unclear about it or you just need to do it?

wgaylord:
Mostly just need to do it, wanted to document exactly what was still needed, incase I don't get the time to work on it.

extremeheat:
Ok, should be passing now. Required changes were:

* server.on `login` (packet) -> `playerJoin` (nmp event) to handle play state
* named_entity_spawn is now done with spawn_entity packet
* dimension codec is sent outside login packet now (before play state)

test/internalTest.js
```diff
@@ -632,17 +632,34 @@ for (const supportedVersion of mineflayer.testedVersions) {
...
-            playerUUID: '1-2-3-4',
-            x: 1,
-            y: 2,
-            z: 3,
-            yaw: 0,
-            pitch: 0,
-            currentItem: -1,
-            metadata: []
-          })
+          if (bot.registry.version['>=']('1.20.2')) {
```
rom1504, line 635: would be good to make that a feature

rom1504:
great

test/internalTest.js
extremeheat: done, looks like removing dupe https://github.com/PrismarineJS/minecraft-data/commit/587a1341c83ef4fc1cf3353d35b09435a11f8217 caused range to be incorrect, just fixed that

rom1504:
Remaining here, test some examples manually
* [ ] Inventory.js
* [ ] digger
* [ ] chest
* [ ] check previous PRs

rom1504:
I tested digger, chest, inventory, jumper, viewer and pathfinder, all working

PR #3263, opened 2023-12-30
Add Chunk Batch support

wgaylord:
Adds support for 1.20.2 chunk batches.

### DIFF
--- a/lib/plugins/blocks.js
+++ b/lib/plugins/blocks.js
@@ -274,7 +274,7 @@ function inject (bot, { version, storageBuilder, hideErrors }) {
   })
 
   bot._client.on('chunk_batch_finished', (packet) => {
-    let chunkPerMill = packet.batchSize / (Date.now() - chunkBatchStartTime) // Get chunks per millisecond
+    const chunkPerMill = packet.batchSize / (Date.now() - chunkBatchStartTime) // Get chunks per millisecond
     bot._client.write('chunk_batch_received', {
       chunksPerTick: 25 / chunkPerMill
     })
--- a/lib/plugins/blocks.js
+++ b/lib/plugins/blocks.js
@@ -278,9 +278,9 @@ function inject (bot, { version, storageBuilder, hideErrors }) {
   })
 
   bot._client.on('chunk_batch_finished', (packet) => {
-    const MilliPerChunk = (Date.now() - chunkBatchStartTime) / packet.batchSize // Gets millisecond per chunk
-    const ClampedMilliPerChunk = Math.min(Math.max(MilliPerChunk, weightedAverage / 3.0), weightedAverage * 3.0) // Prevents the MilliPerChunk from being hugely different then the average.
-    weightedAverage = ((weightedAverage * oldSampleWeight) + ClampedMilliPerChunk) / (oldSampleWeight + 1)
+    const milliPerChunk = (Date.now() - chunkBatchStartTime) / packet.batchSize // Gets millisecond per chunk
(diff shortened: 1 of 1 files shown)


lib/plugins/blocks.js
```diff
@@ -268,6 +268,17 @@ function inject (bot, { version, storageBuilder, hideErrors }) {
     }
   })
 
+  let chunkBatchStartTime = 0
+  bot._client.on('chunk_batch_start', (packet) => {
+    chunkBatchStartTime = Date.now()
+  })
+
+  bot._client.on('chunk_batch_finished', (packet) => {
```
rom1504, line 276: can you add a comment explaining what is the purpose of this system ?

lib/plugins/blocks.js
```diff
@@ -268,6 +268,19 @@ function inject (bot, { version, storageBuilder, hideErrors }) {
...
+  // Chunk batches are used by the server to throttle the chunks per tick for players based on their connection speed.
+  let chunkBatchStartTime = 0
+  bot._client.on('chunk_batch_start', (packet) => {
+    chunkBatchStartTime = Date.now()
+  })
+
+  bot._client.on('chunk_batch_finished', (packet) => {
+    const chunkPerMill = packet.batchSize / (Date.now() - chunkBatchStartTime) // Get chunks per millisecond
+    bot._client.write('chunk_batch_received', {
+      chunksPerTick: 25 / chunkPerMill
```
rom1504, line 281: what is 25 ?

wgaylord: A constant the client uses in that calculation.

rom1504: do you know why?

rom1504: / what is the source of this ?

wgaylord: Well, now I feel dumb and have learned to always double check wiki.vg's info since there is a good chance it could be wrong or slightly incorrect. 

Normal client uses this insanity to get its chunksPerTick.  

https://github.com/extremeheat/extracted_minecraft_data/blob/client1.20.2/client/net/minecraft/client/multiplayer/ChunkBatchSizeCalculator.java 

While the input to the onBatchFinished is the batchsize.  No sure where someone got the 25 from and put it on wiki.vg going to probably correct that information later.

lib/plugins/blocks.js
wgaylord: Added a comment and comments each part of the calculation now that I am doing it correctly.

lib/plugins/blocks.js
wgaylord: Updated to used the same calculation as 1.20.2 Vanilla client uses.  (Although converted to millisecond instead of the nanoseconds they use.)

lib/plugins/blocks.js
```diff
@@ -268,6 +268,24 @@ function inject (bot, { version, storageBuilder, hideErrors }) {
...
+  let chunkBatchStartTime = 0
+  let weightedAverage = 2 // The Vanilla client uses nano seconds with its weighted average starting at 2000000 converted to milliseconds that is 2
+  let oldSampleWeight = 1 // This is used for keeping track of the weight of the old average when updating it.
+
+  bot._client.on('chunk_batch_start', (packet) => {
+    chunkBatchStartTime = Date.now() // Get the time the chunk batch is starting.
+  })
+
+  bot._client.on('chunk_batch_finished', (packet) => {
+    const MilliPerChunk = (Date.now() - chunkBatchStartTime) / packet.batchSize // Gets millisecond per chunk
```
rom1504, line 281: Should start variable named with lower case

wgaylord: I will fix that when I get home, surprised the linter didnt carch that.

wgaylord: Both now start with lower case

PR #3264, opened 2023-12-30
Make explicit supported versions in readme

rom1504:


rom1504:
what do you think @extremeheat ?

PR #3277, opened 2024-01-05
fix types: Add `maxCatchupTicks`, improve `supportFeature` typing

zardoy:


index.d.ts
```diff
@@ -26,6 +27,8 @@ export interface BotOptions extends ClientOptions {
   difficulty?: number
   chatLengthLimit?: number
   physicsEnabled?: boolean
+  /** @default 4 */
+  maxCatchupTicks?: number
```
zardoy, line 31: fyi will be using 0 in p web client because of timer issues in browsers

PR #3279, opened 2024-01-08
Update recommended Node.js version

Nyaasu66:
Correct recommended Node.js version in README and `package.json` to reduce misunderstandings.


Nyaasu66:
According to this line of code:
https://github.com/PrismarineJS/mineflayer/blob/3a6ce543b4ba8a3d0f55777670d142968af14571/index.js#L3C84-L3C84
The recommended version number of Node.js should be upgraded to 18.

PR #3280, opened 2024-01-08
Deprecate and rename `respawn` option to `autoRespawn`

### LINKED ISSUE #3276: rename `respawn` option to `autoRespawn`
for clarity, let's deprecate `respawn` option in favor of more clear `autoRespawn`, also there is no way to disable `respawn`  disable ability in general so this might be confusing

from https://github.com/PrismarineJS/mineflayer/pull/3206#issuecomment-1737921758

zardoy:
fixes #3276

### DIFF
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -70,7 +70,7 @@ function createBot (options = {}) {
   options.loadInternalPlugins = options.loadInternalPlugins ?? true
   options.client = options.client ?? null
   options.brand = options.brand ?? 'vanilla'
-  options.autoRespawn = options.autoRespawn ?? options.autoRespawn ?? true
+  options.autoRespawn = options.autoRespawn ?? options.respawn ?? true
   const bot = new EventEmitter()
   bot._client = options.client
   bot.end = (reason) => bot._client.end(reason)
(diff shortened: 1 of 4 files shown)


lib/loader.js
```diff
@@ -70,7 +70,7 @@ function createBot (options = {}) {
   options.loadInternalPlugins = options.loadInternalPlugins ?? true
   options.client = options.client ?? null
   options.brand = options.brand ?? 'vanilla'
-  options.respawn = options.respawn ?? true
+  options.autoRespawn = options.autoRespawn ?? options.autoRespawn ?? true
```
extremeheat, line 73: Incorrect logic

Also, the doc seems already pretty clear. Is there anyone that's been confused by this?

zardoy: oops, fixed

rom1504:
why is it worth it to do such a breaking change for a minor renaming?

zardoy:
> why is it worth it to do such a breaking change for a minor renaming?

where did you find the breaking change?

the pr is here because of https://github.com/PrismarineJS/mineflayer/issues/3276#issuecomment-1881884816, if for some reason you don't like it feel free to close it...

rom1504:
Removing the option `options.respawn` is a breaking change

zardoy:
> Removing the option options.respawn is a breaking change

But I didn't remove support for options.respawn, old option should still be supported. am I missing something?

rom1504:
ah yeah I see, yeah then it's not breaking

not sure if it's really useful though, but maybe...

zardoy:
Closing this so you can prioritize a lot of other more important prs

PR #3285, opened 2024-01-18
Updated digging code to account for raycasted tall grass checks

GenerelSchwerz:
Currently, prismarine-world's implementation of raycasting does not account for blocks with no shapes. If attempting to dig a block such as tall grass, the current raycast implementation fails due to not detecting an intersection and assuming the raycast failed.

This check does boundary checking for found blocks and considers their distance away from the eyePos. If the collision is closer than the expected targetPos intersect, then that viewing angle is invalid.

If no visible faces are detected via the raytrace, then we check for if there were any closer collisons. If there weren't and the block we are checking also has no faces, assume that we can look at the block from any tried face and continue with the dig accordingly.

There is a small TODO to calculate the digging face based on an AABB + ray intersection check, but that is not currently implemented in prismarine-physics's AABB code. An example can be found here: https://github.com/nxg-org/mineflayer-util-plugin/blob/af6612c809e040409365b9f1238aa0af96422f66/src/calcs/aabb.ts#L231

### DIFF
--- a/lib/plugins/digging.js
+++ b/lib/plugins/digging.js
@@ -14,12 +14,11 @@ function inject (bot) {
   bot.targetDigBlock = null
   bot.lastDigTime = null
 
-
   /**
-   * 
-   * @param {require('prismarine-block').Block} block 
-   * @param {boolean | 'ignore'} forceLook 
-   * @param {Function | 'auto' | 'raycast'} digFace 
+   *
+   * @param {require('prismarine-block').Block} block
+   * @param {boolean | 'ignore'} forceLook
+   * @param {Function | 'auto' | 'raycast'} digFace
    */
   async function dig (block, forceLook, digFace) {
     if (block === null || block === undefined) {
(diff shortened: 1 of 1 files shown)


lib/plugins/digging.js
```diff
@@ -95,6 +102,11 @@ function inject (bot) {
           }
           await bot.lookAt(closest.targetPos, forceLook)
           diggingFace = closest.face
+        } else if (closerBlocks.length === 0 && block.shapes.length === 0) {
+          // no other blocks were detected and the block has no shapes.
+          // The block in question is replaceable (like tall grass) so we can just dig it
+          // TODO: do AABB + ray intercept check to this position for diggingFace.
+          await bot.lookAt(block.position.offset(0.5, 0.5, 0.5), forceLook)
```
extremeheat, line 109: Ok, my understanding here is if the user doesn't specify a digging direction, mineflayer defaults to doing a raytrace to find the direction to look at before digging.

You want to:
1. prevent ray trace from matching a block closer than the expected target
2. handle raycast failures because the block has no shape, as raycast detection needs shapes to determine a collision by preventing the exception in the else condition (code already defaults to a hardcoded constant)

minecraft-assets should have the block model data, so prismarine-world should be able to refer to that when checking for collisions as opposed to the blockCollisions.json in minecraft-data

But otherwise LGTM for a temporary fix in mineflayer

extremeheat (approved):

rom1504:
I think we should remove ray casting call from digging
This should be called by the user independently

PR #3293, opened 2024-01-27
Timeout for bot.tabComplete()

Ynfuien:
I added a timeout, because without it, a promise was never* resolved, if provided command/string had no completions.

*Never, meaning till next successful tab completion, when it would just double the newer result.

I set the default timeout to, what I think is enough low/high, 100 ms. But it can be changed with a function argument, so everyone will set whatever they want.


lib/plugins/chat.js
```diff
@@ -182,7 +182,15 @@ function inject (bot, options) {
       lookedAtBlock: position
     })
 
-    const [packet] = await once(bot._client, 'tab_complete')
+    // Doing promise race, because of 'tab_complete' event never happening,
+    // in situation when there was no completions for tabbed command.
+    const eventPromise = once(bot._client, 'tab_complete')
```
extremeheat, line 187: We already have an onceWithCleanup function that handles timeouts. You should use that instead here.

Ynfuien: I think now is good?

lib/plugins/chat.js
```diff
@@ -1,4 +1,4 @@
-const { once } = require('events')
+const { onceWithCleanup } = require('mineflayer/lib/promise_utils')
```
extremeheat, line 1: `require('../promise_utils')`

lib/plugins/chat.js
```diff
@@ -165,7 +165,7 @@ function inject (bot, options) {
     })
   }
 
-  async function tabComplete (text, assumeCommand = false, sendBlockInSight = true) {
+  async function tabComplete (text, assumeCommand = false, sendBlockInSight = true, timeout = 100) {
```
extremeheat, line 168: 100ms seems rather low of a timeout for high latency connections

Ynfuien: Then someone can change it for their use case. But if you think that e.g. 200 would be better, then I can change it

Ynfuien: How about now? It's based on bot's ping

extremeheat: No, that's worse. Ideally we stick to something at least 1 second, high latency connections are perfectly valid, they should not be an error case. Also, we shouldn't have every function having inconsistent timeout values, that makes things very messy.

I just checked the mineflayer code for existing timeouts, for blocks.js and command blocks and uses [5000ms timeout](https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/place_block.js#L14), so I'd say use that here also.

Ynfuien: I don't see why dynamic timeout would be worse, but alright, I will change it to 5s

extremeheat (approved):
LGTM, but I'm not sure returning an empty array is any better than letting an exception propagate to the user

lib/plugins/chat.js
```diff
@@ -182,8 +182,12 @@ function inject (bot, options) {
       lookedAtBlock: position
     })
 
-    const [packet] = await once(bot._client, 'tab_complete')
-    return packet.matches
+    try {
+      const [packet] = await onceWithCleanup(bot._client, 'tab_complete', { timeout })
+      return packet.matches
+    } catch {
+      return []
```
rom1504, line 189: Yes I think let's not try catch here. The user should handle the error appropriately (for example maybe by retrying)

Ynfuien: Done

PR #3302, opened 2024-02-08
fix world typing

GenerelSchwerz:
The world typing for mineflayer is incorrect.


GenerelSchwerz:
See line [here](https://github.com/PrismarineJS/mineflayer/blob/cc98f1307e3ab48477d2a9ff29da4447f42b30bc/lib/plugins/blocks.js#L524), We are using the Sync version of prismarine-world, not Async. As such, the typing of world.World is incorrect in favor of world.WorldSync

PR #3304, opened 2024-02-09
Update api.md - addChatPattern example of usage

boly38:
Update api.md - addChatPattern example of usage

Update api.md - addChatPatternSet example of usage

If we/You dont want to maintain example per API (which would reduce the number of user questions..)
another way to suggest example is to add a link to the test : example [chat_parsing](https://github.com/PrismarineJS/mineflayer/blob/b9491ae508dc52e4538dd66eb6f47a639facde20/examples/chat_parsing.js#L17-L36) IDK

anyway, thanks for this wonderfull library 👍

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1691,6 +1691,8 @@ the event will be called `"chat:name"`, with name being the name passed
 
 returns a number which can be used with bot.removeChatPattern() to only delete this pattern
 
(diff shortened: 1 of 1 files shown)


rom1504:
What about instead pointing to the existing chat pattern example?

boly38:
Ok I ll amend tomorrow ^^

boly38:
I added 👀 next to usage link
image

PR #3306, opened 2024-02-15
Deleted scoreboards not being updated

Ynfuien:
Deleted scoreboards were deleted from `bot.scoreboards`, but not from `bot.scoreboard` (`ScoreBoard.positions`). In result, a scoreboard existed till another one would take it's place.


rom1504:
thanks

PR #3320, opened 2024-03-01
Fix 1.20.4 server resource pack error

TerminalCalamitas:
I have updated the code to allow for a bot to join a 1.20.4 server with a specified resource pack. There is still an error in the bot.emit() method or the minecraft-protocol/src/datatypes/compiler-minecraft.js:48:25 because compiler-minecraft.js is unable to read the uuid length.

### DIFF
--- a/lib/plugins/resource_pack.js
+++ b/lib/plugins/resource_pack.js
@@ -2,7 +2,8 @@ module.exports = inject
 
 function inject (bot) {
   let latestHash
-  const activeResourcePacks = new Set();
+  let latestUUID
+  const activeResourcePacks = {};
   const TEXTURE_PACK_RESULTS = {
     SUCCESSFULLY_LOADED: 0,
     DECLINED: 1,
@@ -10,21 +11,21 @@ function inject (bot) {
     ACCEPTED: 3
   }
 
-  bot._client.on('add_resource_pack', (data) => { // Emits the same as resource_pack_send but also uuid because that's how active packs are tracked
-    bot.emit('resourcePack', data.url, data.hash, data.uuid)
+  bot._client.on('add_resource_pack', (data) => { // Emits the same as resource_pack_send but sends uuid rather than hash because that's how active packs are tracked
+    bot.emit('resourcePack', data.url, data.uuid)
     // Adding the pack to a set by uuid
-    activeResourcePacks.add(data.uuid)
-    latestHash = data.hash
+    latestUUID = data.uuid
+    activeResourcePacks[data.uuid] = data.url
   })
 
   bot._client.on('remove_resource_pack', (data) => { // Doesn't emit  anything because it is removing rather than adding
     // if uuid isn't provided remove all packs
     if (data.uuid === undefined) {
-      activeResourcePacks.clear();
+      activeResourcePacks = {};
     } else {
       // Try to remove uuid from set
       try {
-        activeResourcePacks.remove(data.uuid)
+        delete activeResourcePacks[data.uuid]
       } catch (error) {
          console.error("Tried to remove UUID but it was not in the active list.");
       }
@@ -45,6 +46,11 @@ function inject (bot) {
       bot._client.write('resource_pack_receive', {
         result: TEXTURE_PACK_RESULTS.SUCCESSFULLY_LOADED,
         hash: latestHash
+      })
+    } else if (bot.version == '1.20.3' || bot.version == '1.20.4') {
+      bot._client.write('resource_pack_recieve', {
+        uuid: latestUUID,
+        result: TEXTURE_PACK_RESULTS.ACCEPTED
       })
     } else {
       bot._client.write('resource_pack_receive', {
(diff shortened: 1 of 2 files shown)


wgaylord:
This may be because the packet expects the uuid to be the UUID type not a string.

wgaylord:
Actually could you share the actual error your getting? Want to make sure I am not saying something incorrect.

TerminalCalamitas:
Here's the error
```
TypeError: Serialization error for configuration.toServer : Write error for undefined : Cannot read properties of undefined (reading 'length')
    at module.exports.Write.UUID (/home/caleb/node_modules/minecraft-protocol/src/datatypes/compiler-minecraft.js:48:25)
    at Object.packet_resource_pack_receive (eval at compile (/home/caleb/node_modules/protodef/src/compiler.js:258:12), <anonymous>:783:26)
    at eval (eval at compile (/home/caleb/node_modules/protodef/src/compiler.js:258:12), <anonymous>:801:82)
    at packet (eval at compile (/home/caleb/node_modules/protodef/src/compiler.js:258:12), <anonymous>:804:9)
    at CompiledProtodef.write (/home/caleb/node_modules/protodef/src/compiler.js:76:12)
    at e.message (/home/caleb/node_modules/protodef/src/compiler.js:102:25)
    at tryCatch (/home/caleb/node_modules/protodef/src/utils.js:50:16)
    at CompiledProtodef.createPacketBuffer (/home/caleb/node_modules/protodef/src/compiler.js:102:5)
    at Serializer.createPacketBuffer (/home/caleb/node_modules/protodef/src/serializer.js:12:23)
    at Serializer._transform (/home/caleb/node_modules/protodef/src/serializer.js:18:18) {
  field: 'configuration.toServer'
}
```

TerminalCalamitas:
Oh, this is the packet I'm basing the code off of in case it helps: https://wiki.vg/Protocol#Registry_Data

lib/plugins/resource_pack.js
```diff
@@ -24,6 +47,11 @@ function inject (bot) {
         result: TEXTURE_PACK_RESULTS.SUCCESSFULLY_LOADED,
         hash: latestHash
       })
+    } else if (bot.version === '1.20.3' || bot.version === '1.20.4') {
```
rom1504, line 50: Please use support feature function

lib/plugins/resource_pack.js
```diff
@@ -24,6 +47,11 @@ function inject (bot) {
...
+      bot._client.write('resource_pack_recieve', {
```
rom1504, line 51: Typo + check protocol.json for packet definition

TerminalCalamitas:
I will fix 1.20.4 not passing by adding more verbose testing once my pr for adding a minecraft-data feature gets merged

TerminalCalamitas:
The code is still waiting for a PR to be merged in the PrismarineJS/minecraft-data repository but  the bot still times out when connecting to a server, this time only giving an error after the bot times out.
[currentLog.log](https://github.com/PrismarineJS/mineflayer/files/14476156/currentLog.log)

wgaylord:
> The code is still waiting for a PR to be merged in the PrismarineJS/minecraft-data repository but the bot still times out when connecting to a server, this time only giving an error after the bot times out. [currentLog.log](https://github.com/PrismarineJS/mineflayer/files/14476156/currentLog.log)

Looks like it doesn't time out, it gets disconnected by the server for a internal connection error. (Probably something unhappy in their Velocity proxy) The error is because https://github.com/PrismarineJS/node-minecraft-protocol/blob/master/src/client/versionChecking.js#L2 hasn't been updated to handle the disconnects during 1.20.4 play and config state. Which use the NBT tags instead of JSON. (While disconnects during login state still use the JSON for compatibility)

TerminalCalamitas:
Running a local paper server, I get something different but the bot still doesn't spawn so I don't think the immediate problem is with the versionChecking.js (although it would still be good to update it).
(note: I cut this log short but It will continue to send keep alive packets until the program is stopped)
[LocalHost.log](https://github.com/PrismarineJS/mineflayer/files/14476564/LocalHost.log)

TerminalCalamitas:
It seems like the problem that I might be running into might be caused by this: https://github.com/PrismarineJS/minecraft-data/pull/853 
I will see if I can work on that to fix the problem.

TerminalCalamitas:
Never mind, it seems like that issue was talking about the prismarine server rather than client.

TerminalCalamitas:
Looking through more logs, it seems like the client.on('configuration.finish_configuration') never gets triggered when using the new resource pack packets, I tried looking through code, but I couldn't find where that was emitted, is there something I'm missing for that to get triggered?

wgaylord:
Thats a packet. Which means the server isnt telling the client to leave
configuration state.

TerminalCalamitas:
I figured out that the problem was that I forgot to send a TEXTURE_PACK_RESULTS.SUCCESSFULLY_LOADED packet

TerminalCalamitas:
Technically the code won't work until this https://github.com/PrismarineJS/minecraft-data/pull/854 minecraft-data pull request is merged but from the mineflayer side, everything is finished.

rom1504:
this is a lot of untested code that will definitely break in the next update from mojang

rom1504:
the code does seem valid though; merging

PR #3324, opened 2024-03-04
Fix scoreboard delete handler not first checking if scoreboard exists

Ynfuien:
Forgot to check if scoreboard on a given position exists.


extremeheat (changes requested):

extremeheat (approved):

extremeheat:
Thanks

PR #3328, opened 2024-03-12
Remove invalid sign check

zardoy:
... or add a way to skip this check, because for some characters like `|` this is simply not correct (e.g. you can have 45 bars on the line): `|||||||||||||||||||||||||||||||||||||||||||||`


extremeheat:
Many of these checks are designed to prevent the bot from getting flagged/kicked by the server. Have you checked how this is handled my vanilla?

zardoy:
> Many of these checks are designed to prevent the bot from getting flagged/kicked by the server. Have you checked how this is handled my vanilla?

Of course, it does handle normally. Actually, I didn't see any restrictions on the line length of the sign and its not stated in the protocol docs either. In the vanilla, the sign editor calculates the width of each character to get the maximum length of the line. For example one `| ' , . ;` character is equal to 1 / 3 of the `a` character. Or `( ) [ ] { }` and space is equal to 1 / 1.46 of regular character and so on. So, the maximum length of the line is still 15 characters but its only if you take into account the specific width of these characters

rom1504:
did you check this for minecraft 1.8 ?

zardoy:
@rom1504 Of course! It does work without any issues. The maximum length you can get is 45 characters on the line

zardoy:
GitHub Action says:

```js
1 failing
1) mineflayer_external 1.20.4v
     sign:
   Uncaught TypeError: Cannot read properties of undefined (reading 'trimEnd')
    at Timeout._onTimeout (test/externalTests/sign.js:23:42)
    at listOnTimeout (node:internal/timers:569:17)
    at process.processTimers (node:internal/timers:512:7)
```

The 400k lines log is crazy, but i don't think its current PR breaks the pipeline

rom1504:
If the max is 45 characters then why not adapt the check instead of removing it ?

zardoy:
> If the max is 45 characters then why not adapt the check instead of removing it ?

Okay, not a problem.

---

As said in https://github.com/PrismarineJS/mineflayer/pull/3328#issuecomment-1996263771 In the original client, if you type only `a` the maximum length is 15, if you type only `.` the maximum length is 45, but again, nothing bad happens if you send 20 `a` to the vanilla server

zardoy:
Is this good to go?

zardoy:
@extremeheat could you look into this as well? Don't see anything that blocks the pr

extremeheat (approved):

PR #3335, opened 2024-03-24
Update README.md

ChrisArbogast:
Added a more detailed code example, showing how to make a command and bot response.


extremeheat:
The point of the example is to be simple. Your change seems to make it more complex by wrapping one event in another event and removing the recursion check, so I don't see the improvement here

ChrisArbogast:
> The point of the example is to be simple. Your change seems to make it more complex by wrapping one event in another event and removing the recursion check, so I don't see the improvement here
@extremeheat 

Understandable enough. I thought it might be nice to give a more in depth example, because even though it is slightly more confusing, it is more detailed and better. Apologies, have a great day.

PR #3337, opened 2024-03-27
refactor/simplifying_code

SnowRunescape:
Subject: Simplification of Code - Enhancing Efficiency

### DIFF
--- a/lib/plugins/ray_trace.js
+++ b/lib/plugins/ray_trace.js
@@ -17,12 +17,7 @@ module.exports = (bot) => {
   }
 
   bot.blockAtCursor = (maxDistance = 256, matcher = null) => {
-    const { position, height, pitch, yaw } = bot.entity
-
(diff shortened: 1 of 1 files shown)


zardoy:
I just wonder why maxDistance was chosen to be 255 instead of 5? It's seems to be more logical to replicate the vanilla behaviour …

SnowRunescape:
> I just wonder why maxDistance was chosen to be 255 instead of 5? It's seems to be more logical to replicate the vanilla behaviour …

god question, i removed only the code repetition

wgaylord:
> I just wonder why maxDistance was chosen to be 255 instead of 5? It's seems to be more logical to replicate the vanilla behaviour …

I think its because its used for finding a block in line of sight? Such as it your looking around, 256 seems like it would allow for seeing what block your looking towards out to 15 chunks away from the player at max. 

That is just my thoughts on it.

PR #3355, opened 2024-04-20
Added the serverBrand property to index.d.ts

Khaogamermain01:


Khaogamermain01:
Hmm seems like a sign error. It doesn't look like it was affected by my commit.

PR #3366, opened 2024-05-02
Update typescript and fix dts syntax

mallusrgreatv2:
Updates typescript version and makes changes in the definition file to fix syntax.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,6 @@
     "mocha": "^10.0.0",
     "standard": "^17.0.0",
     "standard-markdown": "^7.1.0",
-    "typescript": "^5.0.3"
+    "typescript": "^5.4.5"
   }
 }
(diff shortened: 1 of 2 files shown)


extremeheat (approved):
LGTM, but we should have a way to test out the types to make sure they're valid at least

package.json
```diff
@@ -47,6 +47,6 @@
     "mocha": "^10.0.0",
     "standard": "^17.0.0",
     "standard-markdown": "^7.1.0",
-    "typescript": "^5.0.3"
+    "typescript": "^5.4.5"
```
extremeheat, line 50: since we are using the `^` (caret) operator, npm will automatically install the latest version of `typescript@5.* `, see https://docs.npmjs.com/cli/v6/using-npm/semver#caret-ranges-123-025-004

PR #3382, opened 2024-05-19
Player hitbox fixes

AreaDenial:
Initialize width and height on the bot entity during login
Add eyeHeight property to the bot entity, which updates on crouch/uncrouch
Change raytracing and visibility functions to use bot.entity.eyeHeight instead of bot.entity.height


extremeheat:
Alot of this entity logic should really be moved to prismarine-entity I think over the messy dynamic object extension code at the moment. But the present PR seems to be ok.

extremeheat (approved):

PR #3386, opened 2024-05-25
fix gamemode incorrect reset

yanyancode:
When the packet's gamemode is undefined, the bot's gamemode resets to survival (when gamemode in server is unchanged at respawn).

### DIFF
--- a/lib/plugins/game.js
+++ b/lib/plugins/game.js
@@ -24,8 +24,10 @@ function inject (bot, options) {
 
   function handleRespawnPacketData (packet) {
     bot.game.levelType = packet.levelType ?? (packet.isFlat ? 'flat' : 'default')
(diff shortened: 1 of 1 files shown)


lib/plugins/game.js
```diff
@@ -24,8 +24,10 @@ function inject (bot, options) {
 
   function handleRespawnPacketData (packet) {
     bot.game.levelType = packet.levelType ?? (packet.isFlat ? 'flat' : 'default')
-    bot.game.hardcore = packet.isHardcore ?? Boolean(packet.gameMode & 0b100)
-    bot.game.gameMode = parseGameMode(packet.gameMode)
+    if (packet.gameMode != undefined) {
```
extremeheat, line 27: This should check either for `packet.gameMode != null` or use !== (strict equality).

rom1504:
As far as I can tell this is already done in master

PR #3391, opened 2024-06-06
feat: Add block prediction placement

zardoy:
This is what native client does. It is important if the ping is low and after placement, calculates physics correctly. If the placement is not allowed the block is still updated to air by the server.

Maybe it's worth adding it as a setting to not break existing behaviors, but I don't see a reason why it shouldn't be enabled by default .


lib/plugins/place_block.js
```diff
@@ -8,6 +8,14 @@ function inject (bot) {
     let oldBlock = bot.blockAt(dest)
     await bot._genericPlace(referenceBlock, faceVector, options)
 
+    // optimistic update
+    const heldItem = bot.heldItem
+    const mcData = require('minecraft-data')(bot.version)
+    const defaultState = mcData.blocksByName[heldItem?.name]?.defaultState
+    if (defaultState) {
+      bot.world.setBlockStateId(dest, defaultState)
+    }
```
extremeheat, line 17: 1. This should be using the registry
2. This is not a correct way to deduce correct block state from an item
3. If world state is mutated then you must ensure that the data is fully in sync with the server (how does this handle the case where the server rejects?)

extremeheat (changes requested):
> This is what native client does

What version does this apply to and do you have a source for that?

lib/plugins/place_block.js
zardoy: I believe I already answered 3rd question in the description. If the server rejects a block update would be emitted (even if it's air) as this is what the native client also does.

Can you elaborate on the 2nd question?

zardoy:
> What version does this apply to and do you have a source for that?

I don't know how to find things in the native client source I could check it by launching the client. This was always the behaviour of minecraft client.

rom1504:
Can you explain what this is supposed to do if you think this is correct? I don't understand it from the code

zardoy:
> Can you explain what this is supposed to do if you think this is correct? I don't understand it from the code

Minecraft vanilla client does optimistic updates:
- what happens on right click: https://github.com/extremeheat/extracted_minecraft_data/blob/0535398931e0827a5bf4a9c1bd19d617b56aa33d/client/net/minecraft/world/item/BlockItem.java#L62
- then https://github.com/extremeheat/extracted_minecraft_data/blob/0535398931e0827a5bf4a9c1bd19d617b56aa33d/client/net/minecraft/world/item/BlockItem.java#L143
- and then https://github.com/extremeheat/extracted_minecraft_data/blob/158aff8ad2a9051505e05703f554af8e50741d69/client/net/minecraft/client/multiplayer/ClientLevel.java#L165

otherwise, it would make position calculation incorrect on a big ping (eg when building up)

PR #3392, opened 2024-06-06
fix: Fix `heldItemChanged` was not fired in all cases when the held item was changed

zardoy:
it should not be in set_slot event, it should be updated whenever the slot is actually updated (e.g. by inventory manipulation)


rom1504:
If this is correct, please add a test to check what is not working before this PR

rom1504:
Old closing

zardoy:
No let me work on it please!

rom1504:
Nothing prevents you to work on it. You can continue in his branch and ask to reopen here or open a new branch and PR

PR #3398, opened 2024-06-15
fix `inventory.clickWindow` `mode` assertion

zardoy:
This is the right assertion:

https://github.com/PrismarineJS/prismarine-windows/blob/42122d1c70f668b850bd4dfc647dacd2f52e8252/lib/Window.js#L28


zardoy:
my bad, it throws that it's unimplemented, need to implement it first and then I re-open this

PR #3454, opened 2024-09-04
Added `switchWorld` event

SinanAkkoyun:
This PR implements the `switchWorld` event, which only fires when the world is being switched (like overworld => end or lobby -> minigame) and not upon death


lib/plugins/blocks.js
```diff
@@ -560,6 +560,7 @@ function inject (bot, { version, storageBuilder, hideErrors }) {
       dimension = packet.dimension
       worldName = packet.worldName
     }
+    if (!packet.copyMetadata) bot.emit('switchWorld')
```
rom1504, line 563: This field doesn't exist in older versions 

Please add a test

PR #3460, opened 2024-09-17
Fix window.deposit requiring item.count

Pix3lPirat3:
Patches window.deposit throwing an error if "count" arg is undefined, figured this out with WhoTho since you should be able to only specify window.deposit(item.type), as the other args can be "null", however as WhoTho pointed out--

> undefined == null
but
undefined !== null


u9g (approved):
LGTM

rom1504:
No let's not use == null, it includes too many cases 

Let's use === undefined || === null

PR #3489, opened 2024-10-27
Support 1.21.3.

rom1504:


MohmmadQunibi:
Minecraft seems to love changing its protocols alot

boly38:
@rom1504 thanks for you huge work around this amazing project 🙏 

is it possible in some way to maintain/provide in home page (ex. root readme) the latest version tested & supported
and this to avoid question/discussion about current supported version &/or error message (like `No data available for version 1.21.3` & https://github.com/PrismarineJS/mineflayer/discussions/3513) ?

I know that there is already 1.21 but it would be great to know current minor to install without having to check one by one 🙏

rom1504:
@boly38 see https://github.com/PrismarineJS/mineflayer/blob/master/lib/version.js#L1

rom1504:
current error is 

TypeError: SizeOf error for undefined : Cannot read properties of undefined (reading '_value')
    at Object.MovementFlags (eval at compile (/home/runner/work/mineflayer/mineflayer/node_modules/protodef/src/compiler.js:262:12), <anonymous>:2320:31)
    at Object.packet_position_look (eval at compile (/home/runner/work/mineflayer/mineflayer/node_modules/protodef/src/compiler.js:262:12), <anonymous>:2351:34)


just need to change how this is sent

extremeheat:
We could do a basic PR to mineflayer so basic things work like chat without blocking on https://github.com/PrismarineJS/minecraft-data/pull/948.

The things that are currently broken (like Slots in https://github.com/PrismarineJS/mineflayer/issues/3526) would remain broken as they are pending that PR. But we don't have to block on basically a refactor

extremeheat:
Pretty much everything is working now minus setCreativeSlot. No idea why this is failing, there is nothing I see unusual in code diffs. I checked out vanilla clients with flying squid and didn't see anything abnormal there either (worked fine). It's possibly something is wrong with bot environment. 

Otherwise if people want try 1.21.3 support can do

```
npm install prismarinejs/mineflayer#1.21.3
```

I'm not going to have much more time to work on this but I'll try to look into this some more when if I get the time.

lib/plugins/creative.js
```diff
@@ -21,15 +21,38 @@ function inject (bot) {
...
 
     if (Item.equal(bot.inventory.slots[slot], item, true)) return
+
+    if (bot.registry.version['>=']('1.21.3')) {
+      bot._client.write('set_creative_slot', {
+        slot,
+        item: Item.toNotch(item)
+      })
+      // No ack
+      bot._setSlot(slot, item)
```
rom1504, line 39: the update slot event is emitted sync here; so the once update slot should be created before this function call

rom1504: ah nevermind you're waiting for a network update for the slot below

rom1504:
Looks like still need an update on https://github.com/PrismarineJS/mineflayer/blob/master/test/internalTest.js#L560 and then it's done

extremeheat:
Yeah, I looked into the internal failing test and it seems to be because it's expecting a display name, so it sends a display name update but then immediately erases the display name in another player_info packet. Not sure why old versions are passing. But I think we can disable the display name check for now. Or make a new test just this that doesn't have this problem.

Either way the player_info packet is going to be refactor with https://github.com/PrismarineJS/minecraft-data/pull/948 to be same across versions so it will be revisited

extremeheat:
This is the problem:
```
2025-01-07T23:22:51.9383127Z {
2025-01-07T23:22:51.9383380Z   action: {
2025-01-07T23:22:51.9383574Z     _value: 53,
2025-01-07T23:22:51.9383789Z     add_player: true,
2025-01-07T23:22:51.9384025Z     initialize_chat: false,
2025-01-07T23:22:51.9384280Z     update_game_mode: true,
2025-01-07T23:22:51.9384534Z     update_listed: false,
2025-01-07T23:22:51.9384785Z     update_latency: true,
2025-01-07T23:22:51.9385051Z     update_display_name: true,
2025-01-07T23:22:51.9385314Z     update_priority: false
2025-01-07T23:22:51.9385544Z   },
2025-01-07T23:22:51.9385714Z   data: [
2025-01-07T23:22:51.9385907Z     {
2025-01-07T23:22:51.9386293Z       uuid: '00000000-0000-0000-0000-000000000000',
2025-01-07T23:22:51.9386876Z       player: { name: 'bot5', properties: [] },
2025-01-07T23:22:51.9387295Z       chatSession: undefined,
2025-01-07T23:22:51.9387625Z       gamemode: 0,
2025-01-07T23:22:51.9387839Z       listed: undefined,
2025-01-07T23:22:51.9388070Z       latency: 0,
2025-01-07T23:22:51.9388282Z       displayName: {
2025-01-07T23:22:51.9388504Z         type: 'compound',
2025-01-07T23:22:51.9388837Z         value: { text: { type: 'string', value: 'wvffle' } }
2025-01-07T23:22:51.9389156Z       },
2025-01-07T23:22:51.9389352Z       listPriority: undefined
2025-01-07T23:22:51.9389602Z     }
2025-01-07T23:22:51.9389777Z   ]
2025-01-07T23:22:51.9390067Z }
2025-01-07T23:22:51.9390259Z {
2025-01-07T23:22:51.9390436Z   action: {
2025-01-07T23:22:51.9390625Z     _value: 53,
2025-01-07T23:22:51.9390837Z     add_player: true,
2025-01-07T23:22:51.9391073Z     initialize_chat: false,
2025-01-07T23:22:51.9391320Z     update_game_mode: true,
2025-01-07T23:22:51.9391569Z     update_listed: false,
2025-01-07T23:22:51.9391802Z     update_latency: true,
2025-01-07T23:22:51.9392021Z     update_display_name: true,
2025-01-07T23:22:51.9392242Z     update_priority: false
2025-01-07T23:22:51.9392436Z   },
2025-01-07T23:22:51.9392579Z   data: [
2025-01-07T23:22:51.9392729Z     {
2025-01-07T23:22:51.9392933Z       uuid: '00000000-0000-0000-0000-000000000000',
2025-01-07T23:22:51.9393226Z       player: { name: 'bot5', properties: [] },
2025-01-07T23:22:51.9393487Z       chatSession: undefined,
2025-01-07T23:22:51.9393696Z       gamemode: 0,
2025-01-07T23:22:51.9393878Z       listed: undefined,
2025-01-07T23:22:51.9394070Z       latency: 0,
2025-01-07T23:22:51.9394253Z       displayName: undefined,
2025-01-07T23:22:51.9394472Z       listPriority: undefined
2025-01-07T23:22:51.9394675Z     }
2025-01-07T23:22:51.9394822Z   ]
2025-01-07T23:22:51.9394965Z }
```

`update_display_name: true` is set to true twice because value 53 is being written in the internal tests displayName

lib/plugins/entities.js
```diff
@@ -583,8 +598,65 @@ function inject (bot) {
   })
 
   bot._client.on('player_info', (packet) => {
+    console.dir(packet, { depth: null })
```
rom1504, line 601: Remove

lib/plugins/physics.js
```diff
@@ -45,10 +45,14 @@ function inject (bot, { physicsEnabled, maxCatchupTicks }) {
...
     z: 0,
+    dx: 0, // 1.21.3
+    dy: 0,
+    dz: 0,
     yaw: 0,
     pitch: 0,
     onGround: false,
-    time: 0
+    time: 0,
+    flags: bot.registry.version['>=']('1.21.3') ? {} : 0
```
rom1504, line 55: Should probably be a feature

lib/plugins/physics.js
```diff
@@ -330,30 +345,59 @@ function inject (bot, { physicsEnabled, maxCatchupTicks }) {
...
-
-    const newYaw = (packet.flags & 8 ? conv.toNotchianYaw(bot.entity.yaw) : 0) + packet.yaw
-    const newPitch = (packet.flags & 16 ? conv.toNotchianPitch(bot.entity.pitch) : 0) + packet.pitch
+
+    // TODO: this current mineflayer logic maybe incorrect. New (maybe even older) versions of minecraft have flag values for
+    // dx/dy/dz but mineflayer is assuming that 0b111 applies for both velocity and position.
+
+    let newYaw, newPitch
+
+    if (bot.registry.version['>=']('1.21.3')) {
```
rom1504, line 370: Should be a feature

rom1504:
Left a few minor comments but otherwise LGTM

rom1504:
I hope 1.21.4 is more minor than this

lib/plugins/physics.js
extremeheat: The below features can be removed after https://github.com/PrismarineJS/minecraft-data/pull/948 so if there is a feature added here it would just be temporary

lib/plugins/physics.js
extremeheat: I reverted this, I messed up the client and serverbound position packets (serverbound doesn't have dx/dy/dz and changed onGround bool to a packed flag int so we dont need a version check)

extremeheat:
Mojang said earlier they plan to do breaking changes in minor versions now so I think they will probably not worry much about big changes anymore. Probably won't get much better until LLMs can take the (java diff, current protocol spec) and apply the updates directly to the protocol spec and lib code. I looked at 1.21.4 earlier and there are some big changes to the internals for refactoring but network changes are not that bad. But there might still be behavior updates (like the one related to creative set slot) that could be annoying.

But I think since this is the new normal i think it's probably a good time to simplify as much as possible/add abstractions where needed so the code doesn't get too messy with lots of version/feature based branching

PR #3508, opened 2024-11-21
Fix gamemode test and implementation

IceTank:
This fixes the broken gamemode test and also fixes the gamemode implementation for pre 1.10


IceTank:
Tests are not working. Will fix.

lib/plugins/game.js
```diff
@@ -25,10 +25,11 @@ function inject (bot, options) {
   function handleRespawnPacketData (packet) {
     bot.game.levelType = packet.levelType ?? (packet.isFlat ? 'flat' : 'default')
     bot.game.hardcore = packet.isHardcore ?? Boolean(packet.gameMode & 0b100)
-    if (bot.registry.isOlderThan('1.10')) { // gamemode is used pre 1.10 and post 1.20 but in between it's gameMode
```
extremeheat: This should be a feature

IceTank: Gamemode in 1.20.6 is now of type "mapper" https://prismarinejs.github.io/minecraft-data/?d=protocol&v=1.20.6#toClient_undefined. What is that? @extremeheat

IceTank: Is that like a string then?

extremeheat: Mapper is a enum. The enum values can be found in the protocol files inside minecraft-data for the relevant version. 

The current Mcpc mcdata protocol doc gen doesn't support the mapper type. Can be fixed in future by swapping it with the protodef-yaml html gen that bedrock is currently using.

IceTank:
Related https://github.com/PrismarineJS/mineflayer/pull/3508

IceTank:
Do we wait for mc data to update, or can we merge with just the version check and add the feature later?

extremeheat:
What change to the protocol was made for this with 1.21? I don't see any changes. The last change was with 1.20.5 with the new world state field. It doesn't seem mineflayer is handling that in the linked PR.

IceTank:
> What change to the protocol was made for this with 1.21?

The login packet in 1.21 now has gamemode as string instead of a number. This means the mineflayer needs to use `bot.game.gameMode = packet.gamemode` for 1.21 and `bot.game.gameMode = parseGameMode(packet.gamemode)` for all lower versions.

IceTank:
Looks like the mc protocol did not change. But minecraft-protocol now maps the game mode from a number to a string by itself. Wiki.vg still has the gamemode value as a number https://wiki.vg/Protocol#Login_.28play.29. But there is a clear change in minecraft-protocol. `packet.gamemode` gives a number pre 1.21 and a string post or at 1.21. Without a version check, mineflayer won't be able to handle gamemode coming from minecraft-protocol.

extremeheat:
Yes, as you can see in the mcdata link the data is encapsulated inside the world state (SpawnInfo) field. This is not handled in mineflayer and there is an existing feature for that change, `spawnRespawnWorldDataField`.

IceTank:
The `packet_login` now has the `worldState` field then I guess? https://github.com/PrismarineJS/minecraft-data/blob/9cfa6ea955af24d6f481484ebe5d8221a2265421/data/pc/1.21.1/protocol.json#L4993

How does that match up with Wiki.vg login (play) packet? Or is login (play) at wiki.vg not the same as packet_login? https://wiki.vg/Protocol#Login_.28play.29

extremeheat:
1. Wiki.vg is not reliable on its own, you need to look at the vanilla source code to be correct. It's manually updated and sometimes is missing data from being out of date or oversight (it's not validated against anything). But can be helpful for general reference. All the structures have to be perfectly correct or there will be a crash.
2. Some packets encode shared data structures. To avoid duplicating packet fields, minecraft-data may similarly use shared data types between for packet fields. This has many benefits including that should there be a change to the shared data type, it's less likely to see desync (eg, one packet is updated and another is not). 
3. Inside the protocol.yml files for the versions, atop of all the packet definitions is an "MC:" header that contains the official class name that you can use to lookup the packet def in the official mc code.

zardoy:
It might be a little off topic but shouldn’t parseGamemode function also fixed? packet with game mode -1 is valid but Minecraft client would go to survival mode if the number is out of range (not to spectator or any other)

IceTank:
Ok, I updated the code to use the spawnRespawnWorldDataField feature instead of the version check. I also implemented @zardoy suggestion with gamemode out of bounds checking.

zardoy:
> Ok, I updated the code to use the spawnRespawnWorldDataField feature instead of the version check. I also implemented @zardoy suggestion with gamemode out of bounds checking.

thank you very much! A lot of people would benefit from your improvements

IceTank:
Failing test is unrelated to the changes.

lib/plugins/game.js
```diff
@@ -25,7 +30,12 @@ function inject (bot, options) {
   function handleRespawnPacketData (packet) {
     bot.game.levelType = packet.levelType ?? (packet.isFlat ? 'flat' : 'default')
     bot.game.hardcore = packet.isHardcore ?? Boolean(packet.gameMode & 0b100)
-    bot.game.gameMode = packet.gamemode || parseGameMode(packet.gameMode)
+    // Either a respawn packet or a login packet. Depending on the packet it can be "gamemode" or "gameMode"
+    if (bot.supportFeature('spawnRespawnWorldDataField')) { // 1.20.5
+      bot.game.gameMode = packet.gamemode
+    } else {
+      bot.game.gameMode = parseGameMode(packet.gamemode ?? packet.gameMode)
+    }
```
extremeheat, line 38: I don't understand how this is different from the current code. It does the same thing.

IceTank: Pre 1.20.5, the respawn packet and the login packet used numbers to indicate what gamemode the player is in. The problem is that with one packet, it's `gamemode`, and with the other, it's `gameMode`.

IceTank: `packet.gamemode` could be a number. But game.gameMode needs to be a string. Also,`||` is wrong as it will be falsy if the gamemode is 0. And 0 is a valid gamemode in some versions where the packet has gamemode as a number.

IceTank: This line was broken before `bot.game.gameMode = packet.gamemode || parseGameMode(packet.gameMode)`. It only works correctly in versions after 1.20.5. In versions pre 1.20.5, it would set game.gameMode to a number for gamemodes other than survival. No one noticed that because most people use survival mode and the test was not working, and the change was only made recently.

extremeheat: `gamemode` lower case is only defined on 1.20.5+ because this function is called with a SpawnInfo object. `gameMode` is the previous number field. `||` can be replaced with `??`, but otherwise for the else condition it should be this:

```js
      bot.game.gameMode = parseGameMode(packet.gameMode)
```

`parseGameMode` can likely just be set to `gameModes[gameModeBits] || 'survival']`

I think it may be worthwhile to just backport this change inside minecraft-data to 1.20.5 to all the old versions so we can get rid of the logic here

IceTank: > `gamemode` lower case is only defined on 1.20.5+ because this function is called with a SpawnInfo object.

No, the two different cases have been used before. Handle respawn is used by the respawn packet and the login packet. They both use different cases for the same field pre 1.20.5.

See https://prismarinejs.github.io/minecraft-data/?v=1.12.2&d=protocol#toClient_respawn

And

https://prismarinejs.github.io/minecraft-data/?v=1.12.2&d=protocol#toClient_login

The same protocol uses different versions of the field gamemode.

IceTank: I think it's better to focus on the tests instead of the implementation. The protocol is inconsistent and changes a lot across versions. Trying to figure out what the implementation should look like by looking really hard at the code always fails in mineflayer. That's why there are tests. They tell you if something works or not. The current implementation does not work for pre 1.20.5. Your changes to game.js to update past 1.20.5 only when through review because the gamemode test was not working and did not catch the bug with the different cased gamemode fields.

extremeheat: Yes what I mean is there are things we can fix in minecraft-data to avoid the need to do special handling in mineflayer. 

Can you reopen the PR against the 1.21.2 branch?

IceTank:
@extremeheat  What's the progress on this?

PR #3542, opened 2025-01-06
Patches isInWater for digSpeed

Pix3lPirat3:
If the bot is standing in water it should not be slowed, however passing isInWater is slowing it, while digSpeed should only be slowed if the block at eyeHeight is water

### DIFF
--- a/lib/plugins/digging.js
+++ b/lib/plugins/digging.js
@@ -242,7 +242,7 @@ function inject (bot) {
     return block.digTime(
       type,
       creative,
-      bot.entity.isInWater,
+      bot.blockAt(bot.entity.position.offset(0, bot.entity.eyeHeight, 0)).name === 'water',
       !bot.entity.onGround,
       enchantments,
       bot.entity.effects


lib/plugins/digging.js
```diff
@@ -242,7 +242,7 @@ function inject (bot) {
     return block.digTime(
       type,
       creative,
-      bot.entity.isInWater,
+      bot.blockAt(bot.entity.position.offset(0, bot.entity.eyeHeight, 0)).name === 'water',
```
extremeheat, line 245: I think this would be better if you put it into a function for clarity

Eg
```js
bot._getBlockAtEyeLevel = () => bot.entity.position && bot.blockAt(bot.entity.position.offset(0, bot.entity.eyeHeight, 0))
```

lib/plugins/digging.js
```diff
@@ -242,13 +242,16 @@ function inject (bot) {
...
+      _getBlockAtEyeLevel()?.name === 'water',
       !bot.entity.onGround,
       enchantments,
       bot.entity.effects
     )
   }
 
+  _getBlockAtEyeLevel = () => bot.entity.position && bot.blockAt(bot.entity.position.offset(0, bot.entity.eyeHeight, 0))
+
+  bot._getBlockAtEyeLevel = _getBlockAtEyeLevel;
```
extremeheat, line 254: This can be inlined to

```js
bot._getBlockAtEyeLevel = () => bot.entity.position && bot.blockAt(bot.entity.position.offset(0, bot.entity.eyeHeight, 0))
```

if it's not being called many times

extremeheat: What `  _getBlockAtEyeLevel = () => bot.entity.position && bot.blockAt(bot.entity.position.offset(0, bot.entity.eyeHeight, 0))` does is create a global variable as you are not binding it with `let` or `const`. That is causing lint to fail

Pix3lPirat3:
> Where is in water defined and used ? Should we change its meaning or add another similar attribute?

https://github.com/PrismarineJS/mineflayer/blob/c7ce4ef650a165444fd2ddeb8535262830911ee6/lib/plugins/digging.js#L245

isDrowning would *technically* fit the description but would be wrong at the same time.. isHeadUnderwater boolean would be the direct translation, the existing mineflayer code needs to be changed from "checking the block at the bot's feet" to "checking the block at the bot's head", as that's the only time the action is slowed.

Nekiplay:
also in lava?

Pix3lPirat3:
> also in lava?

I tried this in vanilla Minecraft, your digSpeed does not get affected if you're *standing underneath* lava, only if you're "swimming" (feet not on ground) which is why only water is being specified

rom1504:
Added tests and merging

PR #3548, opened 2025-01-09
Implement skinUsesKeyedProperties for 1.19.3+

Pix3lPirat3:
player_info packet uses the key field for skin properties instead of name, Applies to Minecraft 1.19.3 and above.
awaiting response to https://github.com/PrismarineJS/minecraft-data/pull/971

Also added a fix on servers that send extra properties which results in a JSON crash


rom1504:
Can you add a test ?

rom1504:
Stuck on https://github.com/PrismarineJS/minecraft-data/pull/971

Pix3lPirat3:
obsolete and fixed

PR #3561, opened 2025-01-16
Fix silent failure when digging unbreakable blocks

BBpezsgo:
The `bot.digTime` function returns `Infinity` if the bot cannot break the block (such as "bedrock" or "barrier"), but if you pass `Infinity` to `setTimeout` it will produce the following warning:
```
TimeoutOverflowWarning: Infinity does not fit into a 32-bit signed integer.
Timeout duration was set to 1.
```
This will result in a silent failure and the `bot.dig` function will resolve.

With this change, it will throw an error in such cases.


BBpezsgo:
This will still fail if the dig time is more than 2147483647 milliseconds, but Infinite is more readable than that.

PR #3570, opened 2025-01-26
Use node 22

rom1504:
Node 22 is now LTS, switch to that in CI and supported node version.


rom1504:
https://nodejs.org/en/about/previous-releases

PR #3584, opened 2025-02-14
Fix vehicle passengers array in entities being null

Dueel:
Fixes `TypeError: Cannot read properties of undefined (reading 'passengers')` by adding passengers array in `setEntityData`


extremeheat (approved):

extremeheat (changes requested):
This will not fix the error you state in the PR. That does not state that passengers is missing, but that `passengers` is being read from an null/undefined object. Also, this will erase the `passengers` field. That doesn't seem intentional.

Issue #3576, opened 2025-01-31
TypeError: Cannot read properties of undefined (reading 'passengers')


TheRealPerson98:
## Versions
 - mineflayer: Latest
 - server: vanilla/spigot/paper Latest
 - node: 22.13.1

When joining some servers that have books / or nothing sometimes on join u get

lib\plugins\entities.js:841
```
TypeError: Cannot read properties of undefined (reading 'passengers')
    at Client.<anonymous> (C:\ops/mineflayer\lib\plugins\entities.js:841:17)
    at Client.emit (node:events:517:28)
    at emitPacket (C:\opsminecraft-protocol\src\client.js:84:12)
    at Array.forEach (<anonymous>)
    at FullPacketParser.<anonymous> (C:REDACTEDnode_modules\minecraft-protocol\src\client.js:99:26)
    at FullPacketParser.emit (node:events:517:28)
    at addChunk (C:REDACTEDnode_modules\readable-stream\lib\internal\streams\readable.js:323:12)
    at readableAddChunk (C:REDACTEDnode_modules\readable-stream\lib\internal\streams\readable.js:300:9)
    at Readable.push (C:REDACTEDnode_modules\readable-stream\lib\internal\streams\readable.js:246:10)
    at FullPacketParser._transform (C:REDACTEDnode_modules\protodef\src\serializer.js:89:10)
```

also a null on getting the servers dim but i don't have that error

closed my pull i don't have the time to dig deeper into the issue

bowoflare:
I am running into this same issue, do you know any  fixes?

TheRealPerson98:
I just forked it and added null checks to it bc nothing im doing is using passengers

notegorkaa:
which minecraft version?

mrab54:
I'm seeing this on 1.21 java

hilbix:
Same for MC 1.20.4, Mineflayer 4.26.0, Node.js v18.19.0.  Downgrade to 4.25 works:

`package.json`:
```
{
  "dependencies":
    { "mineflayer": "=4.25.0"
    }
}
```

Following code:

```
#!/usr/bin/env nodejs
'use strict';

const mineflayer = require('mineflayer');
const B = mineflayer.createBot({ host:'127.0.0.1', username:'t', hideErrors:false });
const D = console.error;
B.once('spawn', () => D('spawn', B.version));
B.once('end',   () => process.exit(D('END')));
```

[...]

PR #3585, opened 2025-02-14
Patch vehicles & passengers (Tested: 1.8.9, 1.20.4)

Pix3lPirat3:
So I found an issue where passengers were infinitely being added (unless I did that myself), an issue with passenger becoming undefined, and an issue with set_passengers running into the same issues. been working on this the last two nights and got it to a point of *working* (from my own testing) in 1.8.9 and 1.20.4, requires further testing. (Don't believe there are any built-in testing)

My test code was;
```JS
  setInterval(function() {
    let boat = bot.nearestEntity(ent => ent.name == 'Boat');
    if(!boat) return bot.chat('No boat');
    let player = bot.nearestEntity(ent => ent.type == 'player');
    if(!player) return bot.chat('No player.');
    let spider = bot.nearestEntity(ent => ent.name == 'Spider');
    if(!spider) return bot.chat('No spider.');

    bot.chat('Boat Passengers: ' + boat.passengers.length + ' Player Vehicle: ' + player.vehicle?.name + ' Spider Passengers: ' + spider.passengers?.length)
  }, 1000)
```

I don't believe the mount/unmount events work currently, was curious if that should still remain as `bot.emit('mount/unmount')` or added into the `set_passenger/attach_entity` events and emitted directly from the passenger/vehicle entities. However that's a separate thing.

### DIFF
--- a/docs/history.md
+++ b/docs/history.md
@@ -1,3 +1,9 @@
+## 4.26.0
+* [Use node 22 (#3570)](https://github.com/PrismarineJS/mineflayer/commit/dbff9314418d30df203c32fe83f2a1b56653d0a6) (thanks @rom1504)
+* [Fix infinity setTimeout by throwing error (#3561)](https://github.com/PrismarineJS/mineflayer/commit/69539494c88c2ca718330142839a35414ecd3bda) (thanks @BBpezsgo)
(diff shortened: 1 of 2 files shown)


Pix3lPirat3:
nope broken for spiders on set_passengers (so other mobs aswell) bot player seems to be working

DatArnoGuy:
Does this fix https://github.com/PrismarineJS/mineflayer/issues/3576 ?

rom1504:
Please add tests

Pix3lPirat3:
> Does this fix #3576 ?

Yes, and mount/unmount needs to be patched aswell but figured that'd be a separate PR



> Please add tests

Definitely need to, was running into quite a bit of issues with 1.20.4 (set_passengers)

PR #3586, opened 2025-02-14
fix vehicle/passengers

Pix3lPirat3:
Re-opening because I accidentally closed my last one, still need to figure out how to create a test for this.

So I found an issue where passengers were infinitely being added (unless I did that myself), an issue with passenger becoming undefined, and an issue with set_passengers running into the same issues. been working on this the last two nights and got it to a point of *working* (from my own testing) in 1.8.9 and 1.20.4, requires further testing. (Don't believe there are any built-in testing)

My test code was;
```JS
  setInterval(function() {
    let boat = bot.nearestEntity(ent => ent.name == 'Boat');
    if(!boat) return bot.chat('No boat');
    let player = bot.nearestEntity(ent => ent.type == 'player');
    if(!player) return bot.chat('No player.');
    let spider = bot.nearestEntity(ent => ent.name == 'Spider');
    if(!spider) return bot.chat('No spider.');

    bot.chat('Boat Passengers: ' + boat.passengers.length + ' Player Vehicle: ' + player.vehicle?.name + ' Spider Passengers: ' + spider.passengers?.length)
  }, 1000)
```

I don't believe the mount/unmount events work currently, was curious if that should still remain as `bot.emit('mount/unmount')` or added into the `set_passenger/attach_entity` events and emitted directly from the passenger/vehicle entities. However that's a separate thing.

### DIFF
--- a/lib/plugins/entities.js
+++ b/lib/plugins/entities.js
@@ -380,12 +380,9 @@ function inject (bot) {
     // attach entity
     const entity = fetchEntity(packet.entityId)
     if (packet.vehicleId === -1) {
-      const vehicle = entity.vehicle
-      delete entity.vehicle
-      bot.emit('entityDetach', entity, vehicle)
+      bot.emit('entityDetach', entity, entity.vehicle)
     } else {
-      entity.vehicle = fetchEntity(packet.vehicleId)
-      bot.emit('entityAttach', entity, entity.vehicle)
+      bot.emit('entityAttach', entity, fetchEntity(packet.vehicleId))
     }
   })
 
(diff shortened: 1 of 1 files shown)


lib/plugins/entities.js
```diff
@@ -380,12 +380,9 @@ function inject (bot) {
     // attach entity
     const entity = fetchEntity(packet.entityId)
     if (packet.vehicleId === -1) {
-      const vehicle = entity.vehicle
-      delete entity.vehicle
-      bot.emit('entityDetach', entity, vehicle)
+      bot.emit('entityDetach', entity, entity.vehicle)
```
rom1504, line 383: Well this is not updating the state to detach anymore

lib/plugins/entities.js
```diff
@@ -380,12 +380,9 @@ function inject (bot) {
...
     } else {
-      entity.vehicle = fetchEntity(packet.vehicleId)
-      bot.emit('entityAttach', entity, entity.vehicle)
+      bot.emit('entityAttach', entity, fetchEntity(packet.vehicleId))
```
rom1504, line 385: This is not updating the state to attach anymore

rom1504:
Fixed in #3601

Issue #3537, opened 2024-12-30
Casting rod always casts directly south


Zetaphor:
- [X ] The [FAQ](https://github.com/PrismarineJS/mineflayer/blob/master/docs/FAQ.md) doesn't contain a resolution to my issue 

## Versions
 - mineflayer: #.#.#
 - server: vanilla 1.21.1
 - node: 22.12.1

## Detailed description of a problem
Whenever I use the `bot.fish` command, the rod is always cast directly south, regardless of what direction the bot is looking (head) or facing (body).

## What did you try yet?

I've tried using both a look and a lookAt before using either the `bot.fish` or the `bot.activate` command.

It appears as though [the direction is being hardcoded into the activate command](https://github.com/PrismarineJS/mineflayer/blob/f6187f66c16dd122165287be7864c78b2fe7c32c/lib/plugins/inventory.js#L121)

## Your current code
```js
const nearestWaterBlock = this.bot.helpers.world.findNearestWaterBlock(this.bot)
if (nearestWaterBlock) {
  console.log('Nearest water block:', nearestWaterBlock)
  const dx = nearestWaterBlock.x - currentPos.x
  const dy = nearestWaterBlock.y - currentPos.y
  const dz = nearestWaterBlock.z - currentPos.z
  const yaw = Math.atan2(-dx, dz)
  const groundDistance = Math.sqrt(dx * dx + dz * dz)
  const pitch = Math.atan2(dy, groundDistance)
  await this.bot.look(yaw, pitch, true)
  await new Promise(resolve => setTimeout(resolve, 1000))
} else {
  this.bot.chat('No water block found nearby.')
  return
}

this.bot.fish
```

In the above code the bot will look at the water block as expected, then after the 1 second delay it will cast the rod and change its look direction to the south.

## Expected behavior
I expect the fish/activateItem function to not override the look direction

Zetaphor:
I've identified the issue, it's not the direction but the hardcoded rotation

Zetaphor:
I've found that passing `false` to the rotation value in the `client.write` resolves this issue. I'm not sure if that means this value can be omitted or if the boolean is valid

PR #3588, opened 2025-02-17
Update inventory.js

gamersindo1223:
Fixes
* https://github.com/PrismarineJS/mineflayer/issues/3537
* Makes it possible to specify where to activate item


gamersindo1223:
If there's anything that's need to be changed, please let me know

DatArnoGuy:
Should we just get the existing rotation and use that instead of setting new rotation?

Issue #3600, opened 2025-03-08
entity.passengers not correct


llmlover:
- [x] The [FAQ](https://github.com/PrismarineJS/mineflayer/blob/master/docs/FAQ.md) doesn't contain a resolution to my issue 

## Versions
 - mineflayer: ^4.26.0
 - server: vanilla/spigot/paper 1.21.1 - vanilla
 - node: v22.13.1

## Detailed description of a problem
Display entity passengers are not being parsed/read

## Your current code
```ts
bot.once('spawn', async () => {
const blockDisplays = Object.values(bot.entities).filter(e => e.name === 'block_display')
  const getGun = () => blockDisplays.find(e => e.getCustomName()?.toString() === 'GUN') 

  let gun = getGun()
  if(!gun) {
    bot.chat('Summoning gun')
[...]

PR #3601, opened 2025-03-08
Entity Passengers get pushed wrongly

SinanAkkoyun:
```js
originalVehicle.passengers = originalVehicle.passengers.splice(index, 1)
```
Hey! If I understood this correctly: Reassignment is unnecessary, splice already modifies the array itself, which resulted in https://github.com/PrismarineJS/mineflayer/issues/3600

### DIFF
--- a/lib/plugins/entities.js
+++ b/lib/plugins/entities.js
@@ -809,7 +809,7 @@ function inject (bot) {
     const originalVehicle = passenger.vehicle
     if (originalVehicle !== null) {
       const index = originalVehicle.passengers.indexOf(passenger)
-      originalVehicle.passengers = originalVehicle.passengers.splice(index, 1)
+      originalVehicle.passengers.splice(index, 1)
     }
     passenger.vehicle = vehicle
     vehicle.passengers.push(passenger)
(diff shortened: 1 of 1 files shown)


extremeheat:
It's an unnecessary operation, but unless the old reference is being incorrectly used somewhere over access via `bot`, it's not a bug that should break anything.

rom1504:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice#return_value looks like splice returns an array of removed elements so indeed this is a behavior change

rom1504:
Can you please fix the failing tests ?

SinanAkkoyun:
> Can you please fix the failing tests ?

I fixed the tests now, I believe my PR was not the issue
The original code `if(vehicle !== null)` did not test for vehicle being `undefined`; I replaced it with `if(vehicle)`

I don't understand the new test fail for 1.13.2:
```
54 passing (4m)
  1 failing

  1) mineflayer_external 1.13.2v
       useChests:

      AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:

27 !== 305

      + expected - actual

      -27
      +305
      
      at /home/runner/work/mineflayer/mineflayer/test/externalTests/useChests.js:86:10
      at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
```

I think this isn't due to my PR, however please let me know if that's the case and I will try to investigate

PR #3610, opened 2025-03-17
Added `switchWorld` event

SinanAkkoyun:
This PR implements the `switchWorld` event, which only fires when the world is being switched (like overworld => end or lobby -> minigame) and not upon death

(original PR: [3454](https://github.com/PrismarineJS/mineflayer/pull/3454))

image

Hi, I'm sorry I don't quite understand what you mean, but I now replaced it with 
```js
if (packet.copyMetadata === false)
```
as found a couple of lines above in the original codebase

Please let me know if you meant something else


SinanAkkoyun:
I can't get a reliable general solution for this. If anyone knows world switch detection beyond checking copyMetadata of respawn packets please let me know

PR #3611, opened 2025-03-18
fix incorrect rotation during useItem()

hilaolu:
For 1.21, #3480 introduces a new field `rotation` for `useItem`. However, the field may not be correct. According to [Minecraft Protocol](https://minecraft.wiki/w/Java_Edition_protocol#Use_Item), `rotation` should have fields `yaw` and `pitch`, but not `x` and `y`. In practice, lacking the two field will lead to bot facing default direction when `useItem` is called.

### DIFF
--- a/lib/plugins/inventory.js
+++ b/lib/plugins/inventory.js
@@ -128,7 +128,7 @@ function inject (bot, { hideErrors }) {
       bot._client.write('use_item', {
         hand: offHand ? 1 : 0,
         sequence,
-        rotation: { x: 0, y: 0 }
+        rotation: { yaw: 0, pitch: 0 }
       })
     }
   }


extremeheat:
The source for field names is http://github.com/prismarinejs/minecraft-data as that is what the serializer is using when encoding and decoding packets. 

You should check against there and not the wiki. If the serialization was wrong, it would lead to a crashing error here or in Minecraft so structure is most likely correct (you can verify against source code)

extremeheat:
Looks correct, rotation is a vec2f

https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/latest/proto.yml#L3357C17-L3357C22

PR #3619, opened 2025-03-23
Added if (vehicle) check

SinanAkkoyun:
In line 816 of entities.js, a null/undefined check for vehicle was missing.

Noticed it while dismounting from a horse on an `mcproxy` 1.8.8 instance


SinanAkkoyun:
Related to https://github.com/PrismarineJS/mineflayer/pull/3601

rom1504:
Please add a non regression test

rom1504:
Code looks correct but without test this will definitely break again

PR #3624, opened 2025-03-29
fix: fix entities leak on world switch, add worldSwitch event

zardoy:
Emit removing the entity before actually removing it from the list (when clearing the object)

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -165,6 +165,7 @@
       - ["messagestr" (message, messagePosition, jsonMsg, sender, verified)](#messagestr-message-messageposition-jsonmsg-sender-verified)
       - ["inject_allowed"](#inject_allowed)
       - ["login"](#login)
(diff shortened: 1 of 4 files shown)


rom1504:
Please add a test, see pending PR in switch world

zardoy:
> see pending PR in switch world

That pr seems reliable enough for me, do you have any suggestion? Should I remove it here?

rom1504:
Either finish the initial PR so the feature work including passing tests or do not use this feature yes
It's just about having a working implementation or not

zardoy:
> It's just about having a working implementation or not

Ok, not a problem. I'll remove the event emitting then (since I'm not aware of the issues of that pr and I don't really need it)

zardoy:
I have no idea how to finish it, should I close it?

rom1504:
Yes

PR #3625, opened 2025-03-29
fix firing heldItemChanged, add type

zardoy:
Make sure to *always* fire `heldItemChanged` when its actually changed from windows API or any other external call that messes with the prismarine windows


rom1504:
Current tests fail and no new one is introduced

zardoy:
> Current tests fail and no new one is introduced

Ok yes it totally makes sense to add a test here, will do as I understand the change itself is fine

zardoy:
Also seems tests fails are unrelated…

PR #3631, opened 2025-04-02
Add Entity Attacker Detection into EntityHurt emitter

1tsMeMario:
Capture

### DIFF
--- a/lib/plugins/entities.js
+++ b/lib/plugins/entities.js
@@ -374,7 +374,8 @@ function inject (bot) {
 
   bot._client.on('damage_event', (packet) => { // 1.20+
     const entity = bot.entities[packet.entityId]
-    bot.emit('entityHurt', entity)
+    const source = bot.entities[packet.sourceCauseId - 1] // damage_event : SourceCauseId : The ID + 1 of the entity responsible for the damage, if present. If not present, the value is 0
+    bot.emit('entityHurt', entity, source)
   })
 
   bot._client.on('attach_entity', (packet) => {
--- a/index.d.ts
+++ b/index.d.ts
@@ -81,7 +81,7 @@ export interface BotEvents {
   health: () => Promise<void> | void
   breath: () => Promise<void> | void
   entitySwingArm: (entity: Entity) => Promise<void> | void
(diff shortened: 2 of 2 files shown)


1tsMeMario:
side note. with the pvp and pathfinding its still buggy when fighting against enemies with projectiles but this isnt anything this can fix. this just gets the exact entity which damaged the bot or any other entity

1tsMeMario:
recommended detection code:
`bot.on('entityHurt', (entity, source) => {
    if (entity === bot.entity && (source.type === 'mob' || source.type === 'player')) // ignore projectiles because current combat system for ranged sucks
    {
            console.log(`${master}`, `${bot.username} was attacked by ${source.displayName}. Enabling combat system.`);
    } else {
            console.log(`${master}`, `${bot.username} was hurt by an unknown cause`);
    }
});`

you can do distance checks or whatever too

lib/plugins/entities.js
```diff
@@ -374,7 +374,8 @@ function inject (bot) {
 
   bot._client.on('damage_event', (packet) => { // 1.20+
     const entity = bot.entities[packet.entityId]
-    bot.emit('entityHurt', entity)
+    const source = bot.entities[packet.sourceCauseId - 1] // damage_event : SourceCauseId : The ID + 1 of the entity responsible for the damage, if present. If not present, the value is 0
+    bot.emit('entityHurt', entity, source)
```
rom1504, line 378: source will be empty for all old versions

1tsMeMario:
also another thing i realized. the documentation says its + 1. but reading the packets its - 1

lib/plugins/entities.js
1tsMeMario: well of course. as the comment said for that event. 1.20+

1tsMeMario: bot._client.on('damage_event', (packet) => { // 1.20+ <--- right here. this was already there before i added my code.

rom1504:
https://github.com/PrismarineJS/mineflayer/blob/1e982e16f9f345725e8f39c9c641a0289f5fbdb4/lib/plugins/entities.js#L372 looks like it was working for other versions 

Would be good to have a test to make sure

PR #3665, opened 2025-05-28
activateItem: Added argument for degree, default fallback 

De1p0:
Allows users to specify the degrees to which the bot should turn, defaulting to the bot’s current rotation if no values are provided.

Problem: 
By default, activateItem aims at coordinates (0, 0) if no degrees are specified.
Referenced here: https://github.com/PrismarineJS/mineflayer/blob/f6187f66c16dd122165287be7864c78b2fe7c32c/lib/plugins/inventory.js#L131

### DIFF
--- a/lib/plugins/inventory.js
+++ b/lib/plugins/inventory.js
@@ -21,7 +21,7 @@ const ALWAYS_CONSUMABLES = [
   'golden_apple'
 ]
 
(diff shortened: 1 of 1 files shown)


De1p0:
> Please adapt the doc and ideally the tests and examples 

I added docs

Issue #3667, opened 2025-05-31
Bot's not able to place item frames


Plumbiller:
## Is your feature request related to a problem? Please describe.
It's Unimplemented.

## Describe the solution you'd like
The expected behaviour to be implemented.

## Describe alternatives you've considered
bot.placeBlock(referenceBlock, faceVector)

## Additional context
```
try {
              await bot.placeEntity(referenceBlock, { x: 0, y: 0, z: 0 });
              console.log('Placed item_frame on block at', referenceBlock.position);
            } catch (err) {
              console.log('Failed to place item_frame:', err.message);
            }
```
When trying to spawn an item frame, the intuitive function I used is bot.placeEntity() but the error message I get when doing so is: 

> Failed to place item_frame: Unimplemented. 

So it's kinda sad item frames are not imlpemented in mineflayer.

PR #3673, opened 2025-06-11
Add support for placing item frames

jlucaso1:
fix Bot's not able to place item frames #3667


jlucaso1:
only failing on 1.21.3 🤔

PR #3740, opened 2025-09-09
Fixed buckets

MaxRobinsonTheGreat:
Fixed buckets by fixing `activateItem`.

Previously, the `rotation` parameters were hard-coded and forced the wrong rotation, so I convert current bot.entity.rotation to the expected x, y. 

Added an example script `bucketer.js` to test and to document bucket usage, which isn't obvious. `placeBlock` and `activateBlock` don't work with buckets.

### DIFF
--- a/examples/bucketer.js
+++ b/examples/bucketer.js
@@ -1,7 +1,7 @@
 const mineflayer = require('mineflayer')
 
 // put the bot on grass, give it a water bucket, and say "water_bucket grass_block"
-// this just uses any item type on any target block, but is not bot.useOnBlock
+// this just uses any item type on any target block, but is not bot.activateBlock
 
 const bot = mineflayer.createBot({
   host: 'localhost',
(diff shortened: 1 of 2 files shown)


rom1504:
The bucket test CI failure is fixed in branch `fix-bucket-test-3740`. The fix is also available at `rom1504/mineflayer:fix-bucket-test`.

**Root cause:** The test flew 4 blocks up into air and tried to use `activateItem()` with a water bucket while looking down at air. Water bucket placement requires the server's ray trace to hit a solid block face — with no solid block in range, the water was never placed.

**Fix:**
- Place a stone block at a known absolute position (`(2, groundY, 2)`)
- Fly near it and look at the top face of the stone block
- Use `blockUpdate` event (via `onceWithCleanup` with 5s timeout) instead of `setTimeout(1000)` for reliable detection
- Use `bot.test.groundY` for version-independent Y coordinates (supports both regular and tall worlds)

@MaxRobinsonTheGreat please cherry-pick commit `fix-bucket-test-3740` into your branch, or I can open a PR against your fork.

PR #3746, opened 2025-09-22
make physics plugin robust when blocks plugin disabled

Callanplays:
When loadInternalPlugins=false and the blocks plugin is not loaded, physics crashes when calling bot.blockAt. This patch guards those calls, allowing the plugin to continue operating when physicsEnabled=false (sending necessary position packets to prevent desyncs), without altering behavior when the world plugin is present.


Callanplays:
wtf 😠

PR #3767, opened 2025-10-21
Limit rendering chunks

obseletecode72:
Some servers even when you send the viewDistance setting to it, they can send the default chunk amount that has been set, so you must check how much chunks you are receiving before saving them in to memory

### DIFF
--- a/lib/plugins/blocks.js
+++ b/lib/plugins/blocks.js
@@ -2,7 +2,6 @@ const { Vec3 } = require('vec3')
 const assert = require('assert')
 const Painting = require('../painting')
 const { onceWithCleanup } = require('../promise_utils')
-
 const { OctahedronIterator } = require('prismarine-world').iterators
 
 module.exports = inject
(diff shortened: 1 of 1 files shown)


rom1504:
Please undo irrelevant changes and fix lint

obseletecode72:
be more specific please.

lib/plugins/blocks.js
```diff
@@ -2,7 +2,6 @@ const { Vec3 } = require('vec3')
 const assert = require('assert')
 const Painting = require('../painting')
 const { onceWithCleanup } = require('../promise_utils')
-
```
rom1504: Undo

lib/plugins/blocks.js
```diff
@@ -20,44 +19,48 @@ const dimensionNames = {
...
+    const viewDist = bot.settings.viewDistance ?? 8
+    const botChunkX = Math.floor(bot.entity.position.x / 16)
+    const botChunkZ = Math.floor(bot.entity.position.z / 16)
+    const dx = args.x - botChunkX
+    const dz = args.z - botChunkZ
+    const dist = Math.max(Math.abs(dx), Math.abs(dz))
+    if (dist > viewDist) return
+
     if (!args.bitMap && args.groundUp) {
-      // stop storing the chunk column
```
rom1504: Undo

PR #3770, opened 2025-10-27
allow the use of plugins/physics without plugins/blocks as a dep

Callanplays:
I choose to use the built-in functionality of setting loadInternalPlugins to false so I can pick and choose which ones to load (namely, disabling the blocks plugin since the block change packets are using too much throughput for what I'm doing and are causing async lag and CPU overload).
I'd still like to use the physics plugin without requiring the bots' knowledge of the world state, so the solution I've found is to pass an air block into the physics module.


extremeheat:
I believe we discussed removing loadInternalPlugins option in first place. 

What likely makes more sense is specific flags to disable specific functionality in deterministic and stable way (as opposed to some undefined behaviors as mineflayer is not intended to work without other internal modules)

Callanplays:
Seems like a good idea

PR #3773, opened 2025-11-01
Adjust game height and minimum Y values

Erdef119:
The minimum Y was wrong for versions >1.17, and height was wrong for >1.18

### DIFF
--- a/lib/plugins/game.js
+++ b/lib/plugins/game.js
@@ -56,8 +56,8 @@ function inject (bot, options) {
       bot.registry.loadDimensionCodec(packet.dimensionCodec)
     }
 
-    bot.game.minY = 0
-    bot.game.height = 256
+    bot.game.minY = -64
+    bot.game.height = 320
 
     if (bot.supportFeature('dimensionDataInCodec')) { // 1.19+
       // pre 1.20.5 before we consolidated login and respawn's SpawnInfo structure into one type,


olcxja:
so cool now I can bot my server

extremeheat:
This is supposed to be set by the server in the codec and specific dimension data, why are you hard coding it ? The default values are set simply for old versions.

Erdef119:
well it didn't work on a server i was playing on and there is a check for onGround below y 0, idk why it doesn't work but when i changed these values it worked

extremeheat:
Can you provide log output after adding this code :
```js
bot._client.on('login', (d) => console.dir(d, { depth: null })
bot._client.on('respawn', (d) => console.dir(d, { depth: null })
bot.on('spawn', () => console.dir(bot.registry.dimensionsByName, { depth: null })
```

Erdef119:
[out.txt](https://github.com/user-attachments/files/23506253/out.txt)
Note that this data could be from before the bot gets transferred to the server with onGround checks (this is after the code changes)

Erdef119:
I have just looked into the logs, and there is data from the other servers

extremeheat:
Could be an issue with mismatched field names between `login` and `respawn`. I notice worldType and dimension have unequal naming between the two

plainprince:
Made a new PR fixing all the CI's and using correct per-world lookup, #3822

PR #3786, opened 2025-12-14
Add Velocity proxy support for Minecraft 1.20.2+ configuration phase

StarWeizz:
# Fix Velocity Proxy Support for Minecraft 1.20.2+ Configuration Phase

## Summary

This PR fixes bot disconnections when connecting through Velocity proxy servers with Minecraft 1.20.2+ versions. The issue occurred during server transfers when the bot sent gameplay packets (movement, physics) during the configuration phase, causing the server to kick the bot with error: "Disconnected, while in configuration phase".

Fixes #3764

### Background

Minecraft 1.20.2 introduced a new "configuration phase" that occurs:
- During initial login
- During server transfers via Velocity proxy
- When changing resource packs

During this phase:
- **Only configuration packets are allowed** (registry_data, resource_pack, finish_configuration, etc.)
- **Gameplay packets are NOT allowed** (movement, position, look, etc.)
- Sending gameplay packets causes immediate disconnection

### What Was Broken

Mineflayer was sending physics/movement packets during the configuration phase, causing:
1. **Server kicks** with error: "Disconnected, while in configuration phase server_resource_pack"
2. **Failed transfers** between Velocity backend servers
3. **Resource pack deadlocks** - server waiting for SUCCESSFULLY_LOADED before sending finish_configuration

### 1. New Velocity Plugin (`lib/plugins/velocity.js`)

A new plugin that:
- **Tracks configuration phase state** via `bot.inConfigurationPhase` flag
- **Detects configuration phase** through multiple events:
  - `start_configuration` (1.20.2+)
  - `select_known_packs` (1.21+)
  - `registry_data` (1.20.2+)
  - `transfer` (1.20.5+)
- **Disables physics** during configuration to prevent movement packets
- **Automatically accepts resource packs** during configuration phase (sends ACCEPTED + SUCCESSFULLY_LOADED immediately)
- **Re-enables physics** after configuration completes (after finish_configuration)
- **Emits events** for tracking: `bot.on('configurationPhase', (phase) => {})`

### 2. Physics Plugin Modification (`lib/plugins/physics.js`)

Added checks to prevent sending movement packets during configuration:
- `sendPacketPosition()` - Returns early if `bot.inConfigurationPhase`
- `sendPacketLook()` - Returns early if `bot.inConfigurationPhase`
- `sendPacketPositionAndLook()` - Returns early if `bot.inConfigurationPhase`
- `updatePosition()` - Returns early if `bot.inConfigurationPhase`

[...]

PR #3794, opened 2025-12-28
Send client brand correctly during configuration state for MC 1.20.2+

ODAKAB:
Handling configuration state should be considered with #3786

### DIFF
--- a/lib/plugins/game.js
+++ b/lib/plugins/game.js
@@ -27,6 +27,11 @@ function inject (bot, options) {
     throw new Error('Unsupported brand channel name')
   }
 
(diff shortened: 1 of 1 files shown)


extremeheat:
This looks like it belongs in node-minecraft-protocol

ODAKAB:
> This looks like it belongs in node-minecraft-protocol

I don't know, based on rest of the code and also based on how it handled sending client brand inside mineflayer, it made sense to me. If you say past commits were also inappropriate to do them inside mineflayer, then it might be better if someone more familiar with all these projects and their relationship does it.

PR #3754, opened 2025-10-04
🎈 1.21.9/1.21.10 support

rom1504bot:
This automated PR sets up the relevant boilerplate for Minecraft version 1.21.9.

Ref: 1096

* You can help contribute to this PR by opening a PR against this <code branch>pc1_21_9</code> branch instead of <code>master</code>.

PR #3811, opened 2026-01-20
Pc 1 21 9

HenkjanvDiermen:
Using SuperGamerTron's PR https://github.com/PrismarineJS/mineflayer/pull/3807

Fix example tests failing due to protocol auto-detection - pass version from test harness to example bots


extremeheat:
Done in #3754

PR #3822, opened 2026-02-07
Fix dimension type lookup for 1.19-1.20.4 on proxy/modded servers

plainprince:
## Summary

- For versions 1.19-1.20.4, the dimension codec lookup (`dimensionsByName`) used `packet.worldName` (the world/level name) instead of the dimension **type** name. On vanilla servers these are identical, but on proxy or modded servers they can differ (e.g. world `"lobby"` with dimension type `"overworld"`), causing the lookup to fail and leaving `minY=0, height=256` as incorrect defaults.
- Uses the correct field for the dimension type: `worldType` from login packets, `dimension` from respawn packets. This also fixes the `skyLightSent` check in `blocks.js` which compares `bot.game.dimension === 'overworld'`.
- Removes a dead-code fallback (`packet.worldType && !bot.game.dimension`) that was guarded by a condition that could never be true, since `dimensionIsAWorld` always set `bot.game.dimension` first.

Related to #3773 -- this is the proper fix for the underlying issue (rather than hardcoding `-64`/`320` as defaults, which breaks pre-1.17 versions).

## Test plan

- [x] All 353 internal tests pass (`npx mocha test/internalTest.js`)
- [x] Verified on a local 1.21.4 vanilla server: `minY=-64, height=384` correctly resolved from the dimension codec
- [x] CI should pass for all supported versions (no change to defaults for pre-1.19)
- [ ] Ideally tested on a BungeeCord/Velocity proxy with custom world names to confirm the lookup now resolves correctly

### DIFF
--- a/test/externalTests/dimensionData.js
+++ b/test/externalTests/dimensionData.js
@@ -1,5 +1,4 @@
 const assert = require('assert')
-const { onceWithCleanup } = require('../../lib/promise_utils')
 
 module.exports = () => {
   const tests = []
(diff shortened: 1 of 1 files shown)


plainprince:
Do you prefer with comments to show the bugfix or should I remove them?

extremeheat:
Comments are good when explaining non obvious logic, it doesn't seem excessive here

plainprince:
Welp, all CI's pass! That's great!

plainprince:
I can try to test it on a custom paper server with a custom spawn dimension and see if it works, but I feel like the testing here is slightly excessive (and I already tested it with all vanilla dimensions) as I see no reason why it should fail, but it's no problem if you want me to.

plainprince:
Can we merge?

PR #3827, opened 2026-02-08
Fix flaky time and nether tests

plainprince:
## Summary
Fixes two flaky tests that cause spurious CI failures on unrelated PRs (e.g. #3821).

- **time test**: The daylight cycle was left enabled during `/time set` assertions, so time would drift between the command and the check (e.g. expected 18000, got 21695 -- a delta of 3695, well beyond the 510-tick tolerance). Fix: disable daylight cycle before time transition tests and re-enable only for the progression/toggle sections.
- **nether test**: On slow CI runners, `placeBlock` would time out waiting for `blockUpdate` (5s hard limit) immediately after chunk loading. Fix: add a 1s stabilization wait after `waitForChunksToLoad()`.

## Test plan
- [ ] CI time test passes consistently across all versions
- [x] CI nether test passes on 1.15.2
- [x] No regressions on other tests

Made with [Cursor](https://cursor.com)

### DIFF
--- a/test/externalTests/plugins/testCommon.js
+++ b/test/externalTests/plugins/testCommon.js
@@ -192,9 +192,16 @@ function inject (bot) {
       })
       childBotName = message.json.with[0].insertion
       bot.chat(`/tp ${childBotName} 50 ${bot.test.groundY} 0`)
-      // The server processes commands in order, so the TP is applied
-      // before this chat message is delivered to the child bot. The
-      // child's waitForChunksToLoad() then waits at the new position.
+      // Wait for the child entity to arrive at the teleport target,
+      // confirming the server has processed the TP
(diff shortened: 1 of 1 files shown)


plainprince:
Just a heads up -- #3828 includes these flaky test fixes alongside the `@types/node` bump from #3821. If #3828 gets merged, this PR can be closed as the changes are already covered there.

plainprince:
The reason the last test fails is probably due to a corrupted packet which is just a problem with the way this CI works at it's lowest level, so it can be safely ignored :)

rom1504:
Please rebase

plainprince:
Closing in favor of #3828, which includes these flaky test fixes alongside the @types/node bump.

PR #3828, opened 2026-02-08
Bump @types/node to 25.2.1 + fix flaky time & nether tests

plainprince:
## Summary
- Bumps \`@types/node\` from \`^24.0.6\` to \`^25.2.1\` (supersedes #3821)
- Fixes two flaky tests that were causing CI failures on #3821 (and other PRs)

## Flaky test fixes
- **time test**: Daylight cycle was left enabled during \`/time set\` assertions, so time would drift between the command and the check (e.g. expected 18000, got 21695). Fix: disable daylight cycle before time transition tests, re-enable only for progression/toggle sections.
- **nether test**: \`placeBlock\` timed out on slow CI waiting for \`blockUpdate\` right after chunk loading. Fix: add a 1s stabilization wait after \`waitForChunksToLoad()\`.

## Test plan
- [x] Lint passes
- [x] Time test passes consistently across all versions
- [x] Nether test passes on 1.15.2
- [x] No regressions on other tests

Made with [Cursor](https://cursor.com)

### DIFF
--- a/test/externalTests/plugins/testCommon.js
+++ b/test/externalTests/plugins/testCommon.js
@@ -192,9 +192,16 @@ function inject (bot) {
       })
       childBotName = message.json.with[0].insertion
       bot.chat(`/tp ${childBotName} 50 ${bot.test.groundY} 0`)
-      // The server processes commands in order, so the TP is applied
-      // before this chat message is delivered to the child bot. The
-      // child's waitForChunksToLoad() then waits at the new position.
+      // Wait for the child entity to arrive at the teleport target,
+      // confirming the server has processed the TP
+      const targetPos = new Vec3(50, bot.test.groundY, 0)
+      while (!bot.players[childBotName]?.entity ||
+             bot.players[childBotName].entity.position.distanceTo(targetPos) > 5) {
+        await sleep(100)
+      }
+      // Let the child's physics engine initialize at the new position
+      // (ground detection, chunk processing) before starting the test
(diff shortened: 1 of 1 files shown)


plainprince:
flaky tests are still failing, working on them.

plainprince:
Nice, CI is passing and the flaky tests are seemingly fixed.

rom1504:
Please rebase

test/externalTests/plugins/testCommon.js
```diff
@@ -188,14 +188,21 @@ function inject (bot) {
...
       bot.chat(`/tp ${childBotName} 50 ${bot.test.groundY} 0`)
-      setTimeout(() => {
-        bot.chat('loaded')
-      }, 5000)
+      // Wait for the child entity to arrive at the teleport target,
+      // confirming the server has processed the TP
+      const targetPos = new Vec3(50, bot.test.groundY, 0)
+      while (!bot.players[childBotName]?.entity ||
+             bot.players[childBotName].entity.position.distanceTo(targetPos) > 5) {
+        await sleep(100)
```
rom1504: Much better to use events than this kind of checking every 100ms

rom1504:
Merging as it looks like a strict improvement 

However the while wait for 100ms check is an anti pattern for mineflayer and we should use events or introduce appropriate ones if there are none. Would be great to fix in other PR or at least we should not reproduce this pattern in other places of the code base

Issue #3833, opened 2026-02-13
Throwing wind charges down


SwedZou:
I have not yet found any way to make the bot throw a wind charge down at the ground. If i use activateItem(), the bot will look up and then throw the wind charge. And bot._client.write doesn't work in 4.34.0
I would like a similar function as ActivateItem(), but that doesn't looks up when running. Or a function just for throwing things.

Issue #3621, opened 2025-03-28
allow rotations in activateItem


boxyfour:
## Is your feature request related to a problem? Please describe.

i've been trying to implement a feature that allows the bot to rotate downwards and use an xp bottle, but the default implementation for activateItem assumes the rotation of the player to be 0 0.

## Describe the solution you'd like

a second parameter expecting a vector2 for the rotation, or pitch / yaw. default it to 0 0 otherwise

## Describe alternatives you've considered

sucking it up and throwing xp at a wall

cnbdhr62:
Our system attempted to address this issue:

Added rotation parameter to activateItem function to allow specifying pitch/yaw when using items. This enables more precise control over item usage direction, particularly useful for throwing experience bottles downward. The rotation parameter is optional and defaults to 0,0 to maintain backward compatibility.

You can examine the suggested modifications here: https://github.com/cnbdhr62/PrismarineJS-mineflayer-3621/commit/1f6f4899ff6c5835fbb6b7a077011ba0a6577556.


> [!CAUTION]
> Advisory: This solution was generated using automated tools, and we strongly recommend reviewing the code thoroughly before implementation. The solution may be partial - please consider it as inspiration for your own implementation.

---

http://missionfixinternet.com

Nekiplay:
and i need this

PR #3840, opened 2026-02-21
Fix activateItem always having the same rotation

SuperGamerTron:
Correctly uses the bot's rotation in the `activateItem` function instead of having a (0, 0) rotation.
Fixes:
https://github.com/PrismarineJS/mineflayer/issues/3537
https://github.com/PrismarineJS/mineflayer/issues/3621
https://github.com/PrismarineJS/mineflayer/issues/3833

### DIFF
--- a/test/externalTests/activateItem.js
+++ b/test/externalTests/activateItem.js
@@ -1,0 +1,76 @@
+const assert = require('assert')
+
+module.exports = () => async (bot) => {
+  const Item = require('prismarine-item')(bot.registry)
+
+  // Test that activateItem sends the bot's actual rotation (PR #3840).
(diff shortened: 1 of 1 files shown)


osyra42:
Could this be related to why my bot always faced south when trying to go fishing? The bot will look at the right water block and then turn south quickly, cast, and then look back at the block.

DatArnoGuy:
> Could this be related to why my bot always faced south when trying to go fishing? The bot will look at the right water block and then turn south quickly, cast, and then look back at the block.

Yes

osyra42:
has this been pulled to main?

rom1504:
Added tests

Issue #3882, opened 2026-04-02
Bot freezes after taking any damage


Skygawker:
## Versions
 - mineflayer: 4.37.0
 - server: vanilla/spigot/paper Vanilla and paper 1.21.4
 - node: 22.22.2

## Detailed description of a problem
A clear and concise description of what the problem is, with as much context as possible.
What are you building? What problem are you trying to solve?

I am trying to build a bot and it works okay, but when something deals him damage bot freezes and cannot do nothing untill he is killed or rejoining the server

## What did you try yet?

Did you try any method from the API?
Did you try any example? Any error from those?

I write any code, download code from examples or write my own, downgrade and upgrade java, node, mineflayer and still no fix

## Your current code
```js


const mineflayer = require('mineflayer')

const bot = mineflayer.createBot({
  host: 'localhost',
  port: 25565,
  auth: 'microsoft'
})



```

## Expected behavior
bot that handles damage and doesnt get stuck after 1 hit

NikkoPlayz:
Yea i get the same problem..

stgov:
Got the same bug

jeremy46231:
can reproduce on 1.12.11 (paper), the root cause seems to be that the velocity packet from the server is parsed wrong (giving NaN values)? that's why #3883 stops the physics from breaking but doesn't make the physics work right, it just ignores the packets that lead to NaN values.

(tw: ai slop; my llm managed to make it work correctly by changing the endianness of minecraft-protocol's lpVec3 parsing and parsing the velocity in this packet as lpVec3 (not dividing by 8000), and also maybe entityVelocityIsLpVec3 is set wrong in minecraft-data?, but I don't know what I'm doing with this codebase and it probably just made a mess lol)

Scorpwanna:
Same, it just stands there and takes a beating.

NeilRuffell:
Identical issue.

Minecraft Version: 1.21.1 (Fabric)

TiredFox65:
same issue in forge 1.20.1

luisgbr1el:
I solve this problem in Vanilla 1.21.11 simply tracking the last valid position and restoring it when NaN is detected.

```js
let lastValidBotPosition = null;

// Add this after loading plugins
bot.on('physicsTick', () => {
    if (bot.entity?.position) {
        if (Number.isFinite(bot.entity.position.x) && Number.isFinite(bot.entity.position.z))
            lastValidBotPosition = bot.entity.position.clone();
[...]

PR #3883, opened 2026-04-04
Fix for physics lock-up in 1.21.4

sampritchard03:
During my testing (and also documented here: https://github.com/PrismarineJS/mineflayer/issues/3882), bots on at least 1.21.4 will permanently lock up when hit by an arrow. This is because prismarine physics returns NaN for pos.x and pos.z. This edit just catches and rejects the illegal position, shouldn't affect anything that already works.


MICKYcyber:
well whats the fucking fix

GenerelSchwerz:
this fix is a stop measure, review #3889 for the actual fix.

PR #3890, opened 2026-04-21
add dynamic tick interval support

GenerelSchwerz:
Currently, the default execution of ticks is 50ms. This is representative of normal minecraft behavior. However, being able to change that tickrate is helpful for debugging various applications, as we can see the bot's movement in-game, step by step. Further decreasing the time between tick executions also opens the door for machine learning code using an actual nmp client as an agent.

### DIFF
--- a/lib/plugins/physics.js
+++ b/lib/plugins/physics.js
@@ -11,8 +11,8 @@ module.exports = inject
 
 const PI = Math.PI
 const PI_2 = Math.PI * 2
-const PHYSICS_INTERVAL_MS = 50
-const PHYSICS_TIMESTEP = PHYSICS_INTERVAL_MS / 1000 // 0.05
+const DEFAULT_PHYSICS_INTERVAL_MS = 50
+const DEFAULT_PHYSICS_TIMESTEP = DEFAULT_PHYSICS_INTERVAL_MS / 1000 // 0.05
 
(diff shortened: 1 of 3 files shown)


lib/plugins/physics.js
```diff
@@ -202,6 +210,22 @@ function inject (bot, { physicsEnabled, maxCatchupTicks }) {
...
+      if (!Number.isFinite(value) || value <= 0) {
+        throw new RangeError('physics.physicsIntervalMs must be a finite number greater than 0')
+      }
+      physicsIntervalMs = value
+      physicsTimestep = value / 1000
+      restartPhysicsTimer()
+      return value
+    }
+  })
+
```
extremeheat, line 228: Why is this not a function? Setting props for side effects is not good idiomatic API

extremeheat (changes requested):

lib/plugins/physics.js
GenerelSchwerz: Yeah I'll change it tomorrow.

GenerelSchwerz: Oh actually, it's to maintain the same interface that various other settings have/had. Seemed right at the time, can change it if you want.

GenerelSchwerz: fix has been applied.

PR #3891, opened 2026-04-21
Look sync fix

GenerelSchwerz:
Update look sync to:

1. be stricter on pitch changes as well.
2. warn the user if multiple bot.look calls are being made at the same time, as simply returning completion leaves the end result ambiguous. Ideally, we'd throw an error instead to dissuade bad coding practice, but that feels too opinionated.

Implementation based on: https://github.com/nxg-org/mineflayer-util-plugin/blob/master/src/movementFunctions.ts , where this code does correctly sync sent packet data and expected look return timing.


extremeheat:
CI failing but also
> warn the user if multiple bot.look calls are being made at the same time, as simply returning completion leaves the end result ambiguous. Ideally, we'd throw an error instead to dissuade bad coding practice, but that feels too opinionated.

Does the current logic not simply cancel the previous call and change the position target to the new call's parameters? Why would deprecating this working functionality be better than existing handling?

GenerelSchwerz:
> CI failing but also
> 
> > warn the user if multiple bot.look calls are being made at the same time, as simply returning completion leaves the end result ambiguous. Ideally, we'd throw an error instead to dissuade bad coding practice, but that feels too opinionated.
> 
> 
> 
> Does the current logic not simply cancel the previous call and change the position target to the new call's parameters? Why would deprecating this working functionality be better than existing handling?

Because if you have logic that requires the bot to be looking at a certain point in a plugin (for example, aiming a bow), and the end user overrides that look with their own code, it causes Stange behavior that is a headache for both sides. Having the bot error or warn the user that their code is overriding existing calls is strictly good, as subsequent calls to look/lookAt on the same tick makes at least one code path largely irrelevant.

DatArnoGuy:
Should the api have something like bot.isPerformingLookingTask for end users to check if bot is already changing its angle? This would help users to avoid errors

extremeheat:
> Having the bot error or warn the user that their code is overriding existing calls is strictly good, as subsequent calls to look/lookAt on the same tick makes at least one code path largely irrelevant.

Yes, but rejecting this is actually a behavior change from current API. I don't see why this is necessary if it's already handled by synchronous overriding of any previous calls in orderly manner

GenerelSchwerz:
> > Having the bot error or warn the user that their code is overriding existing calls is strictly good, as subsequent calls to look/lookAt on the same tick makes at least one code path largely irrelevant.
> 
> Yes, but rejecting this is actually a behavior change from current API. I don't see why this is necessary if it's already handled by synchronous overriding of any previous calls in orderly manner

It's not necessary, it's preferential. Silently overriding a previous lookAt call may make that previous code fail silently. For example: if a plugin requests to look in a specific direction for aiming a bow, but another plugin requests to look in a slightly different direction (say, at the player the bow is shooting at), the bow code will silently fail and the end user will spend far more time debugging than what it's worth. In contrast, throwing the warn/error will immediately let the user know that a code path is being overridden.

GenerelSchwerz:
An alternative to this is implementing a priority system for looking. But at that point, a more robust fix for implementing priority inputs per-tick would be better.

extremeheat:
Unless there is an actual error or bug in the code (or some kind of deprecation warning that behavior will change), we should not be printing warnings to the console.

What you said can apply to many other bot functions, we don't print warnings on them. If there is an error, the API should explicitly throw an exception, not a half way "allowed" but logs spam to the console

extremeheat:
For debugging purposes it may make sense. However `.lookAt` is called in many places sometimes within mineflayer's own code. So a random log message is not helpful in figuring out where it's being called from. So if we did need to log, it would likely make sense to do as a `console.trace` than a log message (would have to check if mineflayer has a mechanism for debug messages at the moment)

GenerelSchwerz:
> Unless there is an actual error or bug in the code (or some kind of deprecation warning that behavior will change), we should not be printing warnings to the console.
> 
> 
> 
> What you said can apply to many other bot functions, we don't print warnings on them. If there is an error, the API should explicitly throw an exception, not a half way "allowed" but logs spam to the console

I only recommended warning to console due to the physicTick warning implementation. I would heavily suggest throwing the exception over printing a warning.

extremeheat:
I'd also add that in most cases, you do want to override the past lookAt. If you start a .dig and then want to look away from that job immediately after, you should be able to just call .lookAt and have it override that last action.

If the user does not desire that, they should be able to check with API if there is some existing action and cancel that. A queue system could be added if needed yes but this can also be implemented on the user side. If we added this queue system to mineflayer, it'd have to be something generalized as opposed to some special mechanism just for `.lookAt` calls

GenerelSchwerz:
> I'd also add that in most cases, you do want to override the past lookAt. If you start a .dig and then want to look away from that job immediately after, you should be able to just call .lookAt and have it override that last action.
> 
> 
> 
> If the user does not desire that, they should be able to check with API if there is some existing action and cancel that. A queue system could be added if needed yes but this can also be implemented on the user side. If we added this queue system to mineflayer, it'd have to be something generalized as opposed to some special mechanism just for `.lookAt` calls

For this digging example, it is unlikely that you'd want to do this.

The behavior you are describing, overriding the dig's lookAt, would only occur if the dig behavior hasn't actually begun yet. You would override the look, and you would never look at that block.

However, the dig packet would still be sent out because the code path assumes that once its lookAt is complete, it is looking at the block. This is blatantly wrong behavior. That dig function should throw an error detailing that it wasn't able to even begin digging due to the lookAt being overridden.

extremeheat:
It may not be "correct" but from gameplay standpoint, but from game logic it's valid hence we have an option whether to begin the lookAt. In any case, mineflayer isn't supposed to block the user from doing illogical things (beyond things that would result in real error or undefined behaviors), for that a higher level library would make more sense.

What you are proposing is essentially a behavior change to block overriding the previous lookAt which I don't understand

GenerelSchwerz:
> It may not be "correct" but from gameplay standpoint, but from game logic it's valid hence we have an option whether to begin the lookAt. In any case, mineflayer isn't supposed to block the user from doing illogical things (beyond things that would result in real error or undefined behaviors), for that a higher level library would make more sense.
> 
> 
> 
> What you are proposing is essentially a behavior change to block overriding the previous lookAt which I don't understand

This is incorrect. What I am proposing is that the previous lookAt that would normally be overridden would throw the error, not the current call. I believe I implemented that correctly; if not, that was my intention. 

What I just described would be a failure of the bot.dig function, not a failure of the user. Performing that dig where the code expects to be looking at the block when in reality it is not is a failure of the library. If a lookAt is called that overrides the bot.dig call, bot.dig should throw an error so it does not continue execution.

GenerelSchwerz:
Can I get a reply on my most recent comment

extremeheat (changes requested):
Yes, I already addressed that. A behavior change is not necessary there, my opinion is not changed

GenerelSchwerz:
> Yes, I already addressed that. A behavior change is not necessary there, my opinion is not changed

If you could please explain further, that would be appreciated. My current understanding of your stance is found here:

> I'd also add that in most cases, you do want to override the past lookAt. If you start a .dig and then want to look away from that job immediately after, you should be able to just call .lookAt and have it override that last action.

This logic makes sense. My most recent change allows this behavior to occur. The issue that my PR fixes is that the code for bot.dig would still execute, even if we wanted to cancel the lookAt code for dig. Given that the code for bot.dig assumes that the bot's server reported yaw and pitch is looking at said block (unless "ignore" is explicitly given in bot.dig's options for looking), interrupting the lookAt behavior called by bot.dig in favor of a newer lookAt should cause bot.dig to throw an error.

If the user wants to avoid this issue and silently override the behavior as before, they would simply need to call bot.dig in a try/catch loop. The function that most recently called bot.lookAt would never throw an error.

> In any case, mineflayer isn't supposed to block the user from doing illogical things (beyond things that would result in real error or undefined behaviors), for that a higher level library would make more sense.

The scenario for digging that I just described falls in line with what mineflayer's current code does and expects. If the user wants to look somewhere immediately after calling bot.dig, then they should specifically request bot.dig with the option `ignore` for looking. The intended behavior of bot.dig is to be looking at the block when starting the dig.

> It may not be "correct" but from gameplay standpoint, but from game logic it's valid hence we have an option whether to begin the lookAt.

Considering that the default option is to be looking at the block, having external code cause the bot to not be looking at the block when beginning digging should throw an error, not continue bot.dig's execution.

> What you are proposing is essentially a behavior change to block overriding the previous lookAt which I don't understand

I am proposing that tasks that require lookAt to complete with the server-sided reported state being what was initially requested should fail if the server-sided reported value is not what was initially requested upon lookAt's completion/cancellation. I am not proposing any change to how overriding previous bot.lookAt calls is handled.

if lookAt 1 is cancelled, lookAt 2 will still succeed as intended. The current behavior of lookAt 2 is entirely unchanged.

extremeheat:
> If the user wants to avoid this issue and silently override the behavior as before, they would simply need to call bot.dig in a try/catch loop. The function that most recently called bot.lookAt would never throw an error.

Yes, and this is a breaking behavior change. Your new error would propagate and affect other calls like bot.dig.

> if lookAt 1 is cancelled, lookAt 2 will still succeed as intended. The current behavior of lookAt 2 is entirely unchanged.

The error is definitely a behavior change and it will crash existing code that doesn't handle the error

Right now, lookAt can gracefully preempt any previous calls and there is no state corruption doing so. For example if you enter a console and do `bot.lookAt(vec1)`, and you want to suddenly look somewhere else, typing in `bot.lookAt(vec2)` will just cancel any previous transition and start transition in other direction. This is pretty normal for animated behavior in software. Via some API you might be able to get a handle and cancel previous operation but there is no reason to change the existing behavior that works as is

Issue #3220, opened 2023-10-30
Description of argument `yaw` in `bot.look` in possiblely wrong


szdytom:
https://github.com/PrismarineJS/mineflayer/blob/08208e2f110af2c6de41fac9a389597aac916412/docs/api.md?plain=1#L1820-L1821

The document says `yaw` starts from due east(i.e. +X) but I doubt it. I tested myself by calling `bot.look(0, 0, true)` and check the bot's rotation angle in another client with F3+B, it says the bot is look at due north(i.e. -Z). I then called `bot.setControlState('forward', true)` and the bot indeed moved north.

My experiment on the Minecraft 1.20.1 server turns out with the following result:

```
                -Z(North)             
                    ^ yaw = 0           
                    |            
 yaw = 0.5*PI       |            
   -X(West) --------|-------> +X(East)     
                    |       yaw = 1.5*PI
                    |            
                    | yaw = PI         
                +Z(South)                 
```

TAOtxi:
Unbelievable, this problem still exists after 3 years and has not been fixed yet. [Wiki](https://minecraft.wiki/w/Rotation) as mentioned above, yaw should start from the south direction, but it is obvious that there are obvious errors in both the bot and the documentation.

PR #3901, opened 2026-05-12
Add look2 and getNotchYawPitch methods to adapt to the game's rotation

TAOtxi:
```js
// The result is the same
bot.look2(90, 0)  // use degree, more detail see https://minecraft.wiki/w/Rotation#Entity_format
bot.look(Math.PI / 2, 0)
```

Related to #3220 
I can't understand why the perspective system of the game needs to be transformed into other systems, which makes it even more complex.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -297,6 +297,14 @@ export interface Bot extends TypedEmitter<BotEvents> {
     force?: boolean
   ) => Promise<void>
 
+  look2: (
+    notchYaw: number,
+    notchPitch: number,
+    force?: boolean
+  ) => Promise<void>
+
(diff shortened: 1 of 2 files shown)


extremeheat:
What problem are you trying to solve?

TAOtxi:
> What problem are you trying to solve?

Make it easier to change the bot's perspective.

When I want the bot to face due east, the F3 information in the game tells me that the yaw and pitch in the due east direction are (-90,0). But when making the bot look in this direction, it needs to convert (-90,0) for bot. Actually, mineflayer did not expose the conversions method externally, so I implement the look2 method.

Issue #3910, opened 2026-05-25
bot.dig() applies vanilla 5x airborne penalty when bot.entity.onGround=false at packet-send time; no internal guard or warning


VasilisDragon:
When `bot.dig(block)` is called while `bot.entity.onGround === false`, or when
`onGround` flips to false during `dig()`'s internal `lookAt` awaits, the
`block_dig.start` packet is sent with the bot airborne and the server applies
its documented 5x airborne dig-speed penalty. The actual dig then takes ~5x
longer than `bot.digTime(block)` reported. There's no internal guard, no
warning event, and the behaviour isn't documented.

### Results

Same pattern on Paper 1.21.11 and vanilla 1.21.11:

| Trial | onGround at dig call | digTimeEstimate | actual elapsed | delta |
|---|---|---|---|---|
| 1: on-ground (Paper)        | true  | 4550 ms | 4565 ms  | +15 ms |
| 2: jumped (Paper)           | false | 4550 ms | **22524 ms** | **+17974 ms** |
| 3: on-ground rerun (Paper)  | true  | 4550 ms | 4560 ms  | +10 ms |
| 1: on-ground (vanilla)      | true  | 4550 ms | 4552 ms  | +2 ms |
| 2: jumped (vanilla)         | false | 4550 ms | **22512 ms** | **+17962 ms** |
| 3: on-ground rerun (vanilla)| true  | 4550 ms | 4550 ms  | +0 ms |

The 4 550 ms baseline reflects the unrelated `incorrect_for_wooden_tool`
material issue tracked at PrismarineJS/minecraft-data#1193; the 5x multiplier
above it is the bug here. Running the same trial shape on stone (unaffected
by the material issue) shows the same 5x ratio, so the two are independent.

### Repro

Standalone repro and full packet traces:
**https://gist.github.com/VasilisDragon/114bf4b7d700064f92701696ad5937f3**

Three files do the work: `setup_{paper,vanilla}.js` (RCON helpers that
build a clean diamond_ore platform and prep the bot), `repro_dig_airborne.js`
(three trials, with an optional `--trace` flag), and `packet_trace.js`
(wire-level capture hook).

### Wire-level evidence (vanilla 1.21.11, trial 2)

```
t=8381  trial.about-to-dig  digTimeEstimate=4550  onGround=true
t=8454  out block_dig.START loc=(251,100,500) face=1
                            onGround=FALSE pos=(251.5, 100.753, 499.5) velY=0.248
[...]

PR #3912, opened 2026-05-27
digging: recompute waitTime after lookAt resolves

VasilisDragon:
## Problem

`lib/plugins/digging.js` captures `const waitTime = bot.digTime(block)` at the top of `dig()`, then awaits one or more `bot.lookAt()` calls before sending the `block_dig.START` packet. If `bot.entity.onGround` flips during those awaits (jump released, walked off a ledge, mid-fall), the captured `waitTime` reflects the pre-await ground state while the server applies the vanilla airborne 5× dig-time penalty against the actual post-await state.

Result: `finishDigging()` fires roughly 5× too early on airborne digs. The local timeout sends `block_dig.FINISH` before the server has applied the actual break, producing visible "swung but didn't break" symptoms.

Repro and packet-trace evidence: #3910.

## Fix

Keep the early `bot.digTime(block)` call for the existing `Infinity` check only (fail-fast on unmineable blocks before any look/swing work happens). Recompute `waitTime` immediately before the `block_dig.START` packet is written, so it reflects bot state at packet-send time.

Two-line diff in `lib/plugins/digging.js`:
- Remove `const waitTime =` from the early Infinity check; inline the call.
- Add `const waitTime = bot.digTime(block)` between `createTask()` and the `block_dig.START` write.

## Test

Added to the existing `digTime` suite in `test/internalTest.js` (fake server + bot harness, runs for every tested version), per review:

- `recomputes digTime after lookAt resolves so the timer reflects post-await onGround state`: stubs `bot.lookAt` to flip `bot.entity.onGround` to false before resolving, spies on `bot.digTime`, and asserts the final call (the one that sets `waitTime`) observes the post-await ground state. Fails with the fix reverted (verified locally).

`npm test` passes locally (`standard` lint plus mocha).

## Scope

Intentionally narrow. The optional `waitForGround` flag and `digTimeDesync` diagnostic event also discussed in #3910 are not included here; they belong in separate follow-up PRs once this minimal fix lands.

## Precedent

Parallels #3542 (`Patches isInWater for digSpeed`) — same module, same flavor of fix: the local timer now reflects the actual server-side mechanic at packet-send time rather than at function-entry time.


test/diggingWaitTimeTest.js
extremeheat, line 1: This should be a proper internal test otherwise it cannot be maintained

VasilisDragon: Moved into test/internalTest.js in existing suite, runs via fake server + bot harness. verified it catches regression when fix reverted. CI shows unrelated mineflayer_external 1.19.3 server-jar download timeout.

vRXDk5U:
>  Add const waitTime = bot.digTime(block) between createTask() and the block_dig.START write.

This is an improvement but still doesn't match the vanilla behavior. Vanilla updates the mining speed whenever the conditions change. Here I start mining slowly because I am mid-air but then resume my normal speed when I touch the ground:

https://github.com/user-attachments/assets/0cc0d355-a350-43f0-bdea-2769104f8769

vRXDk5U:
>  Vanilla updates the mining speed whenever the conditions change.

That being said, most of the scenarios when this would happen to a real player don't happen to mineflayer bots; they usually stand still while mining and don't mine from the side of ladders or underwater.

VasilisDragon:
> > Vanilla updates the mining speed whenever the conditions change.
> 
> That being said, most of the scenarios when this would happen to a real player don't happen to mineflayer bots; they usually stand still while mining and don't mine from the side of ladders or underwater.

Right, and no single up-front estimate can track a mid-dig change. This PR keeps mineflayer's single-estimate model on purpose and only fixes when the estimate is taken: at packet-send time instead of function entry, which is the stale-state case from #3910 (~5x early FINISH). Matching vanilla with per-tick progress accumulation would be a larger change than this minimal fix, and as you note it rarely matters for bots that mine standing still. If it shows up in practice it can land as a follow-up alongside the waitForGround and digTimeDesync items in the scope note.

extremeheat (approved):

VasilisDragon:
On September 11, I reran the internal tests on 1.12.2, 1.13.2, 1.14.4 and 1.15.2. All passed, and the regression failed without the fix on all four.

The old CI logs have expired, so this doesn't explain that failure.

PR #3919, opened 2026-06-07
Fix isAlive state for non-death respawns

### LINKED ISSUE #3905: In respawn event, bot.isAlive is always set to false
## Versions
 - mineflayer: latest
 - server: leaf, velocity 1.21.11

The server I am playing is base on velocity. When I enter the lobby server, `spawn` event always appears earlier than `respawn` event (Perhaps it's a bug in leaf).
Sometimes when transmitting to other sub servers, the `spawn` event is not appeared, but the `respawn` event appeared (Maybe it's another bug of leaf).

This leads to `bot.isAlive` set to false([#master/lib/plugins/health.js#L7](https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/health.js#L7)).

In fact, the bot is still alive, But bot will not update its position and rotation([#master/lib/plugins/physics.js#L149-L154](https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/physics.js#L149-L154)).

## Solution
Replace `bot.isAlive` with `bot.health > 0` in [#master/lib/plugins/physics.js#L149-L154](https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/physics.js#L149-L154).

Or..
In the `respawn` event callback, `bot.isAlive` is only set to false when the player dies.
```js
bot._client.on('respawn', (packet) => {
    // bot.isAlive = bot.supportFeature('spawnRespawnWorldDataField') ?
    //   packet.copyMetadata !== 0 : packet.copyMetadata
    bot.isAlive = Boolean(packet.copyMetadata)
    bot.emit('respawn')
 })
```
But I'm not sure if `packet.copyMetadata` is always set to 0 when the player dies

0rionix:
Fixes #3905.

This PR prevents `bot.isAlive` from being set to `false` for every `respawn` packet.

Some servers/proxies can send `respawn` packets for non-death transitions, such as server transfers or world changes. In those cases the bot is still alive, but Mineflayer marked it as dead, which could stop physics position updates.

Now `bot.isAlive` is only set to `false` on `respawn` if `bot.health <= 0`.

Added an internal regression test to verify healthy respawn packets keep `bot.isAlive === true`.

Tested with:

- `npm run mocha_test -- test/internalTest.js -g "keeps isAlive true"`
- `npm run lint`

### DIFF
--- a/lib/plugins/health.js
+++ b/lib/plugins/health.js
@@ -4,7 +4,7 @@ function inject (bot, options) {
   bot.isAlive = true
 
   bot._client.on('respawn', (packet) => {
-    bot.isAlive = false
+    if (bot.health <= 0) bot.isAlive = false
     bot.emit('respawn')
   })
 
--- a/lib/plugins/health.js
+++ b/lib/plugins/health.js
@@ -6,6 +6,7 @@ function inject (bot, options) {
   bot._client.on('respawn', (packet) => {
     if (bot.health <= 0) bot.isAlive = false
     bot.emit('respawn')
+    if (bot.health > 0) bot.emit('spawn')
   })
 
   bot._client.once('update_health', (packet) => {
--- a/test/internalTest.js
+++ b/test/internalTest.js
@@ -525,7 +525,7 @@ for (const supportedVersion of mineflayer.testedVersions) {
     })
 
     describe('health', () => {
-      it('keeps isAlive true for non-death respawn packets', (done) => {
+      it('keeps isAlive true and emits spawn for non-death respawn packets', (done) => {
         const loginPacket = bot.test.generateLoginPacket()
         let respawnPacket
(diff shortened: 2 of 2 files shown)


0rionix:
Hi @extremeheat Could you plz review my PR?

lib/plugins/health.js
```diff
@@ -4,8 +4,9 @@ function inject (bot, options) {
   bot.isAlive = true
 
   bot._client.on('respawn', (packet) => {
-    bot.isAlive = false
+    if (bot.health <= 0) bot.isAlive = false
     bot.emit('respawn')
+    if (bot.health > 0) bot.emit('spawn')
```
extremeheat, line 9: what is this change? this seems breaking

PR #3922, opened 2026-06-11
Prevent lookingTask resolving when pitch incorrect.

vRXDk5U:
lookingTask was getting resolved when the yaw was correct regardless of the pitch. This caused bot.look to resolve immediately when only the pitch changed.
In the following image, bot.dig thinks it is already looking at the block above and started mining it early:
image

### DIFF
--- a/lib/plugins/physics.js
+++ b/lib/plugins/physics.js
@@ -304,7 +304,7 @@ function inject (bot, { physicsEnabled, maxCatchupTicks }) {
   let lookingTask = createDoneTask()
 
   bot.on('move', () => {
-    if (!lookingTask.done && Math.abs(deltaYaw(bot.entity.yaw, lastSentYaw)) < 0.001) {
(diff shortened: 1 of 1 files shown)


VasilisDragon:
Can confirm the root cause in the source. bot.look applies the target to bot.entity.yaw and bot.entity.pitch immediately, and the move handler only checks yaw convergence before finishing lookingTask, so a pitch-only look resolves on the first move event while lastSentPitch is still ramping at maxDeltaPitch per tick. Digging straight up or down is the everyday pitch-only case, which matches your screenshot.

Verified the fix mechanics while reading: both values are radians (the notchian conversion only happens at packet write), pitch is clamped so the plain abs diff is correct where yaw needs deltaYaw wrap handling, and lastSentPitch converges exactly so there is no hang risk.

One thing the description undersells: this affects default digs. dig() passes forceLook through to lookAt unchanged, so bot.dig(block) takes the non-force path and awaits lookingTask. The forceLook = true default in the docs does not exist in the code. It also composes with #3912: once the look await has its real duration, the window where onGround can flip mid-await gets wider, which is exactly the stale state that PR recomputes waitTime after.

For reference, #3912 got asked for an internal test in test/internalTest.js for the same kind of timing change.

extremeheat (approved):

PR #3923, opened 2026-06-12
docs: correct bot.dig forceLook description

VasilisDragon:
## Problem

`docs/api.md` shows `bot.dig(block, [forceLook = true], [digFace])`, but the parameter has no default in `lib/plugins/digging.js` (the signature is `dig (block, forceLook, digFace)`, back to at least 3.0.0). Calling `bot.dig(block)` therefore takes the non-force path: the bot turns its head at the normal look rate and the turn is awaited before digging starts. The current text also says `forceLook` accepts `'raycast'`, but that value belongs to `digFace`, which already documents it.

## Fix

Docs only, no code change. Signature and TOC: `[forceLook = true]` -> `[forceLook]`. The `forceLook` bullet now describes the actual omitted/false behavior, keeps `'ignore'`, and drops the misattributed `'raycast'`. Correcting the code instead (adding the documented default) would silently turn every default dig into an instant head snap, so the docs are corrected to match the code.

English docs only; translations are maintained separately.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -299,7 +299,7 @@
       - [bot.tossStack(item)](#bottossstackitem)
       - [bot.toss(itemType, metadata, count)](#bottossitemtype-metadata-count)
       - [bot.elytraFly()](#botelytrafly)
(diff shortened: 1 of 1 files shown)


extremeheat:
Thanks

PR #3931, opened 2026-06-21
Synchronize inventory when received close_window packet

TAOtxi:
Fix: When the bot opens a container, if the bot's inventory items are updated but the container is suddenly damaged, the updated items during this period will not be synchronized to the `bot.inventory`

### DIFF
--- a/lib/plugins/inventory.js
+++ b/lib/plugins/inventory.js
@@ -709,6 +709,8 @@ function inject (bot, { hideErrors }) {
   bot._client.on('close_window', (packet) => {
     // close window
     const oldWindow = bot.currentWindow
+    if (!oldWindow) return
+    copyInventory(oldWindow)
     bot.currentWindow = null
     bot.emit('windowClose', oldWindow)
   })


Vinciepincie:
I encountered this issue too, not in a damaged container but a trade window plugin.

Is there a specific reason you are doing an early return instead of this?  It changes the emit behavior too, i dont think preventing the emit is necessary.

```
    if (oldWindow) copyInventory(oldWindow);
```

I accidentally made a duplicate and found your merge request, so i closed it. I think the description may still be useful.


---
https://github.com/PrismarineJS/mineflayer/pull/3942
**Issue:**
When the server closes a window (`close_window `packet), the handler discards the window without calling` copyInventory()`, so any changes in the player-inventory section of that window are lost and bot.inventory is left stale. The client-initiated path `bot.closeWindow()` already calls `copyInventory()` before discarding but if the server initiates close it doesn't.

**When this is needed:**
Any time the server force-closes a container the bot is using: shop/GUI plugins that close the window after a click, containers closed on teleport, anti-cheat closing windows, etc. After such a close, `bot.inventory.items() `reports the pre-window contents even though the bot's actual inventory changed while the window was open.

Fixed by using` copyInventory(oldWindow)` in the `close_window `packet handler, mirroring what `closeWindow()` does.

**Example:**
Bot puts items in a [SafeTrade ](https://www.spigotmc.org/resources/safetrade.13885/)plugin window and both players accept the trade. After the trade succeeds the items are gone from the bot's actual inventory, but logging bot.inventory still shows the old items until the bot opens and closes any random container. With this fix it gets synced properly.

TAOtxi:
@Vinciepincie There's no particular reason, I just referenced the following code:😁
```js
bot._client.on('login', () => {
    // close window when switch subserver
    const oldWindow = bot.currentWindow
    if (!oldWindow) return
    bot.currentWindow = null
    bot.emit('windowClose', oldWindow)
  })
```

PR #3937, opened 2026-07-01
fix: EntityAttach/entityDetach and bot dismount for set_passengers

### LINKED ISSUE #2819: entityAttach and entityDetach events not triggering post 1.8

arielvino:
Fixes #2819.

## Problem
Since 1.9, a vehicle's passenger changes arrive via the `set_passengers`
packet, which carries the vehicle's *complete* passenger list on every change.
Unlike 1.8's `attach_entity` it never sends a `-1`, and it is not resent when
the vehicle is destroyed. The existing handler didn't account for this:

- `dismount` for the bot was gated on an `entityId === -1` that never arrives,
  so it effectively never fired.
- `entityAttach` / `entityDetach` were never emitted for other entities.
- The passenger list was mutated by pushing/splicing individual entries rather
  than reconciled against the packet, so seat order (front seat = driver) and
  departures weren't tracked.

## Fix
Rewrite the `set_passengers` handler to diff the incoming list against the
vehicle's previous `vehicle.passengers` (which mineflayer already maintains):

- Entities in the old list but not the new one **left**: clear their
  `.vehicle` and emit `dismount` (bot) or `entityDetach` (others).
- Entities in the new list but not the old one **joined**: set their
  `.vehicle`, remove them from any prior vehicle whose own packet hasn't
  arrived yet, and emit `mount` (bot) or `entityAttach` (others).
- Commit the new list in server order, so `passengers[0]` is the front seat
  and a driver change is detectable.
- Bail out early if the vehicle entity is unknown.

Also emit `entityDetach` for remaining riders in the `entityGo`
passengers of a *destroyed* vehicle (which never generates a `set_passengers`) 
are detached too.

## Tests
Two in-process tests in `test/internalTest.js`, run across every tested
version except 1.8 (which uses the separate `attach_entity` pa

- mount then dismount of the bot via a re-sent passenger list
  old `-1` check missed).
- `entityAttach`/`entityDetach` for other entities, plus seat-
  a second rider boards, then the driver leaves and the remaining rider is
  promoted to the front seat.

A `vehicleTypeId(registry)` helper picks a mountable entity ty
in every tested version (entity names were CamelCase before the 1.11 flattening).

### DIFF
--- a/lib/plugins/entities.js
+++ b/lib/plugins/entities.js
@@ -771,31 +771,50 @@ function inject (bot) {
   })
 
   bot._client.on('set_passengers', ({ entityId, passengers }) => {
-    const passengerEntities = passengers.map((passengerId) => fetchEntity(passengerId))
+    // set_passengers carries a vehicle's *complete* passenger list whenever it
+    // changes. Unlike 1.8's attach_entity it never sends -1 and is not resent when
+    // the vehicle is destroyed (entityGone handles that). To recover who actually
+    // mounted or dismounted we diff the incoming list against the vehicle's previous
+    // passenger list, which mineflayer already keeps in `vehicle.passengers`.
     const vehicle = entityId === -1 ? null : bot.entities[entityId]
+    if (!vehicle) return // unknown vehicle: nothing to diff against
 
-    for (const passengerEntity of passengerEntities) {
-      const originalVehicle = passengerEntity.vehicle
-      if (originalVehicle) {
-        const index = originalVehicle.passengers.indexOf(passengerEntity)
-        originalVehicle.passengers.splice(index, 1)
-      }
-      passengerEntity.vehicle = vehicle
-      if (vehicle) {
-        vehicle.passengers.push(passengerEntity)
-      }
-    }
+    const newPassengers = passengers.map((passengerId) => fetchEntity(passengerId))
+    const oldPassengers = vehicle.passengers ?? []
 
-    if (passengers.includes(bot.entity.id)) {
-      const originalVehicle = bot.vehicle
-      if (entityId === -1) {
+    const left = oldPassengers.filter((entity) => !newPassengers.includes(entity))
+    const joined = newPassengers.filter((entity) => !oldPassengers.includes(entity))
+
+    for (const passenger of left) {
+      passenger.vehicle = null
+      if (passenger === bot.entity) {
         bot.vehicle = null
-        bot.emit('dismount', originalVehicle)
+        bot.emit('dismount', vehicle)
       } else {
-        bot.vehicle = bot.entities[entityId]
-        bot.emit('mount')
+        bot.emit('entityDetach', passenger, vehicle)
       }
     }
+
+    for (const passenger of joined) {
+      // If the passenger was riding a different vehicle, drop it from that one's
+      // list first: that vehicle's own set_passengers may not have arrived yet.
+      const previous = passenger.vehicle
+      if (previous && previous !== vehicle) {
+        const index = previous.passengers.indexOf(passenger)
+        if (index !== -1) previous.passengers.splice(index, 1)
+      }
+      passenger.vehicle = vehicle
+      if (passenger === bot.entity) {
+        bot.vehicle = vehicle
+        bot.emit('mount')
+      } else {
+        bot.emit('entityAttach', passenger, vehicle)
+      }
+    }
+
+    // Commit the new list in the server's order, so passengers[0] is the front seat
+    // (the one steering a boat) and a driver change is detectable by comparing index 0.
+    vehicle.passengers = newPassengers
   })
 
   // dismounting when the vehicle is gone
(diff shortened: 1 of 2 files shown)


arielvino:
--- edit ---
CI now passing without flakyness.

arielvino:
Hi, I'm new to mineflayer. My first contribution here.
Who do I need to ask for a review?
@extremeheat @rom1504

TAOtxi:
In fact, there are too few active administrators

PR #3947, opened 2026-07-19
Send horizontal collision state in movement packets

GenerelSchwerz:
## Summary

- populate the 1.21.3+ `hasHorizontalCollision` movement bit from the physics state
- send a status-only movement packet when horizontal collision changes
- preserve the collision bit in immediate and delayed server-position responses

## Rationale

Vanilla includes horizontal collision in every serverbound movement packet and sends a status-only packet when that state changes. Mineflayer currently serializes it as undefined, including when acknowledging a server position correction.

The value is gated to 1.21.3+, so legacy protocols do not gain extra collision-only movement packets. This improves vanilla packet fidelity; it does not change collision resolution or claim to resolve server-side wall-contact rejection by itself.

## Tests

- StandardJS lint
- focused internal physics test on 1.8.8, 1.21.3, and 1.21.11

### DIFF
--- a/lib/plugins/physics.js
+++ b/lib/plugins/physics.js
@@ -60,6 +60,11 @@ function inject (bot, { physicsEnabled, maxCatchupTicks }) {
   // see: https://discord.com/channels/413438066984747026/519952494768685086/901948718255833158
   let timeAccumulator = 0
   let catchupTicks = 0
+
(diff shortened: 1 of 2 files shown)


GenerelSchwerz:
Closed until I make a Minecraft data PR.

PR #3926, opened 2026-06-16
🎈 26.2

rom1504bot:
This automated PR sets up the relevant boilerplate for Minecraft version 26.2.

Ref: https://github.com/PrismarineJS/minecraft-data/pull/1198

* You can help contribute to this PR by opening a PR against this <code branch>pc26_2</code> branch instead of <code>master</code>.

PR #3834, opened 2026-02-14
Add Velocity proxy support for 1.20.2+ configuration phase

plainprince:
## Summary

Picks up #3786 by @StarWeizz and fixes the CI failures.

Adds a velocity plugin that handles the re-configuration phase during Velocity proxy server transfers (1.20.2+). Without this, the bot sends gameplay packets (movement, physics) during configuration, causing disconnections.

### What the plugin does
- Tracks when the bot re-enters configuration phase (server transfers)
- Disables physics during configuration to prevent movement packets
- Auto-accepts resource packs during configuration (required for transfer to complete)
- Re-enables physics after configuration finishes

### CI fix
The original PR's velocity plugin was activating during the **initial login** configuration phase (which node-minecraft-protocol already handles), causing:
- Duplicate `finish_configuration` acknowledgement packets
- Physics disabled during login, preventing spawn
- Timeouts on all 1.20.2+ test versions

Fixed by only activating after the bot's first spawn -- the initial login configuration is left entirely to node-minecraft-protocol.

### Other fixes
- Lint errors (missing newlines at EOF, object-shorthand)
- Removed redundant `registry_data`/`select_known_packs` listeners (minecraft-protocol handles these)

Fixes #3764
Supersedes #3786

## Test plan
- [ ] CI passes for all versions (1.8.8 through 1.21.8)
- [ ] Lint passes
- [ ] No regressions on existing tests
- [ ] Velocity proxy transfers work on 1.20.2+ (manual test)

Made with [Cursor](https://cursor.com)

PR #3963, opened 2026-08-14
fix(anvil): use 50-char name limit on 1.17+

### LINKED ISSUE #3945: Wrong anvil name length limit in 1.17+

## Versions
 - mineflayer: 4.37.1
 - server: vanilla/spigot/paper 1.17+
 - node: v25.8.2

In minecraft 1.17 the anvil name limit was increased to 50 while in the code it's 35.
This causes a problem with my bot which can potentially rename over 35 character long items

## Expected behavior
In 1.17+ it correctly asserts the name to be under 50
https://minecraft.fandom.com/wiki/Java_Edition_1.17#:~:text=Anvils,to%2050%20characters.

Xuxchloris:
Fixes #3945.

## Problem

Minecraft 1.17 raised the anvil custom-name limit from 35 to 50 characters, but `openAnvil` still rejected names longer than 35 characters on every version, breaking bots that need to rename items with names between 36 and 50 characters.

## Change

`lib/plugins/anvil.js`: compute the limit once per opened anvil from the `anvilNameLengthIsFifty` minecraft-data feature (`bot.supportFeature`):

- 1.17+ → 50 characters
- 1.16 and below → 35 characters (unchanged)

Applies to both `anvil.combine` and `anvil.rename`.

## Tests

`test/externalTests/anvil.js`: version-gated test that renames an item with a 40-character name (skipped on versions where the limit is still 35). The existing rename tests keep covering the short-name path on every version.

## Verification

- `standard` lint clean on both changed files.
- The new external test runs in the per-version CI matrix on 1.17+; on older versions it early-returns (35-char limit is preserved).

### DIFF
--- a/lib/plugins/anvil.js
+++ b/lib/plugins/anvil.js
@@ -15,6 +15,9 @@ function inject (bot) {
       throw new Error('Not a anvil-like window: ' + JSON.stringify(anvil))
     }
 
+    // 1.17+ raised the anvil custom-name limit from 35 to 50 characters
+    const nameLengthLimit = bot.supportFeature('anvilNameLengthIsFifty') ? 50 : 35
+
     function err (name) {
       anvil.close()
       throw new Error(name)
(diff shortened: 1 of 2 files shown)


Xuxchloris:
The failing "MC 1.12.2 1.13.2 1.14.4 1.15.2" job is a pre-existing flake unrelated to this PR:

- The new anvil test early-returns on versions without `anvilNameLengthIsFifty` (1.16 and below), so it does not execute in this job at all.
- The same job is failing on other open PRs (#3926, #3834) that touch completely different code paths.

Could you re-run it? If it still fails I'll dig into the logs.

PR #3969, opened 2026-08-17
fix(inventory): don't copy unconfirmed click predictions into inventory on window close

u9g:
## Background

Diagnosed from the packet trace uploaded by CI run [31992963697](https://github.com/PrismarineJS/mineflayer/actions/runs/31992963697) (same commit as failing master run [31992922399](https://github.com/PrismarineJS/mineflayer/actions/runs/31992922399)): the `creative` external test failed 4/4 attempts with `10 !== 9` because `ladder x3` was stuck in inventory slot 10.

What the trace shows (26.1):

1. `crafting` test crafts ladders at a crafting table, deposits them (`window_click` slot 11) and sends `close_window` 1ms later.
2. The server never echoes the click — **zero packets mentioning the crafted item after the deposit** (no `set_slot` for the mapped inventory slot, ever).
3. `closeWindow` → `copyInventory` copied the *client-predicted* window slot into `bot.inventory`.
4. `/clear` can't fix it: the item only exists client-side, so the server never sends a correction for that slot. The phantom survives every `resetState` and fails the count assertion in `creative`.

## Fix

Track the last **server-confirmed** value of each window slot (`_serverSlots`), updated from the authoritative paths:

- `set_slot` (`bot._setSlot`)
- `window_items` (both the immediate handler and the deferred replay at open)
- accepted transactions on pre-1.17.1 (`onAccepted` — there the prediction is only applied on server acceptance, so `window.slots` is server truth)

`copyInventory` at close now copies that confirmed view instead of `window.slots`, which may hold optimistic predictions the server dropped. Trailing syncs after the client-side close (the `lastClosedWindow` handling from #3854) still apply on top as usual.

## Vanilla parity (26.1 decompile)

- Vanilla predicts clicks locally too: `MultiPlayerGameMode#handleContainerInput` snapshots, applies `containerMenu.clicked(...)`, and sends the diff as `changedSlots` + `stateId` — same shape as mineflayer's `window.acceptClick` + `window_click`.
- Server truth arrives through the same three paths this PR tracks: `handleContainerSetSlot` → `setItem`, `handleContainerContent` → `initializeContents`, `handleSetPlayerInventory` → `inventory.setItem`.
- Vanilla's `closeContainer` does no copy-back at all — its window inventory-region slots wrap the *same* `Inventory` object, so dropped-click phantoms are transient and get overwritten by the next authoritative sync. `bot.inventory` is a separate model, so closing on the confirmed view lands directly on the state vanilla would converge to.

## Testing

- New internal test `does not copy unconfirmed click predictions into the inventory on window close`: opens a crafting window, picks an item up and places it one slot over with the server never echoing, closes, and asserts the server-confirmed placement survives with no phantom. Fails on master (item vanishes into the predicted slot), passes on all `stateIdUsed` versions (1.17.1+), skipped on transaction-era versions where prediction is only applied on acceptance.
- Full internal suite: 604 passing, 0 failing. `npm run lint` clean.

### DIFF
--- a/lib/plugins/inventory.js
+++ b/lib/plugins/inventory.js
@@ -422,10 +422,23 @@ function inject (bot, { hideErrors }) {
     bot.emit('windowClose', window)
   }
 
+  function setServerSlot (window, slot, item) {
+    // Track the last server-confirmed value of each slot. Optimistic click
+    // predictions update window.slots but not this copy, so a window closed
+    // before the server echoes the click still copies what the server last
+    // said — a dropped click (e.g. on a stale stateId) is never corrected
+    // by any packet, and copying the prediction leaves phantom items.
+    if (!window._serverSlots) window._serverSlots = new Array(window.slots.length).fill(null)
+    window._serverSlots[slot] = item
+  }
+
   function copyInventory (window) {
     const slotOffset = window.inventoryStart - bot.inventory.inventoryStart
+    // Prefer the server-confirmed view over window.slots, which may hold
+    // optimistic click predictions the server never echoed.
+    const slots = window._serverSlots ?? window.slots
     for (let i = window.inventoryStart; i < window.inventoryEnd; i++) {
-      const item = window.slots[i]
+      const item = slots[i]
       const slot = i - slotOffset
       if (item) {
         item.slot = slot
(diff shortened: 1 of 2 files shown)


u9g:
This is a bandaid fix around the fact that we are getting phantom items, we should fix the root cause.

extremeheat:
We could auto enable debug mode packet logging for when CI is reran in debug mode (check RUNNER_DEBUG env var), that'd help LLM parse through errors

Issue #3893, opened 2026-04-22
Add support for Minecraft Java Edition 26.1.2


EMIRKEREM12:
**Add support for Minecraft Java Edition 26.1.2**

Hello,

Currently, attempting to connect to a 26.1.2 server results in the following error:

```
Error: unsupported protocol version: 26.1.2.
```

Setting `version: '1.21.4'` or `version: '1.21.2'` in the bot options does not resolve the issue — the server rejects the connection entirely.

Thank you for all the work on this project!

extremeheat:
Correct, at the moment the latest supported mineflayer version is 1.21.11

osyra42:
If you are the owner of the server you can use ViaVersion and ViaBackwards, I have had some luck doing this

Zoooobee:
@osyra42 thanks a lot, It works for me §

Siigari:
We've submitted a PR to minecraft-data with protocol data for 26.1.2 (protocol 775): https://github.com/PrismarineJS/minecraft-data/pull/1184

The packet ID mappings were extracted from the official server jar using `--reports` flag. We've verified that a mineflayer bot can connect, complete login, pass through configuration, and enter the play state on a 26.1.2 server using this data.

Additional patches needed in other prismarine packages (prismarine-chunk, prismarine-physics) are version aliases — happy to submit those as separate PRs if the minecraft-data PR is merged.

walidkorchi:
> If you are the owner of the server you can use ViaVersion and ViaBackwards, I have had some luck doing this

The workaround stops at the `spawn` event, I tried it and indeed the bot joins the server, but when using `bot.simpleClick.leftMouse (slot)` which is an abstraction of `bot.clickWindow(slot, 0, 0)`, it just does not work after plenty of debugging. The packet size buffers simply do not match with what the servers expects I guess, and I feel like packets sent fail silently specifically from `packet_update_time`. I tried the 26.1.2 mappings by fetching the protocol from the PR sent to minecraft-data by @Siigari , and as expected, ran into unsupported version error from prismarine-chunk.

[...]

PR #3971, opened 2026-08-22
fix: don't crash on open_window with an unrecognised window type

liamnschermerhorn-code:
Opening an unrecognised window type kills the bot outright.

`windows.createWindow()` returns `null` when it does not recognise the window type and cannot synthesise a layout — it can only synthesise one when the packet supplies a slot count, and `open_window` no longer sends `slotCount`. So any unrecognised type takes that path, and `prepareWindow()` then dereferences the null:

```
TypeError: Cannot read properties of null (reading 'id')
    at prepareWindow (lib/plugins/inventory.js:690)
    at Client.<anonymous> (lib/plugins/inventory.js:710)
```

This is reachable on any server with a mod- or plugin-registered menu, so the bot dies rather than degrading.

### Reproduction

Against a NeoForge 26.2 server running Valkyrien Skies Eureka, sneak-right-clicking a ship helm sends:

```
open_window { windowId: 1, inventoryType: 74, windowTitle: { translate: "gui.vs_eureka.ship_helm" } }
```

which reliably terminates the bot.

### Change

Emit `unknownWindow` with the raw packet and return, instead of dereferencing the null. Consumers that care about custom menus can listen for it; everyone else simply keeps a live bot. Behaviour for all recognised window types is unchanged.

After the change, the same interaction yields:

```
[unknownWindow] type=74 title="gui.vs_eureka.ship_helm"
BOT ALIVE: yes
```

`npx standard` passes on the changed file.

### DIFF
--- a/lib/plugins/inventory.js
+++ b/lib/plugins/inventory.js
@@ -705,8 +705,16 @@ function inject (bot, { hideErrors }) {
   bot._client.on('open_window', (packet) => {
     // open window
     lastClosedWindow = null
-    bot.currentWindow = windows.createWindow(packet.windowId,
+    const window = windows.createWindow(packet.windowId,
       packet.inventoryType, packet.windowTitle, packet.slotCount)
+    if (!window) {
+      // createWindow() can only synthesise a layout for an unrecognised window type when
+      // the packet carries a slot count, and open_window stopped sending one. Any window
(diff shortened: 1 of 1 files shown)


extremeheat:
Does this also happen on 26.1 ? 26.2 is not supported at the moment, see #3893

PR #3974, opened 2026-08-22
Fix client-side inventory desyncs behind the 26.1 creative test failure

u9g:
Investigating the `creative` external test failing on 26.1 with `10 !== 9` at `creative.js:53` ([this run](https://github.com/PrismarineJS/mineflayer/actions/runs/31992963697) on #3968) turned up two client-side inventory desyncs. Replaying the uploaded packet trace showed the server held exactly 9 items at assert time — the 10th item existed only in mineflayer's state.

### 1. `set_player_inventory` slot ids were applied without mapping

The packet (1.21.2+) uses vanilla `Inventory` indices (0–8 hotbar, 9–35 main, 36–39 armor, 40 offhand), but the handler passed `packet.slotId` straight to `bot._setSlot`, which uses window-0 numbering (hotbar 36–44). The trace shows the server sending `set_player_inventory slotId=0` and `set_slot slot=36` with identical contents — mineflayer wrote the former into the crafting-result slot. Server-moved hotbar items (e.g. `/give`) became phantom items in slots the server never corrects.

### 2. `bot.craft` raced ahead of the server on the result pickup

This is what actually broke the creative test. `clickWindow` already waits for the server's recomputed result slot after every crafting-**grid** click, but not after clicking the **result** slot. The trace shows the result pickup, the deposit click, and `close_window` all sent within ~1 ms; the 26.1 server never applied the craft — it returned the 7 grid sticks on window close (and the between-test `/clear` removed 11 items: 7 sticks + 4 planks, no ladders) — while the client-side click simulation kept 3 phantom ladders in `bot.inventory.slots[10]`. Since `/clear` only corrects slots the server believes are occupied, the ghost survived every mocha retry, so `creative` (which runs right after `crafting`) always counted 9 + 1 = 10. Extending the existing result-slot wait to result clicks keeps the client in lockstep with the server.

### Verification

Against local vanilla servers, `crafting` + `creative`:
- **26.1**: both pass (previously `creative` failed every retry)
- **1.21.11** and **1.16.5** (regression check for both the client-predicted and transaction-ack click paths): both pass

---

**Update:** now includes a regression test (`test/externalTests/playerInventory.js`) covering the slot-mapping bug directly: pick an item onto the cursor in a chest, close the window, and assert the server's `set_player_inventory` return lands in inventory slot 36 (hotbar 0) and does **not** corrupt crafting-result slot 0. Verified on 1.8.8 (legacy return path) and 26.1.

Note: on 26.1 the test's chest interaction requires `player_loaded` (#3986, below this PR in the stack) — without it the server ignores the interaction indefinitely.

### DIFF
--- a/lib/plugins/craft.js
+++ b/lib/plugins/craft.js
@@ -21,6 +21,9 @@ function inject (bot) {
       }
 
       if (windowCraftingTable) {
+        // The last put-away click is unconfirmed; closing on a wrong model
+        // leaves items the server never placed.
+        await bot._syncWindow(windowCraftingTable)
         bot.closeWindow(windowCraftingTable)
         windowCraftingTable = undefined
       }
--- a/lib/plugins/creative.js
+++ b/lib/plugins/creative.js
@@ -55,11 +55,14 @@ function inject (bot) {
       })
     }
 
-    await onceWithCleanup(bot.inventory, `updateSlot:${slot}`, {
-      timeout: 5000,
-      checkCondition: (oldItem, newItem) => item === null ? newItem === null : newItem?.name === item.name && newItem?.count === item.count && newItem?.metadata === item.metadata
-    })
-    creativeSlotsUpdates[slot] = false
+    try {
+      await onceWithCleanup(bot.inventory, `updateSlot:${slot}`, {
+        timeout: 5000,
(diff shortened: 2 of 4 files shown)


rom1504:
CI fails

rom1504:
Lgtm let's wait for CI to pass

PR #3975, opened 2026-08-22
fix: don't let a failed external test poison the ones after it

u9g:
## Problem

[CI run 32586764293](https://github.com/PrismarineJS/mineflayer/actions/runs/32586764293) (1.16.5) failed two tests and killed the bot for the rest of the job. Three separate problems chained up:

- `examples/digger.js` crashed on an uncaught chunk-load timeout. Neither wait in `runExample` settles when the child dies, so the test hung until mocha's 90s timeout with its listeners still armed on the shared bot.
- The retry then ran on top of those listeners: the previous attempt's `run()` fired again on the new child's "Ready!", so every command went out twice and the example answered twice.
- The duplicate answer tripped `assert.fail('Unexpected message: ...')` inside a `checkCondition`, which runs inside the client's socket read path. The throw unwound it and the client stopped receiving packets entirely — the trace shows no S2C packet after that assertion, and the bot died 30s later on `keepAliveError`, failing everything downstream.

## Changes

- Catch a throwing `checkCondition` in `onceWithCleanup` and reject the promise instead of letting it escape into the emitter.
- Fail `runExample` as soon as the example process exits, instead of waiting for a mocha timeout that leaves listeners armed.
- Abort a still-running example before each attempt (via a new `AbortSignal` option on `onceWithCleanup`), so a retry never shares the bot with the attempt before it.

Both `onceWithCleanup` behaviors are covered by new internal tests.

### DIFF
--- a/lib/promise_utils.js
+++ b/lib/promise_utils.js
@@ -42,20 +42,44 @@ function createDoneTask () {
  * @param event - The name of the event you want to listen for
  * @param [timeout=0] - An amount, in milliseconds, for which to wait before considering the promise failed. <0 = none.
  * @param [checkCondition] - A function which matches the same signature of an event emitter handler, and should return something truthy if you want the event to be handled. If this is not provided, all events are handled.
+ * @param [signal] - An AbortSignal which, when aborted, removes the listener and rejects the promise with the signal's reason.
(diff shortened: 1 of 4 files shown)


rom1504:
CI fails

PR #3978, opened 2026-08-22
test(time): cut the time test from 13 broadcast waits to 3

u9g:
`time` costs ~12.5s on 1.8.8–1.21.3, ~5.4s on 1.21.4–1.21.11, and ~155ms on 26.1 — measured across two CI runs, n=56, very consistent within each tier. It is about 6.5% of the entire 28-version matrix.

### Why it is slow

From `MinecraftServer` in the 1.8.8 server jar:

```
121: getfield  y:I                  // tickCounter
124: bipush    20
126: irem                           // tickCounter % 20
127: ifne      190                  // not a multiple of 20 -> skip entirely
...
144: new       hu                   // S03PacketTimeUpdate
150: le.K()J                        // getTotalWorldTime()
155: le.L()J                        // getWorldTime()
163: ldc       "doDaylightCycle"
169: hu."<init>":(JJZ)V
180: lx.a(packet, dimensionId)      // broadcast
```

`/time set` only calls `World.setWorldTime` and sends nothing, so time reaches clients **only** via that 20-tick broadcast, and the interval is a hardcoded immediate — no gamerule or server property changes it. Newer versions echo commands immediately, which is exactly why 1.21.4+ and 26.1 are fast.

So on older servers every wait costs a full tick interval. The test had 13 of them. A packet trace on 1.8.8 showed **8346 of 8348ms spent waiting for broadcasts** — sending all the commands and running every assertion took 2ms.

### What this changes

The test now only pays for state it actually distinguishes. Dropped:

- `/time set 6000` and `/time set 0` — interior day values on the same path as the `12000` boundary, which is the case worth asserting
- the two dedicated `doDaylightCycle` waits — the broadcasts satisfying the night and day-progression waits already carry the flag, so those assertions ride along for free
- `/time set day|night` — the server resolving a keyword to a number, after which mineflayer's path is identical to a numeric `/time set`

Retained: timeOfDay parsing, both `isDay` branches including the 12000 boundary, both `doDaylightCycle` branches, and day/moonPhase progression. The type and range assertions at the top of the test never cost a wait and are untouched.

Each remaining wait is keyed on the state it verifies rather than on "any time event". That also removes a latent race: a bare `await once(bot, 'time')` resolves on whichever broadcast arrives first, and on servers that echo each command that can be the *gamerule's* echo rather than the one under test. I hit this while writing the patch — two batched commands behind a bare wait made 26.1 fail with `Expected day to be at least 1, got 0`.

`setDaylightCycle` is renamed to `sendSetDaylightCycleCommand`, since it sends and returns immediately and every caller owns its own wait.

### Measurements

Same machine, test duration only:

| version | before | after |
|---|---|---|
| 1.8.8 | 12737ms | **2769ms** |
| 26.1 | — | 876ms |

13 waits to 3. Roughly 10s on each of the 21 versions in the slow tier, about 3.5 minutes off the full matrix.

### Two things worth a maintainer's eye

**The `day`/`night` aliases.** Dropping them gives up integration coverage of the server keywords. It proves nothing about mineflayer, but it is a canary if a future server version changes what they resolve to. Happy to restore them at ~2s/version if you would rather keep that.

**26.1 at 876ms.** I could not A/B that on the machine I was testing on (26.1 needs more memory than it could reliably give), so I cannot separate machine overhead from a real cost on the fast-echo tier. The CI logs on this PR will show the true per-version numbers.

### DIFF
--- a/test/externalTests/time.js
+++ b/test/externalTests/time.js
@@ -48,8 +48,8 @@ module.exports = () => async (bot) => {
     }
   }
 
-  // Helper to set gamerule using the correct name for the version
-  const setDaylightCycle = (value) => {
+  // The gamerule is renamed on versions with gameRuleUsesResourceLocation.
+  const sendSetDaylightCycleCommand = (value) => {
     if (bot.supportFeature('gameRuleUsesResourceLocation')) {
       bot.test.sayEverywhere(`/gamerule minecraft:advance_time ${value}`)
     } else {
@@ -61,7 +61,7 @@ module.exports = () => async (bot) => {
   // time from drifting between /time set and the assertion
   const originalDaylightCycle = bot.time.doDaylightCycle
   // Commands apply in the order sent: the first /time set below lands after this.
-  setDaylightCycle(false)
+  sendSetDaylightCycleCommand(false)
 
   // Test time transitions
   const timeTests = [
@@ -82,7 +82,7 @@ module.exports = () => async (bot) => {
   // Must be read before the commands below mutate them.
   const currentDay = bot.time.day
   const currentPhase = bot.time.moonPhase
-  setDaylightCycle(true)
+  sendSetDaylightCycleCommand(true)
   bot.test.sayEverywhere('/time add 24000')
   // Two commands are in flight. A server that answers each one sends two
   // broadcasts, and the first still has the old day, so wait on the day itself.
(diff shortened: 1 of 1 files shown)


rom1504:
Ok

PR #3979, opened 2026-08-22
test(useChests): assert the click transitions instead of clicking randomly; fix rejected quick-move clicks on 1.12–1.16

u9g:
Stacked on #3974 (`fix-26.1-crafting-desync`); the base retargets to `master` once that merges.

`useChests` made **366 window clicks**, and roughly 350 of them asserted nothing.

### What the clicks were doing

| source | clicks | asserted |
|---|---|---|
| `createRandomLayout(0.95)` — ~51 items x `moveSlotItem` | ~102–120 | nothing; pure setup |
| `testMouseClick(250)` | 250 | nothing |
| deposit/withdraw x 8 | ~14 | deposit/withdraw across chest vs trapped, single vs double |

The fuzz loop was:

```js
await bot.clickWindow(~~(Math.random() * window.inventoryStart), 0, 0)
```

Every iteration was `mouseButton = 0`, `mode = 0`, varying only the slot, and the loop had no assertions — it failed only if `clickWindow` threw. That throw lives inside `if (bot.supportFeature('transactionPacketExists'))`, so from 1.17 onward there is no accept/reject to throw on at all, only `waitForWindowUpdate`. On 19 of the 28 tested versions it verified nothing beyond "did not time out".

The random layout existed only to give that loop something to click on.

### What replaces them

A left or right click resolves differently depending on whether the cursor is holding something and what is in the target slot, so a random click only ever reaches whichever case the current state happens to allow — repeating it does not change that. `testClickTransitions` drives each case from its own precondition and asserts the resulting slot and cursor contents:

| case | asserts |
|---|---|
| empty cursor + occupied slot | cursor takes the type and count, slot emptied |
| holding + empty slot, right click | exactly one lands, cursor keeps the rest |
| holding + matching stack | slot returns to the full count, cursor emptied |
| holding + different type | contents exchanged both ways |
| holding + empty slot | whole stack dropped |
| empty cursor + stack, right click | the halves account for the whole stack |
| shift click | chest slot emptied |

That covers right click and shift-click quick move — neither of which the old mode-0 loop reached.

The layout is now three fixed items, which is what those preconditions need. They are written server side (`/item replace block` on 1.17+, `/replaceitem block` on 1.13+, `slot.container` before that) and asserted once the window opens, so a wrong command fails loudly instead of silently under-filling.

### What the new assertions surfaced

Shrinking the layout and asserting on it exposed three things the old test could not see, each fixed in its own commit:

1. **Filling by clicking lost moves on 26.1.** `/give` + `moveSlotItem` depends on the client's predicted state, and from 1.17 a click is not confirmed per transaction. On 26.1 two of the three items never arrived while the client reported all three placed; with ~51 items that margin hid it. Hence the server-side fill above.
2. **Which half of a double chest the window lists first depends on the version.** On 1.8.8 it is the lower-coordinate block; on 1.13+ it follows the chest's facing, so the filled block's slots land at 27–29 instead of 0–2. The fill assertion now searches the container instead of assuming slots 0–2.
3. **mineflayer's shift click was rejected by every 1.12–1.16 vanilla server.** For `actionIdUsed` versions the server accepts a click only if the packet's `item` matches the result of its own `slotClick`; for a quick move that is the moved stack up to 1.11 and empty from 1.12, and mineflayer always sent the clicked stack. The old loop never sent mode 1, so nothing noticed. `lib/plugins/inventory.js` now sends an empty item for mode 1 from 1.12 on. (An earlier revision of this PR drew the line at 1.9; CI on 1.9.4–1.11.2 rejected that — those servers want the stack — and the harness retries turned the error into `Event close did not fire`.)
4. **Most of the test was the bot turning its head.** `openContainer` and `placeBlock` face the block before acting, and a non-forced look turns at 3 rad/s, so each 180° swing between the small chest behind the bot and the large chest in front costs ~1.05s. On 1.8.8 that was ~8.1s of the 13.4s — ten `openContainer` calls averaging 0.8s of rotation around ~100ms of server round trip. Nothing in the test asserts on looking and turn speed is not tested anywhere (the external tests that look pass `force=true`), so the test now sets `bot.physics.yawSpeed`/`pitchSpeed` high enough that a look resolves in one physics tick.

### Measurements

Test duration only:

| version | before | click rewrite | + instant look |
|---|---|---|---|
| 1.8.8 | 38323ms | 14746ms | **3596ms** |
| 1.9.4 | | | 3499ms |
| 1.11.2 | | | 3414ms |
| 1.12.2 | | | 3418ms |
| 1.16.5 | | | 1589ms |
| 26.1 | ~13.5s (CI) | 14376ms | CI |

366 clicks to 33 on 1.8.8.

Where the 14746ms went on 1.8.8, from a `mark()` after every awaited step: `placeBlock` ×6 3.3s, chest-lid test 1.1s (`wait(500)` + one open), deposit/withdraw ×8 8.1s — of which 7.2s was the eight `openContainer` turns and 0.9s the actual clicks — fill + open 0.3s, the nine asserted clicks 0.45s at one transaction tick each. The click saving is concentrated on old versions because each `clickWindow` waits for a `transaction` that arrives on the next tick (median gap 50ms up to 1.16, 0ms from 1.17); the turning cost is the physics tick and so the same on every version.

### Verification

Run against real servers locally, all passing:

| version | covers |
|---|---|
| 1.8.8 | `slot.container` fill, quick move acknowledged with the stack |
| 1.9.4, 1.11.2 | `slot.container` fill, quick move acknowledged with the stack (1.9–1.11) |
| 1.12.2 | `slot.container` fill, quick move acknowledged empty |
| 1.16.5 | `/replaceitem block` fill |
| 26.1 | `/item replace block` fill, no transaction acknowledgement — passed locally before the instant-look commit; the box used (8 GB) can no longer keep the 26.1 server ahead of its ticks, so CI is the check on it |

CI covers the remaining tiers.

### DIFF
--- a/lib/plugins/inventory.js
+++ b/lib/plugins/inventory.js
@@ -624,7 +624,7 @@ function inject (bot, { hideErrors }) {
         action: actionId,
         mode,
         // Must match what the server's own click handling returns.
-        item: Item.toNotch((mode === 2 || mode === 4 || (mode === 1 && bot.registry.version['>=']('1.12'))) ? null : click.item)
+        item: Item.toNotch((mode === 2 || mode === 4 || (mode === 1 && bot.supportFeature('quickMoveClickSendsEmptyItem'))) ? null : click.item)
       })
     } else { // 1.17
       bot._client.write('window_click', {
--- a/lib/plugins/inventory.js
+++ b/lib/plugins/inventory.js
@@ -623,8 +623,8 @@ function inject (bot, { hideErrors }) {
         mouseButton,
         action: actionId,
         mode,
-        // protocol expects null even if there is an item at the slot in mode 2 and 4
-        item: Item.toNotch((mode === 2 || mode === 4) ? null : click.item)
+        // Must match what the server's own click handling returns.
+        item: Item.toNotch((mode === 2 || mode === 4 || (mode === 1 && bot.registry.version['>=']('1.12'))) ? null : click.item)
       })
     } else { // 1.17
       bot._client.write('window_click', {
--- a/test/externalTests/useChests.js
+++ b/test/externalTests/useChests.js
@@ -133,19 +133,15 @@ module.exports = () => async (bot) => {
     1: ['fishing_rod', 'bow']
   }
 
-  function getRandomStackableItem () {
-    if (Math.random() < 0.75) {
-      return itemsWithStackSize[64][~~(Math.random() * itemsWithStackSize[64].length)]
-    } else {
-      if (Math.random() < 0.5) {
-        return itemsWithStackSize[16][~~(Math.random() * itemsWithStackSize[16].length)]
-      } else {
-        return itemsWithStackSize[1][~~(Math.random() * itemsWithStackSize[1].length)]
-      }
-    }
-  }
+  // The transition cases need one stack of more than 2, a second stack of a
+  // different type, and a free slot. Nothing asserts on a fuller chest.
+  const layout = [
+    { slot: 0, name: itemsWithStackSize[64][0], count: 8 },
+    { slot: 1, name: itemsWithStackSize[64][1], count: 8 },
+    { slot: 2, name: itemsWithStackSize[16][0], count: 4 }
+  ]
 
-  async function createRandomLayout (window, slotPopulationFactor) {
+  async function createLayout (window) {
     await bot.test.becomeCreative()
 
     // Wait for the given item to show up in hotbar.0. Depending on version and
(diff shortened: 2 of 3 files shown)


lib/plugins/inventory.js
```diff
@@ -626,8 +626,8 @@ function inject (bot, { hideErrors }) {
         mouseButton,
         action: actionId,
         mode,
-        // protocol expects null even if there is an item at the slot in mode 2 and 4
-        item: Item.toNotch((mode === 2 || mode === 4) ? null : click.item)
+        // Must match what the server's own click handling returns.
+        item: Item.toNotch((mode === 2 || mode === 4 || (mode === 1 && bot.registry.version['>=']('1.12'))) ? null : click.item)
```
rom1504: Let's use feature support

test/externalTests/useChests.js
```diff
@@ -133,93 +135,120 @@ module.exports = () => async (bot) => {
...
+
+  // Write the slots server side. Filling them by clicking leaves the result at
+  // the mercy of the client's predicted state, which from 1.17 is not confirmed
+  // per click, and a lost move is invisible until an assertion reads the slot.
+  function fillChest (pos) {
+    const at = `${pos.x} ${pos.y} ${pos.z}`
+    for (const { slot, name, count } of layout) {
+      const item = bot.registry.itemsByName[name]
+      assert.ok(item, `${name} should exist on this version`)
+      if (bot.registry.version['>=']('1.17')) {
```
rom1504: Feature support

lib/plugins/inventory.js
rom1504: Still

u9g:
Switched both checks to `bot.supportFeature` in b87fe1af. The three features (`quickMoveClickSendsEmptyItem`, `hasItemCommand`, `replaceItemSlotIsPrefixed`) are in PrismarineJS/minecraft-data#1262; this needs that released before CI here is green again, since an unknown feature resolves to false.

PR #3983, opened 2026-08-23
test(trade): two trades per offer, and shift click in putAway

u9g:
Carries #3979's 2-line `inventory.js` fix as its first commit (identical patch, so git drops it when either lands): the shift click below needs the right mode-1 `item` on 1.12–1.16, gated on minecraft-data 3.115.0's `quickMoveClickSendsEmptyItem` feature, which master's `^3.114.0` already resolves to.

`trade` took ~32s on every version up to 1.13 and ~1.7s from 1.14. A packet trace on 1.8.8 put 21.3s of the 23.2s at exactly **one transaction tick per `clickWindow`** — 426 clicks × 50ms — and 1.9s in setup; there are no other waits in it. From 1.14 the server fills the input slots on `select_trade`, so the click loop never runs there.

Three changes, one commit each, after the carried fix:

**`putAway`: shift click** instead of pick-up-and-drop — the same move in one click, for every caller (furnace, enchantment table, anvil, the crafting-grid cleanup, `disrobe`, villager's input leftovers). Two carve-outs keep the old path: result slots, where a shift click repeats the craft or trade while the inputs cover it (`craftingResultSlot`, merchant slot 2), and a full inventory, where a shift click does nothing and the old path tosses the stack. Sending the right `item` for a mode-1 click on 1.12–1.16 is #3979's `inventory.js` fix, carried here as the first commit.

**`villager`: shift click the output** on <1.14, where `trade` deposits exactly the price per iteration so the repeat cannot happen; from 1.14 the server pre-fills whole stacks and it stays on `putAway`. This is the one place that knows the inputs hold a single trade, which is why it is not folded into `putAway`.

**`test(trade)`: two trades per offer instead of eleven.** The test checks `nbTradeUses`, `maximumNbTradeUses`, the item arithmetic after trading, and that an exhausted offer is refused; none of that needs eleven rounds. `maxUses: 3` with one use already spent gives two, and the refusal at the end still fires. Two emerald stacks cover the 86 needed, so eight of the ten `setInventorySlot` calls go too (each is 400ms on 1.21.3+).

### Measurements

1.8.8, test duration only, from the traces:

| | clicks | duration |
|---|---:|---:|
| before | 604 | 32.2s |
| after | 101 | 7.1s |

Where the 6.6s traced goes now: 101 clicks 5.05s (trade #4 alone 67 — its 36-emerald deposit is still one right click per item; #3982 halves that independently), command block `wait(500)` ×2 1.04s, redstone → `entitySpawn` 0.35s, `setInventorySlot` ×3 0.14s.

### Verification

`trade` on 1.8.8 (7113ms), 1.9.4 (6856ms), 1.12.2 (7173ms), 1.16.5 (1909ms), 1.21.1 (1690ms) — the stack-ack tier, the empty-ack tier, the 1.14–1.16 `putAway` tier, and the no-ack tier. With the `putAway` change: full external suite on 1.8.8 and 1.21.1 (49/49 each), `trade`/`furnace`/`crafting` on 1.12.2, internal tests.

After rebasing on master: `trade` on 1.8.8, 1.12.2, 1.16.5, 1.21.4; `furnace` and `crafting` on 1.12.2.

### DIFF
--- a/lib/plugins/inventory.js
+++ b/lib/plugins/inventory.js
@@ -657,10 +657,19 @@ function inject (bot, { hideErrors }) {
   async function putAway (slot) {
     const window = bot.currentWindow || bot.inventory
     const promisePutAway = once(window, `updateSlot:${slot}`)
-    await clickWindow(slot, 0, 0)
     const start = window.inventoryStart
     const end = window.inventoryEnd
-    await putSelectedItemRange(start, end, window, null)
+    const item = window.slots[slot]
+    // A shift click on a result slot repeats the craft or trade while the
+    // inputs cover it, and does nothing when the inventory has no room.
+    const resultSlot = slot === window.craftingResultSlot || (window.type === 'minecraft:merchant' && slot === 2)
+    const room = item && (window.findItemRange(start, end, item.type, item.metadata, true, item.nbt) || window.firstEmptySlotRange(start, end) !== null)
+    if (room && !resultSlot) {
+      await clickWindow(slot, 0, 1)
+    } else {
+      await clickWindow(slot, 0, 0)
+      await putSelectedItemRange(start, end, window, null)
+    }
     await promisePutAway
   }
 
--- a/lib/plugins/inventory.js
+++ b/lib/plugins/inventory.js
@@ -623,8 +623,8 @@ function inject (bot, { hideErrors }) {
         mouseButton,
         action: actionId,
         mode,
-        // protocol expects null even if there is an item at the slot in mode 2 and 4
-        item: Item.toNotch((mode === 2 || mode === 4) ? null : click.item)
+        // Must match what the server's own click handling returns.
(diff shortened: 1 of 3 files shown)


rom1504:
Approved

.github/workflows/ci.yml
```diff
@@ -66,6 +66,18 @@ jobs:
...
       - name: Install Dependencies
         run: npm install
+      # Unreleased upstream fixes the external tests depend on:
+      # - minecraft-wrap 1.9.0 persists the version manifest to a path shared by
+      #   the concurrent per-version test processes, which tear each other's
+      #   writes (PrismarineJS/node-minecraft-wrap#111).
+      # - minecraft-protocol's ping never rejects when the server closes the
+      #   status connection, so the retry in externalTest.js waits out the
+      #   whole closeTimeout per attempt (PrismarineJS/node-minecraft-protocol#1514).
+      - name: Install unreleased minecraft-wrap and minecraft-protocol fixes
```
rom1504: Remove

PR #3987, opened 2026-08-23
test: always write the packet trace, defaulting to a timestamped file in tmpdir

u9g:
Stacked on #3973.

The TRACE jsonl was opt-in, so a flaky local run left nothing to look at unless you remembered to set `TRACE` beforehand — the same gap CI closed by always setting it. Recording is cheap (<1% runtime), so default it on:

- `TRACE=<file>` still overrides the path; otherwise the trace goes to `<tmpdir>/mineflayer-trace-<ISO timestamp>-<pid>.jsonl`. The pid suffix keeps concurrent mocha processes (one per version in CI) apart.
- The path is printed (`trace: /tmp/...`) when the stream first opens, i.e. right under the `mineflayer_external <v>` title.
- With no disabled state the `if (process.env.TRACE)` guards go away; `bot.test.dumpMarker` is now always defined.

CI is unchanged — it still sets `TRACE` explicitly so the artifact copy keeps working.

### DIFF
--- a/lib/plugins/health.js
+++ b/lib/plugins/health.js
@@ -8,9 +8,16 @@ function inject (bot, options) {
     bot.emit('respawn')
   })
 
+  // 1.21.4+ servers ignore block and item interactions until the client has
+  // sent player_loaded (or 60 ticks have passed)
+  function spawn () {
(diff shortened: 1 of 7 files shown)


rom1504:
CI fails

PR #3989, opened 2026-08-23
test(fishing): give the rod Lure III to cut the bite wait

u9g:
The fishing test is the slowest test in the external suite: it casts an unenchanted rod and waits out the full vanilla bite delay of 5–30 seconds. Lure III takes 15 seconds off that delay, so the wait becomes 0–15 seconds.

Measured on 26.1: 33s before; 13s, 14.5s, 19s across three runs after.

`/enchant` on the held rod works on every tested version (1.8–26.1), unlike prismarine-item's `enchants` setter, which writes NBT that `set_creative_slot` no longer carries on 1.20.5+ (components). `awaitItemReceived` waits for the server's `set_slot` echo so the rod is enchanted before the cast.


rom1504:
CI fails

PR #3992, opened 2026-08-24
Poll for the sign text echo in the nether test instead of sleeping 500ms

u9g:
After `bot.updateSign`, the nether test slept a fixed 500ms before asserting the sign text. That's a lose-lose: the server usually echoes the new text within a tick or two (so ~450ms is wasted on every version in the matrix), and on a slow CI box 500ms may not be enough and the assert fails.

This polls `blockAt(...).signText` every 50ms until it matches, capped at 5s. There's no event to await instead: the `update_sign` and `tile_entity_data` handlers write via `column.setBlock`/`setBlockEntity`, which bypasses the world's `blockUpdate` emit — #3993 (stacked on this) adds the missing events, after which this loop becomes a plain await.

It also flattens the test into a linear async flow with no raw `bot._client` listener: by the time `placeBlock`'s block update echo has arrived the server has already opened the sign editor, so the text can be sent right away — and assertion failures now reject the test naturally instead of throwing inside a callback (previously they surfaced as uncaught exceptions from a `setTimeout`).

The poll condition uses `signText?.trimEnd()` because before the echo arrives `signText` is `undefined` — a state the old fixed sleep always slept through.

Verified locally by running the nether external test against 26.1 and 1.8.8; both pass. Measured on 1.8.8: 6655ms → 6043ms.

### DIFF
--- a/test/externalTests/nether.js
+++ b/test/externalTests/nether.js
@@ -13,31 +13,6 @@ module.exports = () => async (bot) => {
   }
   assert.notStrictEqual(signItem, null)
 
-  const p = new Promise((resolve, reject) => {
-    bot._client.once('open_sign_entity', (packet) => {
-      console.log('Open sign', packet)
-      const sign = bot.blockAt(new Vec3(packet.location))
-      bot.updateSign(sign, '1\n2\n3\n')
-
-      setTimeout(() => {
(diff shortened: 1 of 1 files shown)


rom1504:
CI fails

PR #3993, opened 2026-08-24
Emit blockEntityData and signOpen events and use them in the nether test

u9g:
Stacked on #3992.

The server's block entity updates (modern `block_entity_data` and legacy `update_sign`) were stored in the world silently, and `open_sign_entity` was not surfaced at all — so consumers had to poll `bot.blockAt` or reach into `bot._client`.

- **Commit 1** adds two public events, documented in api.md + index.d.ts:
  - `blockEntityData` (block) — emitted with the updated block whenever block entity data arrives
  - `signOpen` (block) — emitted with the placed sign when the server opens the sign editor
- **Commit 2** rewrites the nether test on top of them: the sign-echo polling loop from #3992 becomes a wait for `blockEntityData`, and the placed sign is taken from `signOpen` instead of being read out of the world by position.

Verified locally by running the nether external test against 26.1 (modern packet path) and 1.8.8 (legacy `update_sign` path); both pass.


rom1504:
CI fails

PR #3996, opened 2026-08-24
test: confirm superflat fills via chat echo instead of sleeping 100ms

u9g:
The six `/fill` repairs in `resetBlocksToSuperflat` are fire-and-forget chat commands, synchronized only by a fixed `wait(100)` — an assumption that a slow CI tick can exceed (and that fast ticks waste).

A marker chat message sent on the same ordered connection is only echoed back after the fills have executed server-side, and the fills' `multi_block_change` packets sit earlier in the same ordered stream — so by the time the echo arrives, the client's world is provably up to date.

Measured on 1.8.8 (instrumented reset, full suite): the fills step went from ~102ms avg to ~47ms avg (min 4ms, max 66ms), all tests passing identically. Combined with #3995 this puts the whole pre-test reset around ~3.5s per version, mostly the /clear-confirm roundtrips.

### DIFF
--- a/lib/plugins/chat.js
+++ b/lib/plugins/chat.js
@@ -142,6 +142,25 @@ function inject (bot, options) {
     if (data.positionId === 2) bot.emit('actionBar', msg, null)
   })
 
+  let send = (message) => bot._client.chat(message)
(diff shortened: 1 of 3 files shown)


rom1504:
CI fails

PR #3997, opened 2026-08-24
test(fishing): use the maximum viable Lure level per version

u9g:
Stacked on #3989.

From reading the decompiled server code: the bite wait is rolled as `nextInt(100, 900)` on 1.8.x and `nextInt(100, 600)` on 1.9+ (verified in 1.16.5/1.21.8/26.1 sources), then reduced by Lure level × 100 ticks with **no clamp** — a non-positive roll simply rerolls next tick. That means:

- **Lure 8 on 1.8.x / Lure 5 on 1.9+** are the maximum viable levels: only rolls in the top 100 ticks survive, so the wait becomes 1–100 ticks — expected ~2.5s, **hard cap 5s** (vs 5–30s vanilla, ~15s expected under Lure III).
- One level higher and the roll can never go positive: **the hook never bites** and the test hangs — which is why the level is version-conditional and why `/enchant` (capped at III) is replaced with setting the enchant via item NBT (prismarine-item handles the per-version encoding).

The timers themselves can't be fast-forwarded directly: the 1.8.8 bobber saves no timer fields to NBT and the modern `FishingHook.addAdditionalSaveData` is empty, so `/entitydata` / `/data merge` have nothing to touch; `/tick sprint` would also sprint past the 10–30-tick catchable window.

Measured (fishing test, multiple runs):
- 1.8.8: **7.4s / 7.4s** (was 6.9–22.1s across runs tonight)
- 1.16.5: **7.3s**
- 26.1: 9.2–17.9s — modern runs still sometimes miss the first nibble and pay a second (now-capped) cycle; that reel-timing issue predates this change and is a separate fix.

### DIFF
--- a/test/externalTests/fishing.js
+++ b/test/externalTests/fishing.js
@@ -3,24 +3,23 @@ module.exports = () => async (bot) => {
 
   bot.test.sayEverywhere('/fill ~-10 ~-1 ~-10 ~10 ~-1 ~10 water')
   bot.test.sayEverywhere('/weather rain') // rain shortens the vanilla bite wait
-  // The server rolls the bite wait as nextInt(100, 900) on 1.8.x and
-  // nextInt(100, 600) on 1.9+, then subtracts Lure level x100 ticks; a
-  // non-positive roll just rerolls next tick. These levels are the maximum
-  // viable: they leave only 1-100 tick waits (expected ~2.5s, hard cap 5s).
-  // One level higher and the roll can never go positive - the hook never
-  // bites. /enchant caps at Lure III, so the level is set via item NBT.
-  // From 1.20.5 the servers silently strip enchantments written by the
-  // client into a creative slot, so the rod must come from a server /give.
-  // The enchantments component lost its levels wrapper in 1.21.5.
-  if (bot.registry.version['>=']('1.21.5')) {
+  // The server rolls the bite wait as nextInt(100, fishingBiteDelayMaxTicks)
+  // then subtracts Lure level x100 ticks; a non-positive roll just rerolls
+  // next tick. This Lure level is the maximum viable: it leaves only 1-100
+  // tick waits (expected ~2.5s, hard cap 5s). One level higher and the roll
+  // can never go positive - the hook never bites. /enchant caps at Lure III.
+  // Component-era servers silently strip enchantments written by the client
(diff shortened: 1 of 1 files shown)


u9g:
Follow-up commit: the modern-version variance turned out not to be bite detection at all. Packet capture showed exactly one bite per run, detected instantly — the time went into the wait itself, because **1.20.5+ servers silently strip enchantments written by the client into a creative slot**: the Lure 5 rod was actually unenchanted on those versions. The rod now comes from a server `/give` with the enchantments component (`levels` wrapper for 1.20.5–1.21.4, flattened from 1.21.5).

Measured after: 1.8.8 4.6s · 1.20.6 3.1s · 1.21.4 4.4s · 1.21.8 5.6s · 26.1 4.0s — fishing is now ~3–6s and bounded on every era.

rom1504:
CI fails

test/externalTests/fishing.js
```diff
@@ -3,8 +3,26 @@ module.exports = () => async (bot) => {
...
+  // The server rolls the bite wait as nextInt(100, 900) on 1.8.x and
+  // nextInt(100, 600) on 1.9+, then subtracts Lure level x100 ticks; a
+  // non-positive roll just rerolls next tick. These levels are the maximum
+  // viable: they leave only 1-100 tick waits (expected ~2.5s, hard cap 5s).
+  // One level higher and the roll can never go positive - the hook never
+  // bites. /enchant caps at Lure III, so the level is set via item NBT.
+  // From 1.20.5 the servers silently strip enchantments written by the
+  // client into a creative slot, so the rod must come from a server /give.
+  // The enchantments component lost its levels wrapper in 1.21.5.
+  if (bot.registry.version['>=']('1.21.5')) {
```
rom1504: Feature version

rom1504: Still

u9g:
Switched to features: `itemsWithComponents` already existed; `enchantmentsComponentIsFlat` and `fishingBiteDelayMaxTicks` are added in PrismarineJS/minecraft-data#1261 (the Lure level now derives from the bite delay bound rather than being hardcoded per era). CI here needs that release + bump before it goes green.

The earlier CI failure was the 26.1 nether `blockUpdate:(0, 128, 0)` flake, unrelated to this PR — rerun.

PR #3999, opened 2026-08-24
test(exampleBee): await the op confirmation instead of sleeping 2s

u9g:
Stacked on #3984, same pattern: the fixed 2s sleep after `/op bee` was waiting for the op — and with it the spawn-protection bypass — to apply. The server confirms exactly that with a chat message (`Opped bee` / `Made bee a server operator`), so await it instead.

Measured (instrumented runs):

| | before | after |
|---|---|---|
| op wait | 2000ms fixed | 50ms (1.8.8) / 11ms (26.1) |
| exampleBee total | 5.8s | 3.85s (1.8.8) / 4.3s (26.1) |

What remains is ~2.9s of the actual flight (11 `flyTo` hops — the behavior being demonstrated) plus ~0.9s of child-bot setup.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,11 +66,18 @@ jobs:
           java-package: jre
       - name: Install Dependencies
         run: npm install
-      # minecraft-wrap 1.9.0 persists the version manifest to a path shared by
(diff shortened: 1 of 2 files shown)


rom1504:
CI fails

PR #4004, opened 2026-08-27
ci: cache server jars

u9g:
Each CI job currently spends its first stretch downloading server jars, and — worse — `download()` in the external tests fetches Mojang's launcher manifest **before** noticing the jar is already on disk. Those Mojang round-trips are the flake behind runs like [this one](https://github.com/PrismarineJS/mineflayer/actions/runs/32682944608), where 6 of 8 MC jobs died in the `before all` hook with truncated/corrupt JSON (`SyntaxError: Bad control character in string literal` / `Unterminated string`) from `minecraft-wrap`'s launcher download — mocha `--retries` can't save a failed hook.

- **Cache `server_jars/`** per matrix group (`actions/cache@v4`, keyed by the group's version list — jars are immutable per version), and skip `download()` entirely when the jar exists. On a cache hit, test startup makes no Mojang requests at all, so the download flake is off the hot path and the download time is gone.

Verified locally: with the jar present, the 1.8.8 external suite boots the server and runs with no network access to Mojang. The download path (cache miss) is untouched and still sha1-validates what it fetches. A restored jar is trusted as-is: GitHub cache scoping only lets a run restore caches written by its own ref, its base branch, or the default branch, so every possible writer of a restorable entry could already run arbitrary code in the reading context — re-hashing the jar would add no boundary.

First run on this PR populates the caches; subsequent runs get the benefit.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,6 +66,16 @@ jobs:
           java-package: jre
       - name: Install Dependencies
         run: npm install
+      # server jars are immutable per version, so the group's version list is
+      # the whole key; a hit lets the tests start without touching Mojang
+      - name: Compute server jar cache key
+        id: jar-cache-key
+        run: echo "key=server-jars-$(echo '${{ matrix.versions }}' | tr ' ' '_')" >> "$GITHUB_OUTPUT"
+      - name: Cache server jars
+        uses: actions/cache@v4
+        with:
+          path: server_jars
+          key: ${{ steps.jar-cache-key.outputs.key }}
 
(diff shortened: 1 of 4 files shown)


rom1504:
So instead of testing the cache provided by Mojang that people use while using mineflayer we test the cache provided by GitHub which nobody uses except our tests

I am not sure that's a win

u9g:
> So instead of testing the cache provided by Mojang that people use while using mineflayer we test the cache provided by GitHub which nobody uses except our tests
> 
> I am not sure that's a win

Good point

PR #4006, opened 2026-08-27
ci: install unreleased minecraft-wrap and minecraft-protocol fixes

u9g:
Two upstream fixes the external tests depend on are not released yet; install them from their commits in CI until they are, then revert this.

**minecraft-wrap** 1.9.0 writes `version_manifest.json` to `server_jars/versions/`, which the four per-version test processes in each CI job share, and reads it straight back. They tear each other's writes, so every job has been failing in the `before all` hook with `SyntaxError ... JSON` at `launcher_download.js:42`. PrismarineJS/node-minecraft-wrap#111 keeps the manifest in memory and writes downloads to a temp file + rename.

**minecraft-protocol**'s `ping` never rejects when the server closes the status connection before answering, so each failed attempt of the ping retry in `externalTest.js` (#3976) waits out the full `closeTimeout`. PrismarineJS/node-minecraft-protocol#1514 rejects on close.

The other open PRs are rebased onto this branch so their CI reflects only their own changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,11 +66,18 @@ jobs:
           java-package: jre
       - name: Install Dependencies
         run: npm install
-      # minecraft-wrap 1.9.0 persists the version manifest to a path shared by
-      # the concurrent per-version test processes, which tear each other's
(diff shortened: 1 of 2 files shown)


rom1504:
Not merging until deps are released

u9g:
This was just a temp branch

PR #4007, opened 2026-08-27
chat: on 1.19.0, wait for queued commands before sending a chat message

u9g:
Stacked on #3996, which fails deterministically on 1.19 (and only 1.19): the bot is kicked with `multiplayer.disconnect.out_of_order_chat` on the very first reset, and every test after that fails with `Event move did not fire`.

Packet trace from the failed job:

```
C2S chat_command  fill … -56 … air        timestamp …753
C2S chat_command  fill … (×5)             timestamp …776
C2S chat_message  superflat-reset-done    timestamp …776
S2C kick_disconnect  multiplayer.disconnect.out_of_order_chat
```
Server log: `flatbot sent out-of-order chat: 'fill ~-5 -56 ~-5 ~5 -56 ~5 air'` — it rejected the *first* fill, the only packet stamped earlier than the marker.

In 1.19.0 `ServerGamePacketListenerImpl`, `handleChatCommand` calls `ensureRunningOnSameThread` (re-queues to the main thread) before `tryHandleChat`, while `handleChat` calls `tryHandleChat` straight on the netty thread. So the chat message updates `lastChatTimeStamp` before the main thread reaches the first `/fill`, whose earlier timestamp then fails `updateChatOrder`. 1.19.1+ replaced this with the ordered signed-message chain, and pre-1.19 has no timestamps, so no other version is affected.

Fix in `lib/plugins/chat.js`, gated to 1.19.0: `handleCustomCommandSuggestions` is queued to the main thread the same way and always replies, so a plain message that follows an unacknowledged command first awaits a `tab_complete` roundtrip; that reply proves every earlier command has been processed. All sends on that version go through one chain so order is preserved. Other versions are untouched, and `bot.chat` stays fire-and-forget.

### DIFF
--- a/lib/plugins/chat.js
+++ b/lib/plugins/chat.js
@@ -143,7 +143,7 @@ function inject (bot, options) {
   })
 
   let send = (message) => bot._client.chat(message)
-  if (bot.registry.version['==']('1.19')) {
+  if (bot.supportFeature('chatCommandsQueuedToMainThread')) {
     // 1.19.0 rejects a queued command as out-of-order if a chat message overtakes
     // it; a tab_complete reply proves every earlier command has been processed.
     let sendChain = Promise.resolve()
--- a/lib/plugins/chat.js
+++ b/lib/plugins/chat.js
@@ -142,6 +142,25 @@ function inject (bot, options) {
     if (data.positionId === 2) bot.emit('actionBar', msg, null)
   })
 
+  let send = (message) => bot._client.chat(message)
+  if (bot.registry.version['==']('1.19')) {
+    // 1.19.0 rejects a queued command as out-of-order if a chat message overtakes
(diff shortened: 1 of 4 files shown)


lib/plugins/chat.js
```diff
@@ -142,6 +142,25 @@ function inject (bot, options) {
     if (data.positionId === 2) bot.emit('actionBar', msg, null)
   })
 
+  let send = (message) => bot._client.chat(message)
+  if (bot.registry.version['==']('1.19')) {
```
rom1504: Support version

u9g: Done — switched to `bot.supportFeature('chatCommandsQueuedToMainThread')`, feature added in PrismarineJS/minecraft-data#1257. Needs a minecraft-data bump here once that's released.

PR #4009, opened 2026-08-27
test: summon trade villager with the components-era count key on 1.20.5+

u9g:
Item stacks became components in 1.20.5 and the NBT-era `Count` key in `/summon` is silently ignored, so every trade in the `trade` test cost 1 on 1.20.5+ and `expectAmount` had been loosened to accept anything ≥ 1. This emits `count` on those versions and makes `expectAmount` a strict `assert.strictEqual` everywhere, so a regression in trade pricing fails instead of passing quietly.

Side effect worth knowing: the 36-emerald trade now actually costs 36 on 1.20.5+, which means the test exercises the same input-slot transitions there as on older versions. Verified locally: `trade` passes on 1.20.6, 1.21.1, 1.21.11 (log now shows `36x Emerald and 1x Book`), and 1.19.2 unchanged.

### DIFF
--- a/test/externalTests/trade.js
+++ b/test/externalTests/trade.js
@@ -2,24 +2,16 @@ const assert = require('assert')
 const { once } = require('../../lib/promise_utils')
 
 module.exports = () => async (bot) => {
-  function expectAmount (amount, greaterThan) {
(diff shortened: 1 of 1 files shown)


rom1504:
CI fails

PR #4021, opened 2026-08-29
placeBlock: reject an occupied target up front; nether test clears its sign spot

u9g:
## Problem

The `nether` external test intermittently fails the whole `MC 1.21.8 … 26.1` job with

```
Error: Event blockUpdate:(0, 128, 0) did not fire within timeout of 5000ms
    at placeBlockWithOptions (lib/plugins/place_block.js:13:36)
```

From the CI packet traces of two failing runs (one 26.1, one 1.21.8): chunk (0,0) already had a `red_mushroom` / `brown_mushroom` at (0,128,0) when it loaded. Nether Wastes mushroom features use heightmap placement, which in the nether resolves to the top of the bedrock roof, so mushrooms generate at y=128 — and on some seeds exactly where the test places its sign. The server rejected the placement and re-sent both blocks unchanged (vanilla's resync after a failed use-item), so `placeBlock`'s wait for a *type change* at the target never resolved. All retries failed identically because only the overworld portal location varies per retry.

## Changes

- **`placeBlock`** checks the destination block before sending anything and rejects with `Cannot place block at (x, y, z): it is occupied by <name>` unless the block is one the server lets a placement replace. Previously this surfaced as a 5 s timeout with a misleading "event did not fire" message.
  - The replaceable set comes from the server: vanilla 1.20+ sends the `minecraft:replaceable` block tag on join (play-phase `tags` on 1.20, configuration-phase on 1.20.2+, both already parsed by nmp), which is the same property `canBeReplaced()` consults. That makes it correct for datapacks and modded servers and needs no updating as versions add blocks. Verified on real 1.20.1 and 1.21.8 servers that the entries map straight onto minecraft-data block ids.
  - Servers older than 1.20 don't have the tag, so they use a built-in list derived per version from `Material.isReplaceable()` in the decompiled sources (1.8.8 through 1.19.4), minus tall flowers, whose `canBeReplaced()` is false from 1.13 on. The list is split at the flattening because pre-1.13 `snow` is the full snow block while the layer is `snow_layer`.
- **`nether.js`** checks the sign target and, if it isn't air, clears it with `/setblock` and awaits the block update before placing. Zero cost on the common path.
- Unit tests in `internalTest.js`: the rejection on every tested version, and on 1.20+ that a tag sent by the server overrides the built-in list; docs note in `api.md`.

## Verification

- `internalTest.js`: 729 passing (41 for `placeBlock` across the 28 tested versions).
- `nether` + `sign` external tests on 1.21.8: pass.
- Forced the failure locally (`/setblock 0 128 0 red_mushroom` before the guard) on 1.21.8: guard cleared it, sign placed and text echoed, test passed without a retry.

### DIFF
--- a/lib/plugins/place_block.js
+++ b/lib/plugins/place_block.js
@@ -2,9 +2,9 @@ const { onceWithCleanup } = require('../promise_utils')
 
 module.exports = inject
 
-// Fallback for servers that never send the minecraft:replaceable block tag
-// (pre-1.20): blocks whose Material.isReplaceable() holds, minus tall flowers,
-// whose canBeReplaced() is false from 1.13 on. Names a version lacks are skipped.
+// Servers before 1.20 send no minecraft:replaceable block tag. These are the blocks whose
+// Material.isReplaceable() holds, minus tall flowers, whose canBeReplaced() is false from
+// 1.13 on. A name the version lacks is skipped.
 const REPLACEABLE_LEGACY = ['air', 'flowing_water', 'water', 'flowing_lava', 'lava', 'tallgrass', 'deadbush', 'double_plant', 'vine', 'fire', 'snow_layer', 'structure_void']
 const REPLACEABLE_FLAT = [
   'air', 'cave_air', 'void_air', 'water', 'lava', 'bubble_column', 'structure_void', 'light',
--- a/lib/plugins/place_block.js
+++ b/lib/plugins/place_block.js
@@ -2,16 +2,28 @@ const { onceWithCleanup } = require('../promise_utils')
 
 module.exports = inject
 
-// Vanilla lets a placed block overwrite these; anything else at the target
-// makes the server reject the placement and only echo the unchanged block.
-const REPLACEABLE = new Set([
+// Fallback for servers that never send the minecraft:replaceable block tag
+// (pre-1.20): blocks whose Material.isReplaceable() holds, minus tall flowers,
+// whose canBeReplaced() is false from 1.13 on. Names a version lacks are skipped.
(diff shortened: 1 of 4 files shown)


rom1504:
Failing CI

lib/plugins/place_block.js
```diff
@@ -2,10 +2,22 @@ const { onceWithCleanup } = require('../promise_utils')
 
 module.exports = inject
 
+// Vanilla lets a placed block overwrite these; anything else at the target
+// makes the server reject the placement and only echo the unchanged block.
+const REPLACEABLE = new Set([
```
rom1504, line 7: Is this really constant across versions?

u9g: Done now via block tags on newer versions and checked list on legacy versions

lib/plugins/place_block.js
```diff
@@ -2,10 +2,34 @@ const { onceWithCleanup } = require('../promise_utils')
 
 module.exports = inject
 
+// Fallback for servers that never send the minecraft:replaceable block tag
+// (pre-1.20): blocks whose Material.isReplaceable() holds, minus tall flowers,
+// whose canBeReplaced() is false from 1.13 on. Names a version lacks are skipped.
+const REPLACEABLE_LEGACY = ['air', 'flowing_water', 'water', 'flowing_lava', 'lava', 'tallgrass', 'deadbush', 'double_plant', 'vine', 'fire', 'snow_layer', 'structure_void']
+const REPLACEABLE_FLAT = [
```
rom1504, line 9: Yeah I don't really like this.... Should this be a property of blocks in Minecraft data and then have prismarine registry update it dynamically for recent versions?

u9g: It comes from the tags packet on newer versions, but we can move this list to mcdata for older versions, sure

PR #4023, opened 2026-08-29
inventory: sync the server's inventory bookkeeping after closing a window on pre-1.17

u9g:
Follow-up to #3974, which fixed the 1.17+ half of the `creative` `10 !== 9` failure. This is the ≤1.16.5 half, seen on #4021's CI run (`MC 1.16.5`, all retries).

### What the server does (from decompiled 1.16.5 / 1.17.1 / 1.8.8)

`ServerPlayer.doCloseContainer()` on ≤1.16.5 is just `containerMenu.removed(this); containerMenu = inventoryMenu;` — the player inventory menu's `lastSlots` (the per-slot "last thing I sent the client" list that `broadcastChanges()` diffs against) stays as it was when the container *opened*. Normally the next `ServerPlayer.tick()` diffs and resends the changed slots. But `/clear` also calls `inventoryMenu.broadcastChanges()`; when it runs in the same task drain as the close, a slot the container filled and `/clear` emptied diffs as empty→empty and no `set_slot` is ever sent. 1.17 added `inventoryMenu.transferState(containerMenu)` in `doCloseContainer`, which closes the hole — the same boundary as `supportFeature('stateIdUsed')`.

In the failing trace: `crafting` puts 3 ladders in slot 10 and sends `close_window`; `resetState`'s `/clear` follows 29 ms later; the server reports removing 7 items (planks + ladders) but sends `set_slot` only for 9, 36 and 0. The client-predicted ladders in slot 10 survive every retry. A passing run shows the tick's `set_slot 0/10` resync arriving 45 ms after the close, before the `/clear`.

### Fix

`handleContainerClick`'s accepted path runs `broadcastChanges()` under `ignoreSlotUpdateHack` (`skipPacketSlotUpdates` on 1.8.8), which refreshes `lastSlots` silently, and its `transaction` ack is a real round trip. So `closeWindow` now follows `close_window` with `clickWindow(-999, 0, 0)` — a no-op on both sides with an empty cursor — on versions without state ids, and returns a promise that settles once the server has answered (immediately on 1.17+). `craft()` awaits it and `window.close()` returns it. Every call site inside an async function now awaits it (`tossStack`, the external tests — which hand off to `resetState`'s `/clear` right after closing — and the `anvil`/`trader` examples). `anvil.js`'s `err()` and `craft()`'s catch path stay fire-and-forget: they throw next, and the click reaches the server either way.

### Verification

- New internal test asserts the click follows `close_window` on ≤1.16.5 and is absent on 1.17+ (all 28 tested versions).
- Real 1.16.5 server: `crafting`, `creative`, `furnace`, `useChests` pass; the packet trace shows each close followed by the `-999` click and `transaction accepted`.
- Internal suite: 632 passing.

### DIFF
--- a/lib/plugins/inventory.js
+++ b/lib/plugins/inventory.js
@@ -352,8 +352,9 @@ function inject (bot, { hideErrors }) {
 
   function extendWindow (window) {
     window.close = () => {
-      closeWindow(window)
+      const closed = closeWindow(window)
       window.emit('close')
+      return closed
     }
 
     window.withdraw = async (itemType, metadata, count, nbt) => {
@@ -412,6 +413,11 @@ function inject (bot, { hideErrors }) {
     bot.emit('heldItemChanged', bot.heldItem)
   }
 
+  // Pre-1.17 servers keep the player inventory's slot bookkeeping from before
+  // the container opened until their next tick or an accepted click on it;
+  // until then a change to a slot the container touched (e.g. by /clear) is
+  // never sent. A rejected click already carries the full inventory, so the
+  // sync click's outcome is not an error.
   function closeWindow (window) {
     bot._client.write('close_window', {
       windowId: window.id
@@ -420,6 +426,8 @@ function inject (bot, { hideErrors }) {
     lastClosedWindow = window
     bot.currentWindow = null
     bot.emit('windowClose', window)
+    if (bot.supportFeature('stateIdUsed')) return Promise.resolve()
+    return clickWindow(-999, 0, 0).catch(() => {})
   }
 
   function copyInventory (window) {
--- a/docs/api.md
+++ b/docs/api.md
@@ -484,6 +484,8 @@ This function returns a `Promise`, with `void` as its argument when done withdra
 
 #### window.close()
 
+Close the `window`; returns the `Promise` from [bot.closeWindow(window)](#botclosewindowwindow).
+
 ### Recipe
 
 See [prismarine-recipe](https://github.com/PrismarineJS/prismarine-recipe)
--- a/docs/api.md
+++ b/docs/api.md
@@ -2186,7 +2186,9 @@ Put the item at `slot` in the inventory.
 
 #### bot.closeWindow(window)
 
-Close the `window`.
+This function returns a `Promise`, with `void` as its argument once the server has acknowledged the close.
+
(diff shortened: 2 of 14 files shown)


lib/plugins/inventory.js
```diff
@@ -420,6 +426,8 @@ function inject (bot, { hideErrors }) {
     lastClosedWindow = window
     bot.currentWindow = null
     bot.emit('windowClose', window)
+    if (bot.supportFeature('stateIdUsed')) return Promise.resolve()
+    return clickWindow(-999, 0, 0).catch(() => {})
```
rom1504, line 430: Why do we ignore failures ?

u9g:
**The bug**

Minecraft servers don't tell every player about every inventory change constantly,  that would be wasteful. Instead, the server keeps notes: "here's what I last told this player each of their inventory slots contains." Whenever a slot's real contents differ from the notes, the server sends the player an update and corrects its notes.

Here's the flaw: while you have a chest or crafting table open, the server only maintains its notes for that screen. Its notes about your main inventory freeze at whatever things looked like when you opened it.

On Minecraft 1.16.5 and older, when you close the chest, the server does not update those frozen notes. It normally doesn't matter, because a tick later the server does a routine check, notices its notes are out of date, and sends corrections. There's a tick where the server's notes are wrong.

Our test failure lands exactly in that gap:

1. The bot crafts 3 ladders. They land in an inventory slot. The server's frozen notes still say that slot is empty.
2. The bot closes the crafting screen. The notes are still frozen and wrong.
3. Before the routine check happens, the test cleanup runs /clear, which empties the bot's inventory and then asks the server to send out any updates. The server compares: the slot is actually empty, and my notes say it's empty, so nothing to report. It never tells the bot the ladders are gone.
5. But the bot had already drawn the ladders into its own picture of its inventory when it crafted them (the game works that way to feel responsive, you assume your action worked until told otherwise). No correction ever comes. The bot believes in 3 ladders that don't exist, forever. The next test counts its items, gets 10 instead of 9, and fails on every retry, because /clear can never remove imaginary items.

This is a genuine bug in Minecraft itself. Mojang fixed it in 1.17: from then on, closing a screen properly carries the notes over.

lib/plugins/inventory.js
u9g: see comment below

PR #4024, opened 2026-08-29
placeBlock: reject as soon as the server refuses the placement

u9g:
## Problem

`useChests` failed on 1.9.4 with

```
Error: Event blockUpdate:(1, 4, 1) did not fire within timeout of 5000ms
    at placeBlockWithOptions (lib/plugins/place_block.js:13:36)
```

The packet trace shows the server had answered within 25 ms: `block_place (1,3,1)/top` → `block_change (1,3,1)=grass`, `block_change (1,4,1)=air`, no `set_slot`. Vanilla replies to every `block_place` with a block update for the reference block and then one for the block across the face, whether or not it placed anything. `placeBlock` only waited for the target's *type to change*, so an explicit refusal was discarded and reported five seconds later as a generic "event did not fire".

(Why the server refused in that run: `trade` summons its villager on a command block at (1,4,1); `/kill` only starts the 20-tick death animation, the reset's `/fill` removes the command block, the corpse falls onto (1,4,1), and `useChests`'s second chest is refused there for the ~1 s until the entity is removed. That test race is left for a separate change; this PR is about reporting the server's answer.)

## Changes

- **`placeBlock`** listens for the reference-block update as the server's ack, then treats the next target update as the verdict. A refusal now rejects immediately with `Server refused to place chest at (1, 4, 1): the block is still air`; a genuine no-reply says `Server did not answer the placement at <pos>: …`. Unchanged target updates that arrive *before* the ack are still ignored, which is what #3090 guarded against.
- **`digAndBuild.js`** asserts that placing into the bot's own hitbox rejects with the new message in under 2 s.
- **`api.md`** notes that `placeBlock` rejects when the server refuses.

Complements #4021, which pre-checks a target the *client* can see is occupied; this handles refusals only the server knows about (entities in the way, spawn protection, …). The two touch adjacent lines in `place_block.js` and `api.md`, so whichever lands second needs a trivial rebase.

## Verification

| Version | Tests | Result |
|---|---|---|
| 1.8.8 | digAndBuild, nether, sign | pass |
| 1.9.4 | digAndBuild, nether, sign, useChests | pass |
| 1.21.8 | digAndBuild, nether, sign, useChests | pass |
| 26.1 | digAndBuild, sign | pass |

With the `trade` → `useChests` race reproduced on 1.9.4, the failure now surfaces instantly with the message above instead of timing out.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1948,6 +1948,7 @@ Denies resource pack.
 #### bot.placeBlock(referenceBlock, faceVector)
 
 This function returns a `Promise`, with `void` as its argument when the server confirms that the block has indeed been placed.
+It rejects as soon as the server refuses the placement (for example because an entity is in the way).
 
  * `referenceBlock` - the block you want to place a new block next to
  * `faceVector` - one of the six cardinal directions, such as `new Vec3(0, 1, 0)` for the top face,
--- a/lib/plugins/place_block.js
+++ b/lib/plugins/place_block.js
@@ -7,6 +7,8 @@ function inject (bot) {
     const dest = referenceBlock.position.plus(faceVector)
     const oldBlock = bot.blockAt(dest)
 
+    await bot._genericPlace(referenceBlock, faceVector, options)
+
     // Whether or not it placed anything, the server answers block_place with a
     // block update for the reference block, then one for the block across the
     // face. Updates for dest that arrive before the reference one are stale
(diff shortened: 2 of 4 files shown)


u9g:
CI on the first push failed on two versions:

- **1.16.5 `creative` `10 !== 9`** — the known phantom-ladder flake, fixed by #4023; unrelated.
- **1.17.1 `exampleDigger`** — a real regression in this PR. `examples/digger.js` calls `bot.placeBlock` on every `move` tick while jumping, so several placements for the same spot are in flight at once. The server's replies carry no request id, and the ack listener was keyed on the reference position, so the *first* reply (a refusal while the bot was still in the way) acked every waiting call and they all rejected — including the one the server went on to accept, so the example never chatted `Placing a block was successful` and the test saw `Server refused to place dirt at (50, 4, 0): the block is still dirt` instead.

Fixed by counting in-flight placements per destination: the ack → verdict fast path is used only when a single call is waiting on that destination; otherwise callers settle on a type change as before and share the outcome. Verified locally: 1.17.1 `exampleDigger` ×3, `digAndBuild`/`nether`/`sign`/`useChests`; 1.9.4 `digAndBuild`/`useChests`/`exampleDigger`.

rom1504:
Add more tests for this behavior

PR #3677, opened 2025-06-13
fix editBook sending invalid packet in 1.13+

autowert66:
After Minecraft 1.17, the packet for editing a book changed.
To support writing newer books, a feature `editBookPacketUsesNbt` was added, and #3373 updated the code for writing books to use that feature and send the correct package.

However, after checking for the feature, the wrong code gets executed (uses nbt -> non-nbt packet is sent, and vice versa). This effectively breaks all book functionality since the introduction of the book edit packet.

This PR fixes this by updating the packet sent in each case.

## Test Results
See [source](https://gist.github.com/autowert66/435833eaa9b32c375852713929a7608a).

| version | pr | upstream |
| --- | --- | --- |
| 1.8.9 | ✔ Success | ✔ Success |
| 1.9.4 | ✔ Success | ✔ Success |
| 1.10.2 | ✔ Success | ✔ Success |
| 1.11.2 | ✔ Success | ✔ Success |
| 1.12.2 | ✔ Success | ✔ Success |
[...]

PR #4026, opened 2026-08-29
Re-enable the book test and fix writeBook/signBook on 1.13+

u9g:
The `book` external test has been in `excludedTests` since #3373, and enabling it turned up three problems. It now passes on all 28 tested versions (1.8.8 → 26.1), verified locally with `-g "<v>v book"` for each.

### 1. The `edit_book` branches in `lib/plugins/book.js` were swapped
Same fix as #3677 (credit to @autowert66 for finding it first); this PR additionally re-enables the existing test and fixes what enabling it turned up on 1.20.5+.

#3373 put the 1.17+ `hand/pages/title` packet shape under `editBookPacketUsesNbt` (1.13–1.16) and the 1.13–1.16 `new_book/signing/hand` shape under the else branch, so `writeBook`/`signBook` failed to serialize the packet on every version from 1.13 up. Swapped back.

### 2. 1.20.5+: the first `updateSlot` after `edit_book` is often a stale click echo
The bot's `window_click` sends full item components where the server expects `HashedSlot` hashes, so the server re-sends the affected slots after every click, and it applies the edit asynchronously. The plugin was resolving on the first `updateSlot` for the hand slot, which was frequently a pre-edit echo; the signed book then ended up on the cursor while the caller read the old item. It now waits for an update that actually carries the edit (pages present / `written_book`).

### 3. 1.21.9+: an `edit_book` handled in the same tick as the preceding clicks is never acknowledged
Confirmed with packet traces: with a 100 ms gap before `edit_book`, or with the book already in the hotbar, the server acks with `set_slot`; back-to-back with the move clicks it sends nothing. A `bot._syncWindow` round-trip after the move (same idiom as `craft.js`) forces the edit into a later tick.

The test reads book content from `writable_book_content` / `written_book_content` components on 1.20.5+ instead of `nbt`.

Not addressed here: the hashed-slot mismatch in `clickWindow` behind (2) affects every click on an item with components on 1.21.5+; fixing it properly means hashing component NBT in prismarine-item / minecraft-protocol.

### DIFF
--- a/lib/plugins/book.js
+++ b/lib/plugins/book.js
@@ -1,5 +1,5 @@
 const assert = require('assert')
-const { once } = require('../promise_utils')
+const { onceWithCleanup } = require('../promise_utils')
 
 module.exports = inject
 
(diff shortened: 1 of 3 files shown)


autowert66:
Thanks @u9g for your efforts 👍

Issue #184, opened 2014-05-09
Doesn't support 1.7.6


thiecec:
Would be great if it supported 1.7.6, as https://github.com/andrewrk/node-minecraft-protocol does.

deathcap:
`master` is now targeting 1.8.3. Should we close this since a newer version is (in-progress) supported, or keep open for the (remote, but appealing) possibility of multi-version protocol support? (1.8+1.7,etc.)

roblabla:
For talk about 1.8 support, #198. For talk about multi-protocol support, #199.

PR #4027, opened 2026-08-29
inventory: claim clicked slots by hash on 1.21.5+

u9g:
Depends on PrismarineJS/prismarine-item#184 (`Item.toHashedNotch`); CI will fail until that is released and `prismarine-item` is bumped.

### Problem
Since 1.21.5, `window_click`'s `changedSlots` and `cursorItem` are `HashedSlot`s — item id, count and a CRC32C per component — not full items. `clickWindow` still passes `Item.toNotch` output, so the `hash` field is never set and the server resyncs every slot holding an item with components after every click (`set_slot`/`set_cursor_item`). Those resyncs look exactly like the acknowledgement packets plugins wait for; #4026 shows the resulting race in `writeBook`/`signBook`.

### Change
`clickWindow` and `_syncWindow` use `Item.toHashedNotch` for slot and cursor claims when the protocol defines `HashedSlot`. Nothing else changes: a component prismarine-item can't hash is sent with hash 0, which keeps today's behaviour (a resync) for that slot only.

The switch is keyed on `bot.registry.protocol.types.HashedSlot` rather than a `supportFeature` because minecraft-data has no feature for this; happy to add one there if preferred.

### Verification
- Per-component: `/give`-ing items with each supported component on vanilla 1.21.5, 1.21.11 and 26.1 and moving them between slots produced **no** server resync (packet trace), i.e. the hashes were accepted. Those accepted values are frozen as unit vectors in the prismarine-item PR.
- Regression: the full external test suite passes on 1.21.5, 1.21.11 and 26.1 with this change (74 passing, 0 failing, no retries on each).

### DIFF
--- a/lib/plugins/inventory.js
+++ b/lib/plugins/inventory.js
@@ -549,6 +549,10 @@ function inject (bot, { hideErrors }) {
     return changedSlots
   }
 
+  // 1.21.5+ clicks claim slot contents by hash; the server resends any slot
+  // whose claim doesn't match its own state.
+  const toClickNotch = bot.registry.protocol.types.HashedSlot ? Item.toHashedNotch : Item.toNotch
+
   async function clickWindow (slot, mouseButton, mode) {
     // if you click on the quick bar and have dug recently,
     // wait a bit
@@ -600,7 +604,7 @@ function inject (bot, { hideErrors }) {
       changedSlots = changedSlots.map(slot => {
         return {
           location: slot,
-          item: Item.toNotch(window.slots[slot])
+          item: toClickNotch(window.slots[slot])
         }
       })
     }
@@ -614,7 +618,7 @@ function inject (bot, { hideErrors }) {
         mouseButton,
         mode,
         changedSlots,
-        cursorItem: Item.toNotch(window.selectedItem)
+        cursorItem: toClickNotch(window.selectedItem)
       })
     } else if (bot.supportFeature('actionIdUsed')) { // <= 1.16.5
       bot._client.write('window_click', {
@@ -834,7 +838,7 @@ function inject (bot, { hideErrors }) {
       mouseButton: 2, // end of a drag that never started
       mode: 5,
       changedSlots: [],
-      cursorItem: Item.toNotch(window.selectedItem)
+      cursorItem: toClickNotch(window.selectedItem)
     })
     await synced
   }


lib/plugins/inventory.js
```diff
@@ -549,6 +549,10 @@ function inject (bot, { hideErrors }) {
     return changedSlots
   }
 
+  // 1.21.5+ clicks claim slot contents by hash; the server resends any slot
+  // whose claim doesn't match its own state.
+  const toClickNotch = bot.registry.protocol.types.HashedSlot ? Item.toHashedNotch : Item.toNotch
```
rom1504, line 554: **Astra agent review — AI-generated, not manually written by the maintainer.**

Please raise the minimum `prismarine-item` dependency to the release that provides `toHashedNotch` when the upstream API lands. This PR still allows `^1.17.0`; an allowed installed version such as 1.18.0 has no such method, and prismarine-item #184 is still open. On 1.21.5+, this selects `undefined`, so both `clickWindow` and `_syncWindow` throw `toClickNotch is not a function` before sending their packet. A dependency bump plus a click/sync smoke check using the declared minimum version would make this API requirement explicit.

rom1504 (commented):

Reviewed at the maintainer's request. Validation was static inspection of the current inventory call sites and package requirement, a fresh check of prismarine-item #184 and master, and a local API check with prismarine-item 1.18.0. I did not run a vanilla server.

lib/plugins/inventory.js
u9g: Agreed. prismarine-item#184 isn't released yet, so I can't raise the minimum version. In the meantime 4bac54b falls back to `Item.toNotch` when `toHashedNotch` is missing, so older prismarine-item versions get the pre-hash behaviour (server resyncs) instead of a crash. I'll bump to the release that ships #184 once it's out.

rom1504 (commented):

The fallback in 4bac54b fixes the missing-method crash. With released prismarine-item 1.18.0 (no toHashedNotch), I exercised pickup, _syncWindow and placement through the current inventory plugin and real compiled NMP codec on 1.21.4, 1.21.5 and 1.21.11: all nine packets encode/decode successfully. CI is green.

The original release prerequisite remains for the advertised hashing behavior: land/release prismarine-item#184, raise the minimum item dependency, then check the hashed path with that release. Today's check validates the safe fallback only; it does not establish the candidate hashing implementation or server acceptance. I would keep this conditional until that coordinated landing.

Skills used: `prismarine-item-inventory-review` checked both slot and cursor conversions, including _syncWindow; `prismarine-protocol-data-review` checked the actual codec and released API availability; `prismarine-review` distinguished the repaired crash from the existing release requirement.

rom1504:
https://github.com/PrismarineJS/prismarine-item/pull/184 not ready

PR #4028, opened 2026-08-30
ci: fail a PR if any test gets more than 1.5x slower than master

u9g:
## What

Recent PRs (#3977, #3984, #3999, #3990, ...) cut a lot of fixed sleeps out of the external tests. Nothing stops them creeping back in, and a fixed wall-clock budget doesn't work here since runtime is dominated by server startup and varies per version. So: compare per-test durations against master's last run and fail on a 1.5x regression.

## How

- `test/common/durationsReporter.js`: the spec reporter, plus a `{ "<test title>": ms }` file of passing tests written to `$DURATIONS`. `mocha_test` now uses it; log output is unchanged.
- `test/common/compareDurations.js <baselineDir> <currentDir>`: for each test present in both runs, fail if it took more than **1.5x** its baseline **and** at least 5s longer (so sub-second tests can't fail on jitter). Prints a `baseline -> now` table per version. Tests without a baseline are skipped.
- `ci.yml`: restore the last master baseline for the version group from `actions/cache` (`durations-<job-index>-*`), run the comparison after the tests, and on pushes to master save the run as the new baseline. Rebaselining is automatic, no committed files.

Only tests that passed count, and with `--retries 3` mocha records the passing attempt's duration, so a flaky retry doesn't register as a slowdown.

## Verification

Locally against the internal tests: 632 durations recorded with the spec output intact; the compare script exits 1 when a test goes 1s -> 20s, 0 when identical, and skips cleanly with no baseline. The first CI run on master seeds the cache; PRs before that just print "no baseline yet".

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,6 +67,18 @@ jobs:
       - name: Install Dependencies
         run: npm install
 
+      # Per-test durations from the last master run of this version group, so a PR
+      # can be failed if it makes a test more than 1.5x slower (test/common/compareDurations.js).
+      - name: Restore duration baseline
+        uses: actions/cache/restore@v4
+        with:
+          path: durations
+          key: durations-${{ strategy.job-index }}-${{ github.sha }}
+          restore-keys: durations-${{ strategy.job-index }}-
+      - run: |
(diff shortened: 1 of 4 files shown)


rom1504:
I like it but I think it might fail. Let's try it out and adapt if anything doesn't work

PR #4031, opened 2026-08-30
Confirm creative slot sets with a stats round trip instead of a fixed 400ms wait

u9g:
## Problem

On `noAckOnCreateSetSlotPacket` versions (1.21.9+), `setInventorySlot` resolves success by sleeping a flat 400ms and treating silence as acceptance. Every successful call pays the full wait. Measured on the 26.1 external suite with a counter: **62 calls × 400ms = 24.8s per run, ~19% of the suite's wall time** — and every creative-mode bot pays the same per slot write.

## Change

A rejection is a `set_slot` correction the server sends while processing our packet on its main thread, in receive order. So after writing `set_creative_slot`, send `client_command(request_stats)` and wait for the `statistics` reply: it is generated after our packet was processed, and any correction was already delivered before it on the same ordered connection. Success now resolves in ~1 tick instead of 400ms.

- Statistics replies carry no transaction id, so in-flight confirms resolve FIFO — matching the server answering requests in packet order.
- The existing `waitTimeout` (400ms) now caps the wait for the reply: a server that never answers stats degrades to exactly the old behavior, never a hang.
- `request_stats` is sent as `{ payload: 1 }` on `respawnIsPayload` versions, `{ actionId: 1 }` otherwise (same gate as health.js respawn).

Primitive choice was verified empirically against a vanilla 26.1 server: play-state `ping_request` is answered on the **netty thread in 0–2ms, before earlier queued packets** (10/10 runs — it proves nothing about processing), while a `statistics` reply consistently arrived **after** a just-sent chat echo (10/10 runs, 7–53ms), confirming main-thread FIFO handling.

## Tests

Added an assertion to the creative external test (gated on `noAckOnCreateSetSlotPacket`): three sequential slot sets must finish under 1s. It fails on master's lib at exactly **1204ms** (3×400ms) and passes with this change (~100ms). The creative test drops from ~6.5s to ~1.4s overall; the 26.1 suite loses ~25s.

Ran the creative external test locally on 1.12.2, 1.19.4 and 26.1 — all passing. Old versions are unaffected: the acked path is untouched and `confirmServerProcessed` is only called from the no-ack branch.

### DIFF
--- a/lib/plugins/creative.js
+++ b/lib/plugins/creative.js
@@ -20,6 +20,30 @@ function inject (bot) {
 
   const creativeSlotsUpdates = []
 
+  // FIFO: the server answers stats requests in packet-arrival order, so each
+  // statistics packet settles the oldest waiter, and everything written
+  // before that waiter's request is processed by then.
+  const statsConfirms = []
+  bot._client.on('statistics', () => statsConfirms.shift()?.())
+
+  function confirmServerProcessed (timeoutMs) {
+    return new Promise((resolve) => {
+      // Timing out resolves as success: a server that never answers stats
+      // degrades to the previous fixed-wait behavior, never a hang.
+      const timer = setTimeout(() => {
+        const i = statsConfirms.indexOf(settle)
+        if (i !== -1) statsConfirms.splice(i, 1)
+        resolve()
+      }, timeoutMs)
+      function settle () {
+        clearTimeout(timer)
+        resolve()
+      }
+      statsConfirms.push(settle)
+      bot._client.write('client_command', bot.supportFeature('respawnIsPayload') ? { payload: 1 } : { actionId: 1 })
+    })
+  }
+
   // WARN: This method should not be called twice on the same slot before first promise succeeds
   async function setInventorySlot (slot, item, waitTimeout = 400) {
     assert(slot >= 0 && slot <= 44)
@@ -38,7 +62,9 @@ function inject (bot) {
       // No ack
       bot._setSlot(slot, item)
       if (waitTimeout === 0) return // no wait
-      // allow some time to see if server rejects
+      // A rejection is a set_slot correction the server sends while
+      // processing our packet on its main thread, so a stats round trip on
+      // the same ordered connection is proof the rejection window has passed.
       return new Promise((resolve, reject) => {
         function updateSlot (oldItem, newItem) {
           if (newItem.itemId !== item.itemId) {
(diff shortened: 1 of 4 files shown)


u9g:
Pushed a second commit: making slot sets fast exposed a latent one-tick race in `resetBlocksToSuperflat` — its marker echo proves the fills *executed*, but command feedback is sent immediately while block changes flush at tick end, so the next test could still see pre-fill blocks (the old 400ms waits were absorbing exactly that tick). The reset now waits, event-driven, until the reset region is actually visible client-side, with a loud 5s deadline. Verified with a digAndBuild→placeBlock ordering that reproduced the stale-dirt failure before the hardening.

u9g:
Third commit: the strict reset check caught a second latent leak, this one a client/server desync. The sign test fired `bot.activateBlock` unawaited from its setTimeout, letting the use_item_on land in the same server tick as the next reset's fills — the sign editor's out-of-band re-broadcast then races the fill's air change, leaving the client permanently showing a sign the server no longer has (trace: sign re-asserted after the fills, no air update ever). The test now awaits the interact plus the server's editor-open ack before ending. sign+sound green ×3 runs.

u9g:
CI red was mine: the sign fix waited for an editor-reopen ack that only exists on 1.20+ (editable signs) — every group containing a pre-1.20 version timed out on it, which is also exactly why 1.21.3–1.21.6 was the one green group. Amended: the test just awaits the interact, and the reset now self-heals any client/server block desync (stale past 1.5s → two forced /setblock changes make the server rebroadcast, whichever side is wrong). Verified this round on 1.12.2, 1.19.4, and 26.1×2 including the desync-exercising sequence. The 1.21.8-group failure was an unrelated one-off login timeout (three of its four versions fully passed); the digAndBuild pickup race seen once locally is pre-existing and fixed in #4034.

lib/plugins/creative.js
```diff
@@ -20,6 +20,30 @@ function inject (bot) {
 
   const creativeSlotsUpdates = []
 
+  // FIFO: the server answers stats requests in packet-arrival order, so each
+  // statistics packet settles the oldest waiter, and everything written
+  // before that waiter's request is processed by then.
+  const statsConfirms = []
+  bot._client.on('statistics', () => statsConfirms.shift()?.())
```
DatArnoGuy, line 27: ```statsConfirms.shift()?.()```
What does this do?

rom1504: Agreed let's do something less weird here

u9g: I rewrote it, should make more sense now

PR #4045, opened 2026-09-02
settings: send particleStatus to the server

u9g:
`bot.settings.particleStatus` has defaulted to `'all'` since the field was added, but the `settings` packet mineflayer writes on login never included it. On 1.21.3+ (`particleStatus` is a mapper: `all` / `decreased` / `minimal`) the field was serialized from `undefined`. It only came out as `0` = `all` because protodef's *compiled* mapper falls an unmapped value through to the numeric type, where `NaN` writes as `0`; the interpreted mapper throws on the same input.

ProtoDef-io/node-protodef#176 makes the compiled mapper strict too (throw on a value not in the mappings), which turns this into a serialization error on every 1.21.3+ login. Send the field. The companion fix for node-minecraft-protocol's configuration-phase `settings` is PrismarineJS/node-minecraft-protocol#1519.

No version gate needed: protodef containers only write the fields the version defines, so the extra property is ignored pre-1.21.3.

**Tests:** no test can fail without this change today — with the current protodef the bytes on the wire are identical (`undefined` → `0` → `all`). Once protodef#176 lands, every existing 1.21.3+ external test exercises it (login would throw without it). Ran locally: internal suite (688 passing, all versions), external `chat` on 1.8.8, 1.16.5 and 1.21.8, lint clean.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,8 +67,8 @@ jobs:
       - name: Install Dependencies
         run: npm install
 
-      # Per-test durations from the last master run of this version group, so a PR
-      # can be failed if it makes a test more than 1.5x slower (test/common/compareDurations.js).
+      # Per-test durations from the last master run of this version group, so a PR gets a
+      # comment listing tests it made more than 1.5x slower (test/common/compareDurations.js).
       - name: Restore duration baseline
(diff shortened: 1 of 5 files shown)


rom1504:
CI fails

PR #4046, opened 2026-09-03
Expose cape URL in player skinData

u9g:
The profile `textures` property carries a `CAPE` entry next to `SKIN` for players who have one. This surfaces it as `player.skinData.capeUrl` (undefined when absent), with the type, docs and the existing skin-data internal tests updated.

Needed by PrismarineJS/prismarine-viewer#490 to render capes.

CI note: the earlier red runs were the new duration-comparison step flagging `fishing` (1.12.2, 1.15.2) and `nether` (1.20.2, 1.20.4, 1.21.3) as >1.5x slower than master; every test passed and none of them touch skin data.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -424,7 +424,8 @@ The skin data is stored in the `skinData` property of the player object, if pres
 // player.skinData
 {
   url: 'http://textures.minecraft.net/texture/...',
-  model: 'slim' // or 'classic'
+  model: 'slim', // or 'classic'
+  capeUrl: 'http://textures.minecraft.net/texture/...' // only if the player has a cape
 }
 ```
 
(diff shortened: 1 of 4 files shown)


rom1504:
Share resulting screenshot

u9g:
Screenshot from PrismarineJS/prismarine-viewer#490 rendering `skinData.capeUrl` from this branch. Left is from behind, right from the front. U9G is slim with the Migrator cape, jeb_ is wide with the Mojang cape, Notch is wide with no cape.

player capes in prismarine-viewer

Taken on an offline 1.21.4 server, so `skinData` was set on the three players from their real Mojang profile textures. Rebased on master.

The red checks are the duration-comparison step, not a test failure: every version passes, and the tests it flags are `fishing` and `nether`, which name a different set of versions on each run and touch no skin data.

Issue #4043, opened 2026-09-01
acceptResourcePack() sends an all-zero pack UUID on 1.20.3+ — server ignores the response and the bot never leaves the configuration phase


JensR93:
- [x] The [FAQ](https://github.com/PrismarineJS/mineflayer/blob/master/docs/FAQ.md) doesn't contain a resolution to my issue

## Versions

- mineflayer: 4.37.1 (the relevant code is unchanged on current `master`)
- server: Paper 1.21.11 (resource pack pushed during the CONFIGURATION phase, `required=false`)
- node: 26.x

## Detailed description of a problem

On any 1.20.3+ server that pushes a resource pack, calling `bot.acceptResourcePack()` (or `denyResourcePack()`) does not work: the server never registers the response, the connection stays in the CONFIGURATION phase exchanging `keep_alive` packets forever, and the bot never spawns (or is eventually kicked by the server's own timeout). This looks like the root cause behind the symptom reports in #3659, #3860, #3768 and #3844.

**Root cause — the response packet carries an all-zero pack UUID:**

1. `lib/plugins/resource_pack.js` stores the pack id as a uuid-1345 instance:
   ```js
   const uuid = new UUID(data.uuid)
   latestUUID = uuid
   ```
2. `acceptResourcePack()` then writes it as-is:
   ```js
   bot._client.write('resource_pack_receive', { uuid: latestUUID, result: TEXTURE_PACK_RESULTS.ACCEPTED })
   ```
3. node-minecraft-protocol's UUID writer (`src/datatypes/compiler-minecraft.js`) expects a string:
   ```js
   const buf = value.length === 32 ? Buffer.from(value, 'hex') : UUID.parse(value)
   ```
   A `UUID` **instance** has `.length === undefined`, and `UUID.parse(<UUID instance>)` does not throw — it silently returns an all-zero buffer:
   ```
   > const UUID = require('uuid-1345')
   > UUID.parse(new UUID('f81d4fae-7dec-11d0-a765-00a0c91e6bf6'))
   <Buffer 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00>
   ```
4. So the client answers with pack uuid `00000000-0000-0000-0000-000000000000`. Vanilla/Paper match resource-pack responses by UUID, silently drop the mismatch, and never proceed to `FINISH_CONFIGURATION`.

## What did you try yet?

Measured live against the same Paper 1.21.11 server, same account, same optional pack:

[...]

PR #4051, opened 2026-09-05
send one protocol-shaped decline

Pix3lPirat3:
Make the hash, result-only, and UUID response forms mutually exclusive so UUID-era clients do not emit a second malformed response.

In reference to https://github.com/PrismarineJS/mineflayer/issues/4043

### DIFF
--- a/lib/plugins/resource_pack.js
+++ b/lib/plugins/resource_pack.js
@@ -60,11 +60,11 @@ function inject (bot) {
       })
     } else if (bot.supportFeature('resourcePackUsesUUID')) {
       bot._client.write('resource_pack_receive', {
-        uuid: latestUUID,
+        uuid: latestUUID.toString(),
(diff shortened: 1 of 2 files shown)


rom1504 (commented):

The single-decline change looks ready from this review. I ran all 56 added cases across the 28 configured versions with minecraft-protocol 1.68.0 and minecraft-data 3.116.0: one decline and the two acceptance responses encode/decode correctly across hash, result-only and UUID layouts. Extra object fields are ignored by the selected production schema, so their presence does not introduce another wire variant. This validates the changed response assembly; it is not a new live-server or configuration-transfer test.

Skills used: `prismarine-protocol-data-review` checked the actual compiled packet schemas and both response directions; `prismarine-architecture-review` checked that removing repeated assembly preserves the protocol contract; `prismarine-review` checked current code/discussion and scope.

Issue #3985, opened 2026-08-23
bot.oxygenLevel / breath are driven by any nearby entity's air_supply, not only the bot's


numenus:
- [x] The [FAQ](https://github.com/PrismarineJS/mineflayer/blob/master/docs/FAQ.md) doesn't contain a resolution to my issue

## Versions
 - mineflayer: 4.37.1 (same code on master @ aa8fdfa)
 - server: vanilla 1.21.11 (any server version where `mcDataHasEntityMetadata` holds — 1.19.4+ — on mineflayer ≥ 4.21.0)
 - node: 22.20.0

## Detailed description of a problem
`bot.oxygenLevel` (documented as a value in [0, 20]) and the `breath` event (documented as "Fires when your oxygen level
change") are driven by the `air_supply` metadata of **whatever entity the `entity_metadata` packet is about**, not only the bot. In `lib/plugins/entities.js`
(4.37.1 lines 493–497, identical on master):

```js
// Breathing (formerly in breath.js)
if (metas.air_supply != null) {
  bot.oxygenLevel = Math.round(metas.air_supply / 15)
  bot.emit('breath')
}
```

There is no `packet.entityId === bot.entity.id` check. `air_supply` is metadata slot 1 of the base Entity, so every
living entity in tracking range sends it when its air changes. The legacy `lib/plugins/breath.js` has the guard
(`if (bot.entity.id !== packet.entityId) return`) but returns early at line 7 whenever `mcDataHasEntityMetadata` is
supported, so on every modern version the unguarded block is the only writer.

The server only sends `air_supply` when it changes, so a bot standing on dry land never receives its own value: its
`oxygenLevel` becomes the last nearby swimmer's. I'm building agents whose "am I in danger" check reads `bot.oxygenLevel`;
two bots on dry land reported drowning for three minutes because a cow was swimming in view.

The unguarded `entities.js` path dates from #3412 ("1.20.6", 44fad41, released in 4.21.0), which moved breathing for
`mcDataHasEntityMetadata` versions into entities.js. The symptom itself is older: the 4.20.1 `breath.js` guard it replaced,
`if (!bot?.entity?.id === packet?.entityId) return`, was a no-op (operator precedence), so the legacy branch applied any
entity's value too. #3412 fixed that guard in the legacy branch; the new entities.js path never received one.

## What did you try yet?
Reproduced deterministically on a vanilla 1.21.11 server: a bot standing on dry land, a cow summoned into a sealed
[...]

PR #4052, opened 2026-09-05
Only update oxygen from the bot's metadata

Pix3lPirat3:
Patch for https://github.com/PrismarineJS/mineflayer/issues/3985

### DIFF
--- a/lib/plugins/entities.js
+++ b/lib/plugins/entities.js
@@ -495,7 +495,7 @@ function inject (bot) {
       }
 
       // Breathing (formerly in breath.js)
-      if (metas.air_supply != null) {
+      if (metas.air_supply != null && entity.id === bot.entity?.id) {
         bot.oxygenLevel = Math.round(metas.air_supply / 15)
         bot.emit('breath')
       }
--- a/test/internalTest.js
+++ b/test/internalTest.js
@@ -996,6 +996,64 @@ for (const supportedVersion of mineflayer.testedVersions) {
         })
       })
 
+      it('only updates oxygen level from bot metadata', function (done) {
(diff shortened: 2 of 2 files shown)


VasilisDragon:
Looks good to me. The guard restores the ownership check already used by the legacy breathing path. The added internal test passes on all 14 applicable configured versions, from 1.19.4 through 26.1; restoring the old handler makes all 14 fail. It also checks that the bot's own metadata still updates oxygen and emits `breath`.

Minebot17:
Thanks for this fix! I hit the same bug on a vanilla 26.1.2 server with mineflayer 4.39.0: while the bot was underwater, `oxygenLevel` kept flipping between the real value and ~400 (a nearby axolotl's `air_supply` / 15), and `breath` fired 616 times in 12 s. With the `entity.id === bot.entity?.id` guard applied locally it now decreases smoothly 10 → 0.

Two small related things I noticed while testing, in case you'd like to fold them in (otherwise I can send a follow-up PR once this lands):

1. **Negative values while drowning.** The server sends negative `air_supply` (I observed `-13`), so `oxygenLevel` becomes `-1`, although the docs describe the range as 0–20. Clamping with `Math.max(0, Math.round(metas.air_supply / 15))` (and the same in the legacy `breath.js` path) keeps it in range.
2. **`undefined` after spawn.** The server only sends the bot's `air_supply` metadata when it changes, so on land `oxygenLevel` stays `undefined` until the bot first goes underwater (the new test works around this by setting `bot.oxygenLevel = 20` on login). Initializing it to 20 (full air) in `breath.js` `inject()` fixes that.

PR #4057, opened 2026-09-06
increase speed of tests exampleBlockFinder.js and exampleDigger.js

Pix3lPirat3:
removes a redundant two-second delay after the child has already confirmed its chunks are loaded. replaces its two-second guess with an inventory response that synchronizes /op and /give.


Pix3lPirat3:
Redundant, already a PR in https://github.com/PrismarineJS/mineflayer/pull/4034

PR #4059, opened 2026-09-06
fix: clear a chest's lid open count when its block is replaced; make useChests await its fill commands

u9g:
Fixes the `useChests` failure on 1.21.11 in https://github.com/PrismarineJS/mineflayer/actions/runs/34038030872/job/101499603931 (release PR #4058, which only touches the changelog).

## What happened

Two stacked problems, visible in the uploaded packet trace:

1. **First attempt: a race in the test.** `fillChest` fires three `/item replace` commands and opens the chest immediately. On 1.21.9+ block interaction packets run from the server's `PacketProcessor` queue at the start of the next tick, while chat commands wait on the main task queue and only run while the tick has time left. Under CI load the window snapshot arrived with only the first replace applied, and the mycelium assertion failed.
2. **Every retry: stale open count in `block_actions.js`.** The aborted attempt left the double chest open, so `openCountByPos` held `1` for that position. The harness reset's `/fill` replaced the chest with air, and the server drops a container's closing block event once the block is no longer a chest, so the count was never cleared. On each retry the server's open block action carried the same count, `chestLidMove` was never emitted, and `once(chest, 'close')` timed out three times.

The second problem is a user-facing bug too: any chest broken while open never emits `chestLidMove` at that position again.

## Changes

- `lib/plugins/block_actions.js`: drop the open count on a `blockUpdate` that changes the block's name.
- `test/externalTests/plugins/testCommon.js`: extract the superflat reset's marker echo into `bot.test.awaitCommandsProcessed(marker)`.
- `test/externalTests/useChests.js`: `fillChest` awaits that echo before the chest is opened.
- `test/internalTest.js`: regression test that opens a chest, replaces it via `block_change`, and expects `chestLidMove` on the next open. It times out on master and passes with the fix.

## Verification

- New internal test: passes on 1.8.8, 1.12.2, 1.16.5, 1.21.11 and 26.1 with the fix; fails on unpatched master.
- External `useChests`: passes locally on 1.21.11 and 1.8.8.
- `standard` clean.

### DIFF
--- a/lib/plugins/block_actions.js
+++ b/lib/plugins/block_actions.js
@@ -22,6 +22,13 @@ function inject (bot) {
   // Stores how many players have currently open a container at a certain position
   const openCountByPos = {}
 
+  // The server drops the closing block event of a container that is no
+  // longer there, so a replaced container must not keep its open count.
+  bot.on('blockUpdate', (oldBlock, newBlock) => {
+    if (!newBlock || oldBlock?.name === newBlock.name) return
+    delete openCountByPos[newBlock.position]
+  })
+
   function parseChestMetadata (chestBlock) {
     const chestTypes = ['single', 'right', 'left']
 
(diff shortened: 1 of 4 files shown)


rom1504:
Approved, resolved conflict and merge

Issue #3759, opened 2025-10-09
entity_teleport serialization errors on 1.21.8


DereC4:
- [Yes] The [FAQ](https://github.com/PrismarineJS/mineflayer/blob/master/docs/FAQ.md) doesn't contain a resolution to my issue 

## Versions
 - mineflayer: 4.33.0
 - server: Wolfx 1.21.8
 - node: 20.19.19

## Detailed description of a problem
When connecting to a server running custom software (Wolfx 1.21.8), Mineflayer throws a stream of errors related to incorrect packet chunk sizes immediately after connecting.

The log is filled with the error: Chunk size is 67 but only 33 was read. Was able to join world still

## Your current code
```js

  this.currentBot = mineflayer.createBot({
      host: server.ip,
      port: server.port,
      username: `kmhozcdb@ronaldofmail.com`,
      auth: 'microsoft',
      // version: '1.21.8', // Auto-detect version
      viewDistance: 1,
      hideErrors: false,
  });


```

## Expected behavior
Not sure; is this due to some in-world mods/datapacks not being recognized or parsed correctly?

## Additional context
Logs here:

`0|cobbleGuard  | 2025-10-09T18:32:04: Attempting to connect to 8.138.207.141:25565 Secure chat: false Online Mode: null Version: Wolfx 1.21.8
0|cobbleGuard  | 2025-10-09T18:32:08: Chunk size is 67 but only 33 was read ; partial packet : {"name":"entity_teleport","params":{"entityId":1073741824,"x":-194.5,"y":76,"z":61.5,"yaw":0,"pitch":0,"onGround":false}}; buffer :768080808004c0685000000000004053000000000000404ec00000000000000000000000000000000000000000000000000000000000c33125893ff3abb20000000001
0|cobbleGuard  | 2025-10-09T18:32:08: Chunk size is 67 but only 33 was read ; partial packet : {"name":"entity_teleport","params":{"entityId":1073741825,"x":-191.5,"y":76,"z":59.5,"yaw":0,"pitch":0,"onGround":false}}; buffer :768180808004c067f000000000004053000000000000404dc0000000000000000000000000000000000000000000000000000000000042ba9d02420c9d230000000001
[...]

PR #4062, opened 2026-09-06
entities: read the 1.21.2+ entity_teleport layout (velocity, float rotation, relative flags)

u9g:
Since 1.21.2 `ClientboundTeleportEntityPacket` carries the same fields as the player position packet: position, delta movement, f32 yaw/pitch and a set of relative flags. minecraft-data described the old layout until PrismarineJS/minecraft-data#1273, which makes the packet parse again on 1.21.3 through 26.1 (fixes the `Chunk size is 67 but only 33 was read` partial reads from #3759).

With that data fix the handler gets a `flags` object, so this applies each part as absolute or relative the same way the `position` handler does, and converts the float rotation instead of the byte rotation. `entity_teleport` is a hot packet, so the handler is chosen once at inject time from the `entityTeleportHasRelativeFlags` feature (added in the same minecraft-data PR), the same way `player_info` picks its handler. Without the data fix the feature resolves to false and the old byte path is unchanged, so this is safe to merge in either order.

Verified with the patched data on a 1.21.4 Velocity/ViaVersion network (Jartex), where these packets previously flooded the log with partial reads: no parse errors, and teleported entities land on the coordinates from the packet.

### DIFF
--- a/lib/plugins/entities.js
+++ b/lib/plugins/entities.js
@@ -337,6 +337,24 @@ function inject (bot) {
   bot._client.on('entity_teleport', (packet) => {
     // entity teleport
     const entity = fetchEntity(packet.entityId)
+    if (packet.flags !== undefined) {
+      // 1.21.2+: same layout as the player position packet, each part absolute or relative per flag
+      const { position: pos, velocity: vel, yaw, pitch } = entity
+      pos.set(
+        packet.flags.x ? pos.x + packet.x : packet.x,
+        packet.flags.y ? pos.y + packet.y : packet.y,
+        packet.flags.z ? pos.z + packet.z : packet.z
+      )
+      vel.set(
+        packet.flags.dx ? vel.x + packet.dx : packet.dx,
+        packet.flags.dy ? vel.y + packet.dy : packet.dy,
+        packet.flags.dz ? vel.z + packet.dz : packet.dz
+      )
+      entity.yaw = conv.fromNotchianYaw((packet.flags.yaw ? conv.toNotchianYaw(yaw) : 0) + packet.yaw)
+      entity.pitch = conv.fromNotchianPitch((packet.flags.pitch ? conv.toNotchianPitch(pitch) : 0) + packet.pitch)
+      bot.emit('entityMoved', entity)
+      return
+    }
     if (bot.supportFeature('fixedPointPosition')) {
       entity.position.set(packet.x / 32, packet.y / 32, packet.z / 32)
     }
--- a/lib/plugins/entities.js
+++ b/lib/plugins/entities.js
@@ -334,27 +334,9 @@ function inject (bot) {
     bot.emit('entityMoved', entity)
   })
 
-  bot._client.on('entity_teleport', (packet) => {
+  function handleEntityTeleportLegacy (packet) {
     // entity teleport
     const entity = fetchEntity(packet.entityId)
-    if (packet.flags !== undefined) {
-      // 1.21.2+: same layout as the player position packet, each part absolute or relative per flag
-      const { position: pos, velocity: vel, yaw, pitch } = entity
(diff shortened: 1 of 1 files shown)


rom1504:
Approved can merge once mc data released with the PR merged

lib/plugins/entities.js
```diff
@@ -346,7 +346,28 @@ function inject (bot) {
...
+    const entity = fetchEntity(packet.entityId)
+    const { position: pos, velocity: vel, yaw, pitch } = entity
+    pos.set(
+      packet.flags.x ? pos.x + packet.x : packet.x,
+      packet.flags.y ? pos.y + packet.y : packet.y,
+      packet.flags.z ? pos.z + packet.z : packet.z
+    )
+    vel.set(
+      packet.flags.dx ? vel.x + packet.dx : packet.dx,
+      packet.flags.dy ? vel.y + packet.dy : packet.dy,
```
rom1504, line 362: **Astra agent review — AI-generated, not manually written by the maintainer.**

Could this apply `flags.yawDelta` (ROTATE_DELTA) before adding the flagged velocity components? With an entity facing yaw 0 and moving at (1, 0, 0), a teleport to yaw 90 with `yawDelta`, `dx`, and `dz` set and zero packet velocity currently leaves velocity at (1, 0, 0); it should rotate to approximately (0, 0, 1). I reproduced the retained vector by executing this handler with real Vec3/conversion helpers. This makes the exposed velocity disagree with the teleported entity's motion. Please rotate the old velocity by the pitch/yaw change before the per-axis additions and cover that case, as the player teleport path in #4063 does. The same rotation-before-addition semantics are visible in [Azalea's implementation](https://azalea.matdoes.dev/src/azalea_protocol/common/movements.rs.html).

rom1504 (commented):

At the maintainer's request, I reviewed the current diff, surrounding handlers, and existing discussion. Validation: static inspection plus a focused local reproduction using the current handler code (a stub client/clock for lifecycle behavior; no full server suite run). One correctness finding is inline.

lib/plugins/entities.js
u9g: Fixed in 8b4440b5617cecb31399748d57cf6b809382a19c: with `yawDelta`, the current velocity is now rotated by the pitch/yaw change (vanilla `xRot` then `yRot`) before the flagged dx/dy/dz are added, the same as the player teleport in #4063. I added an internal test covering absolute, relative, and `yawDelta` teleports, including your yaw 0 → 90 case, which now gives (0, 0, 1). The test skips until a minecraft-data release includes `entityTeleportHasRelativeFlags`. Against minecraft-data master it passes on 1.21.3 through 26.1.

VasilisDragon:
The updated teleport regression passes on all eight affected tested versions with the required data changes from pinned minecraft-data master. The mixed pitch/yaw and relative-velocity cases also check out against the official 1.21.4 implementation. The data release gate still applies: the latest published minecraft-data (3.116.0) skips this test.

rom1504 (commented):

The earlier velocity-rotation finding is addressed: a focused execution of the current entity handler confirms yawDelta rotates the old nonzero velocity before relative additions on 1.21.4, 1.21.11 and 26.1. This was a model-level check with the new feature enabled explicitly, not released-data wire validation. Installed minecraft-data 3.116.0 still lacks that feature. **Code looks good, but keep the [existing data-release prerequisite](https://github.com/PrismarineJS/mineflayer/pull/4062#issuecomment-5565399013) before merging.** VasilisDragon's candidate-data network validation is separate evidence, not a run I performed.

Skills used: prismarine-geometry-movement-review checked rotation-before-addition; prismarine-protocol-data-review separated handler correctness from released feature/schema availability. prismarine-review checked current discussions and avoided duplicate findings.

PR #4063, opened 2026-09-06
physics: send the movement input packets the vanilla client sends

u9g:
Compared `lib/plugins/physics.js` against the 1.21.4 client (`LocalPlayer.tick()` / `sendPosition()` / `aiStep()`, Mojang mappings) and made the bot send what the client sends, in the same order per tick: shift key state, `player_input` when the key set changed, `start/stop_sprinting` when the sprinting state changed, the movement packet, then `tick_end`.

**Bug fixed**: on 1.21.3, 1.21.4 and 1.21.5 `bot.setControlState('sneak', true)` only sent `player_input { shift }`. Those servers set the sneak state from `entity_action` PRESS_SHIFT_KEY (`ServerGamePacketListenerImpl.handlePlayerCommand`) and ignore the shift bit of `player_input` (it is only stored as `lastClientInput`; the server started reading it in 1.21.6, `handlePlayerInput` → `setShiftKeyDown`). The vanilla client sends both packets on those versions and only `player_input` from 1.21.6, so the bot never sneaked server-side there: no crouch hitbox, no sneak-placing against interactable blocks, no sneak interactions.

Which packet carries the sneak state is decided by the `sneakUsesEntityAction` feature from PrismarineJS/minecraft-data#1275 (`1.8` through `1.21.5`). On a minecraft-data release without it the feature is `false`, so no `entity_action` is sent for the shift key on any version, the same as before this PR; the sneak fix for 1.21.3–1.21.5 takes effect once that data ships. The range ends at 1.21.5 on purpose: the 1.21.6+ `entity_action` string mapper has no shift entries, so a missing feature must never route the shift key there.

**Parity changes** (1.21.2+ unless stated):
- `player_input` carries the whole key set (`forward/backward/left/right/jump/shift/sprint`) and is sent whenever it changes, not just for sneak; `tick_end` is sent after every physics tick (the server uses it to reset `knownMovement`, which gates sweep attacks and is what anticheats key their tick boundaries on).
- Sprinting follows `LocalPlayer.aiStep` instead of the sprint key: it starts on the ground with a full forward impulse, more than 6 food (or creative), no item in use, no blindness and not in water; it stops when the forward impulse, food or a wall collision is against it or the player sneaks or enters water. `start/stop_sprinting` follow that state. The local simulation is fed the same state, so the bot no longer moves at sprint speed backwards, sideways or with an empty hunger bar (all versions).
- Movement packets carry `hasHorizontalCollision` from the simulation instead of `undefined`, and a status-only (`flying`) packet is sent when it flips, like `sendPosition` does.
- A position is sent once it moved more than 2e-4 blocks since the last one or after 20 ticks (`positionReminder`), instead of on any change / after 1 s of wall time.

Docs: `bot.setControlState` describes when `sprint` actually sprints. Test: `sends the 1.21.2+ input packets like vanilla` in `internalTest.js` checks the packet set and ordering for sneak, the sprint key while sneaking (key set changes, no sprint), sprint after releasing sneak, stop when forward is released, and `tick_end`; it asserts `entity_action` PRESS_SHIFT_KEY up to 1.21.5 and its absence from 1.21.6.

Verified live on a Paper 1.21.4 Velocity network (Jartex) by tracing the bot's outgoing packets while walking.

### DIFF
--- a/lib/plugins/physics.js
+++ b/lib/plugins/physics.js
@@ -22,11 +22,10 @@ function inject (bot, { physicsEnabled, maxCatchupTicks }) {
   const positionUpdateSentEveryTick = bot.supportFeature('positionUpdateSentEveryTick')
   const hasConfigurationState = bot.supportFeature('hasConfigurationState') // 1.20.2+
   // 1.21.2+ sends the key set as player_input whenever it changes and tick_end after every tick.
-  // The shift key still goes through entity_action up to 1.21.5; from 1.21.6 the server reads it
-  // from player_input alone.
+  // The server takes the sneak state from entity_action up to 1.21.5 and from player_input after.
   const hasPlayerInputPacket = bot.supportFeature('newPlayerInputPacket')
   const hasTickEndPacket = 'packet_tick_end' in bot.registry.protocol.play.toServer.types
-  const sneakUsesEntityAction = !bot.registry.version['>=']('1.21.6')
+  const sneakUsesEntityAction = bot.supportFeature('sneakUsesEntityAction')
 
   bot.jumpQueued = false
   bot.jumpTicks = 0 // autojump cooldown
--- a/lib/plugins/physics.js
+++ b/lib/plugins/physics.js
@@ -419,9 +419,22 @@ function inject (bot, { physicsEnabled, maxCatchupTicks }) {
   }
 
   // 1.21.3+
+  // A forced rotation is answered at once with a look carrying onGround false and no collision; the
+  // last-sent record is left alone.
   bot._client.on('player_rotation', (packet) => {
     bot.entity.yaw = conv.fromNotchianYaw(packet.yaw)
     bot.entity.pitch = conv.fromNotchianPitch(packet.pitch)
+    lastSentYaw = bot.entity.yaw
+    lastSentPitch = bot.entity.pitch
+    bot._client.write('look', {
+      x: lastSent.x,
+      y: lastSent.y,
+      z: lastSent.z,
+      yaw: packet.yaw,
+      pitch: packet.pitch,
+      onGround: false,
+      flags: { onGround: false, hasHorizontalCollision: false }
+    })
   })
 
   // player position and look (clientbound)
@@ -436,21 +449,35 @@ function inject (bot, { physicsEnabled, maxCatchupTicks }) {
 
     // Note: 1.20.5+ uses a bitflags object, older versions use a bitmask number
     if (typeof packet.flags === 'object') {
-      // Modern path with bitflags object
-      // Velocity is only set to 0 if the flag is not set, otherwise keep current velocity
-      vel.set(
-        packet.flags.x ? vel.x : 0,
-        packet.flags.y ? vel.y : 0,
-        packet.flags.z ? vel.z : 0
-      )
+      const oldYaw = conv.toNotchianYaw(bot.entity.yaw)
+      const oldPitch = conv.toNotchianPitch(bot.entity.pitch)
       // If flag is set, then the corresponding value is relative, else it is absolute
       pos.set(
         packet.flags.x ? (pos.x + packet.x) : packet.x,
         packet.flags.y ? (pos.y + packet.y) : packet.y,
         packet.flags.z ? (pos.z + packet.z) : packet.z
       )
-      newYaw = (packet.flags.yaw ? conv.toNotchianYaw(bot.entity.yaw) : 0) + packet.yaw
-      newPitch = (packet.flags.pitch ? conv.toNotchianPitch(bot.entity.pitch) : 0) + packet.pitch
+      newYaw = (packet.flags.yaw ? oldYaw : 0) + packet.yaw
+      newPitch = (packet.flags.pitch ? oldPitch : 0) + packet.pitch
+      // The packet carries a velocity only when all three deltas are present.
+      if (Number.isFinite(packet.dx) && Number.isFinite(packet.dy) && Number.isFinite(packet.dz)) {
+        // A dx/dy/dz flag adds the current velocity to the packet's; yawDelta first turns the
+        // current velocity by the rotation change.
+        let v = vel.clone()
+        if (packet.flags.yawDelta) v = rotateVelocity(v, (oldPitch - newPitch) * PI / 180, (oldYaw - newYaw) * PI / 180)
+        vel.set(
+          (packet.flags.dx ? v.x : 0) + packet.dx,
+          (packet.flags.dy ? v.y : 0) + packet.dy,
+          (packet.flags.dz ? v.z : 0) + packet.dz
+        )
+      } else {
+        // Velocity is only set to 0 if the flag is not set, otherwise keep current velocity
+        vel.set(
+          packet.flags.x ? vel.x : 0,
+          packet.flags.y ? vel.y : 0,
+          packet.flags.z ? vel.z : 0
+        )
+      }
     } else {
       // Legacy path with bitmask number
       // Velocity is only set to 0 if the flag is not set, otherwise keep current velocity
@@ -497,7 +524,7 @@ function inject (bot, { physicsEnabled, maxCatchupTicks }) {
       return
     }
 
-    sendPacketPositionAndLook(pos, newYaw, newPitch, bot.entity.onGround)
+    sendTeleportReply(pos, newYaw, newPitch)
 
     shouldUsePhysics = true
     bot.jumpTicks = 0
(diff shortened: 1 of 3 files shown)


u9g:
3425d014 adds three teleport invariants: a 1.21.2+ position packet's dx/dy/dz set the velocity, with a set delta flag adding the current one and `yawDelta` first turning it by the rotation change; the reply to a teleport carries onGround false and no horizontal collision and leaves the last-sent record alone, so the next tick still transmits the teleported position; a clientbound `player_rotation` is answered at once with an equally ungrounded `look`. A packet without deltas keeps the previous velocity handling. Three tests, all versions green.

u9g:
Rebased onto master (7f188fb3) so the branch installs under pnpm; no other change. It merges cleanly with #4107 (reply timing) and #4108 (standing flag), which are independent of it. Vanilla per-tick order this PR mirrors, from u9g/mc-decompiled 26.1: `LocalPlayer.tick` runs `aiStep` (shift key), then `player_input` when the key set changed, then `sendPosition` (`sendIsSprintingIfNeeded` first, then the movement packet), and `Minecraft.tick` sends `tick_end` after the world tick. Physics internal tests pass on all 28 versions after the rebase.

lib/plugins/physics.js
```diff
@@ -175,33 +193,93 @@ function inject (bot, { physicsEnabled, maxCatchupTicks }) {
...
+      }
+      if (lastSentInputs === null || Object.keys(inputs).some(k => inputs[k] !== lastSentInputs[k])) {
+        lastSentInputs = inputs
+        bot._client.write('player_input', { inputs })
+      }
+    }
+    if (sprinting !== lastSentSprinting) {
+      lastSentSprinting = sprinting
+      bot._client.write('entity_action', {
+        entityId: bot.entity.id,
```
rom1504, line 259: **Astra agent review — AI-generated, not manually written by the maintainer.**

Please reset or resynchronize the new sent-input/sprint caches when the player/session is recreated. These caches survive `login` and `respawn`, although the new server player starts with fresh input state. I reproduced holding sneak, sending one tick, then emitting `login`, a new position, and another tick: neither `player_input` nor PRESS_SHIFT_KEY is sent for the still-held control. The bot therefore believes sneak is held while the new server sees it released; retained sprint state has the same cache ownership problem. A focused transfer/respawn regression with unchanged held controls would cover this.

rom1504 (commented):

At the maintainer's request, I reviewed the current diff, surrounding handlers, and existing discussion. Validation: static inspection plus a focused local reproduction using the current handler code (a stub client/clock for lifecycle behavior; no full server suite run). One correctness finding is inline.

lib/plugins/physics.js
```diff
@@ -21,6 +21,11 @@ function inject (bot, { physicsEnabled, maxCatchupTicks }) {
 
   const positionUpdateSentEveryTick = bot.supportFeature('positionUpdateSentEveryTick')
   const hasConfigurationState = bot.supportFeature('hasConfigurationState') // 1.20.2+
+  // player_input is sent only when the key set changes; tick_end after every tick.
+  // Servers before 1.21.6 ignore the shift bit of player_input; from 1.21.6 entity_action has no shift entries.
+  const hasPlayerInputPacket = bot.supportFeature('newPlayerInputPacket')
+  const hasTickEndPacket = 'packet_tick_end' in bot.registry.protocol.play.toServer.types
+  const sneakUsesEntityAction = bot.supportFeature('sneakUsesEntityAction')
```
rom1504, line 28: **Astra agent review — AI-generated, not manually written by the maintainer.**

Please preserve the legacy sneak fallback, or require a data release containing this feature. `package.json` still allows `minecraft-data: ^3.114.0`; both 3.114.0 and the installed 3.116.0 lack `sneakUsesEntityAction`, so this is false. Unlike the behavior described in the PR, this removes working sneaking on older versions: the deleted `setControlState` branch sent `entity_action` when `newPlayerInputPacket` was false. In a real 1.20.4 protocol client/server test with data 3.116.0, setting sneak and waiting three physics ticks produces no sneak action at this head; the identical test passes with the actual base physics code. Keeping the old pre-`player_input` fallback would preserve those installations while the newer sneak fix waits for its data prerequisite.

Skills used: prismarine-protocol-data-review helped check the allowed releases and feature gate; prismarine-behavior-test-review guided the server-observed head/base comparison; prismarine-review separated this regression from the existing cache-reset feedback.

rom1504 (commented):

One additional dependency-compatibility regression is inline. Validation: real 1.20.4 protocol client/server comparison against the current head and its actual base physics plugin, with the same installed dependencies; the head fails and the base passes. The existing focused 1.21.4 physics tests pass. No vanilla server suite was run.

Skills used: prismarine-protocol-data-review checked the declared dependency range and feature availability; prismarine-behavior-test-review guided the observable regression test; prismarine-review prevented duplicating the existing lifecycle finding.

lib/plugins/physics.js
```diff
@@ -81,12 +96,15 @@ function inject (bot, { physicsEnabled, maxCatchupTicks }) {
     if (!bot.entity?.position || !Number.isFinite(bot.entity.position.x)) return // entity not ready
     if (bot.blockAt(bot.entity.position) == null) return // check if chunk is unloaded
     if (bot.physicsEnabled && shouldUsePhysics) {
-      physics.simulatePlayer(new PlayerState(bot, controlState), world).apply(bot)
+      updateSprinting()
```
rom1504, line 99: **Astra agent review — AI-generated, not manually written by the maintainer.**

Please update the derived sprint state even when local physics is disabled. Here `updateSprinting()` runs only inside the `bot.physicsEnabled` branch, while `sendInputPackets()` below continues running. In the same loaded play session, start sprinting, set `bot.physicsEnabled = false`, then call `bot.clearControlStates()`: the key inputs clear, but `sprinting` stays true and no `stop_sprinting` action is sent. A focused real 1.21.4 protocol client/server test receives no stop action at this head (`[]`), while the identical test with the actual base physics implementation receives action `4`. Please keep this state calculation active independently of simulation and cover clearing controls with physics disabled.

Skills used: prismarine-lifecycle-action-review helped separate simulation readiness from outgoing input state; prismarine-geometry-movement-review helped trace the derived sprint state; prismarine-review checked this against the existing feedback.

rom1504 (commented):

One distinct sprint-state regression is inline. A focused real 1.21.4 protocol client/server head/base comparison reproduces the missing stop action after disabling physics and clearing controls. Dependencies: minecraft-protocol 1.68.0, minecraft-data 3.116.0. No vanilla server or full version suite was run.

Skills used: prismarine-lifecycle-action-review helped separate simulation readiness from outgoing input state; prismarine-geometry-movement-review helped trace the derived sprint state; prismarine-review checked this against the existing feedback.

lib/plugins/physics.js
u9g: Fixed in https://github.com/u9g/mineflayer/commit/b296ac574f5d9622ca22df900fce2c380e67aa47: login and respawn now reset the sent-input caches and the sprint state, so held controls are sent again to the new player. Covered by 'sends the held controls again to the player a login creates' (all versions).

lib/plugins/physics.js
u9g: Fixed in https://github.com/u9g/mineflayer/commit/b296ac574f5d9622ca22df900fce2c380e67aa47: when the installed minecraft-data lacks `sneakUsesEntityAction`, sneak goes through entity_action while its `actionId` is a bare varint (up to 1.21.5; 1.21.6 turns it into a mapper without shift entries). That matches the feature data, so older data releases keep working sneak. The held-controls test checks the sneak action on every version.

lib/plugins/physics.js
u9g: Fixed in https://github.com/u9g/mineflayer/commit/b296ac574f5d9622ca22df900fce2c380e67aa47: `updateSprinting()` now runs on every tick that sends inputs, independent of `bot.physicsEnabled`. Covered by 'stops sprinting when the controls clear with physics disabled' (all versions).

lib/plugins/physics.js
```diff
@@ -80,14 +97,26 @@ function inject (bot, { physicsEnabled, maxCatchupTicks }) {
     if (bot._client.state !== 'play') return // do nothing outside of the play state (e.g. server transfer configuration phase)
     if (!bot.entity?.position || !Number.isFinite(bot.entity.position.x)) return // entity not ready
     if (bot.blockAt(bot.entity.position) == null) return // check if chunk is unloaded
-    if (bot.physicsEnabled && shouldUsePhysics) {
-      physics.simulatePlayer(new PlayerState(bot, controlState), world).apply(bot)
+    if (!shouldUsePhysics) return
```
rom1504, line 100: **Astra agent review — AI-generated, not manually written by the maintainer.**

Please keep input synchronization active while riding. `mount` sets `shouldUsePhysics = false`, so this early return now also suppresses `sendInputPackets()`; the previous implementation wrote sneak directly from `setControlState`. With the actual head/base physics plugins and compiled 1.21.6 codecs, a loaded play bot sends `player_input.shift = true` while standing in both versions. After a mount event, setting the same sneak control sends no input at this head, while the base still sends shift=true. The control is held locally but never reaches the server. This also prevents #4093's intended sneak-to-dismount path working once its #4063 prerequisite lands. Please separate mounted input synchronization from walking simulation readiness and add a mounted regression. This was a controlled plugin/codec reproduction, not a network or vanilla-server test.

Skills used: prismarine-lifecycle-action-review separated riding/input lifetime from physics simulation; prismarine-protocol-data-review checked the actual encoded input; prismarine-review checked that this is distinct from the earlier disabled-physics sprint finding.

rom1504 (commented):

The three earlier findings are fixed in this revision: held-input caches reset, legacy sneak has a schema fallback, and sprint state updates with physics disabled. The current physics suite passes locally (215 passing, 93 pending across the version matrix). **Not ready to merge yet:** the new inline finding reproduces lost sneak input while mounted, independently of those fixes.

Skills used: prismarine-lifecycle-action-review separated riding/input lifetime from walking simulation; prismarine-protocol-data-review verified the head/base difference through compiled packet codecs. prismarine-review checked current discussions and avoided duplicate findings.

PR #4064, opened 2026-09-06
health: send player_loaded when the vanilla client would

u9g:
`player_loaded` (1.21.4+) was written on the first `update_health`, before any chunk had arrived. The vanilla client sends it from `LevelLoadStatusManager`: only after the server's `LEVEL_CHUNKS_LOAD_START` game event (sent from `PlayerList.sendLevelInfo` on join and on respawn) and once the chunk section the player stands in is loaded, or the player is outside the build height, a spectator, or dead. The server keeps the player invulnerable and ignores block/item interactions until it receives the packet (or 60 ticks pass), so sending it early ends that grace period before the bot's world exists locally.

The bot now waits for the game event and for the column at `bot.entity.position` (`chunkColumnLoad` / `forcedMove` re-check), then sends it once; the `spawn` event timing is unchanged. Test added in `internalTest.js` (`player_loaded` describe) for 1.21.4+.

### DIFF
--- a/lib/plugins/health.js
+++ b/lib/plugins/health.js
@@ -35,6 +35,9 @@ function inject (bot, options) {
     chunksAnnounced = false
     bot._client.write('player_loaded', {})
   }
+  // Every clientbound login and respawn starts a level whose chunks the server announces once.
+  bot._client.on('login', sendPlayerLoadedWhenReady)
+  bot._client.on('respawn', sendPlayerLoadedWhenReady)
   bot._client.on('game_state_change', (packet) => {
     if (packet.reason === 13 || packet.reason === 'level_chunks_load_start') {
       chunksAnnounced = true
--- a/lib/plugins/health.js
+++ b/lib/plugins/health.js
@@ -15,10 +15,9 @@ function inject (bot, options) {
     bot.emit('spawn')
   }
 
-  // The vanilla client (LevelLoadStatusManager) sends player_loaded once the server has announced
-  // the chunks are coming (game event LEVEL_CHUNKS_LOAD_START, sent on join and on respawn) and the
-  // chunk the player stands in has arrived, or the player is outside the build height, a spectator
-  // or dead. Until then the server keeps the player invulnerable and ignores interactions.
+  // player_loaded is sent at most once per level_chunks_load_start game event (join and every
+  // respawn), never before it, and only once the column under the bot is loaded unless the bot
+  // is outside the build height, a spectator or dead.
   let awaitingLoad = false
   let chunksAnnounced = false
   function sendPlayerLoadedWhenReady () {
--- a/lib/plugins/health.js
+++ b/lib/plugins/health.js
@@ -11,9 +11,39 @@ function inject (bot, options) {
   // 1.21.4+ servers ignore block and item interactions until the client has
   // sent player_loaded (or 60 ticks have passed)
   function spawn () {
-    if (bot.supportFeature('sendsPlayerLoadedPacket')) bot._client.write('player_loaded', {})
+    sendPlayerLoadedWhenReady()
     bot.emit('spawn')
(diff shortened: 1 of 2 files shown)


lib/plugins/health.js
```diff
@@ -11,10 +11,42 @@ function inject (bot, options) {
...
+    if (!pos || !Number.isFinite(pos.x)) return
+    const outsideBuildHeight = pos.y < bot.game.minY || pos.y >= bot.game.minY + bot.game.height
+    if (!(outsideBuildHeight || bot.game.gameMode === 'spectator' || !bot.isAlive || bot.blockAt(pos) !== null)) return
+    awaitingLoad = false
+    chunksAnnounced = false
+    bot._client.write('player_loaded', {})
+  }
+  // Every clientbound login and respawn starts a level whose chunks the server announces once.
+  bot._client.on('login', sendPlayerLoadedWhenReady)
+  bot._client.on('respawn', sendPlayerLoadedWhenReady)
```
rom1504, line 40: **Astra agent review — AI-generated, not manually written by the maintainer.**

Please clear the previous level's `chunksAnnounced` when handling a new login/respawn, before checking readiness. It is currently cleared only after a successful `player_loaded`. I reproduced `login → level_chunks_load_start` with the bot's column still absent, followed by `respawn`: the old announcement survives and this call immediately writes `player_loaded` before any announcement for the new level (the earlier respawn listener also sets `isAlive = false`, satisfying the readiness exception). This defeats the per-level guard during a level change before loading completes. Please separate starting a new level from the later `spawn()` readiness recheck, and add a regression for an unfinished load followed by respawn/transfer.

rom1504 (commented):

lib/plugins/health.js
u9g: Fixed in 0e51bd80be9e036b0aa964b73c130b837fc20067. Login and respawn now start a new level: they set `awaitingLoad` and clear `chunksAnnounced` without checking readiness. `spawn()` only rechecks readiness. Added a regression test (`ignores a previous level's announcement after a respawn`): login, then `level_chunks_load_start` with the bot's column missing, then respawn. It fails on the previous code and passes on this commit on every tested version from 1.21.4 to 26.1.

VasilisDragon:
`!bot.isAlive` also covers the interval after `respawn` and before the next health update. After a completed join, `respawn → position in an unloaded column → fresh level_chunks_load_start` sends `player_loaded` immediately with health 20 and no column. Reproduced through the network harness on 1.21.4, 1.21.11 and 26.1, including with current master merged in. In 1.21.4, [PlayerList sends the announcement](https://github.com/extremeheat/extracted_minecraft_data/blob/43a29ca80bb329f75b92695317a522492e25c2e7/client/net/minecraft/server/players/PlayerList.java#L418-L424) before the later [health update](https://github.com/extremeheat/extracted_minecraft_data/blob/43a29ca80bb329f75b92695317a522492e25c2e7/client/net/minecraft/server/level/ServerPlayer.java#L723-L737). Could the check distinguish actual death from the temporary respawn state?

rom1504 (commented):

The old-level announcement is now cleared, and the focused player_loaded suite passes (21 passing, 63 pending). **Not ready to merge yet:** I independently reproduced [the already reported healthy-respawn issue](https://github.com/PrismarineJS/mineflayer/pull/4064#issuecomment-5744080917) with the actual health plugin on 1.21.4, 1.21.11 and 26.1: after a completed join, respawn plus a fresh announcement sends player_loaded with health 20 and no column, because isAlive is temporarily false. That existing thread covers the needed repair; no duplicate inline finding.

Skills used: prismarine-lifecycle-action-review distinguished actual death from the interval between respawn and the next health update. prismarine-review checked current discussions and avoided duplicate findings.

PR #4065, opened 2026-09-06
Send Client Information once, when the vanilla client does

u9g:
Up to 1.20.1 the vanilla client sends Client Information after login. From 1.20.2 it sends it during the configuration phase, and again only when a setting changes.

node-minecraft-protocol already sends Client Information during the configuration phase, but from its own defaults (view distance 10). mineflayer then re-sent it after login with the bot's settings (view distance 12). Every connection told the server two different things.

This passes the bot's settings to node-minecraft-protocol as `clientSettings`, so the configuration-phase packet carries them, and only re-sends after login on versions without a configuration phase.

`setSettings` is unchanged from the caller's side: a later `bot.setSettings(...)` still writes the packet, on every version.

Scope notes:
- The client brand change that was here has been dropped — see the comment below.
- The `locale` / `enableServerListing` value fixes were split out, since they're about what's in the packet rather than when it's sent.

### DIFF
--- a/lib/plugins/settings.js
+++ b/lib/plugins/settings.js
@@ -21,9 +21,7 @@ const viewDistanceToBits = {
 }
 
 function inject (bot, options) {
-  function setSettings (settings) {
-    extend(bot.settings, settings)
-
+  function clientInformation () {
     // chat
     const chatBits = chatToBits[bot.settings.chat]
     assert.ok(chatBits != null, `invalid chat setting: ${bot.settings.chat}`)
@@ -52,8 +50,7 @@ function inject (bot, options) {
           bot.settings.skinParts.showRightPants << 5 |
           bot.settings.skinParts.showHat << 6
 
-    // write the packet
-    bot._client.write('settings', {
+    return {
       locale: bot.settings.locale || 'en_US',
       viewDistance: viewDistanceBits,
       chatFlags: chatBits,
@@ -63,7 +60,14 @@ function inject (bot, options) {
       enableTextFiltering: bot.settings.enableTextFiltering,
       enableServerListing: bot.settings.enableServerListing,
       particleStatus: bot.settings.particleStatus
-    })
+    }
+  }
+
+  function setSettings (settings) {
+    extend(bot.settings, settings)
+
+    // write the packet
+    bot._client.write('settings', clientInformation())
   }
 
   bot.settings = {
@@ -92,8 +96,12 @@ function inject (bot, options) {
     particleStatus: 'all'
   }
 
+  // On 1.20.2+ node-minecraft-protocol sends Client Information during the configuration
+  // phase; give it the bot's settings so the server is never told two different things.
+  options.clientSettings = options.clientSettings ?? clientInformation()
+
   bot._client.on('login', () => {
-    setSettings({})
+    if (!bot.supportFeature('hasConfigurationState')) setSettings({})
   })
 
   bot.setSettings = setSettings
--- a/test/internalTest.js
+++ b/test/internalTest.js
@@ -535,6 +535,32 @@ for (const supportedVersion of mineflayer.testedVersions) {
       })
     })
 
+    describe('settings', () => {
+      it('sends Client Information after login only without a configuration phase', async function () {
+        // 1.20.2+ sends it during the configuration phase instead, from options.clientSettings.
+        const hasConfigState = bot.supportFeature('hasConfigurationState')
+        const sent = []
+        const originalWrite = bot._client.write.bind(bot._client)
+        bot._client.write = (name, params) => {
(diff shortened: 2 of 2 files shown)


lib/plugins/settings.js
```diff
@@ -1,6 +1,8 @@
 const assert = require('assert')
 
 module.exports = inject
+module.exports.defaults = defaults
+module.exports.clientInformation = clientInformation
```
rom1504: Why do we need this ? Does not follow the plugin interface

u9g:
Dropped the client brand change that was in this PR.

It gated the brand send on `!bot.supportFeature('hasConfigurationState')`, on the premise that node-minecraft-protocol sends the brand itself during the configuration phase from 1.20.2. It doesn't — `minecraft:brand` appears nowhere in node-minecraft-protocol 1.68.0 (the only "brand" hits in `src/` are in `autoVersion.js`, about the server's version string). mineflayer's send on the `login` packet is the only one, so gating it would have meant the brand is never sent at all on 1.20.2+.

Matching vanilla here (brand right after `login_acknowledged`, in the configuration phase) needs node-minecraft-protocol to send it or to expose the configuration-phase entry, so it's a separate change upstream rather than part of this PR.

lib/plugins/settings.js
```diff
@@ -92,8 +96,12 @@ function inject (bot, options) {
     particleStatus: 'all'
   }
 
+  // On 1.20.2+ node-minecraft-protocol sends Client Information during the configuration
+  // phase; give it the bot's settings so the server is never told two different things.
+  options.clientSettings = options.clientSettings ?? clientInformation()
```
rom1504, line 101: **Astra agent review — AI-generated, not manually written by the maintainer.**

This is a snapshot of the initial settings, but `setSettings()` never updates it. With current node-minecraft-protocol master, entering configuration, calling `bot.setSettings({ viewDistance: 'short' })`, then receiving `start_configuration` sends view distances `12 → 8 → 12`; I reproduced that sequence with the two plugins. The server silently returns to the initial distance while `bot.settings.viewDistance` still says `short`. #1522 is still open and proposes suppressing the repeated configuration send, so please either make that behavior an enforced dependency or keep the settings source shared with subsequent updates. A configuration round-trip test after `setSettings()` would catch this.

rom1504 (commented):

Reviewed at the maintainer's request. I inspected the settings and loader paths, existing interface feedback, current node-minecraft-protocol behavior, and #1522's open status. A local execution of the settings and protocol play plugins reproduced a stale settings resend during reconfiguration. This was an in-process reproduction, not a live server test.

lib/plugins/settings.js
u9g: Fixed in f15a5303ac4ed4ff8bd33adbddbe4f37b4cdb518: `options.clientSettings` is now a getter over `bot.settings`, not a snapshot, so whatever node-minecraft-protocol sends on a reconfiguration matches the current settings, whether or not #1522 lands. Added a test (`keeps later settings when the server reconfigures`): it calls `setSettings({ viewDistance: 'short' })` and then sends `start_configuration`. Before the fix it fails on all 12 tested versions with a configuration phase (12 !== 8), and it passes after.

lib/plugins/settings.js
u9g: That was the `module.exports.defaults` / `module.exports.clientInformation` exports from an earlier revision. They're gone now: the plugin only has the usual `inject(bot, options)` export, and the Client Information is built inside it.

VasilisDragon:
The documented `options.client` path still needs a settings send. With `createBot({ client: mc.createClient(...), viewDistance: 'short' })`, the getter is installed on Mineflayer's options, but the supplied client reads its own options. The server now receives distance 10 while the bot reports `short` (8); the baseline sends the correcting 8 after login. Reproduced over TCP on 1.20.2, 1.21.4, 1.21.11 and 26.1. Could this path retain the login update or share settings with the supplied client?

rom1504 (commented):

The live getter fixes settings reconfiguration for the ordinary createBot path; focused tests pass (40 passing, 16 pending). **Not ready to merge yet:** [the supplied-client finding](https://github.com/PrismarineJS/mineflayer/pull/4065#issuecomment-5744083886) remains. Executing the actual settings and protocol play plugins with separate caller/client options on 1.20.2, 1.21.4 and 26.1 confirms a sent view distance of 10 while bot.settings says short (8). This was an in-process check; VasilisDragon's TCP reproduction is separate evidence. Please handle that existing supported path before merging.

Skills used: prismarine-lifecycle-action-review checked re-entry settings; prismarine-architecture-review traced which options object a supplied client actually owns. prismarine-review checked current discussions and avoided duplicate findings.

PR #4066, opened 2026-09-06
Match the vanilla client's interaction packets: entity interact, digging, placing, sequences, resource packs

u9g:
Audit of the interaction plugins against the vanilla 1.21.4 client (`MultiPlayerGameMode`, `Minecraft.startUseItem/startAttack/continueAttack`, `BlockStatePredictionHandler`, `DownloadedPackSource`) and the server's handlers. One commit per fix:

- **Entity activation** (`activateEntity`, `activateEntityAt`): send `use_entity` interact_at followed by interact, as `startUseItem` does, and with the bot's real sneak state instead of `sneaking: false` (the server applies `setShiftKeyDown(packet.isUsingSecondaryAction())`, so the old value silently un-sneaked the player).
- **`bot.useOn` and `bot.mount`** sent a lone interact with the hit location fixed at the entity's feet; a right click on an entity is always interact_at then interact. Both now go through the same sender as `activateEntity`, and `use_entity` no longer has a second, divergent implementation in `entities.js`.
- **`bot.attack`** throws for the bot itself, item entities and experience orbs: a vanilla client can never target them and the server kicks with `multiplayer.disconnect.invalid_entity_attacked`.
- **`bot.dismount`** on 1.21.3+ sent `player_input { jump }`, which the server never reads; vanilla dismounts through the shift key (`Player.rideTick`). It now holds the sneak control for one physics tick (returns a Promise; older versions keep `steer_vehicle`).
- **Placing**: `use_item_on` first, then the arm swing (vanilla swings only after the use succeeded).
- **Digging**: swing every physics tick while destroying (vanilla `continueAttack`), send only START for instant breaks (creative / zero-hardness; vanilla never sends STOP there), and a user-initiated `stopDigging` aborts with face DOWN like `stopDestroyBlock`, while a retarget keeps the new face.
- **Sequence numbers**: one pre-incremented counter shared by dig start/stop, `use_item_on` and `use_item` (`BlockStatePredictionHandler`), first value 1; abort/release use 0. Previously `block_dig` omitted it, `block_place` hardcoded 0 and `use_item` had its own counter.
- **`activateBlock`** defaults the cursor to the centre of the clicked face (the old (0.5, 0.5, 0.5) is inside the block, impossible for a real client); **`activateItem`** no-ops with an empty hand like vanilla.
- **Resource packs**: vanilla reports ACCEPTED → DOWNLOADED (1.20.3+) → SUCCESSFULLY_LOADED; the missing 1.20.3+ statuses are added. `denyResourcePack` wrote a second DECLINED without the pack UUID after the UUID branch, which throws a serialization error on 1.20.3+; it now sends one.
- `game.js` win-game respawn used `action` where the field is `actionId`.

Tests added in `internalTest.js` for the packet sequences (entity interaction, digging incl. instant break and abort faces, placement order, sequence counter, resource pack statuses, attack rejection). Docs updated for `bot.attack` and `bot.dismount`.

### DIFF
--- a/lib/plugins/sequence.js
+++ b/lib/plugins/sequence.js
@@ -1,0 +1,15 @@
+module.exports = inject
+
+// Block prediction sequence shared by dig start/stop, use_item_on and use_item
+// (vanilla BlockStatePredictionHandler): pre-incremented, so the first value
+// sent is 1. Abort / release / drop actions send 0 and do not consume a value.
+function inject (bot) {
+  const hasSequence = bot.registry.isNewerOrEqualTo('1.19')
+  let sequence = 0
+
+  bot._nextSequence = () => {
+    if (!hasSequence) return 0
+    sequence++
+    return sequence
+  }
+}
--- a/lib/plugins/sequence.js
+++ b/lib/plugins/sequence.js
@@ -1,8 +1,7 @@
 module.exports = inject
 
-// Block prediction sequence shared by dig start/stop, use_item_on and use_item
-// (vanilla BlockStatePredictionHandler): pre-incremented, so the first value
-// sent is 1. Abort / release / drop actions send 0 and do not consume a value.
+// One sequence counter for dig start/stop, use_item_on and use_item. Sent values
+// start at 1; abort, release and drop actions send 0 and must not consume a value.
 function inject (bot) {
   const hasSequence = bot.registry.isNewerOrEqualTo('1.19')
   let sequence = 0
--- a/docs/api.md
+++ b/docs/api.md
@@ -2037,6 +2037,8 @@ Attack a player or a mob.
  * `entity` is a type of entity. To get a specific entity use [bot.nearestEntity()](#botnearestentitymatch--entity---return-true-) or [bot.entities](#botentities).
  * `swing` Default to `true`. If false the bot does not swing its arm when attacking.
 
+Throws if `entity` is the bot itself, an item or an experience orb: the server kicks a client that attacks those.
+
 #### bot.swingArm([hand], showHand)
 
 Play an arm swing animation.
--- a/docs/api.md
+++ b/docs/api.md
@@ -2052,7 +2052,9 @@ Mount a vehicle. To get back out, use `bot.dismount`.
 
 #### bot.dismount()
 
-Dismounts from the vehicle you are in.
+This function returns a `Promise`, with `void` as its argument once the dismount has been sent.
+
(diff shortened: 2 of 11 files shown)


u9g:
6dbae033 invariant: the `use_entity` hit vector is where the line from the eye to the entity's middle enters its bounding box, so it lies on the box surface, and it falls back to the middle for an entity of unknown size. `MultiPlayerGameMode.interactAt` sends `entityHitResult.getLocation()` relative to the entity, which is a surface point; the old (0, height/2, 0) is only reachable with the eye inside the entity. New test asserts the near face for an entity approached along one axis. All versions green.

Not changed here: vanilla's `Minecraft.startUseItem` loops both hands and sends a second interact_at/interact pair when the main hand's result is PASS, which for a player-type entity it always is. Reproducing that needs the per-entity client-side interaction result, which mineflayer does not model, and sending an unconditional second pair would double-trigger interactions that do consume the action.

lib/plugins/sequence.js
```diff
@@ -0,0 +1,14 @@
+module.exports = inject
+
+// One sequence counter for dig start/stop, use_item_on and use_item. Sent values
+// start at 1; abort, release and drop actions send 0 and must not consume a value.
+function inject (bot) {
+  const hasSequence = bot.registry.isNewerOrEqualTo('1.19')
```
rom1504, line 6: Support version

rom1504:
This one is really a lot of subtle change all over the place.

It's great that test pass but I think we should double check all the changes make sense some more.

u9g:
Split into separate PRs, one per fix from the audit (this PR's diff is reproduced exactly by the union of them; the resource-pack part is additionally rebased over #4061):

| Fix | PR |
|---|---|
| Shared block prediction sequence counter | #4085 |
| `win_game` respawn `actionId` field | #4086 |
| Resource pack statuses + single DECLINED | #4087 |
| internal tests: wait for plugin injection | #4088 |
| `activateEntity` vanilla interact pair / sneak state / hit point | #4089 |
| Right-click entities through one path (`useOn`/`mount`) | #4095 (based on #4089) |
| `activateBlock` cursor default / `activateItem` empty hand | #4090 |
| Arm swing after `use_item_on` | #4091 |
| Digging: per-tick swing, instant break, abort faces | #4094 (based on #4085) |
| `bot.attack` rejects self / item / xp orb | #4092 |
| `bot.dismount` holds sneak on 1.21.3+ | #4093 |

The branches merge in any order (the two stacked ones retarget to master automatically once their base merges); verified by merging all eleven in a shuffled order locally — the result is identical to this PR (plus #4061) and the full internal test suite passes on 1.8.8 → 26.1.

PR #4070, opened 2026-09-07
fix: answer ping at the start of the next tick, before movement packets

u9g:
Invariants of the ping answer:

- In the play state a `pong` is written at the start of a physics tick, before that tick's movement packets.
- A ping is answered within one tick even when physics is not ticking.
- Pongs are written in ping order, one per ping.

Vanilla reference: `ClientPacketListener.handlePing` calls `PacketUtils.ensureRunningOnSameThread`, so the pong is written while `Minecraft.runTick` drains its task queue, before the tick loop that runs `LocalPlayer.tick` and `sendPosition`.

Tests (`pong is written ...`): with the bot falling, the pong follows the movement packet of the tick that received the ping and is answered once; with no chunk loaded it is still answered within one tick. Pass on every tested version.

### DIFF
--- a/lib/plugins/game.js
+++ b/lib/plugins/game.js
@@ -139,10 +139,37 @@ function inject (bot, options) {
     bot.game.serverBrand = serverBrand
   })
 
-  // mimic the vanilla 1.17 client to prevent anticheat kicks
+  // Pongs are written in ping order; each ping is answered exactly once.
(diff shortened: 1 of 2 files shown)


u9g:
Folded into #4107, whose branch already carries this commit and test. The one-queue change there is what makes the ping timing correct on its own, so reviewing them together is simpler.

PR #4072, opened 2026-09-07
fix: track the container stateId per window

u9g:
Invariants of the container state id:

- Each window keeps the stateId of the last `window_items` or `set_slot` applied to that window.
- A `window_click` carries the stateId of the window it clicks; a window that has never been synced sends -1.
- A packet routed to the player inventory updates the player inventory's id, never an open container's.
- Entries are dropped when the window closes and on login and respawn.
- `_syncWindow` still sends -1 on purpose.

Vanilla reference: `AbstractContainerMenu` holds its own `stateId`; `ClientPacketListener.handleContainerContent` and `handleContainerSetSlot` apply containerId 0 to `player.inventoryMenu` and anything else to the open `containerMenu`; `MultiPlayerGameMode.handleInventoryMouseClick` sends the clicked menu's `getStateId()`.

Test (`clicks carry the stateId of the window they click, not the last one synced`): a container syncs with stateId 5, a player-inventory `set_slot` with stateId 9 arrives while it is open, and the click on the container carries 5 while a later click on the player inventory carries 9. Fails on master, passes with the change, on every tested version.

### DIFF
--- a/lib/plugins/inventory.js
+++ b/lib/plugins/inventory.js
@@ -30,11 +30,10 @@ function inject (bot, { hideErrors }) {
   let sequence = 0
 
   let nextActionNumber = 0 // < 1.17
-  let stateId = -1
-  if (bot.supportFeature('stateIdUsed')) {
-    const listener = packet => { stateId = packet.stateId }
-    bot._client.on('window_items', listener)
-    bot._client.on('set_slot', listener)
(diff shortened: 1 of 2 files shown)


u9g:
Folded into #4103, which carries this commit; the three branches rewrote the same login/respawn/close_window listeners in inventory.js and conflicted pairwise.

PR #4076, opened 2026-09-07
fix: write the enum names newer protocols declare for client_command and entity_action

u9g:
`packet_client_command.actionId` is a string mapper from 26.1 and `packet_entity_action.actionId` from 1.21.6, while older versions declare a plain varint. protodef's interpreted mapper writer has always thrown `0 is not in the mappings value` on the numeric form, and since ProtoDef-io/node-protodef#176 the compiled writer throws too, which ends the connection: on 26.1 that takes out `bot.respawn()` (so a bot never respawns after dying), `creative.getStats()`, and `bot.wake()` on any version from 1.21.6.

`lib/protocol_enum.js` reads which form the loaded version declares. `game.js` also wrote the field as `action`, which no version has, so the win_game respawn threw on every version.

Tests: `internalTest` now asserts the respawn and wake actions match what the version's schema declares, for every tested version.

### DIFF
--- a/lib/plugins/bed.js
+++ b/lib/plugins/bed.js
@@ -1,3 +1,4 @@
+const enumValue = require('../protocol_enum')
 const { Vec3 } = require('vec3')
 
 module.exports = inject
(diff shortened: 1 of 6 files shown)


u9g:
PrismarineJS/mineflayer#4077 does the same for `client_command`/`entity_action` and additionally fixes `use_entity.hand`, which this PR does not touch: on 26.1 that field is a mapper too, so `bot.activateEntity()` throws in the serializer and sends nothing. Whichever of the two lands, that hunk should be in it.

PR #4077, opened 2026-09-07
fix: send the action ids a protocol maps to names

u9g:
On 26.1 `client_command.actionId` and `use_entity.hand` are `mapper` fields:

```json
"packet_client_command": ["container", [{"name": "actionId", "type": ["mapper", {"type": "varint", "mappings": {"0": "perform_respawn", "1": "request_stats", "2": "request_gamerule_values"}}]}]]
"packet_use_entity":     ["container", [{"name": "target", ...}, {"name": "hand", "type": ["mapper", {"type": "varint", "mappings": {"0": "main_hand", "1": "off_hand"}}]}, ...]]
```

Writing the numeric id there throws inside the serializer:

```
Error: SizeOf error for undefined : 0 is not in the mappings value
    at Object.packet_use_entity ...
```

The packet is dropped on the floor, so on 26.1 the bot cannot respawn (`bot.respawn()`, the death handler, the win_game handler), cannot request stats, and `bot.activateEntity()` does nothing at all — right-clicking an NPC on a 26.1 server looked like the server ignoring us. `entity_action` has been string-mapped since 1.21.6 and `physics.js` already handles it, but `bed.js`'s `wake()` still wrote `actionId: 2`, which both fails to serialize there and means `stop_sprinting` under the new mapping.

`game.js` also wrote `{ action: 0 }`, a field name that does not exist in any protocol version, so the win_game respawn never serialized anywhere.

`lib/protocol_ids.js` asks the schema whether the field is a mapper rather than going through a feature, so this is right whether minecraft-data keeps 26.1's mappers (PrismarineJS/minecraft-data#1279) or drops them (PrismarineJS/minecraft-data#1278), and it needs no unreleased data. `bed.js` keeps `supportFeature('entityActionUsesStringMapper')` to match `physics.js`, which already uses it for the same packet.

Verified live: with this change `bot.activateEntity()` on a 26.1 server's NPC produces `use_entity {target, hand: 'main_hand', location, sneaking}` and the server answers, where before nothing went out.

Tests: three cases in `test/internalTest.js` covering every tested version. They resolve the packet out of the server's read path before asserting — an assertion thrown inside the packet listener wedges the connection and shows up as an `afterEach` timeout — and on 1.8-1.11, whose `use_entity` has no `hand` field at all, the assertion is just that the packet arrives.

### DIFF
--- a/lib/plugins/bed.js
+++ b/lib/plugins/bed.js
@@ -70,7 +70,7 @@ function inject (bot) {
     } else {
       bot._client.write('entity_action', {
         entityId: bot.entity.id,
-        actionId: 2,
+        actionId: bot.supportFeature('entityActionUsesStringMapper') ? 'leave_bed' : 2,
         jumpBoost: 0
       })
     }
--- a/lib/plugins/creative.js
+++ b/lib/plugins/creative.js
@@ -46,7 +46,7 @@ function inject (bot) {
         resolve()
       }, timeoutMs)
       pendingStatsRequests.push(request)
-      bot._client.write('client_command', bot.supportFeature('respawnIsPayload') ? { payload: 1 } : { actionId: 1 })
+      bot._client.write('client_command', bot.supportFeature('respawnIsPayload') ? { payload: 1 } : { actionId: bot.supportFeature('clientCommandUsesStringMapper') ? 'request_stats' : 1 })
     })
   }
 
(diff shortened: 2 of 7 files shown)


u9g:
PrismarineJS/mineflayer#4079 is the same fix. If PrismarineJS/minecraft-data#1278 lands, `use_entity.hand` needs no feature check, but `client_command.actionId` (26.1) and `entity_action.actionId` (1.21.6+) still do.

u9g:
PrismarineJS/mineflayer#4076 (opened first) covers `client_command` and `entity_action` with the same schema-introspection approach under a different helper name. What is only here: **`use_entity.hand`** in `inventory.js`, which is the one that actually breaks a bot — on 26.1 `bot.activateEntity()` serializes nothing at all, so right-clicking an NPC looks like the server ignoring you. If these are consolidated into #4076, that hunk and its test need to come along.

PR #4078, opened 2026-09-07
fix: answer a server teleport at the next tick, like the vanilla packet processor

u9g:
`ClientPacketListener.handleMovePlayer` starts with `PacketUtils.ensureRunningOnSameThread`, so the vanilla client does not act on a teleport when it arrives: it queues it and runs the handler when the client thread next drains its packet queue, at the top of a tick. mineflayer answered from inside the packet handler instead, microseconds after the bytes arrived.

That difference matters on a server that sends a position packet in response to a movement packet. The round trip becomes network RTT instead of one tick, so the exchange runs at hundreds of packets a second rather than twenty:

```
26503 <position  (0.5, 100, 0.5) id0
26503 >teleport_confirm id0
26503 >position_look (0.5, 100, 0.5)
26504 <position  (0.5, 100, 0.5) id0
26504 >teleport_confirm id0
26504 >position_look (0.5, 100, 0.5)
26505 <position ...
```

Measured against a BedWars server that pins players with a teleport every tick, over the same 30 s window: 923 teleports answered, peaking at 538 in one second, before the change; 59 after it, peaking at 27 (the tick rate). The anticheat kicks for the former.

The handler body moves into `handleTeleport` and the listener only queues; `tickPhysics` drains the queue before anything else, ahead of the entity and chunk guards so the first teleport of a login is still answered while chunks are loading.

Test in `test/internalTest.js` emits a `position` packet mid-tick the way #4070's ping test does and asserts nothing is written from the handler, then that the answer goes out on the next tick. It fails on master with `teleport_confirm`/`position_look` written synchronously.

### DIFF
--- a/lib/plugins/physics.js
+++ b/lib/plugins/physics.js
@@ -491,6 +491,9 @@ function inject (bot, { physicsEnabled, maxCatchupTicks }) {
   bot.on('respawn', () => { shouldUsePhysics = false })
   bot.on('login', () => {
     shouldUsePhysics = false
+    // A teleport still queued here belongs to the world the bot just left, and its id means nothing
+    // to the server it is about to talk to.
+    pendingTeleports.length = 0
     if (doPhysicsTimer === null) {
       lastPhysicsFrameTime = performance.now()
       doPhysicsTimer = setInterval(doPhysics, PHYSICS_INTERVAL_MS)
--- a/lib/plugins/physics.js
+++ b/lib/plugins/physics.js
@@ -372,9 +372,10 @@ function inject (bot, { physicsEnabled, maxCatchupTicks }) {
   })
 
   // player position and look (clientbound)
-  // The vanilla client hands every play packet to the client thread, which drains the queue at the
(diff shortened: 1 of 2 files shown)


u9g:
PrismarineJS/mineflayer#4080 is the same fix with a different drain point: it drains on `setImmediate` (end of the current socket read), this one drains at the top of the physics tick, which is where `Minecraft.runTick` calls `packetProcessor.processQueuedPackets()`.

u9g:
The `MC 1.16.5 1.17.1 1.18.2 1.19` shard is a flake on this branch, not a regression: master fails the identical assertion in the same shard (run 34086113235, `mineflayer_external 1.18.2v — Event spawn did not fire within timeout of 30000ms`), and the version that times out moves between runs — 1.19 on the first run here, 1.18.2 on the re-run. Every internal test passes on all seven shards, including 26.1.

u9g:
Folded into #4107, whose branch already carries this commit and its tests (including the every-queued-teleport one from the #4080 discussion). The shared arrival-ordered queue there replaces the per-packet queue this PR added.

PR #4080, opened 2026-09-07
physics: handle a clientbound teleport from the packet queue, not from the read

u9g:
Vanilla's `ClientPacketListener.handleMovePlayer` starts with `PacketUtils.ensureRunningOnSameThread`, so a teleport is applied and answered while `Minecraft.runTick` drains `packetProcessor.processQueuedPackets()` — never from the network read that carried it, and never between the packets of a tick.

mineflayer answered inline from the socket read. When a server sends several teleports in one read, each reply is written the moment its packet is parsed, so the replies interleave with the movement packet the physics tick is producing, in an order the vanilla client never sends.

This queues the packets and drains them on `setImmediate`: every packet of the current read is handled first, then the teleports are applied and answered together. The reply is still one `teleport_confirm` + `position_look` per teleport, still before the next tick's movement packet, so nothing about the packets themselves changes — only when they are written.

### Tests

- New: `answers a teleport from the packet queue, not from the read that carried it` — a `prependListener` on the clientbound `position` marks the window in which its listeners run, and a `bot._client.write` wrapper records whether each movement packet was written inside it. Asserts the teleport is answered and that nothing is written from that window. Passes on every tested version.
- `absolute position & relative position (velocity)` now reads velocity and position inside the `forcedMove` listener. The values it asserts are the ones set by the teleport; awaiting the event lets the next tick apply gravity before the assertion runs, which is a property of the `await`, not of the handler.
- `switchWorld respawn` waits for `forcedMove` instead of `bot.waitForTicks(1)`. It needs the teleport to have landed, and physics only starts ticking once that packet is handled, so a tick count is the wrong thing to wait on.

`npm run mocha_test` (internal): 784 passing, 0 failing, twice; `standard` clean.

### Where this showed up

On a 1.8 server behind ViaVersion (JartexNetwork BedWars, protocol 26.1) the client-side reply ordering feeds a server setback loop: the server resends `position` with `teleportId: 0` at the bot's own coordinates, the rate climbs, and an anticheat kicks. Standing still in one lobby, a 40 s window took **683** clientbound teleports with the old handler and **1** over 129 s with this one. Being straight about it: that server still storms in other windows, so this is a parity fix and not a cure for that kick — the reason to take it is that it is what the vanilla client does.

### DIFF
--- a/lib/plugins/physics.js
+++ b/lib/plugins/physics.js
@@ -96,6 +96,9 @@ function inject (bot, { physicsEnabled, maxCatchupTicks }) {
   })
 
   function cleanup () {
+    clearImmediate(positionDrain)
+    positionDrain = null
+    positionQueue.length = 0
     clearInterval(doPhysicsTimer)
     doPhysicsTimer = null
   }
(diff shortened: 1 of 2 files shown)


u9g:
superseded by #4078

Issue #1302, opened 2020-09-10
Block definition problem for digging


Denaryplanet:
- [x] I have carefully read the [FAQ](https://github.com/PrismarineJS/mineflayer/blob/master/docs/FAQ.md) and it does not contain an answer to my question.
- [x] I have carefully read the [API documentation](https://github.com/PrismarineJS/mineflayer/blob/master/docs/api.md) and it does not contain an answer to my question.
- [x] I have carefully read the [examples](https://github.com/PrismarineJS/mineflayer/tree/master/examples) and they do not contain an answer to my question.

## Versions
 - mineflayer: 2.28.1
 - server: vanilla/spigot/paper 1.12.2
 - node: 12.18.3

## Clear question
When trying to dig at the look, the bot gives out 3 blocks as undefined. There is cobblestone there, though. How can this problem be corrected?

When digging by coordinates everything is fine
## What did you try yet?

bot.blockInSight()
bot.dig()
bot.targetDigBlock

## Your current code

```js

function dig() {
	var block = bot.blockInSight(5, 5/16)
	if(bot.targetDigBlock){
		dig()
	}else{
		if (block == undefined) {
			console.log(block)
			dig()
		}else{
			console.log(block)
			bot.dig(block, onDiggingCompleted)
		}
	}
	function onDiggingCompleted (err) {
		if (err) {
			console.log(err.stack)
			return
		}
		dig()
	}
}

```

## Additional context
P.s cobblestone generator 5 blocks up

flymer5:
everything worked out for me, but I ran into this error. How can you solve it?
![image](https://user-images.githubusercontent.com/58649986/92715373-37c0cd80-f37f-11ea-8144-a2e65bcb00d5.png)

wvffle:
You're calling the dig function recursively, in case of `bot.targetDigBlock` or `block == undefined` you'll most likely loop yourself

flymer5:
> You're calling the dig function recursively, in case of `bot.targetDigBlock` or `block == undefined` you'll most likely loop yourself

I already found this and fixed it, but here is a new problem:
![image](https://user-images.githubusercontent.com/58649986/92734696-96477500-f39a-11ea-9571-9998390f18ca.png)

wvffle:
No, you havent. This is obviously the case. If you have 'fixed' it then show us your code again.

flymer5:
> No, you havent. This is obviously the case. If you have 'fixed' it then show us your code again.
```js
[...]

PR #4081, opened 2026-09-07
fix: give teamRemoved the team that was removed

u9g:
`teamRemoved` reads `teams[teamName]` after `delete teams[teamName]`, so the event has always fired with `undefined`:

```js
delete teams[teamName]
bot.emit("teamRemoved", teams[teamName])
```

`docs/api.md` documents it as `"teamRemoved" (team)`, and the sibling events (`teamCreated`, `teamUpdated`, `teamMemberAdded`, `teamMemberRemoved`) all pass a live team, so a listener has no way to tell which team went away. The team object is still fully populated at that point, so the fix is to emit the local `team` that the handler already holds — its `members` and `name` are readable in the listener, and `bot.teams` / `bot.teamMap` are already cleaned up by then.

Test `teamRemoved carries the team that was removed` creates a team with a member, removes it, and asserts the event argument is the removed team with its members intact while `bot.teams`/`bot.teamMap` no longer mention it. It fails on master on all 28 tested versions (`AssertionError: teamRemoved is given the team, not undefined`) and passes with the fix.

Depends on PrismarineJS/minecraft-data#1302, which maps 1.21.5's `nameTagVisibility`/`collisionRule` to their enum names like every other version and adds a `teamModeUsesStringMapper` feature for the 1.21.6+ `mode` string mapper. The test uses that feature instead of a version compare, so CI needs the minecraft-data bump before the 1.21.6+ cases pass. With that data the teams suite passes on all 28 versions locally.


test/internalTest.js
```diff
@@ -1484,6 +1449,50 @@ for (const supportedVersion of mineflayer.testedVersions) {
...
+          nameTagVisibility: enumRules ? 0 : 'always',
+          collisionRule: enumRules ? 0 : 'always',
+          color: 0,
+          formatting: 0,
+          players
+        }
+      }
+
+      it('teamRemoved carries the team that was removed', async () => {
+        const packetName = bot.supportFeature('teamUsesScoreboard') ? 'scoreboard_team' : 'teams'
```
extremeheat, line 1476: Seems like this is also something that should be fix upstream in data, if the packets didn't change much

u9g: done in https://github.com/PrismarineJS/minecraft-data/pull/1302

u9g: Upstreamed in PrismarineJS/minecraft-data#1302: 1.21.5 now maps `nameTagVisibility`/`collisionRule` to their names like every other version, plus a `teamModeUsesStringMapper` feature for the 1.21.6+ `mode` mapper. The test now uses that feature and has no version special-casing. The `mode` number-to-string and `friendlyFire`-to-`flags` differences are renames across every pre-1.21.6 version, so I left those for a separate breaking data change.

VasilisDragon:
The event fix looks right. One more fixture to update with #1302: `scoreboard reset on login` still sends `0` for both 1.21.5 rules. The [upstream ProtoDef compiler](https://github.com/ProtoDef-io/node-protodef/tree/0d7b6e171fe22aaa237a62e7643a5b2ca0dbde2c) rejects them after the mapper change; the published 1.19.0 compiler still accepts them. Both can be `'always'`, like the new team test.

PR #4085, opened 2026-09-07
use_item and use_item_on like the vanilla client: shared prediction sequence, cursor on the clicked face, no empty-hand use, swing after the packet

u9g:
The `use_item` / `use_item_on` packets match what the vanilla client sends (`Minecraft.startUseItem`, `MultiPlayerGameMode.useItemOn` / `useItem`, `BlockStatePredictionHandler`):

- **One block prediction sequence.** Vanilla keeps a single counter in `BlockStatePredictionHandler` that dig start/stop, `use_item_on` and `use_item` all draw from (first value 1); abort, release and drop actions send 0. mineflayer only counted `use_item` and sent 0 elsewhere, so the server's block change acks did not line up. Adds `lib/plugins/sequence.js` (`bot._nextSequence()`; versions before 1.19 have no sequence field, so the value is simply not serialized there) and wires it into `block_dig` start/stop, `use_item_on` (`generic_place`, `activateBlock`) and `use_item`.
- **`activateBlock`'s cursor defaults to the centre of the clicked face** (`0.5 + face * 0.5` per axis): the cursor is where the crosshair hits the block, and the old `(0.5, 0.5, 0.5)` is inside the block, impossible for a real client.
- **`activateItem` no-ops with an empty hand**, like `startUseItem` (`!itemStack.isEmpty()`). The pre-existing `activateItem rotation` test now holds an item.
- **The arm swings after `use_item_on` when placing**, not before: vanilla only swings once `useItemOn` returns a success.

Tests: `shares one pre-incremented counter across use_item and use_item_on, 0 on release`, `defaults the cursor to the centre of the clicked face and swings after use_item_on`, `does nothing with an empty hand`, `swings the arm after use_item_on`.

#4094 (digging) is based on this branch.

Split from #4066 (the interaction-parity audit); absorbs #4090 and #4091.

### DIFF
--- a/lib/loader.js
+++ b/lib/loader.js
@@ -42,7 +42,8 @@ const plugins = {
   anvil: require('./plugins/anvil'),
   place_entity: require('./plugins/place_entity'),
   generic_place: require('./plugins/generic_place'),
-  particle: require('./plugins/particle')
+  particle: require('./plugins/particle'),
+  sequence: require('./plugins/sequence')
 }
 
 const minecraftData = require('minecraft-data')
--- a/lib/plugins/sequence.js
+++ b/lib/plugins/sequence.js
@@ -1,0 +1,14 @@
+module.exports = inject
+
+// One sequence counter for dig start/stop, use_item_on and use_item. Sent values
+// start at 1; abort, release and drop actions send 0 and must not consume a value.
+function inject (bot) {
+  const hasSequence = bot.registry.isNewerOrEqualTo('1.19')
+  let sequence = 0
+
+  bot._nextSequence = () => {
+    if (!hasSequence) return 0
+    sequence++
+    return sequence
+  }
+}
--- a/test/internalTest.js
+++ b/test/internalTest.js
@@ -69,6 +69,9 @@ for (const supportedVersion of mineflayer.testedVersions) {
         port: PORT
       })
       bot.test = {}
+      // Plugins are injected on a timer after createBot, which can lose the
+      // race against the mock server's playerJoin
+      bot.test.pluginsLoaded = new Promise(resolve => bot.once('inject_allowed', resolve))
 
       bot.test.buildChunk = () => {
         if (bot.supportFeature('tallWorld')) {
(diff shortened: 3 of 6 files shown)


lib/plugins/sequence.js
```diff
@@ -0,0 +1,14 @@
+module.exports = inject
```
extremeheat, line 1: Does this need its own plugin file ? It doesn't make sense to put anywhere else?

test/internalTest.js
```diff
@@ -1514,6 +1595,93 @@ for (const supportedVersion of mineflayer.testedVersions) {
...
     })
 
+    describe('block prediction sequence', () => {
+      it('shares one pre-incremented counter across use_item and use_item_on, 0 on release', function (done) {
+        const useItemFields = registry.protocol?.play?.toServer?.types?.packet_use_item?.[1]
+        if (!useItemFields?.some(f => f.name === 'sequence')) {
+          this.skip()
+          return
+        }
+        const Item = require('prismarine-item')(registry)
```
extremeheat, line 1605: Declare this higher up

test/internalTest.js
```diff
@@ -1404,6 +1462,29 @@ for (const supportedVersion of mineflayer.testedVersions) {
       })
     })
 
+    describe('generic place', () => {
+      it('swings the arm after use_item_on', (done) => {
+        const Item = require('prismarine-item')(registry)
```
extremeheat, line 1467: Declare this higher up

lib/plugins/sequence.js
```diff
@@ -0,0 +1,14 @@
...
+
+// One sequence counter for dig start/stop, use_item_on and use_item. Sent values
+// start at 1; abort, release and drop actions send 0 and must not consume a value.
+function inject (bot) {
+  const hasSequence = bot.registry.isNewerOrEqualTo('1.19')
+  let sequence = 0
+
+  bot._nextSequence = () => {
```
rom1504, line 9: **Astra agent review — AI-generated, not manually written by the maintainer.**

The shared counter still misses the boat `use_item` in `lib/plugins/place_entity.js`: `placeEntity()` calls `_genericPlace()` (consuming sequence N) and then writes `use_item` with only `hand`, so the second prediction does not consume N+1. #4114 repairs that payload but currently supplies a constant 0, so merging both as written still leaves this gap. Please wire that sender into `_nextSequence()` when integrating the two PRs, and cover an interleaved boat placement/item-use sequence through the real serializer (including both packets emitted by boat placement).

rom1504 (commented):

Reviewed at the maintainer’s request using static inspection of the current changes, interaction call sites, tests, and existing review discussion. One remaining shared-counter integration gap is noted inline; no runtime tests were run for this PR.

lib/plugins/sequence.js
u9g: Moved it into `createBot` in https://github.com/PrismarineJS/mineflayer/commit/67be68e. Digging, placing and item use all share it, and as a plugin it could be disabled while the others still call it.

test/internalTest.js
u9g: Done in https://github.com/PrismarineJS/mineflayer/commit/b0967e2, `Item` is declared once per version at the top.

test/internalTest.js
u9g: Done in https://github.com/PrismarineJS/mineflayer/commit/b0967e2, same as above.

lib/plugins/sequence.js
u9g: The boat's `use_item` now takes `_nextSequence()` and sends the rotation (https://github.com/PrismarineJS/mineflayer/commit/3d2d913). There's a new test for item use, release, boat placement and item use again through the real serializer, expecting 1, 0, 2, 3, 4 (https://github.com/PrismarineJS/mineflayer/commit/b0967e2). This replaces the `sequence: 0` in #4114.

rom1504 (commented):

Looks ready to merge from this review at ec0d114. The shared counter now lives in createBot, and the boat sender consumes its own value and includes rotation, addressing the earlier missing-consumer finding. I ran the focused internal interaction suites across all 28 tested versions: 127 passing, 41 version skips, including the boat interleave through the real compiled serializer. Current CI is green. These were local NMP/codec checks, not a new vanilla-server placement run.

Skills used: `prismarine-lifecycle-action-review` traced every sequence consumer and the zero-valued release; `prismarine-protocol-data-review` checked the real codec; `prismarine-architecture-review` checked the shared owner and optional-plugin concern; `prismarine-review` deduplicated the repaired findings.

lib/loader.js
```diff
@@ -77,6 +77,11 @@ function createBot (options = {}) {
   const bot = new EventEmitter()
   bot._client = options.client
   bot.end = (reason) => bot._client.end(reason)
+  // One block prediction sequence shared by dig start/stop, use_item_on and use_item. Sent values
```
rom1504: This file is completely generic today and I think it should stay this way. Can we put that in a relevant plugin?

u9g: Moved it to `lib/plugins/sequence.js` in 63411b6 — `createBot` is back to exactly what it was, and the counter lives with the rest of the protocol handling.

It was a plugin at first; I pulled it into `createBot` because no single plugin owns it (dig, place and item use all draw from it) and because as a plugin it can be disabled while its callers stay loaded. But that is already true of the rest — `digging` needs `blocks` for `bot.world`, `generic_place` needs `inventory` — so it is not a reason to single this one out, and keeping `loader.js` generic is worth more.

lib/plugins/sequence.js
rom1504: This should be in dig plugin probably, but I think it's ok in its own plugin

Issue #1284, opened 2020-08-31
help with windows


laskgay:
- [x] I have carefully read the [FAQ](https://github.com/PrismarineJS/mineflayer/blob/master/docs/FAQ.md) and it does not contain an answer to my question.
- [x] I have carefully read the [API documentation](https://github.com/PrismarineJS/mineflayer/blob/master/docs/api.md) and it does not contain an answer to my question.
- [x] I have carefully read the [examples](https://github.com/PrismarineJS/mineflayer/tree/master/examples) and they do not contain an answer to my question.

## Versions
 - mineflayer: 1.12.2
 - server: vanilla/spigot/paper     spigot
 - node: 12.16.1

## Clear question
A clear question, with as much context as possible.
What are you building? What problem are you trying to solve?

## What did you try yet?

Did you try any method from the API?
Did you try any example? Any error from those?

## Your current code

Please put here any custom code you tried yet.

```js

/*
Some code here, replace this
*/

```

## Additional context

How can I find out the list and description of items in the window when I try to output, undefined

rom1504:
Please fill the template

PR #4086, opened 2026-09-07
Send the client_command respawn after win_game with the actionId field

u9g:
The packet field is `actionId` (`payload` on 1.8), as `health.js` already sends; the `action` key was ignored and the varint happened to encode as 0.

### DIFF
--- a/lib/plugins/game.js
+++ b/lib/plugins/game.js
@@ -123,7 +123,7 @@ function inject (bot, options) {
 
   bot._client.on('game_state_change', (packet) => {
     if ((packet.reason === 4 || packet.reason === 'win_game') && packet.gameMode === 1) {
-      bot._client.write('client_command', { action: 0 })
+      bot._client.write('client_command', bot.supportFeature('respawnIsPayload') ? { payload: 0 } : { actionId: 0 })
     }
     if ((packet.reason === 3) || (packet.reason === 'change_game_mode')) {
       bot.game.gameMode = parseGameMode(packet.gameMode)


rom1504 (commented):

Reviewed at the maintainer's request. I inspected the current change and discussion, checked minecraft-data #1284's current status and the current 26.1 schema, and locally reproduced the numeric-versus-named respawn serialization with ProtoDef. One compatibility finding below; I did not run a live server test.

lib/plugins/game.js
```diff
@@ -123,7 +123,7 @@ function inject (bot, options) {
 
   bot._client.on('game_state_change', (packet) => {
     if ((packet.reason === 4 || packet.reason === 'win_game') && packet.gameMode === 1) {
-      bot._client.write('client_command', { action: 0 })
+      bot._client.write('client_command', bot.supportFeature('respawnIsPayload') ? { payload: 0 } : { actionId: 0 })
```
rom1504, line 126: **Astra agent review — AI-generated, not manually written by the maintainer.**

Could this select the mapped respawn value on 26.1 as well? The current minecraft-data 26.1 schema still maps `actionId` to `perform_respawn` / `request_stats` / `request_gamerule_values`; #1284 is still open. Serializing this PR's `{ actionId: 0 }` against that schema fails with `0 is not in the mappings value`, whereas `{ actionId: 'perform_respawn' }` produces `00`. Thus the win-game path still fails to send its respawn on 26.1. Please retain the legacy numeric/payload branches and cover 26.1 with a serialized win-game regression using the supported mapper contract.

rom1504: **Astra agent review — AI-generated, not manually written by the maintainer.**

Correction: I withdraw the serialization-failure finding above. It came from the isolated interpreted ProtoDef mapper, which does not match node-minecraft-protocol's default compiled path. Rechecking this PR's actual `win_game` handler with released `minecraft-protocol@1.67.0`, `minecraft-data@3.114.0` and `protodef@1.19.0`, the 26.1 packet containing `actionId: 0` encodes successfully and decodes as `perform_respawn`. The compiled mapper explicitly permits raw numeric values. The corresponding 1.8.9 and 1.21.4 packet checks also pass. This was production packet encoding/decoding, not a vanilla-server run.

Skills used: prismarine-protocol-data-review prompted checking the real compiled codec and allowed dependency versions; prismarine-review required correcting the earlier objection when that evidence disproved its premise.

PR #4089, opened 2026-09-07
Right-click entities like the vanilla client: interact pair, real sneak state, hit point; useOn and mount share the path

u9g:
`activateEntity` / `activateEntityAt` / `useOn` / `mount` send what the vanilla client sends on a right click (`MultiPlayerGameMode`, `Minecraft.startUseItem`):

- Before 26.1: `use_entity` interact_at with the hit point relative to the entity, then interact. Both carry the bot's **real** sneak state, since the server applies the packet's secondary-action flag as the player's shift state.
- 26.1+: one `use_entity` carrying the `lpVec3 location`, chosen once at inject through `supportFeature('useEntityHasLocation')` (PrismarineJS/minecraft-data#1282). The hand is the numeric id on every version; the compiled 26.1 codec already accepts it.
- The hit vector is where the line from the eye to the entity's middle enters its bounding box (`interactAt` sends `entityHitResult.getLocation()` minus the entity position, a point on the box surface). An entity type minecraft-data has no size for gets its position. `activateEntity` looks at the hit point it sends.
- `activateEntityAt` no longer mutates the caller's position vector.
- `bot.useOn` and `bot.mount` used to send a lone interact with the hit location fixed at the entity's feet. They now live beside `activateEntity` in `inventory.js` and go through the same sender, so `entities.js` no longer has a second, divergent `use_entity` writer. `attack` gets its own sender carrying only the sneak state (its own packet on 26.1), chosen once at inject from `attackUsesOwnPacket`.

Tests run the recorded params through the real serializer, so a version's packet shape is checked: `activateEntity sends the vanilla interact packets with the sneak state`, `useOn and mount send the same interact pair as activateEntity`, the hitbox-face and unknown-size cases.

Depends on minecraft-data#1282 (merged): 26.1 fails until a minecraft-data release carries it.

Split from #4066 (the interaction-parity audit); absorbs #4095.

### DIFF
--- a/lib/plugins/inventory.js
+++ b/lib/plugins/inventory.js
@@ -245,71 +245,67 @@ function inject (bot, { hideErrors }) {
     bot.swingArm()
   }
 
-  // Where the line from the eye to the entity's middle enters its bounding box, relative to the
-  // entity. Falls back to the middle itself for an entity of unknown size.
+  // Where the line from the eye to the middle of the entity's bounding box enters the box, relative
+  // to the entity: the hit vanilla reports for a player looking at the entity's middle.
   function hitVector (entity) {
-    const middle = new Vec3(0, (entity.height ?? 0) / 2, 0)
-    const width = entity.width ?? 0
-    const height = entity.height ?? 0
-    if (width <= 0 || height <= 0) return middle
+    // An entity type minecraft-data does not know has no size, so its position stands in.
+    if (entity.width == null || entity.height == null) return new Vec3(0, 0, 0)
+    const middle = new Vec3(0, entity.height / 2, 0)
     const eye = bot.entity.position.offset(0, bot.entity.eyeHeight, 0).minus(entity.position)
-    const dir = middle.minus(eye)
-    const half = width / 2
-    let near = 0
-    for (const [from, to, o, d] of [[-half, half, eye.x, dir.x], [0, height, eye.y, dir.y], [-half, half, eye.z, dir.z]]) {
-      if (d === 0) continue
-      const t1 = (from - o) / d
-      const t2 = (to - o) / d
-      near = Math.max(near, Math.min(t1, t2))
+    const toMiddle = middle.minus(eye)
+    // Slab test: parameters along the line run from 0 at the eye to 1 at the middle, each axis bounds
+    // the box to a parameter interval, and the line enters the box at the latest interval start. An
+    // axis the line does not move along already has the middle's coordinate, inside the box.
+    const half = entity.width / 2
+    let entry = 0
+    for (const [min, max, at, step] of [[-half, half, eye.x, toMiddle.x], [0, entity.height, eye.y, toMiddle.y], [-half, half, eye.z, toMiddle.z]]) {
+      if (step === 0) continue
+      entry = Math.max(entry, Math.min((min - at) / step, (max - at) / step))
     }
-    if (!(near > 0) || near > 1) return middle
-    return eye.plus(dir.scaled(near))
+    // Every interval starts behind the eye: the eye is inside the box.
+    if (entry === 0) return middle
+    return eye.plus(toMiddle.scaled(entry))
   }
 
+  // 26.1 folded interact_at into interact: one use_entity carries the hit point. Earlier versions send
+  // interact_at, then interact. The server applies the packet's sneak flag as the player's shift state.
+  const interactEntity = bot.supportFeature('useEntityHasLocation')
+    ? (entity, hit) => {
+        bot._client.write('use_entity', {
+          target: entity.id,
+          hand: 0, // main hand
+          location: hit,
+          sneaking: bot.getControlState('sneak')
+        })
+      }
+    : (entity, hit) => {
+        const sneaking = bot.getControlState('sneak')
+        bot._client.write('use_entity', {
+          target: entity.id,
+          mouse: 2, // interact at
+          x: hit.x,
+          y: hit.y,
+          z: hit.z,
+          hand: 0, // main hand
+          sneaking
+        })
+        bot._client.write('use_entity', {
+          target: entity.id,
+          mouse: 0, // interact
+          hand: 0, // main hand
+          sneaking
+        })
+      }
+
   async function activateEntity (entity) {
-    await bot.lookAt(entity.position.offset(0, (entity.height ?? 0) / 2, 0), false)
-    interactEntity(entity, hitVector(entity))
+    const hit = hitVector(entity)
+    await bot.lookAt(entity.position.plus(hit), false)
+    interactEntity(entity, hit)
   }
 
   async function activateEntityAt (entity, position) {
     await bot.lookAt(position, false)
     interactEntity(entity, position.minus(entity.position))
-  }
-
-  // 26.1+ has a single interact packet carrying the hit point; earlier
-  // versions send interact_at (hit point relative to the entity) followed by
-  // interact, and both must carry the same sneak state
-  const useEntityFields = bot.registry.protocol.play.toServer.types.packet_use_entity[1]
-  const useEntityHasLocation = useEntityFields.some(field => field.name === 'location')
-  // 26.1 declares the hand as a mapper; every earlier version takes the id.
-  const handType = useEntityFields.find(field => field.name === 'hand')?.type
-  const mainHand = Array.isArray(handType) && handType[0] === 'mapper' ? 'main_hand' : 0
-  function interactEntity (entity, hit) {
-    const sneaking = bot.getControlState('sneak')
-    if (useEntityHasLocation) {
-      bot._client.write('use_entity', {
-        target: entity.id,
-        hand: mainHand,
-        location: hit,
-        sneaking
-      })
-      return
-    }
-    bot._client.write('use_entity', {
-      target: entity.id,
-      mouse: 2, // interact at
-      x: hit.x,
-      y: hit.y,
-      z: hit.z,
-      hand: mainHand,
-      sneaking
-    })
-    bot._client.write('use_entity', {
-      target: entity.id,
-      mouse: 0, // interact
-      hand: mainHand,
-      sneaking
-    })
   }
 
   async function transfer (options) {
--- a/lib/plugins/inventory.js
+++ b/lib/plugins/inventory.js
@@ -248,8 +248,9 @@ function inject (bot, { hideErrors }) {
   // Where the line from the eye to the middle of the entity's bounding box enters the box, relative
   // to the entity: the hit vanilla reports for a player looking at the entity's middle.
   function hitVector (entity) {
-    // An entity type minecraft-data does not know has no size, so its position stands in.
-    if (entity.width == null || entity.height == null) return new Vec3(0, 0, 0)
+    // prismarine-entity starts every entity at a zero size and only a type minecraft-data knows is
+    // given one, so a flat box means the bot does not know the size: the entity position stands in.
+    if (!entity.width || !entity.height) return new Vec3(0, 0, 0)
     const middle = new Vec3(0, entity.height / 2, 0)
     const eye = bot.entity.position.offset(0, bot.entity.eyeHeight, 0).minus(entity.position)
     const toMiddle = middle.minus(eye)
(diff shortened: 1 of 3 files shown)


lib/plugins/inventory.js
```diff
@@ -245,30 +245,70 @@ function inject (bot, { hideErrors }) {
...
+  function hitVector (entity) {
+    const middle = new Vec3(0, (entity.height ?? 0) / 2, 0)
+    const width = entity.width ?? 0
+    const height = entity.height ?? 0
+    if (width <= 0 || height <= 0) return middle
+    const eye = bot.entity.position.offset(0, bot.entity.eyeHeight, 0).minus(entity.position)
+    const dir = middle.minus(eye)
+    const half = width / 2
+    let near = 0
+    for (const [from, to, o, d] of [[-half, half, eye.x, dir.x], [0, height, eye.y, dir.y], [-half, half, eye.z, dir.z]]) {
```
u9g: better names here for o and d

lib/plugins/inventory.js
```diff
@@ -245,30 +245,70 @@ function inject (bot, { hideErrors }) {
     bot.swingArm()
   }
 
+  // Where the line from the eye to the entity's middle enters its bounding box, relative to the
+  // entity. Falls back to the middle itself for an entity of unknown size.
+  function hitVector (entity) {
+    const middle = new Vec3(0, (entity.height ?? 0) / 2, 0)
+    const width = entity.width ?? 0
+    const height = entity.height ?? 0
+    if (width <= 0 || height <= 0) return middle
```
u9g: Whats a case that this can happen? If no, remove

lib/plugins/inventory.js
u9g: Need a comment explaining what this loop does

lib/plugins/inventory.js
```diff
@@ -245,30 +245,70 @@ function inject (bot, { hideErrors }) {
...
-    // TODO: tell the server that we are not sneaking while doing this
-    await bot.lookAt(entity.position.offset(0, 1, 0), false)
-    bot._client.write('use_entity', {
-      target: entity.id,
-      mouse: 0, // interact with entity
-      sneaking: false,
-      hand: 0, // interact with the main hand
-      location: new Vec3(0, 0, 0)
-    })
+    await bot.lookAt(entity.position.offset(0, (entity.height ?? 0) / 2, 0), false)
```
u9g: isnt entity.height??0)/2 problematic?

lib/plugins/inventory.js
```diff
@@ -245,30 +245,70 @@ function inject (bot, { hideErrors }) {
...
-    // TODO: tell the server that we are not sneaking while doing this
     await bot.lookAt(position, false)
+    interactEntity(entity, position.minus(entity.position))
+  }
+
+  // 26.1+ has a single interact packet carrying the hit point; earlier
+  // versions send interact_at (hit point relative to the entity) followed by
+  // interact, and both must carry the same sneak state
+  const useEntityFields = bot.registry.protocol.play.toServer.types.packet_use_entity[1]
+  const useEntityHasLocation = useEntityFields.some(field => field.name === 'location')
```
u9g: Supportfeature

lib/plugins/inventory.js
```diff
@@ -245,30 +245,70 @@ function inject (bot, { hideErrors }) {
...
+  // 26.1 declares the hand as a mapper; every earlier version takes the id.
+  const handType = useEntityFields.find(field => field.name === 'hand')?.type
```
u9g: Support feature

lib/plugins/inventory.js
```diff
@@ -245,30 +245,70 @@ function inject (bot, { hideErrors }) {
...
+  const mainHand = Array.isArray(handType) && handType[0] === 'mapper' ? 'main_hand' : 0
```
u9g: Supportfeature

lib/plugins/inventory.js
```diff
@@ -245,30 +245,70 @@ function inject (bot, { hideErrors }) {
...
+  function interactEntity (entity, hit) {
+    const sneaking = bot.getControlState('sneak')
+    if (useEntityHasLocation) {
```
u9g: Should we just change what function is pointed to once based on the boolean rather than if’ing every time?

test/internalTest.js
```diff
@@ -69,6 +69,9 @@ for (const supportedVersion of mineflayer.testedVersions) {
         port: PORT
       })
       bot.test = {}
+      // Plugins are injected on a timer after createBot, which can lose the
+      // race against the mock server's playerJoin
+      bot.test.pluginsLoaded = new Promise(resolve => bot.once('inject_allowed', resolve))
```
u9g: This is not the right fix, let’s think of a better one

lib/plugins/inventory.js
u9g: Renamed: the tuple is now `[min, max, at, step]` — the slab's bounds on that axis, the eye's coordinate, and how far the coordinate moves over the whole line. https://github.com/PrismarineJS/mineflayer/blob/b4f2603f41faf25ee4156dbc9fe0d20250c56112/lib/plugins/inventory.js#L257-L264

u9g: Added — it's a slab test, and the comment above the loop says what the parameter means and why an axis with `step === 0` is skipped. https://github.com/PrismarineJS/mineflayer/blob/b4f2603f41faf25ee4156dbc9fe0d20250c56112/lib/plugins/inventory.js#L256-L264

lib/plugins/inventory.js
u9g: Good catch — it can happen, but not the way I'd written it. prismarine-entity constructs every entity with `width = 0, height = 0` and only a type minecraft-data knows gets a real size, so an unknown type arrives with a flat box, never a missing field. The `== null` guard I replaced it with never fired, and the unknown-size test asserted on a hand-built entity literal with no size field — the one shape a spawned entity never has — so it passed without covering the case. Now `!entity.width || !entity.height`, and the test runs both shapes. https://github.com/PrismarineJS/mineflayer/blob/b4f2603f41faf25ee4156dbc9fe0d20250c56112/lib/plugins/inventory.js#L250-L253

lib/plugins/inventory.js
u9g: Gone. `activateEntity` now looks at the point it sends: `hitVector` is computed once, the look target is `entity.position.plus(hit)`, and the unknown-size fallback is a single `Vec3(0, 0, 0)` inside `hitVector` instead of a `?? 0` at each use. https://github.com/PrismarineJS/mineflayer/blob/b4f2603f41faf25ee4156dbc9fe0d20250c56112/lib/plugins/inventory.js#L301-L305

lib/plugins/inventory.js
u9g: Done — `bot.supportFeature('useEntityHasLocation')`, added in PrismarineJS/minecraft-data#1282. The protocol-shape sniffing is gone. https://github.com/PrismarineJS/mineflayer/blob/b4f2603f41faf25ee4156dbc9fe0d20250c56112/lib/plugins/inventory.js#L271-L273

lib/plugins/inventory.js
u9g: Done. The `handType` sniffing is gone entirely: the hand is the numeric id on every version once PrismarineJS/minecraft-data#1278 makes 26.1's field a plain varint, so it's just `hand: 0`.

lib/plugins/inventory.js
u9g: Same — the mapper detection is gone, see above.

lib/plugins/inventory.js
u9g: Yes. `interactEntity` is now picked once at inject from the feature, and `attack` got the same treatment with `attackUsesOwnPacket` so `entities.js` no longer carries a second `use_entity` writer. https://github.com/PrismarineJS/mineflayer/blob/b4f2603f41faf25ee4156dbc9fe0d20250c56112/lib/plugins/inventory.js#L271-L299 and https://github.com/PrismarineJS/mineflayer/blob/b4f2603f41faf25ee4156dbc9fe0d20250c56112/lib/plugins/entities.js#L836-L839

test/internalTest.js
u9g: Agreed, and it's dropped — no `inject_allowed` hook, no `bot.test.pluginsLoaded`. The tests wait on the login they themselves trigger (`once(bot, 'login')` around `client.write('login', …)`), which is the same pattern the rest of the file uses, so there's no race to lose.

That scaffolding is now one helper shared by the four interaction tests, and the `try` wraps the whole body rather than just the assertions. Before, a write the version can't serialize rejected the async `playerJoin` handler, `done()` was never called, and mocha reported a bare 10s timeout; now it reports the serializer error at the line that sent the packet. That's what turns the current 26.1 failure from "Timeout of 10000ms exceeded" into "SizeOf error for undefined" — still the minecraft-data#1282 blocker, but legible. https://github.com/PrismarineJS/mineflayer/blob/b4f2603f41faf25ee4156dbc9fe0d20250c56112/test/internalTest.js#L756-L777

rom1504 (commented):

Reviewed at the maintainer’s request. Validation: Static inspection of current PR and current minecraft-data 26.1 schema; refreshed #1278/#1282 status; local ProtoDef reproduction of exact hand mapper. Full 26.1 serializer was unavailable in local installed minecraft-data. One actionable compatibility change is noted inline.

lib/plugins/inventory.js
```diff
@@ -245,31 +245,76 @@ function inject (bot, { hideErrors }) {
...
+    return eye.plus(toMiddle.scaled(entry))
+  }
+
+  // 26.1 folded interact_at into interact: one use_entity carries the hit point. Earlier versions send
+  // interact_at, then interact. The server applies the packet's sneak flag as the player's shift state.
+  const interactEntity = bot.supportFeature('useEntityHasLocation')
+    ? (entity, hit) => {
+        bot._client.write('use_entity', {
+          target: entity.id,
+          hand: 0, // main hand
```
rom1504, line 277: **Astra agent review — AI-generated, not manually written by the maintainer.**

Please use the existing 26.1 hand enum here (`hand: 'main_hand'`) while keeping numeric hands in the older-version branch. The current minecraft-data 26.1 schema maps the varint to `main_hand`/`off_hand`; a local ProtoDef reproduction of that exact mapper rejects `0` with “0 is not in the mappings value”, whereas `main_hand` encodes as byte `00`. #1282 is now merged, while #1278 remains open. This can work with the current schema without requiring its mapper to be removed; keep the real-serializer test against that schema to verify the branch.

rom1504: **Astra agent review — AI-generated, not manually written by the maintainer.**

Correction: I withdraw the claim that `hand: 0` fails in this PR's production serializer. node-minecraft-protocol defaults to compiled ProtoDef, whose released mapper permits raw numeric values; my earlier isolated interpreted-mapper check did not represent that path. With `minecraft-protocol@1.67.0` / `minecraft-data@3.114.0` / `protodef@1.19.0`, a valid 26.1 `use_entity` packet using `0` or `main_hand` produces identical bytes and decodes as `main_hand`; `1` and `off_hand` also agree. The same checks pass with protocol 1.68.0 and data 3.116.0. This verifies the hand encoding; the separately documented `useEntityHasLocation` data-release prerequisite still needs to be accounted for when exercising the complete interaction path.

Skills used: prismarine-protocol-data-review guided the actual compiled-codec and released-dependency comparison; prismarine-review required retracting the earlier finding after its production-failure premise was disproved.

u9g: Thanks for rechecking against the compiled codec. `hand: 0` stays, since it is the numeric id on every version and, as the correction shows, the compiled 26.1 mapper already takes it. That also means this PR no longer waits on minecraft-data#1278; I updated the description.

The remaining 26.1 failures (the four entity-interaction tests and the external `trade` test, which activates a villager) are all `SizeOf error … reading 'x'`. They come from `useEntityHasLocation` being false: #1282 merged on 2026-09-14, and the latest minecraft-data release, 3.116.0, is from 2026-09-05. They should pass once a release carries #1282.

PR #4090, opened 2026-09-07
Default activateBlock's cursor to the clicked face and skip activateItem with an empty hand

u9g:
- `activateBlock`'s cursor defaults to the centre of the **clicked face** (`0.5 + face * 0.5` per axis): the cursor is where the crosshair hits the block, and the old `(0.5, 0.5, 0.5)` is inside the block, impossible for a real client.
- `activateItem` no-ops with an empty hand, like vanilla `startUseItem` (`!itemStack.isEmpty()`). The pre-existing `activateItem rotation` test now holds an item, since the no-op would otherwise skip its `use_item`.

Tests: `defaults the cursor to the centre of the clicked face and swings after use_item_on`, `does nothing with an empty hand`.

### DIFF
--- a/lib/plugins/inventory.js
+++ b/lib/plugins/inventory.js
@@ -118,6 +118,8 @@ function inject (bot, { hideErrors }) {
   }
 
   function activateItem (offHand = false) {
+    // use_item must not be sent for an empty hand
+    if (!(offHand ? bot.inventory.slots[45] : bot.heldItem)) return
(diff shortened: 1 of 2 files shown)


u9g:
Folded into #4085 (cherry-picked onto its branch, b142e6cd): same `activateItem` / `activateBlock` lines as the sequence change, so they conflicted as separate PRs.

PR #4091, opened 2026-09-07
Swing the arm after use_item_on when placing, not before

u9g:
Vanilla only swings once `useItemOn` returns a success, so the `arm_animation` follows the `use_item_on` packet.

Test: `swings the arm after use_item_on`.


u9g:
Folded into #4085 (cherry-picked onto its branch, fb1d35fd): same `_genericPlace` packet block as the sequence change.

PR #4092, opened 2026-09-07
Throw from bot.attack for targets the server kicks for

u9g:
The server disconnects with `multiplayer.disconnect.invalid_entity_attacked` (`ServerGamePacketListenerImpl.onAttack`) when the attacked entity is the player itself, an item or an experience orb — a vanilla client can never target those. `bot.attack` now refuses them up front.

Docs updated for `bot.attack`.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -2037,6 +2037,8 @@ Attack a player or a mob.
  * `entity` is a type of entity. To get a specific entity use [bot.nearestEntity()](#botnearestentitymatch--entity---return-true-) or [bot.entities](#botentities).
  * `swing` Default to `true`. If false the bot does not swing its arm when attacking.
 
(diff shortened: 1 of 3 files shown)


u9g:
Closing rather than merging. Checked `processUseEntity` / `handleInteract` on 1.8.9 through 26.1 in [u9g/mc-decompiled](https://github.com/u9g/mc-decompiled): the kick set is the player itself, items, experience orbs **and arrows** (`EntityArrow` / `AbstractArrow`, which covers spectral arrows and tridents; from 1.21.4 only arrows whose `isAttackable()` is false). This PR misses arrows, and the name check is wrong on 1.8–1.12, where minecraft-data calls the orb `XPOrb` / `xp_orb`, so the guard only works for the names the test made up. The guard is a caller convenience rather than packet parity, so it is not worth a per-version entity-class table; the correct list is recorded here if anyone wants it.

PR #4093, opened 2026-09-07
Dismount by holding sneak for one tick on 1.21.3+

u9g:
The server only leaves a vehicle when the player's shift key is down (`Player.rideTick` → `wantsToStopRiding`); the `player_input { jump }` that was sent before is never read for this. `bot.dismount` now holds the sneak control for one physics tick and returns a `Promise` (`void` once the dismount has been sent); older versions keep the `steer_vehicle` unmount flag, which the server reads directly.

Docs and `index.d.ts` updated (`dismount` returns `Promise<void>`).

Test: `holds sneak for one tick on 1.21.3+ and sends the steer_vehicle unmount flag before`.

### DIFF
--- a/lib/plugins/entities.js
+++ b/lib/plugins/entities.js
@@ -882,23 +882,26 @@ function inject (bot) {
     }
   }
 
-  function dismount () {
-    if (bot.vehicle) {
-      if (bot.supportFeature('newPlayerInputPacket')) {
-        bot._client.write('player_input', {
-          inputs: {
-            jump: true
-          }
-        })
-      } else {
-        bot._client.write('steer_vehicle', {
-          sideways: 0.0,
-          forward: 0.0,
-          jump: 0x02
-        })
+  async function dismount () {
+    if (!bot.vehicle) {
+      bot.emit('error', new Error('dismount: not mounted'))
+      return
+    }
+    if (bot.supportFeature('newPlayerInputPacket')) {
+      // Only the shift key in player_input dismounts (Player.rideTick); the jump
+      // flag is never read for this
+      bot.setControlState('sneak', true)
+      try {
+        await bot.waitForTicks(1)
+      } finally {
+        bot.setControlState('sneak', false)
       }
     } else {
-      bot.emit('error', new Error('dismount: not mounted'))
+      bot._client.write('steer_vehicle', {
+        sideways: 0.0,
+        forward: 0.0,
+        jump: 0x02 // unmount flag
+      })
     }
   }
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -2050,7 +2050,9 @@ Mount a vehicle. To get back out, use `bot.dismount`.
 
 #### bot.dismount()
 
-Dismounts from the vehicle you are in.
+This function returns a `Promise`, with `void` as its argument once the dismount has been sent.
+
+Dismounts from the vehicle you are in. On 1.21.3+ this holds the sneak control for one physics tick, which is how vanilla leaves a vehicle.
 
 #### bot.moveVehicle(left,forward)
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -350,7 +350,7 @@ export interface Bot extends TypedEmitter<BotEvents> {
 
   mount: (entity: Entity) => void
 
-  dismount: () => void
+  dismount: () => Promise<void>
 
   moveVehicle: (left: number, forward: number) => void
 
--- a/test/internalTest.js
+++ b/test/internalTest.js
@@ -69,6 +69,9 @@ for (const supportedVersion of mineflayer.testedVersions) {
         port: PORT
       })
       bot.test = {}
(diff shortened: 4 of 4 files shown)


u9g:
Note on the 1.21.3–1.21.5 range, checked in [u9g/mc-decompiled](https://github.com/u9g/mc-decompiled): `Player.rideTick` leaves the vehicle on `isShiftKeyDown()`, but those servers set the player's shift state only from `entity_action` PRESS_SHIFT_KEY (`handlePlayerInput` just stores `lastClientInput`); `player_input.shift` feeds `setShiftKeyDown` from 1.21.6. Master sends sneak through `player_input` from 1.21.3, so the held sneak here only dismounts on 1.21.3–1.21.5 once #4063 (which routes sneak through `sneakUsesEntityAction`) is in. The test mocks `setControlState`, so it does not see the difference. Nothing to change in this PR; it should merge after #4063.

extremeheat (commented):
This code looks messy

lib/plugins/entities.js
```diff
@@ -882,23 +882,26 @@ function inject (bot) {
...
+    }
+    if (bot.supportFeature('newPlayerInputPacket')) {
+      // Only the shift key in player_input dismounts (Player.rideTick); the jump
+      // flag is never read for this
+      bot.setControlState('sneak', true)
+      try {
+        await bot.waitForTicks(1)
+      } finally {
+        bot.setControlState('sneak', false)
       }
```
rom1504, line 898: **Astra agent review — AI-generated, not manually written by the maintainer.**

`waitForTicks(1)` depends on simulation being active, so this changes dismount into a timeout when `bot.physicsEnabled === false`. The physics plugin only emits `physicsTick` inside `physicsEnabled && shouldUsePhysics` (and mounting itself sets `shouldUsePhysics = false`). Using the actual entities and physics plugins with physics disabled, `await bot.dismount()` rejected after 5050 ms despite the sneak packet already having been sent. Could the hold/release use a lifecycle that works without simulation, and be covered with physics disabled and a real mount/dismount path? Stubbing `waitForTicks` here hides the failure.

lib/plugins/entities.js
```diff
@@ -882,23 +882,26 @@ function inject (bot) {
...
     } else {
-      bot.emit('error', new Error('dismount: not mounted'))
+      bot._client.write('steer_vehicle', {
```
rom1504, line 900: **Astra agent review — AI-generated, not manually written by the maintainer.**

Please preserve the caller's existing sneak state. If `getControlState('sneak')` was already true before `dismount()`, the unconditional cleanup changes it to false on either success or timeout, releasing a control the caller never asked this operation to release. I reproduced that with the real control-state implementation. Capture the previous value before the temporary hold and restore it in cleanup, with coverage for initially-held sneak and a rejected wait.

rom1504 (commented):

Reviewed at the maintainer's request. I inspected the current entities/physics code, tests and existing discussion, and locally executed the actual plugins to reproduce the disabled-physics timeout and loss of an already-held sneak control. Two focused lifecycle findings below. No live server test was run.

lib/plugins/entities.js
u9g: Fixed in b8466c6e4865a80d3bfaa4eef6b883a2fa27718d. It was broader than physics being disabled: mounting clears `shouldUsePhysics`, so `physicsTick` never fires while riding even with physics on, and `waitForTicks(1)` always timed out on a real mount. `dismount()` now holds sneak until the server's dismount arrives (5 s timeout). That needed `set_passengers` to notice the bot leaving: the server announces a dismount as the vehicle's new passenger list, which no longer contains the bot, so `bot.vehicle` was never cleared. The stubbed test is replaced by one that mounts through spawn and `set_passengers` packets from the mock server with physics disabled, and checks the server receives shift on then off.

lib/plugins/entities.js
u9g: Fixed in b8466c6e4865a80d3bfaa4eef6b883a2fa27718d: the sneak state from before the call is saved and restored on both success and timeout. Covered by `leaves an already held sneak held when the server does not dismount`, which holds sneak, lets the wait reject, and checks sneak is still held.

rom1504 (commented):

Both earlier findings are addressed: the operation waits for the server's dismount event and restores the caller's previous sneak state. The actual internal tests pass across the version matrix (56 passing), including mounted/physics-disabled behavior and held-sneak timeout. **Keep this conditional on #4063**, as the author already explained for 1.21.3–1.21.5. Our new #4063 review also finds that its tick guard suppresses inputs while riding, which would prevent this implementation's sneak hold reaching the server after those branches are combined. Repair and integration-check that prerequisite before merging this dependent change.

Skills used: prismarine-lifecycle-action-review checked server-observed completion, control restoration and the dependency's mounted-input lifetime. prismarine-review checked current discussions and avoided duplicate findings.

PR #4094, opened 2026-09-07
Swing every tick while digging, skip STOP on instant breaks and abort with face DOWN

u9g:
Audit of the digging packets against vanilla `continueAttack` / `stopDestroyBlock`:

- Swing the arm once per physics tick while a dig is in progress (vanilla `continueAttack`), instead of on a 350 ms timer.
- An instant break (creative or zero hardness) must not send STOP_DESTROY_BLOCK: vanilla sends only START and predicts the break locally.
- A user abort carries face DOWN (vanilla `stopDestroyBlock`), while the abort caused by retargeting carries the **new** block's face — the old code had these two swapped, and let a new dig overwrite the face of the abort it was still sending.
- `bot.stopDigging` is a plain function again (it was being reassigned per dig).

**Based on #4085** (the use_item / use_item_on parity PR, which carries the shared block prediction sequence counter this builds on) — the PR base is that branch; it retargets to master once #4085 merges.

Tests: instant break sends only START, swings every physics tick, abort faces.

### DIFF
--- a/lib/plugins/digging.js
+++ b/lib/plugins/digging.js
@@ -15,7 +15,8 @@ function inject (bot) {
   bot.targetDigFace = null
   bot.lastDigTime = null
 
-  // A dig in progress swings the arm once per physics tick
+  // A dig in progress swings the arm once per physics tick: vanilla Minecraft.continueAttack
+  // swings whenever continueDestroyBlock returns true, and the local player sends the packet unconditionally
   bot.on('physicsTick', () => {
     if (bot.targetDigBlock) bot.swingArm()
   })
@@ -129,7 +130,8 @@ function inject (bot) {
 
     // In vanilla the client will cancel digging the current block once the other block is at the crosshair.
     // Todo: don't wait until lookAt is at middle of the block, but at the edge of it.
-    // An abort caused by retargeting must carry the new block's face
+    // An abort caused by retargeting carries the new block's face: vanilla startDestroyBlock sends
+    // ABORT_DESTROY_BLOCK with the face it was given
     if (bot.targetDigBlock) abortDigging(targetDigFace)
 
     diggingTask = createTask()
@@ -210,7 +212,7 @@ function inject (bot) {
     await diggingTask.promise
   }
 
-  // A user abort must carry face DOWN
+  // A user abort carries face DOWN: vanilla stopDestroyBlock sends ABORT_DESTROY_BLOCK with Direction.DOWN
   function stopDigging () {
     if (bot.targetDigBlock) abortDigging(0)
   }
--- a/lib/plugins/digging.js
+++ b/lib/plugins/digging.js
@@ -130,7 +130,8 @@ function inject (bot) {
     bot._client.write('block_dig', {
       status: 0, // start digging
       location: block.position,
-      face: bot.targetDigFace // default face is 1 (top)
+      face: bot.targetDigFace, // default face is 1 (top)
+      sequence: bot._nextSequence()
     })
     waitTimeout = setTimeout(finishDigging, waitTime)
     bot.targetDigBlock = block
@@ -149,7 +150,8 @@ function inject (bot) {
         bot._client.write('block_dig', {
           status: 2, // finish digging
           location: bot.targetDigBlock.position,
-          face: bot.targetDigFace // always the same as the start face
+          face: bot.targetDigFace, // always the same as the start face
+          sequence: bot._nextSequence()
         })
       }
       bot.targetDigBlock = null
@@ -178,7 +180,8 @@ function inject (bot) {
       bot._client.write('block_dig', {
         status: 1, // cancel digging
         location: bot.targetDigBlock.position,
-        face: cancellationDiggingFace
+        face: cancellationDiggingFace,
+        sequence: 0
       })
       const block = bot.targetDigBlock
       bot.targetDigBlock = null
--- a/lib/plugins/digging.js
+++ b/lib/plugins/digging.js
@@ -6,14 +6,19 @@ const { Vec3 } = require('vec3')
 module.exports = inject
 
 function inject (bot) {
-  let swingInterval = null
   let waitTimeout = null
+  let abortDigging = null // closure of the dig in progress, takes the abort face
 
   let diggingTask = createDoneTask()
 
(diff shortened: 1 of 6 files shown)


lib/plugins/digging.js
```diff
@@ -6,15 +6,20 @@ const { Vec3 } = require('vec3')
...
   let waitTimeout = null
+  let abortDigging = null // closure of the dig in progress, takes the abort face
 
   let diggingTask = createDoneTask()
 
   bot.targetDigBlock = null
   bot.targetDigFace = null
   bot.lastDigTime = null
 
+  // A dig in progress swings the arm once per physics tick
```
u9g: check this against all tested mc versions to make sure this has always been the case

lib/plugins/digging.js
```diff
@@ -212,6 +210,11 @@ function inject (bot) {
     await diggingTask.promise
   }
 
+  // A user abort must carry face DOWN
```
u9g: Why is this?

lib/plugins/digging.js
u9g: Checked on every tested version in [u9g/mc-decompiled](https://github.com/u9g/mc-decompiled) (1.8.9 stands in for 1.8.8; 1.20.1, 1.20.2, 1.20.4, 1.20.6, 1.21.1 and 1.21.3 were decompiled and pushed for this). Same on all 28:

- `Minecraft.sendClickBlockToController` (1.8.9–1.13.2) / `Minecraft.continueAttack` (1.14.4+) runs once per client tick and calls `player.swingItem()` / `swingArm(MAIN_HAND)` / `swing(MAIN_HAND)` whenever `onPlayerDamageBlock` / `continueDestroyBlock` returns true.
- `onPlayerDamageBlock` / `continueDestroyBlock` returns true on every tick the same block is still being hit, including the `blockHitDelay` / `destroyDelay` ticks.
- `EntityPlayerSP.swingItem` / `swingArm` and `LocalPlayer.swing` send the animation packet unconditionally on every call; the swing-in-progress check in the entity superclass gates only the local animation.

The comment now names the vanilla methods.

lib/plugins/digging.js
u9g: `MultiPlayerGameMode.stopDestroyBlock` (`PlayerControllerMP.resetBlockRemoving` before 1.14) sends `ABORT_DESTROY_BLOCK` with `Direction.DOWN` hardcoded. The only other ABORT is the retarget in `startDestroyBlock` / `clickBlock`, which passes the new block's face. Same on all 28 tested versions in [u9g/mc-decompiled](https://github.com/u9g/mc-decompiled). The comment now says so.

lib/plugins/digging.js
```diff
@@ -160,35 +176,21 @@ function inject (bot) {
...
-    const currentBlock = block
-    bot.stopDigging = () => {
-      if (!bot.targetDigBlock) return
-
-      // Replicate the odd vanilla cancellation face value.
-      // When the cancellation is because of a new dig request on another block it's the same as the new dig start face. In all other cases it's 0.
-      const stoppedBecauseOfNewDigRequest = !currentBlock.position.equals(bot.targetDigBlock.position)
-      const cancellationDiggingFace = !stoppedBecauseOfNewDigRequest ? bot.targetDigFace : 0
-
+    abortDigging = (face) => {
```
rom1504, line 179: **Astra agent review — AI-generated, not manually written by the maintainer.**

The new instant-break path calls `_updateBlockState` synchronously before this assignment. That can emit `diggingCompleted`, whose listener can start the next `bot.dig(..., 'ignore')`; when the first call resumes here, it overwrites the second dig’s abort closure with its own. I reproduced an instant first dig followed by a noninstant second dig from `diggingCompleted`: `stopDigging()` sends the second block’s ABORT but removes the first block’s listener, leaving `blockUpdate:<second position>` installed. Please install the operation’s abort/cleanup state before any synchronous completion can fire, and ensure an older operation cannot replace a successor’s state. Cover this event-handler reentry and cancellation sequence.

rom1504 (commented):

Reviewed at the maintainer’s request. Validation: Local reproduction executing current digging plugin with real promise utilities and EventEmitter block updates; instant-completion reentry leaves successor blockUpdate listener after stopDigging. Static inspection of diff versus PR base. One reproduced cancellation-cleanup issue is noted inline.

lib/plugins/digging.js
u9g: Fixed in 2c53f8a6c808f7fc51f2a80ffde25099144140ee. The task, the wait timer and the abort closure are now locals of each `dig` call. `abortDigging` and the `blockUpdate` listener are set before the START packet and before the instant-break `_updateBlockState`. After anything that can synchronously emit `diggingCompleted` or `diggingAborted`, a dig only touches its own task, so an earlier dig can no longer replace a later dig's abort or resolve its task. This also fixes a second symptom of the same bug: the first dig's `finish()` used to resolve the second dig's promise.

New test `a dig started from diggingCompleted of an instant break keeps its own state`: an instant dig whose `diggingCompleted` listener starts a noninstant dig, followed by `stopDigging()`. It checks that the second dig rejects, the ABORT goes to the second position, and no `blockUpdate` listener remains for either position. It fails on the previous commit and passes on all 28 tested versions.

rom1504 (commented):

The reported instant-completion reentry defect is fixed at 2c53f8a: the task/abort/listener state belongs to its dig before synchronous completion can start another one. I ran all four focused digging tests across all 28 tested versions: 112 passing, including cancellation of a successor started from diggingCompleted and listener cleanup.

No additional introduced blocker found. Keep this conditional on landing/updating the #4085 stack and rerunning the failed CI job: its failure is a 1.18.2 nether spawn timeout, not a digging assertion. The current PR base is still the sequence-counter branch, and its source retains the older sequence plugin while #4085 now puts the counter in createBot. Preserve that shared owner when rebasing. These were local NMP tests, not a new vanilla-server run.

Skills used: `prismarine-lifecycle-action-review` traced synchronous completion, successor cancellation and listener ownership; `prismarine-protocol-data-review` checked the real packet-test path; `prismarine-review` separated the repaired finding, stack dependency and unrelated CI result.

PR #4095, opened 2026-09-07
Right-click entities through one path, so useOn and mount send the vanilla pair

u9g:
`bot.useOn` and `bot.mount` sent a lone interact with the hit location fixed at the entity's feet; a right click on an entity is always interact_at then interact (vanilla `startUseItem`). Both now live next to `activateEntity` in `inventory.js` and go through the same sender, and `use_entity` no longer has a second, divergent implementation in `entities.js`. `attack` gets its own sender carrying only the sneak state (its own packet on 26.1).

**Based on #4089** (the activateEntity packet work this builds on) — the PR base is that branch; it retargets to master once #4089 merges.

Test: `useOn and mount send the same interact pair as activateEntity`.

### DIFF
--- a/lib/plugins/entities.js
+++ b/lib/plugins/entities.js
@@ -836,8 +836,7 @@ function inject (bot) {
   }
 
   function useOn (target) {
-    // TODO: check if not crouching will make make this action always use the item
-    useEntity(target, 0)
+    bot._interactEntity(target, bot._entityHitVector(target))
   }
 
   function attack (target, swing = true) {
@@ -846,9 +845,9 @@ function inject (bot) {
       if (swing) {
         swingArm()
       }
-      useEntity(target, 1)
+      attackEntity(target)
     } else {
-      useEntity(target, 1)
+      attackEntity(target)
       if (swing) {
         swingArm()
       }
@@ -856,8 +855,7 @@ function inject (bot) {
   }
 
   function mount (target) {
-    // TODO: check if crouching will make make this action always mount
-    useEntity(target, 0)
+    bot._interactEntity(target, bot._entityHitVector(target))
   }
 
   function moveVehicle (left, forward) {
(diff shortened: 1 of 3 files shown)


lib/plugins/entities.js
```diff
@@ -902,30 +900,19 @@ function inject (bot) {
     }
   }
 
-  function useEntity (target, leftClick, x, y, z) {
-    const sneaking = bot.getControlState('sneak')
-    if (leftClick && bot.supportFeature('attackUsesOwnPacket')) {
+  // An attack carries no hit location, only the sneak state; 26.1 moved it to its own packet.
+  function attackEntity (target) {
+    if (bot.supportFeature('attackUsesOwnPacket')) {
```
u9g: Should we swap the function based on the boolean once rather than iffing every time

lib/plugins/inventory.js
```diff
@@ -939,6 +939,8 @@ function inject (bot, { hideErrors }) {
   bot.deactivateItem = deactivateItem
 
   // not really in the public API
+  bot._interactEntity = interactEntity
```
u9g: These should be fully internal if possible, not even put here, if not possible, why

lib/plugins/entities.js
u9g: Done in a35c1ff1: `attackEntity` is now chosen once at inject time from `attackUsesOwnPacket`.

lib/plugins/inventory.js
u9g: They were only on `bot` because `useOn`/`mount` lived in `entities.js` while the sender is a closure in `inventory.js`. a35c1ff1 moves `useOn` and `mount` next to `activateEntity` in `inventory.js`, so nothing crosses plugins and the `_` exports are gone.

u9g:
Folded into #4089 (commit 8ccbffee on its branch): the useOn/mount move and the once-chosen attack sender are the same change as the interact pair, and keeping them apart left `entities.js` with a divergent `use_entity` writer in between.

PR #4098, opened 2026-09-08
Retain server-observed block inventories in bot.world

DragonModels:
Connects server-observed block-container contents to `bot.world`, advancing the block-inventory portion of #334. After `openBlock`/`openContainer`, callers can query `bot.world.getObservedBlockInventory(position)` for container slots, kind, observation time and stale status. Server packets update the snapshot; optimistic local clicks do not. Raw block-entity NBT stays separate and retains its existing storage behavior.

Unknown or invalidated contents return null. Closing/replacing the window marks the previous observation stale; block/NBT changes, column replacement/unload, respawn, login and disconnect discard it. Entity, player and personal ender-chest inventories are excluded. Double-chest windows are anchored only to the clicked position; the other half is not inferred. Overlapping block/entity open requests are rejected to avoid ambiguous attribution.

Depends on PrismarineJS/prismarine-world#164. The dependency pins the same feature patch applied to the published 3.7.0 release (commit `281ebd6fae18ff0f2a93d6e3072fca133a841d5b`). This preserves the released `vec3` 0.1 range used by entity/chunk libraries; current world master includes an unrelated, unreleased 0.2 bump that caused incompatible vector prototypes in the first CI run. The upstream world PR is unchanged. Replace the temporary pin with a compatible published upstream release before merging.

Validation:
- Upstream CI run [34277994322](https://github.com/PrismarineJS/mineflayer/actions/runs/34277994322) passed all seven Minecraft groups (all 28 supported versions), lint and version preparation on commit `a8b33d355dc6367766a55adf7685ae0896c5dca5`. This includes the repository's external-server tests.
- Final public dependency install with lifecycle scripts disabled: lint and 128 targeted tests pass. These include a regression that reproduced the initial CI vector-prototype failure, packet/lifecycle cases, and window plus absolute/relative-position protocol coverage across all 28 supported versions.
- Earlier full internal plus observation suite: 810 passing, 73 pending with one retry configured. The first broad run had five respawn timeouts; unchanged and modified code both passed all 28 respawn tests separately. The final nested-buffer-copy adjustment was followed by lint and the 99 focused tests again.
- A strict TypeScript consumer fixture passes with dependency declaration checking skipped. Unrestricted `tsc` still reports errors in dependency declarations.
- Local validation used isolated protocol fixtures; the upstream CI result above provides the external-server test evidence.

### DIFF
--- a/lib/observed_block_inventories.js
+++ b/lib/observed_block_inventories.js
@@ -1,0 +1,105 @@
+const blockContainers = new Set([
+  'chest', 'trapped_chest', 'barrel', 'hopper', 'furnace', 'lit_furnace',
+  'blast_furnace', 'smoker', 'dispenser', 'dropper', 'brewing_stand'
+])
+
+module.exports = function observedBlockInventories (bot) {
+  let pending = null
+  let active = null
+  const observed = new Map()
+  const snapshotItem = item => item === null ? null : structuredClone(item)
+
+  function begin (block) {
+    if (pending) throw new Error('A window is already being opened')
+    const context = block
+      ? {
+          world: bot.world,
+          position: block.position.floored(),
+          type: block.type,
+          kind: block.name,
+          column: bot.world.getColumnAt(block.position)
+        }
+      : { entity: true }
+    pending = context
+    return context
+  }
+
+  function valid (context) {
+    return context.world === bot.world &&
+      context.column === bot.world.getColumnAt(context.position) &&
+      bot.world.getBlock(context.position)?.type === context.type
+  }
+
+  function prepare (window, blockWindow) {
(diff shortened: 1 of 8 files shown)


test/observedBlockInventoriesTest.js
extremeheat, line 1: Remove

test/worldVectorCompatibilityTest.js
extremeheat, line 1: Remove

lib/observed_block_inventories.js
extremeheat, line 1: Please make this fit in with existing mineflayer plugin code style

PR #4102, opened 2026-09-08
Normalize open_window titles to ChatMessage

u9g:
`window.title` had no stable type. The `open_window` title is a chat component on every version mineflayer supports (1.8.9 `S2DPacketOpenWindow.readPacketData` reads it with `readChatComponent`; 26.1 `ClientboundOpenScreenPacket` carries a `Component`) and reaches the client as a JSON string, an NBT string or an NBT compound depending on the version and on what the server put in the packet. On a 26.1 proxy network the same menu arrived as a bare `§`-string from one backend and as a component from another, so `String(window.title)` was `"[object Object]"` and scripts that pick GUI items by title had to guess the shape.

The `open_window` handler now runs the field through `ChatMessage.fromNotch`, the same normalization every other chat field gets. `window.title` is always a `ChatMessage`: `String(title)` is the rendered text, `.json` keeps the structure, and callers that already did `ChatMessage.fromNotch(window.title)` still get the text.

The internal test opens a chest with a component title and with the bare-string form on every tested version and asserts the rendered title.

Follow-up: `prismarine-windows` still declares `Window.title` as `string` in its typings.

### DIFF
--- a/test/internalTest.js
+++ b/test/internalTest.js
@@ -648,6 +648,28 @@ for (const supportedVersion of mineflayer.testedVersions) {
         })
       })
 
+      it('window titles are ChatMessages whatever shape the server sends', async () => {
+        const Item = require('prismarine-item')(registry)
+        // A component title plus the bare-string form third-party servers send.
+        const titles = registry.supportFeature('chatPacketsUseNbtComponents')
+          ? [nbt.comp({ text: nbt.string('Test Chest') }), nbt.string('Test Chest')]
+          : [JSON.stringify({ text: 'Test Chest' }), 'Test Chest']
+        const chest = registry.supportFeature('village&pillageInventoryWindows')
+          ? { inventoryType: 2 }
+          : { inventoryType: 'minecraft:chest', slotCount: 27 }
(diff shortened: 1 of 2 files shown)


test/internalTest.js
```diff
@@ -648,6 +648,28 @@ for (const supportedVersion of mineflayer.testedVersions) {
...
 
+      it('window titles are ChatMessages whatever shape the server sends', async () => {
+        const Item = require('prismarine-item')(registry)
+        // A component title plus the bare-string form third-party servers send.
+        const titles = registry.supportFeature('chatPacketsUseNbtComponents')
+          ? [nbt.comp({ text: nbt.string('Test Chest') }), nbt.string('Test Chest')]
+          : [JSON.stringify({ text: 'Test Chest' }), 'Test Chest']
+        const chest = registry.supportFeature('village&pillageInventoryWindows')
+          ? { inventoryType: 2 }
+          : { inventoryType: 'minecraft:chest', slotCount: 27 }
```
extremeheat, line 659: Unrelated, but data inconsistency that should be fixed upstream

extremeheat (approved):

PR #4104, opened 2026-09-08
Accept play-phase resource packs when no resourcePack listener answers them

u9g:
Resource packs were only auto-accepted in the configuration phase (#4005). A pack sent in the play phase just emitted `resourcePack` and waited for the caller. Some proxies (Jartex on 1.21.4, for one) move players between backends through a play-phase pack request and drop the connection about 20 seconds later when nothing answers it: `end socketClosed`, no kick. Vanilla players see the prompt and accept.

Both pack handlers now go through one `autoAcceptResourcePack`: accept in the configuration phase (unchanged), and in play when no `resourcePack` listener exists. With a listener attached, a play-phase pack still waits for `bot.acceptResourcePack()` or `bot.denyResourcePack()`, so code that denies packs keeps working and is not answered twice.

Tests: `accepts a play-phase resource pack when nobody listens for it` and `leaves a play-phase resource pack to the resourcePack listener`; the first fails on master on every tested version. Docs updated for the `resourcePack` event.

Reported in u9g/bot-harness#65.

The tests drive a real bot against the mock server, which is how two further bugs turned up: `denyResourcePack()` wrote a packet without the version's `hash`/`uuid` field (a serialization error instead of a deny on 1.8/1.9, a stray second reply on 1.20.3+), and `resource_pack_send` recorded the hash/uuid only after emitting `resourcePack`, so a listener answering inside the emit sent an empty one. Accept and deny now share one `answerResourcePack(result)`.

### DIFF
--- a/lib/plugins/resource_pack.js
+++ b/lib/plugins/resource_pack.js
@@ -17,9 +17,7 @@ function inject (bot) {
     activeResourcePacks[data.uuid] = data.url
 
     bot.emit('resourcePack', data.url, data.uuid)
-    // The server holds the configuration phase (e.g. a Velocity server transfer)
-    // open until the pack is answered, so accept it there to let it complete
-    if (bot._client.state === 'configuration') acceptResourcePack()
+    autoAcceptResourcePack()
   })
 
   bot._client.on('remove_resource_pack', (data) => { // Doesn't emit  anything because it is removing rather than adding
@@ -44,10 +42,18 @@ function inject (bot) {
       bot.emit('resourcePack', data.url, data.hash)
       latestHash = data.hash
     }
-    // Accept during the configuration phase (e.g. a Velocity server transfer),
-    // which the server holds open until the pack is answered
-    if (bot._client.state === 'configuration') acceptResourcePack()
+    autoAcceptResourcePack()
   })
+
+  // A pack must be answered: the server holds the configuration phase (e.g. a Velocity
+  // transfer) open until it is, and proxies that move players through a play-phase pack
+  // drop an unanswered connection. Play-phase packs are left to a resourcePack listener
+  // when one exists so it can still deny.
+  function autoAcceptResourcePack () {
+    if (bot._client.state === 'configuration' || bot.listenerCount('resourcePack') === 0) {
+      acceptResourcePack()
+    }
+  }
 
   function acceptResourcePack () {
     if (bot.supportFeature('resourcePackUsesHash')) {
--- a/test/internalTest.js
+++ b/test/internalTest.js
@@ -533,6 +533,38 @@ for (const supportedVersion of mineflayer.testedVersions) {
         assert(buf.includes(expectedBytes),
           'resource_pack_receive must carry the pack UUID bytes, not a zero UUID')
       })
+
+      // Some proxies move players through a play-phase pack request and drop the
+      // connection when it goes unanswered.
+      function playPhasePack (withListener) {
+        // The mock server never sends a pack, so the plugin is driven directly.
+        const client = new EventEmitter()
+        client.state = 'play'
+        const writes = []
+        client.write = (name, params) => { writes.push({ name, params }) }
+        const fakeBot = new EventEmitter()
+        fakeBot._client = client
+        fakeBot.supportFeature = registry.supportFeature.bind(registry)
+        require('../lib/plugins/resource_pack')(fakeBot)
+        if (withListener) fakeBot.on('resourcePack', () => {})
+
+        const pack = { url: 'https://example.invalid/pack.zip', hash: '88b406352dc8a335b1050a4bf9577a878c812012', forced: false }
+        if (registry.supportFeature('resourcePackUsesUUID')) {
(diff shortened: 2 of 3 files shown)


lib/plugins/resource_pack.js
```diff
@@ -44,11 +42,19 @@ function inject (bot) {
...
-    if (bot._client.state === 'configuration') acceptResourcePack()
+    autoAcceptResourcePack()
   })
 
+  // A pack must be answered: the server holds the configuration phase (e.g. a Velocity
+  // transfer) open until it is, and proxies that move players through a play-phase pack
+  // drop an unanswered connection. Play-phase packs are left to a resourcePack listener
+  // when one exists so it can still deny.
+  function autoAcceptResourcePack () {
+    if (bot._client.state === 'configuration' || bot.listenerCount('resourcePack') === 0) {
```
rom1504, line 53: **Astra agent review — AI-generated, not manually written by the maintainer.**

A `once` listener can decline the pack and still trigger automatic acceptance here:

```js
bot.once('resourcePack', () => bot.denyResourcePack())
```

Both handlers call this helper after `bot.emit(...)`. Node removes a `once` listener before invoking it, so `listenerCount('resourcePack')` is already zero when this condition runs.

I reproduced this against the current plugin in play state with the pre-UUID/post-hash response shape: the replies are `[DECLINED, ACCEPTED, SUCCESSFULLY_LOADED]`. A `once` listener that accepts produces the acceptance pair twice. A persistent listener that declines produces only `DECLINED`.

Please capture whether the event had listeners before emission (or use `emit`'s boolean return) and use that value for the play-phase default. Add regressions for `once` listeners that deny, accept, and defer their response; each should retain control of that pack's response.

rom1504 (commented):

Written and posted by an Astra agent at the maintainer's request. One inline finding on the play-phase fallback, verified with a local plugin-level reproduction against commit 217e7f9 (not a live-server test).

lib/plugins/resource_pack.js
u9g: Fixed in 8ccaa360910cb6d3b50a40d7d2d3a747d5ad59b9: both handlers now pass `bot.emit('resourcePack', ...)`'s return value to `autoAcceptResourcePack` instead of checking `listenerCount` after the emit. Added regressions for `once` listeners that defer (no reply), deny (same replies as a persistent denying listener, no ACCEPTED), and accept (`[ACCEPTED, SUCCESSFULLY_LOADED]` exactly once). All three fail on 217e7f9 on every tested version.

VasilisDragon:
Looks good to me. The once-listener defer, deny and accept tests now pass on all 28 tested versions. I also checked that default replies keep the pack UUID/hash and that a deferred once listener retains control of its pack. The previous head reproduces the duplicate-response failures.

rom1504 (commented):

**Looks ready to merge from this review.** Using emit's return value preserves a once listener's ownership whether it accepts, declines or defers. I reran the focused resource-pack tests across the version matrix: 151 passing, 17 pending, including all three once-listener regressions. Current CI is green. These are local protocol client/server checks, not a vanilla-server run.

Skills used: prismarine-lifecycle-action-review checked synchronous listener removal and response ownership; prismarine-protocol-data-review checked version-specific response shapes through the existing network harness. prismarine-review checked current discussions and avoided duplicate findings.

test/internalTest.js
```diff
@@ -533,6 +533,54 @@ for (const supportedVersion of mineflayer.testedVersions) {
         assert(buf.includes(expectedBytes),
           'resource_pack_receive must carry the pack UUID bytes, not a zero UUID')
       })
+
+      // Some proxies move players through a play-phase pack request and drop the
+      // connection when it goes unanswered.
+      function playPhasePack (listen) {
+        // The mock server never sends a pack, so the plugin is driven directly.
+        const client = new EventEmitter()
```
rom1504: Use real client

u9g: Done in 46670ce3. The tests now run against the real bot and the mock server: the bot joins as usual, the server writes `add_resource_pack` (or `resource_pack_send` pre-1.20.3) in the play phase, and the test reads back the `resource_pack_receive` packets the bot actually put on the wire. No stubs left. The "no answer expected" cases end deterministically instead of on a timer — a `keep_alive` is sent right after the pack, and since the bot answers packets in order, its `keep_alive` reply means anything it had to say about the pack has already arrived.

That turned up two real bugs the stub could never have caught, fixed in 5bd171cc:

- `denyResourcePack()` wrote a bare `{ result }` packet. On 1.8/1.9 that is missing the `hash` field and throws a serialization error instead of denying; on 1.20.3+ it followed the correct uuid reply with a second uuid-less one. Accept and deny now share one `answerResourcePack(result)` that fills in whatever field the version's packet has, which also collapses the old per-version duplication in `acceptResourcePack`.
- `resource_pack_send` recorded `latestHash`/`latestUUID` *after* emitting `resourcePack`, so a listener that accepted or denied inside the emit sent an empty hash/uuid. Recorded first now.

All 6 play-phase tests pass on all 28 tested versions, and the full `internalTest.js` run is green (935 passing). I checked they still catch the regressions: reverting either fix fails the deny tests on both old and new versions.

rom1504:
CI fails

PR #4105, opened 2026-09-08
fix: drop the deferred respawn reply when a newer teleport arrives

u9g:
Invariants of the respawn teleport reply:

- After a death, the reply to the next teleport is deferred by 1.5 s (unchanged).
- A clientbound `position` packet cancels a pending deferred reply before it is handled; the newer teleport is answered on its own path.
- At most one deferred reply is pending at a time, and none after `end`.

The server holds only its latest teleport, so a deferred reply with the earlier coordinates is a move away from where the server placed the player. `bot.entity` never held those coordinates at that point either: the timer sent the packet without updating the entity.

Test (`physics › answers only the latest teleport when a second one lands inside the respawn reply delay`): death, teleport A, teleport B 100 ms later, wait 1.7 s; the server receives exactly one `position_look`, with B's coordinates. Fails on master (A's reply arrives after B's) on every tested version; passes with the change. Internal suite green on all 28 tested versions.

Overlaps PrismarineJS/mineflayer#4099, which introduces the same `respawnReply` handle for the configuration-state case; the rebase is mechanical whichever lands first.

### DIFF
--- a/lib/plugins/physics.js
+++ b/lib/plugins/physics.js
@@ -98,6 +98,7 @@ function inject (bot, { physicsEnabled, maxCatchupTicks }) {
   function cleanup () {
     clearInterval(doPhysicsTimer)
     doPhysicsTimer = null
+    cancelRespawnReply()
   }
 
   function sendPacketPosition (position, onGround) {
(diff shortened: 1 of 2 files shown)


u9g:
Folded into #4099: both add the same `respawnReply` handle, and together they state one invariant (the deferred reply is cancelled once its coordinates are stale, whether by a newer teleport, `start_configuration`, `login` or `end`). The commit and its test are on that branch.

PR #4106, opened 2026-09-09
closeWindow(null) is a no-op instead of throwing

u9g:
\`bot.closeWindow(bot.currentWindow)\` is the usual way to close whatever is open, and \`bot.currentWindow\` is null whenever the window closed on its own between the check and the call. \`closeWindow\` read \`window.id\` first, so that raced into:

\`\`\`
TypeError: Cannot read properties of null (reading 'id')
    at EventEmitter.closeWindow (mineflayer/lib/plugins/inventory.js:485:24)
\`\`\`

A null window now resolves immediately: no \`close_window\` packet, no \`windowClose\` event, no inventory sync click. Closing a real window is unchanged. The typing accepts \`Window | null\` to match \`bot.currentWindow\`, and the doc notes the no-op.

Test: \`closeWindow(null) resolves without sending close_window or emitting windowClose\` in the internal suite, which fails with the TypeError above on master.

Reported in u9g/bot-harness#29.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -2196,7 +2196,7 @@ Put the item at `slot` in the inventory.
 
 This function returns a `Promise`, with `void` as its argument once the server has acknowledged the close.
 
-Close the `window`. On 1.16.5 and below the server only learns which inventory slots the window changed on its next tick, so await this before anything else (a command, another player) touches those slots.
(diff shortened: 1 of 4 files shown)


u9g:
Folded into #4103, which carries this commit; the three branches rewrote the same login/respawn/close_window listeners in inventory.js and conflicted pairwise.

PR #4107, opened 2026-09-09
fix: answer pings and teleports at the start of the next tick, in arrival order

u9g:
Invariants of the replies to clientbound `ping` and `position` packets:

- In the play state a reply is written at the start of the next physics tick, before the simulation and before that tick's movement packets. Nothing is written from inside the packet handler.
- Replies go out in the order their packets arrived, through one queue: `pong` for a ping, `teleport_confirm` then `position_look` for a teleport.
- A reply queued while no tick is running (before the login packet starts the timer) is still written within one tick. Outside the play state, or without the physics plugin, `game.js` writes the pong immediately.
- Every queued teleport is answered when the tick drains the queue, ahead of the entity and chunk guards, so the first teleport of a login is answered while chunks are loading. The teleport reply carries onGround false.

Vanilla reference (u9g/mc-decompiled): `ClientCommonPacketListenerImpl.handlePing` and `ClientPacketListener.handleMovePlayer` both start with `PacketUtils.ensureRunningOnSameThread`, which puts the packet on `PacketProcessor.packetsToBeHandled`, a FIFO. `Minecraft.runTick` calls `packetProcessor.processQueuedPackets()` and `runAllTasks()` before the `tick()` loop that runs `LocalPlayer.tick` and `sendPosition`, and each handler writes its reply as it runs. 1.8.9 does the same for the teleport through `PacketThreadUtil.checkThreadAndEnqueue` and `Minecraft.runGameLoop`'s `scheduledTasks` drain (the `ping` packet only exists from 1.17).

Why it matters, measured on a BedWars server that pins players with a teleport every tick:

- Answering from the handler made the round trip network RTT instead of one tick: 923 teleports answered in 30 s, peaking at 538 in one second. Answered at the next tick: 59, peaking at 27. The anticheat kicks for the former.
  ```
  26503 <position  (0.5, 100, 0.5) id0
  26503 >teleport_confirm id0
  26503 >position_look (0.5, 100, 0.5)
  26504 <position  (0.5, 100, 0.5) id0
  26504 >teleport_confirm id0
  ...
  ```
- With separate ping and teleport queues the teleport queue drained first, so a pong for a ping that arrived before a teleport always followed the teleport reply. Transaction-ordering anticheats (Grim and friends) treat that as an out-of-order reply and kicked the bot within a minute while it stood still.

Tests in `test/internalTest.js`, each failing on master:

- `pong is written ...`: with the bot falling, the pong follows the movement packet of the tick that received the ping and is answered once; with no chunk loaded it is still answered within one tick.
- teleport: a `position` packet emitted mid-tick writes nothing from the handler; the answer goes out on the next tick.
- `answers every teleport queued for a tick, in arrival order`.
- a ping, a teleport and a second ping arriving between two ticks are answered as `pong 1, teleport_confirm, position_look, pong 2` (master and the two-queue version answer `pong 1` last).

This PR absorbs #4070 (pong at the next tick) and #4078 (teleport at the next tick); the branch carries their commits plus the shared queue. #4063 and #4108 merge cleanly with it.

### DIFF
--- a/lib/plugins/physics.js
+++ b/lib/plugins/physics.js
@@ -491,6 +491,9 @@ function inject (bot, { physicsEnabled, maxCatchupTicks }) {
   bot.on('respawn', () => { shouldUsePhysics = false })
   bot.on('login', () => {
     shouldUsePhysics = false
+    // A teleport still queued here belongs to the world the bot just left, and its id means nothing
+    // to the server it is about to talk to.
+    pendingTeleports.length = 0
     if (doPhysicsTimer === null) {
       lastPhysicsFrameTime = performance.now()
       doPhysicsTimer = setInterval(doPhysics, PHYSICS_INTERVAL_MS)
--- a/lib/plugins/physics.js
+++ b/lib/plugins/physics.js
@@ -78,6 +78,7 @@ function inject (bot, { physicsEnabled, maxCatchupTicks }) {
 
   function tickPhysics (now) {
     if (bot._client.state !== 'play') return // do nothing outside of the play state (e.g. server transfer configuration phase)
+    while (pendingTeleports.length) handleTeleport(pendingTeleports.shift())
     if (!bot.entity?.position || !Number.isFinite(bot.entity.position.x)) return // entity not ready
     if (bot.blockAt(bot.entity.position) == null) return // check if chunk is unloaded
     if (bot.physicsEnabled && shouldUsePhysics) {
@@ -371,7 +372,13 @@ function inject (bot, { physicsEnabled, maxCatchupTicks }) {
   })
 
   // player position and look (clientbound)
-  bot._client.on('position', (packet) => {
+  // The vanilla client hands every play packet to the client thread, which drains the queue at the
+  // start of a tick (PacketUtils.ensureRunningOnSameThread), so a teleport is answered at most once
+  // per tick however fast the server sends them.
+  const pendingTeleports = []
+  bot._client.on('position', (packet) => { pendingTeleports.push(packet) })
+
+  function handleTeleport (packet) {
     // Is this necessary? Feels like it might wrongly overwrite hitbox size sometimes
     // e.g. when crouching/crawling/swimming. Can someone confirm?
     bot.entity.height = 1.8
@@ -451,7 +458,7 @@ function inject (bot, { physicsEnabled, maxCatchupTicks }) {
     lastSentPitch = bot.entity.pitch
 
     bot.emit('forcedMove')
-  })
+  }
 
   bot.waitForTicks = async function (ticks) {
     if (ticks <= 0) return
--- a/lib/plugins/physics.js
+++ b/lib/plugins/physics.js
@@ -78,7 +78,7 @@ function inject (bot, { physicsEnabled, maxCatchupTicks }) {
 
   function tickPhysics (now) {
     if (bot._client.state !== 'play') return // do nothing outside of the play state (e.g. server transfer configuration phase)
-    while (pendingTeleports.length) handleTeleport(pendingTeleports.shift())
+    flushReplies()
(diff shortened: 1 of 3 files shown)


lib/plugins/physics.js
```diff
@@ -371,7 +375,28 @@ function inject (bot, { physicsEnabled, maxCatchupTicks }) {
...
+    if (bot._client.state !== 'play') return
+    while (pendingReplies.length) pendingReplies.shift()()
+  }
+
+  bot._replyOnNextTick = (reply) => {
+    pendingReplies.push(reply)
+    // While no tick is running (the login packet starts it) a reply waits at most one tick.
+    if (doPhysicsTimer === null && replyTimer === null) replyTimer = setTimeout(flushReplies, PHYSICS_INTERVAL_MS)
+  }
+  bot._client.on('position', (packet) => { bot._replyOnNextTick(() => handleTeleport(packet)) })
```
rom1504, line 397: **Astra agent review — AI-generated, not manually written by the maintainer.**

Queuing the entire teleport handler reverses its state updates relative to packet handlers that still run immediately. I reproduced `position(yaw=30)` followed by `player_rotation(yaw=90)` between ticks: the rotation first becomes 90, then `handleTeleport` overwrites it with 30 when the queue drains. The bot ends facing the earlier direction; a later velocity update has the same ordering risk. Please preserve arrival order for incoming state updates while retaining the shared reply FIFO—for example, apply teleport state immediately and queue replies with captured values, or schedule the interacting handlers consistently. A mixed position/rotation regression would catch this.

lib/plugins/physics.js
```diff
@@ -484,6 +509,9 @@ function inject (bot, { physicsEnabled, maxCatchupTicks }) {
   bot.on('respawn', () => { shouldUsePhysics = false })
   bot.on('login', () => {
     shouldUsePhysics = false
+    // A reply still queued here belongs to the world the bot just left, and its id means nothing
+    // to the server it is about to talk to.
+    pendingReplies.length = 0
```
rom1504, line 514: **Astra agent review — AI-generated, not manually written by the maintainer.**

This cleanup also needs to cover respawn/world replacement. The `respawn` handler above only disables physics, and `flushReplies()` runs before that flag is checked. I reproduced queueing a teleport, emitting `respawn`, then ticking: the old coordinates are applied, the old teleport is confirmed, and `handleTeleport` enables physics again before the new world has positioned the bot. Please discard queued work belonging to the prior level on that transition (and invalidate it when leaving play), with a queued-teleport-then-respawn regression.

rom1504 (commented):

lib/plugins/physics.js
u9g: Fixed in 875ff4fefa3ecb58bda0217805806acaf3505fa7: the teleport's state is applied when the packet arrives, and only the reply is queued, using the position and rotation that teleport set. Added a test for a teleport followed by `player_rotation`.

lib/plugins/physics.js
u9g: Fixed in 875ff4fefa3ecb58bda0217805806acaf3505fa7: a respawn drops queued teleport replies (pongs stay, so pings are still answered in order), and leaving play drops everything queued. Added a test for a teleport queued before a respawn.

rom1504 (commented):

**Looks ready to merge from this review.** Teleport state now applies on arrival, replies capture that teleport's values, and a respawn removes old teleport replies while retaining pongs. Leaving play clears the queue. I reran the focused physics/ping tests: 220 passing, 60 pending across the version matrix, including mixed teleport/rotation ordering and queued-teleport-then-respawn regressions. Current CI is green. Validation used local protocol client/server tests, not vanilla.

Skills used: prismarine-lifecycle-action-review traced arrival order separately from reply order and checked queue ownership across world/session transitions. prismarine-review checked current discussions and avoided duplicate findings.

rom1504:
Resolve conflicts and then you can merge

PR #4108, opened 2026-09-09
physics: a teleport leaves the standing flag alone

u9g:
Invariants:

- A clientbound `position` packet sets the bot's position, velocity and rotation and nothing else, as `ClientPacketListener.handleMovePlayer` does. Whether the bot stands is decided by the physics tick.
- The bot's own entity starts airborne on login (`Entity.onGround` starts false); prismarine-entity defaults it to true.
- With `bot.physicsEnabled = false` nothing simulates `bot.entity.onGround`, so it reads as it was last left; the position reminder still goes out every second and carries that value. `docs/api.md` states this under `bot.physicsEnabled`.

- The reply to a teleport carries `onGround: false` and no collision, whatever the standing flag holds, and leaves the last-sent record alone (the same `sendTeleportReply` as #4063).

Test (`physics › answers a teleport airborne and then reports the standing flag it had while physics is disabled`): the bot lands on a block, physics is disabled, and capture starts before the server teleports it to the same spot. Over the next 1.5 s the reply is the only movement packet with `onGround: false`, and the reminders after it carry `true`. Fails on master for all tested versions (every packet carries `false`).

These are the teleport changes from #4063, on their own.

### DIFF
--- a/lib/plugins/physics.js
+++ b/lib/plugins/physics.js
@@ -417,7 +417,6 @@ function inject (bot, { physicsEnabled, maxCatchupTicks }) {
 
     bot.entity.yaw = conv.fromNotchianYaw(newYaw)
     bot.entity.pitch = conv.fromNotchianPitch(newPitch)
-    bot.entity.onGround = false
 
     if (bot.supportFeature('teleportUsesOwnPacket')) {
       bot._client.write('teleport_confirm', { teleportId: packet.teleportId })
--- a/docs/api.md
+++ b/docs/api.md
@@ -908,7 +908,7 @@ world height
 
 #### bot.physicsEnabled
 
-Enable physics, default true.
+Enable physics, default true. While it is false the position reminder still goes out every second and carries `bot.entity.onGround` as it was left: a teleport does not change it, and nothing simulates it.
 
 #### bot.player
 
--- a/lib/plugins/entities.js
+++ b/lib/plugins/entities.js
@@ -110,6 +110,9 @@ function inject (bot) {
     bot.entity.height = PLAYER_HEIGHT
     bot.entity.width = PLAYER_WIDTH
     bot.entity.eyeHeight = PLAYER_EYEHEIGHT
+    // A player that has just been added to the world is airborne until a tick of physics
+    // says otherwise (Entity.onGround starts false); prismarine-entity defaults it to true.
+    bot.entity.onGround = false
   })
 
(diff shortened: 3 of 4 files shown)


lib/plugins/physics.js
```diff
@@ -417,7 +417,6 @@ function inject (bot, { physicsEnabled, maxCatchupTicks }) {
 
     bot.entity.yaw = conv.fromNotchianYaw(newYaw)
     bot.entity.pitch = conv.fromNotchianPitch(newPitch)
```
rom1504, line 419: **Astra agent review — AI-generated, not manually written by the maintainer.**

Keeping the simulated standing state also changes the immediate teleport reply in this branch: `sendPacketPositionAndLook(..., bot.entity.onGround)` below now sends `onGround: true` if the bot was standing. I reproduced that with physics disabled. The teleport reply is supposed to carry false independently of the retained entity state, as the separate reply helper in #4063 already implements. Please include that bounded reply change here or make landing #4063 first an explicit dependency, and start the test's packet capture before sending the teleport so it checks the immediate reply as well as the later reminders (the current capture begins 200 ms afterward).

rom1504 (commented):

lib/plugins/physics.js
u9g: Included here in 704a251: the teleport handler (and the delayed reply after a respawn) now goes through the same `sendTeleportReply` helper as #4063, so the reply carries `onGround: false` and no collision whatever `bot.entity.onGround` holds, and it leaves the last-sent record alone. The test now starts capturing before the teleport is sent and asserts the reply is the only movement packet that is not grounded, with reminders after it reporting `true`. On the previous commit it fails for all tested versions (the reply carries `true`).

rom1504 (commented):

**Looks ready to merge from this review.** The retained standing state is now independent of the immediate/delayed teleport reply's ungrounded flags, and the regression captures packets before sending the teleport. I reran the physics suite across the version matrix: 135 passing, 33 pending. Current CI is green. These are local protocol client/server checks, not vanilla-server validation.

Skills used: prismarine-geometry-movement-review checked simulated standing state versus teleport reply flags; prismarine-lifecycle-action-review followed immediate and delayed replies. prismarine-review checked current discussions and avoided duplicate findings.

rom1504:
Approved, please resolve conflicts and merge

PR #4112, opened 2026-09-12
scoreboard: read the 1.20.3+ objective and score packets

u9g:
## Problem

1.20.3 reshaped both scoreboard packets and mineflayer still reads the pre-1.20.3 shapes, so the sidebar API is empty on every server from 1.20.3 on.

**`scoreboard_score` lost its `action` field.** The handler guards on `packet.action === 0`, which is never true:

```js
if (scoreboard !== undefined && packet.action === 0) {
  const updated = scoreboard.add(packet.itemName, packet.value)
```

so no score is ever recorded — `bot.scoreboards.<name>.itemsMap` stays `{}` and `scoreUpdated` never fires. Removal moved to its own `reset_score` packet, which nothing listens for, so `scoreRemoved` never fires either.

```js
// 26.1 production server, sidebar objective present and displayed
bot.scoreboard.sidebar.itemsMap   // {}
```

**The objective title became an NBT component** in the same release, and `setTitle` `JSON.parse`s it, so the `catch` leaves the raw prismarine-nbt tag in `scoreboard.title`:

```js
String(bot.scoreboard.sidebar.title)   // '[object Object]'
bot.scoreboard.sidebar.title           // { type: 'compound', value: { color: ..., text: ..., bold: ... } }
```

Even where the parse works (1.13–1.20.2) it keeps only the top-level `text`, dropping the `extra` array that carries most server titles.

## Fix

- `scoreboard_score`: read the score unless the packet is the old action-1 removal.
- `reset_score`: new handler. An absent `objective_name` means the score is dropped from every objective, which is what vanilla does.
- `setTitle`: `ChatMessage.fromNotch`, which already covers the plain string (1.8), the JSON string (1.13) and the NBT component (1.20.3) — the same treatment `open_window` titles get. `ScoreBoard.title` is now a `ChatMessage`; `String(title)` gives the text as before, with the extras included.
- 1.20.3+ can send a component to draw in place of a score's own name (`display_name`); it becomes that entry's `displayName` when present, otherwise the team-formatted name as before.

One small behaviour change beyond that: removing a score that the objective does not hold no longer emits `scoreRemoved` with an `undefined` item. Real 1.20.3+ servers send `reset_score` for entities that were never scored (a lobby sends a burst of them on join), so the event would otherwise fire constantly with nothing in it.

## Tests

Four cases under `describe('scoreboard')` in `internalTest.js`, driving a real server: the objective title (every version), a score recorded and then dropped (through `action: 1` before 1.20.3, through `reset_score` after), the per-score display name, and a `reset_score` with no objective name clearing the score everywhere. The last two are 1.20.3+ only. All pass across the version matrix (1.8.8 through 26.1); the first two fail on master for 1.20.3+.

### DIFF
--- a/lib/plugins/scoreboard.js
+++ b/lib/plugins/scoreboard.js
@@ -38,26 +38,35 @@ function inject (bot) {
     }
   })
 
+  function removeScore (scoreboard, itemName) {
+    if (scoreboard !== undefined) {
+      if (!(itemName in scoreboard.itemsMap)) return
+      const removed = scoreboard.remove(itemName)
+      return bot.emit('scoreRemoved', scoreboard, removed)
+    }
+
+    // No objective named: the score is dropped from every objective that holds it.
+    for (const sb of Object.values(scoreboards)) {
+      if (itemName in sb.itemsMap) {
+        const removed = sb.remove(itemName)
+        bot.emit('scoreRemoved', sb, removed)
+      }
+    }
+  }
+
   bot._client.on('scoreboard_score', (packet) => {
     const scoreboard = scoreboards[packet.scoreName]
-    if (scoreboard !== undefined && packet.action === 0) {
-      const updated = scoreboard.add(packet.itemName, packet.value)
-      bot.emit('scoreUpdated', scoreboard, updated)
-    }
+    // 1.20.3 dropped the action field: the packet only ever sets a score, and a removal arrives as
+    // reset_score. Before that, action 1 meant the score was removed and carried no value.
+    if (packet.action === 1) return removeScore(scoreboard, packet.itemName)
+    if (scoreboard === undefined) return
+    const updated = scoreboard.add(packet.itemName, packet.value, packet.display_name)
+    bot.emit('scoreUpdated', scoreboard, updated)
+  })
 
-    if (packet.action === 1) {
-      if (scoreboard !== undefined) {
-        const removed = scoreboard.remove(packet.itemName)
-        return bot.emit('scoreRemoved', scoreboard, removed)
-      }
-
-      for (const sb of Object.values(scoreboards)) {
-        if (packet.itemName in sb.itemsMap) {
-          const removed = sb.remove(packet.itemName)
-          return bot.emit('scoreRemoved', sb, removed)
-        }
-      }
-    }
+  // 1.20.3+. An absent objective name means every objective.
+  bot._client.on('reset_score', (packet) => {
+    removeScore(packet.objective_name == null ? undefined : scoreboards[packet.objective_name], packet.entity_name)
   })
 
   bot._client.on('scoreboard_display_objective', (packet) => {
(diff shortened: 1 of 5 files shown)


lib/plugins/scoreboard.js
```diff
@@ -38,26 +38,35 @@ function inject (bot) {
...
+    if (packet.action === 1) return removeScore(scoreboard, packet.itemName)
+    if (scoreboard === undefined) return
+    const updated = scoreboard.add(packet.itemName, packet.value, packet.display_name)
+    bot.emit('scoreUpdated', scoreboard, updated)
+  })
+
+  // 1.20.3+. An absent objective name means every objective.
+  bot._client.on('reset_score', (packet) => {
+    removeScore(packet.objective_name == null ? undefined : scoreboards[packet.objective_name], packet.entity_name)
   })
```
rom1504, line 70: **Astra agent review — AI-generated, not manually written by the maintainer.**

An explicitly named unknown objective also becomes `undefined` here, so `removeScore` treats it as a request to clear every objective. I reproduced this with scores for `player` in `kills` and `deaths`: `reset_score { entity_name: 'player', objective_name: 'missing' }` deleted both scores and emitted two `scoreRemoved` events. A reset for an objective that has already been removed should leave the other objectives intact. Could you distinguish an absent `objective_name` from a failed lookup, and add a regression test alongside the existing no-objective case?

rom1504 (commented):

Reviewed at the maintainer's request. I inspected the current scoreboard changes and tests, and locally exercised the current handler with real ScoreBoard/prismarine-chat objects and an EventEmitter packet source. The unknown-objective reset below clears unrelated scores; existing-objective and absent-objective controls behaved as expected. I did not run a live-server test or the full suite.

lib/plugins/scoreboard.js
u9g: Fixed in c3fcdfd3efa8f352c835ec0870c8be14dc99d785. `removeScore` now takes the objective name, not the looked-up scoreboard. It clears every objective only when the name is absent (or empty, which is how pre-1.20.3 `scoreboard_score` action 1 signalled "all"). A named objective the bot doesn't know is a no-op. The same fall-through affected the pre-1.20.3 `action: 1` path, so that's fixed too. I added a regression test ('leaves every objective alone when a removal names an unknown one') that runs on every tested version through whichever removal packet that version uses. It fails on all 28 versions without the fix and passes with it.

rom1504 (commented):

**Looks ready to merge from this review.** An explicitly unknown objective now leaves other objectives alone, while an absent objective still clears all matching scores. I reran the scoreboard tests across all 28 tested versions: 106 passing, including both legacy action-based removal and reset_score. The title/display-name API changes are reflected in declarations and documentation. Current CI is green. Validation used local protocol client/server tests, not vanilla.

Skills used: prismarine-protocol-data-review checked the old/new score packet contracts; prismarine-architecture-review checked the shared ScoreBoard model and its public declarations. prismarine-review checked current discussions and avoided duplicate findings.

rom1504:
Approved, resolve conflicts and merge

PR #4113, opened 2026-09-12
Read the 1.20.3+ scoreboard packets: scores, resets and NBT titles

u9g:
### Problem

1.20.3 changed three things about the scoreboard packets, and mineflayer still reads the 1.20.2 shapes. On every version from 1.20.3 on (so 1.20.3 … 1.21.x, 26.1):

| what | 1.20.2 | 1.20.3+ |
| --- | --- | --- |
| set a score | `scoreboard_score { itemName, action, scoreName, value }` | `scoreboard_score { itemName, scoreName, value, display_name?, number_format? }` — no `action` |
| remove a score | `scoreboard_score` with `action: 1` | `reset_score { entity_name, objective_name? }` |
| objective display text | JSON string | NBT component |

The consequences:

- **Every score is dropped.** The handler adds only when `packet.action === 0`; the field is gone, `undefined === 0` is false, so `bot.scoreboard.sidebar.items` stays `[]` for the life of the bot and `scoreUpdated` never fires. A server-side sidebar — the thing most minigame servers put every piece of game state into — is completely invisible to a bot.
- **No score is ever removed.** `reset_score` had no handler at all, so once the first bug is fixed, entries would accumulate forever.
- **`ScoreBoard.title` is raw NBT.** `setTitle` does `JSON.parse(title).text`, which throws on an object and falls through to keeping the object as-is, so a title prints as `[object Object]`.

Reproduced on a 26.1 production server (jartexnetwork BedWars): `bot.scoreboard.sidebar` is `{ name: 'page0', title: [object Object], items: [] }` while the client shows a full sidebar.

### Fix

- `scoreboard_score` with no `action` is always an update (`packet.action === undefined || packet.action === 0`); the pre-1.20.3 `action: 1` removal path is untouched.
- New `reset_score` handler: removes the entry from the named objective, or from every objective when the packet names none — which is what vanilla's `ResetScorePacket` with an absent objective means.
- `setTitle` goes through `ChatMessage.fromNotch`, which reads an NBT component, a JSON string and a plain string alike. That also fixes the old path silently throwing away everything but the first text run of a multi-part title (`{text:"",extra:[…]}` used to give `""`).
- 1.20.3+ also lets the server override how a single entry renders (`display_name`). `item.displayName` prefers that component when one is sent, and otherwise falls back to the team-formatted entity name exactly as before.

### Compatibility

`ScoreBoard.title` is now a `ChatMessage` instead of a string. `ScoreBoardItem.displayName` has always been a `ChatMessage`, so this makes the class consistent, and `title.toString()` gives the text the field used to hold. `index.d.ts` and `docs/api.md` are updated.

### Tests

`test/scoreboardTest.js` drives the plugin directly off the wire shapes — no server needed, the plugin only wants a registry, a packet source and an event bus. It covers the 1.20.2 action-based add/remove path, the 1.20.3+ actionless add, `reset_score` with and without an objective name, the per-score `display_name`, its fallback, an NBT title, a multi-run NBT title, and a plain-string 1.8 title. Five of the eight fail on master.

`standard` is clean.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -678,7 +678,8 @@ Name of the scoreboard.
 
 #### ScoreBoard.title
 
-The title of the scoreboard (does not always equal the name)
+The title of the scoreboard (does not always equal the name), as a `ChatMessage`. Servers send it
+as a chat component, so `.toString()` for the plain text and `.toAnsi()` for the coloured form.
 
 #### ScoreBoard.itemsMap
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -789,15 +789,15 @@ export interface VillagerTrade {
 
 export class ScoreBoard {
   name: string
-  title: string
+  title: ChatMessage
   itemsMap: { [name: string]: ScoreBoardItem }
   items: ScoreBoardItem[]
 
   constructor (packet: object);
 
-  setTitle (title: string): void;
+  setTitle (title: string | object): void;
 
-  add(name: string, value: number): ScoreBoardItem;
+  add(name: string, value: number, displayName?: string | object): ScoreBoardItem;
 
   remove (name: string): ScoreBoardItem;
 }
--- a/lib/plugins/scoreboard.js
+++ b/lib/plugins/scoreboard.js
@@ -40,9 +40,13 @@ function inject (bot) {
 
   bot._client.on('scoreboard_score', (packet) => {
     const scoreboard = scoreboards[packet.scoreName]
-    if (scoreboard !== undefined && packet.action === 0) {
-      const updated = scoreboard.add(packet.itemName, packet.value)
+    // 1.20.3 moved removal into its own reset_score packet, so this packet lost its action
+    // field and now only ever sets a score.
+    if (packet.action === undefined || packet.action === 0) {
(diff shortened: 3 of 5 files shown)


u9g:
Closing as a duplicate — #4111 and #4112 landed the same two fixes (ChatMessage title, actionless `scoreboard_score`, `reset_score`, per-score `display_name`) about five minutes before this, and #4112's shared `removeScore` helper is the tidier shape. Nothing here is missing from that pair.

The one piece that does not overlap is the test: this branch carries `test/scoreboardTest.js`, which drives the plugin straight off the wire shapes with a stub bot (a registry, an EventEmitter for `_client` and one for the bot) instead of standing up a server, so it can cover 1.8, 1.20.2 and 1.21.4 in the same run — the 1.20.2 action-based add/remove path, the actionless add, `reset_score` with and without an objective name, the per-score `display_name` and its fallback, an NBT title, a multi-run NBT title and a plain-string title. Eight cases, five of which fail on master, and it runs in 140 ms.

Happy to send it as a follow-up on top of #4112 if that is useful; the branch stays at u9g:fix/scoreboard-scores-1-20-3.

PR #4114, opened 2026-09-12
placeEntity: write a complete use_item when placing a boat

u9g:
### Problem

`lib/plugins/place_entity.js` finishes a boat placement with

```js
bot._client.write('use_item', {
  hand: options.offhand ? 1 : 0
})
```

`use_item` grew a `sequence` field in 1.19 and a `rotation` field in 1.21.2. Every other `use_item` in the codebase sends both — see `activateItem` in `plugins/inventory.js`.

So on 1.19 – 1.21.1 the packet carries a garbage sequence, and on **1.21.2 and newer the serializer throws** on the absent rotation:

```
$ node -e "…createSerializer({state:'play',isServer:false,version:'1.21.4'})
           .createPacketBuffer({name:'use_item',params:{hand:0}})"
in packet use_item: SizeOf error for undefined : Cannot read properties of undefined (reading 'x')
```

A throw out of `client.write` takes the connection down with it, so `bot.placeEntity(…)` with a boat in hand disconnects the bot on every current version.

Nothing caught this because `placeEntity` is listed in `excludedTests` in `test/externalTest.js`.

### Fix

Send the same payload `activateItem` does: `hand`, `sequence` and `rotation` from the bot's current yaw/pitch. `sequence: 0` matches what `generic_place.js` already writes; PR #4085 replaces both with a real shared counter, and this change is written so that rebases cleanly.

### Test

`test/placeEntityPacketTest.js` drives `bot.placeEntity` with a stub for `_genericPlace` and a `_client.write` that runs the real `minecraft-protocol` serializer, so an incomplete packet fails the test instead of a live connection. It covers 1.18.2 (which takes the old `block_place` path), 1.19.4, 1.20.4 and 1.21.4, and asserts exactly one item-use packet goes out. The 1.21.4 case fails on master with the `SizeOf error` above.

Found while running a bot on a 26.1 production server, where a throw from `client.write` is not recoverable.

### DIFF
--- a/lib/plugins/place_entity.js
+++ b/lib/plugins/place_entity.js
@@ -1,4 +1,5 @@
 const assert = require('assert')
+const { toNotchianYaw, toNotchianPitch } = require('../conversions')
 
 module.exports = inject
 
@@ -36,8 +37,15 @@ function inject (bot) {
 
     if (type === 'boat') {
       if (bot.supportFeature('useItemWithOwnPacket')) {
+        // Same payload as bot.activateItem(): 1.19 added the prediction sequence and 1.21.2 the
+        // rotation, and leaving either out makes the serializer throw, which ends the connection.
         bot._client.write('use_item', {
-          hand: options.offhand ? 1 : 0
+          hand: options.offhand ? 1 : 0,
+          sequence: 0, // 1.19.0+, as in generic_place
+          rotation: {
+            x: toNotchianYaw(bot.entity.yaw),
+            y: toNotchianPitch(bot.entity.pitch)
+          }
         })
       } else {
         bot._client.write('block_place', {
--- a/test/placeEntityPacketTest.js
+++ b/test/placeEntityPacketTest.js
@@ -1,0 +1,59 @@
+/* eslint-env mocha */
+
+const assert = require('assert')
+const EventEmitter = require('events').EventEmitter
+const Vec3 = require('vec3').Vec3
+const registryLoader = require('prismarine-registry')
+const mc = require('minecraft-protocol')
+
+// Drives bot.placeEntity far enough to see what it puts on the wire, and serializes every packet
+// with the real protocol so a missing field fails here instead of killing a live connection.
+function fakeBot (version) {
+  const registry = registryLoader(version)
+  const bot = new EventEmitter()
+  bot.registry = registry
+  bot.version = version
+  bot.supportFeature = registry.supportFeature.bind(registry)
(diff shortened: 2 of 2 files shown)


test/placeEntityPacketTest.js
```diff
@@ -0,0 +1,59 @@
...
+      // throws exactly as the real client would on an incomplete packet
+      serializer.createPacketBuffer({ name, params })
+      bot.written.push({ name, params })
+    }
+  }
+  bot._genericPlace = async (referenceBlock) => referenceBlock.position
+  require('../lib/plugins/place_entity')(bot)
+  return bot
+}
+
```
extremeheat, line 31: Seems like alot of these PRs are stacking code similar to this. There should be a better way to do this cleanly

lib/plugins/place_entity.js
```diff
@@ -36,8 +37,15 @@ function inject (bot) {
 
     if (type === 'boat') {
       if (bot.supportFeature('useItemWithOwnPacket')) {
+        // Same payload as bot.activateItem(): 1.19 added the prediction sequence and 1.21.2 the
+        // rotation, and leaving either out makes the serializer throw, which ends the connection.
         bot._client.write('use_item', {
-          hand: options.offhand ? 1 : 0
+          hand: options.offhand ? 1 : 0,
+          sequence: 0, // 1.19.0+, as in generic_place
```
rom1504, line 44: **Astra agent review — AI-generated, not manually written by the maintainer.**

#4085 has now merged, and [the merged boat path](https://github.com/PrismarineJS/mineflayer/blob/4fb62af5f8807adbf0c76c3d91221be246f40e8b/lib/plugins/place_entity.js#L40-L49) already writes both rotation and `sequence: bot._nextSequence()`. That covers this PR's missing-payload fix while using the shared prediction counter.

Please rebase and drop the superseded runtime hunk; resolving the overlap by keeping `sequence: 0` would discard the shared-counter behavior. This could be closed as superseded, or narrowed to additional regression coverage. If retaining the test, its current fake bot has no `_nextSequence`, so it also needs adapting to the merged implementation; the existing real-bot NMP fixture would initialize that shared state.

Skills used: `prismarine-review` checked current base/dependency status and existing discussion; `prismarine-architecture-review` checked the shared sequence owner; `prismarine-protocol-data-review` checked what the retained fixture would actually exercise.

rom1504 (commented):

Current-base integration check using `prismarine-review`, `prismarine-architecture-review`, and `prismarine-protocol-data-review`: #4085 has landed since the earlier discussion. One inline update covers the resulting overlap and test adaptation.

PR #4115, opened 2026-09-12
placeBlock: derive the face and the cursor from a raycast, like the client does

u9g:
### Problem

`_genericPlace` decides which face it wants, aims at the centre of that face, and puts both the chosen face and a synthesised cursor in the packet:

```js
// Look at the center of the face
let dx = 0.5 + faceVector.x * 0.5
…
await bot.lookAt(referenceBlock.position.offset(dx, dy, dz), options.forceLook)
…
bot._client.write('block_place', { location: pos, direction: vectorToDirection(faceVector), cursorX: dx, … })
```

The vanilla client cannot do that. `Minecraft.pick` raycasts from the camera once per frame and `useItemOn` is handed exactly that `BlockHitResult` — block, face and hit point together:

```java
this.hitResult = this.player.raycastHitResult(partialTicks, cameraEntity);
…
BlockHitResult blockHit = (BlockHitResult) this.hitResult;
this.gameMode.useItemOn(this.player, hand, blockHit);   // -> new ServerboundUseItemOnPacket(hand, blockHit, sequence)
```

So in vanilla the face and cursor are *derived from the look*, never chosen alongside it, and a `use_item_on` whose face is not the one the player's own ray reaches first is a packet no client can send.

**Aiming at the centre of a side face gets this wrong every time.** The centre of a side face is on the far side of the block's own body from anything looking down at it, so the ray reaches the top face first. Bridging out from an island, laying wool along a row:

| placement | eye | aimed at (claimed face centre) | claimed face | face the ray actually hits |
|---|---|---|---|---|
| `-55,38,3` | `(-55.5, 40.62, 3.5)` | `(-55, 38.5, 3.5)` | `+x` | **`+y`** at `(-55.118, 39, 3.5)` |
| `-54,38,3` | `(-55.5, 40.62, 3.5)` | `(-54, 38.5, 3.5)` | `+x` | **`+y`** at `(-54.354, 39, 3.5)` |
| `-53,38,3` | `(-55.5, 40.62, 3.5)` | `(-53, 38.5, 3.5)` | `+x` | **`+y`** at `(-53.590, 39, 3.5)` |

The bot sends `direction = EAST` with a cursor at `y = 0.5` while the rotation in the same packet puts its crosshair on the top face at `y = 1.0`. A server that checks the hit reverts the placement silently — which is what a bridging loop on a 26.1 production server saw, and it is also why real players bridge by sneaking to the edge and looking *down and back*: that is the only way to get a crosshair onto a side face.

### Fix

Do what the client does:

1. Generate candidate aim points on the requested face — centre first, then a grid working outwards, so a face that is only partly visible still has somewhere to aim. A requested slab `half` biases the vertical candidates into that half, which is how vanilla picks it too.
2. Raycast from the eye through each candidate with `bot.world.raycast` (the same call `blockAtCursor` uses) and keep the first whose hit is that block on that face.
3. Look along it and send **the hit the raycast returned**: its `face` for `direction`, and its intersection point relative to the block as the cursor.
4. If no candidate is visible, the face genuinely cannot be reached from where the bot stands. Throw `Cannot place against the <face> face of <block> at <pos>: it is not visible from <pos>` rather than send a packet no client could have produced — the caller can step to the edge and retry.

`options.delta` and `options.forceLook === 'ignore'` keep their current meaning: a caller that supplies its own cursor, or forbids turning, has said it knows where it is aiming, so those paths are untouched.

### Behaviour change worth calling out

Step 4 turns a class of silently-reverted placements into a thrown error. That is the point — the packet was never going to work on a server that validates it — but callers that stand in the wrong place and got away with it on a lenient server will now see an error instead of "the block is still air" a moment later. The message names the face and both positions, so the fix is usually a one-block step.

### Tests

`test/genericPlaceRaycastTest.js` builds a bot around a real `prismarine-world` chunk (no server) and, for each placement, raycasts the bot's own `yaw`/`pitch` from its eye and asserts the packet names the block that ray hits, the face it hits, and the exact point as the cursor. It covers a side face from the block behind it, a side face from a block up and back, a top face, a side face at eye level, the `delta` and `forceLook: 'ignore'` passthroughs, and a face walled off on every side (which must throw and write nothing). Three of the seven fail on master.

Reported in u9g/bot-harness#90 with the full raycast arithmetic for the bridging case above.

### DIFF
--- a/lib/plugins/generic_place.js
+++ b/lib/plugins/generic_place.js
@@ -1,8 +1,36 @@
 const assert = require('assert')
+const { Vec3 } = require('vec3')
 module.exports = inject
+
+// Points on `faceVector`'s face of a unit cube: the centre first, then a grid working outwards,
+// so a face that is only partly visible still has somewhere to aim. When a slab half is asked for,
+// the candidates on the face's vertical axis start in that half, since that is what picks it.
+function faceAimPoints (faceVector, half) {
+  const axis = faceVector.x !== 0 ? 'x' : faceVector.y !== 0 ? 'y' : 'z'
+  const inFace = ['x', 'y', 'z'].filter(a => a !== axis)
+  const spread = [0.5, 0.3, 0.7, 0.15, 0.85]
+  const vertical = half === 'top'
+    ? [0.75, 0.9, 0.6, 0.55]
+    : half === 'bottom' ? [0.25, 0.1, 0.4, 0.45] : spread
+  const stepsFor = (a) => (a === 'y' ? vertical : spread)
+  const out = []
+  for (const u of stepsFor(inFace[0])) {
+    for (const v of stepsFor(inFace[1])) {
+      const p = { x: 0, y: 0, z: 0 }
+      p[axis] = faceVector[axis] > 0 ? 1 : 0
+      p[inFace[0]] = u
+      p[inFace[1]] = v
+      out.push(new Vec3(p.x, p.y, p.z))
+    }
+  }
+  return out
+}
 
 function inject (bot) {
   const Item = require('prismarine-item')(bot.registry)
+  // When a face turns out not to be in view, refuse rather than send a packet no client could
+  // produce. Off by default: the packets were accepted by vanilla servers before.
+  bot.placeFaceStrict = false
   /**
    *
    * @param {import('prismarine-block').Block} referenceBlock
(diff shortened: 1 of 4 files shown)


lib/plugins/generic_place.js
```diff
@@ -20,21 +48,53 @@ function inject (bot) {
...
+      }
+    } else {
+      // A block with no collision box - air, water, tall grass - is not pickable at all, so there
+      // is nothing to raycast against and those callers keep the old behaviour.
+      const pickable = referenceBlock.shapes && referenceBlock.shapes.length > 0
+      const hit = pickable && options.forceLook !== 'ignore'
+        ? aimAtFace(referenceBlock, faceVector, options.half)
+        : null
+      if (hit) {
+        await bot.lookAt(hit.aim, options.forceLook)
```
rom1504, line 75: **Astra agent review — AI-generated, not manually written by the maintainer.**

The hit is computed before the awaited turn, but movement can change the eye position before this saved cursor is sent. I reproduced this with the PR’s real prismarine-world/chunk fixture: starting above `(0,63,0)`, moving the bot +0.2 on X during `lookAt` leaves the packet cursor at `(0.5,1,0.5)` while the final crosshair hits `(0.7,1,0.5)`, even with `strictFace: true`. A larger move can invalidate the face/block too. Please raycast from the current eye and final rotation after the await, revalidate the requested block/face, and use that hit for the packet (or retry/refuse according to strict/fallback policy). Add the moving-during-look case to the existing world-backed tests.

rom1504 (commented):

Reviewed at the maintainer’s request. Validation: Local reproduction using current PR plugin and existing real prismarine-world/chunk fixture; movement during look produces saved-versus-final cursor mismatch. Static inspection of current tests/docs and fallback policy. One reproduced placement-timing issue is noted inline.

lib/plugins/generic_place.js
```diff
@@ -1,8 +1,36 @@
...
+  const spread = [0.5, 0.3, 0.7, 0.15, 0.85]
+  const vertical = half === 'top'
+    ? [0.75, 0.9, 0.6, 0.55]
+    : half === 'bottom' ? [0.25, 0.1, 0.4, 0.45] : spread
+  const stepsFor = (a) => (a === 'y' ? vertical : spread)
+  const out = []
+  for (const u of stepsFor(inFace[0])) {
+    for (const v of stepsFor(inFace[1])) {
+      const p = { x: 0, y: 0, z: 0 }
+      p[axis] = faceVector[axis] > 0 ? 1 : 0
```
rom1504, line 20: **Astra agent review — AI-generated, not manually written by the maintainer.**

These candidates lie on a unit cube, so a visible face on a partial block can be reported as invisible. With the existing real world/chunk fixture, a bottom `oak_slab` at `(0,63,0)` and a stationary bot at `(0.5,64,-3)` (`eyeHeight: 1.62`) has a clear TOP hit at `(0.5,63.5,0.5)`, but `_genericPlace(..., new Vec3(0,1,0), { forceLook: true, strictFace: true })` throws `not visible`. The candidates use local Y=1 instead of the slab's Y=0.5; with strict mode off, the fallback's final crosshair misses the slab entirely. Please generate aim candidates from the block's actual shape faces and add this non-full-cube case to the world-backed tests. This occurs without movement during `lookAt`.

Skills used: prismarine-behavior-test-review helped test visibility with the real chunk geometry and an independently chosen surface hit; prismarine-review helped distinguish this from the existing movement-timing finding.

rom1504 (commented):

Re-reviewed head 27805567aa88bbc140c5b3d91766fe8a47fd105b and current discussion. All nine existing raycast tests pass; a focused stationary bottom-slab reproduction shows a visible top face rejected by strict mode. No vanilla server was run.

Skills used: prismarine-behavior-test-review guided validation against the real world/chunk geometry; prismarine-review helped identify the separate root cause and avoid repeating prior feedback.

lib/plugins/generic_place.js
u9g: Fixed in 10cd5d25a805602b638ee42553b4ed5b3e89bbd4. After the `lookAt` await, `lookAtFace` now raycasts from the current eye along the current yaw/pitch and sends that hit only if it is still the requested block and face. If it isn't, it re-aims from where the bot now stands (up to 3 tries), then falls back, or throws when `strictFace` is set. I added two world-backed tests: a +0.2 x drift during the turn (the cursor matches the final crosshair) and a move that takes the face out of view (strict throws and writes nothing).

lib/plugins/generic_place.js
u9g: Fixed in 10cd5d25a805602b638ee42553b4ed5b3e89bbd4. The aim candidates now come from `referenceBlock.shapes` (each collision box's face on the requested side), not from a unit cube. A requested `half` narrows the vertical range to that half when the box reaches into it. I added your case as a test: a bottom `oak_slab` at (0,63,0), the bot at (0.5,64,-3), top face, `strictFace: true`. It now places with cursor y = 0.5, and the cursor matches the crosshair's hit.

rom1504 (commented):

The two earlier findings are fixed at 10cd5d2: candidates use actual collision shapes, and the cursor comes from a fresh ray after the awaited turn. All 12 world-backed raycast tests pass locally, including the bottom slab, movement during turning, strict rejection and explicit-cursor/ignore-look controls. I found no additional code blocker.

Please rerun the failed CI jobs before merging. Their logs show a 1.14.4 server bind failure and an unrelated 1.19.2 scoreboard-reset assertion; they do not show a failing placement assertion. I have not established that the latter is a baseline failure. No vanilla-server suite was rerun locally.

Skills used: `prismarine-geometry-movement-review` checked shape surfaces and post-await eye/rotation consistency; `prismarine-review` compared the fixed threads and separated CI evidence from a new defect.

PR #4116, opened 2026-09-12
waitForTicks: say when it is physics that is missing, not the server

u9g:
### Problem

`physicsTick` is only emitted while the simulation runs:

```js
if (bot.physicsEnabled && shouldUsePhysics) {
  …
  bot.emit('physicsTick')
}
```

so `bot.waitForTicks(n)` on a bot with `bot.physicsEnabled = false` never advances. It waits out `n * 50 + 5000` ms and then rejects with

```
Timeout waiting for 2 ticks after 6000ms
```

which reads as a lagging server and sends you looking in the wrong place. Anything that freezes the bot — an anticheat guard that turns physics off in a pre-game lobby is the usual reason on real servers — turns every `waitForTicks` in a script into a multi-second stall with a misleading error. `mineflayer-pathfinder#381` fixed the same trap for `human.walkTo`/`lookAt`; this is the one underneath it.

That example message shows a second bug: it says **2** ticks when the call asked for **20**. The tick listener decrements the same `ticks` variable the message reads, so the number printed is whatever was still outstanding, never the number requested.

### Fix

- Reject up front when physics is already off, with `bot.waitForTicks needs physics: bot.physicsEnabled is false` — the same wording pathfinder#381 uses, so the two read alike.
- If physics is switched off *during* the wait, append `: physics was disabled while waiting` to the timeout message.
- Capture the requested count so the timeout names it.

No behaviour change for a bot with physics running.

### Tests

`test/waitForTicksTest.js` injects the physics plugin against a stub bot (no server, no world) and covers a normal wait, the up-front rejection, and the timeout message. The last two fail on master.

### DIFF
--- a/test/waitForTicksTest.js
+++ b/test/waitForTicksTest.js
@@ -1,0 +1,53 @@
+/* eslint-env mocha */
+
+const assert = require('assert')
+const EventEmitter = require('events')
+const Vec3 = require('vec3').Vec3
+const registryLoader = require('prismarine-registry')
+
+// The physics plugin injects against a stub, so waitForTicks can be exercised without a server.
+function fakeBot () {
+  const registry = registryLoader('1.21.4')
+  const bot = new EventEmitter()
+  bot.registry = registry
+  bot.version = '1.21.4'
+  bot.supportFeature = registry.supportFeature.bind(registry)
+  bot.blockAt = () => null
+  bot._client = new EventEmitter()
+  bot._client.write = () => {}
+  bot.entity = { position: new Vec3(0, 64, 0), velocity: new Vec3(0, 0, 0), yaw: 0, pitch: 0, onGround: true, height: 1.8, eyeHeight: 1.62 }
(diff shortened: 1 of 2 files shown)


test/waitForTicksTest.js
```diff
@@ -0,0 +1,53 @@
...
+  bot.supportFeature = registry.supportFeature.bind(registry)
+  bot.blockAt = () => null
+  bot._client = new EventEmitter()
+  bot._client.write = () => {}
+  bot.entity = { position: new Vec3(0, 64, 0), velocity: new Vec3(0, 0, 0), yaw: 0, pitch: 0, onGround: true, height: 1.8, eyeHeight: 1.62 }
+  bot.game = {}
+  require('../lib/plugins/physics')(bot, {})
+  return bot
+}
+
```
extremeheat, line 23: This code seems quite messy. Is there not way to do this properly with a nmp server/client combo ?

PR #4118, opened 2026-09-12
Expose the interaction ranges the server sends, and the reach check vanilla makes

u9g:
### Problem

An interact packet sent from out of reach is dropped by the server without a reply. So a bot 20 cm too far from a shopkeeper gets no window, no error, and no chat line — and nothing in the API mentions a range, so working out what happened takes a while. Two dead ends in one session came down to this:

```js
// a walled-in BedWars shopkeeper; feet-to-feet distance in the comment
d = 3.15  ->  bot.activateEntity(shop)   // no windowOpen in 6 s, no error
d = 2.80  ->  bot.activateEntity(shop)   // 'Shop - Quickbuy'

d = 3.56  ->  bot.attack(npc)            // nothing
d = 2.84  ->  bot.attack(npc)            // the NPC's menu opens
```

The server has been sending the exact numbers since 1.20.5, and mineflayer already folds them into `bot.entity.attributes` and then never reads them:

```js
bot.entity.attributes['player.entity_interaction_range']  // { value: 3, modifiers: [] }
bot.entity.attributes['player.block_interaction_range']   // { value: 4.5, modifiers: [] }
```

### What this adds

```js
bot.entityInteractionRange()                    // the attribute, else vanilla's 3.0
bot.blockInteractionRange()                     // the attribute, else vanilla's 4.5
bot.canInteractWithEntity(entity, buffer = 0)
bot.canInteractWithBlock(block, buffer = 0)
```

The predicates follow vanilla exactly (26.1 `Player.java`):

```java
public boolean isWithinEntityInteractionRange(final AABB aabb, final double buffer) {
    double maxRange = this.entityInteractionRange() + buffer;
    double distanceToSq = aabb.distanceToSqr(this.getEyePosition());
    return distanceToSq < maxRange * maxRange;
}
```

— squared distance from the **eye** to the target's **bounding box**, not centre to centre, which is a real difference for a 1.95-tall villager you are standing next to. `buffer` is the slack the server gives itself (`isWithinEntityInteractionRange(targetBounds, 3.0)` in `ServerGamePacketListenerImpl`), so a caller can ask either "could a player standing here have clicked that?" (`0`) or "will this particular server accept it?" (`3.0`).

The attribute lookup matches on the last segment of the key, because the same attribute arrives as `generic.armor`, `minecraft:armor` or `armor` depending on the version and protocol build, and the ranges as `player.entity_interaction_range`. `getAttributeValue` moves out of `plugins/explosion.js` into `lib/attributes.js` with it, unchanged, rather than existing twice.

### No behaviour change

Nothing calls the predicates. They are what an out-of-reach check inside `attack`/`activateEntity` would be built on, but making those throw is a breaking change for every combat plugin that swings in a loop, so it belongs in its own PR (and probably behind a flag). This PR is just the numbers and the geometry, so that a caller can do the check today:

```js
if (!bot.canInteractWithEntity(villager)) await bot.pathfinder.goto(new goals.GoalNear(...))
bot.activateEntity(villager)
```

### Tests

`test/interactionRangeTest.js` injects the entities plugin against a stub bot — no server — and pins the vanilla defaults, both attribute spellings, the modifier arithmetic, the `buffer` argument, the block cube, and the eye-to-hitbox geometry against the two live measurements above (3.15 feet-to-feet is 2.85 eye-to-box and reachable; 3.56 is not).

Reported with the full session context in u9g/bot-harness#92.

### DIFF
--- a/lib/attributes.js
+++ b/lib/attributes.js
@@ -1,0 +1,38 @@
+// https://minecraft.wiki/w/Attribute#Operations
+function getAttributeValue (prop) {
+  let X = prop.value
+  for (const mod of prop.modifiers) {
+    if (mod.operation !== 0) continue
+    X += mod.amount
+  }
+  let Y = X
+  for (const mod of prop.modifiers) {
+    if (mod.operation !== 1) continue
+    Y += X * mod.amount
+  }
+  for (const mod of prop.modifiers) {
+    if (mod.operation !== 2) continue
+    Y += Y * mod.amount
+  }
+  return Y
+}
+
+// The key an attribute arrives under is not stable across versions or protocol builds - the same
+// attribute shows up as `generic.armor`, `minecraft:armor` or `armor`, and the interaction ranges
+// as `player.entity_interaction_range` - so match on the last segment, which is.
+function findAttribute (entity, name) {
+  const attributes = entity && entity.attributes
+  if (!attributes) return null
+  if (attributes[name]) return attributes[name]
+  for (const key of Object.keys(attributes)) {
+    if (key.split(/[:.]/).pop() === name) return attributes[key]
+  }
+  return null
+}
+
+function attributeValue (entity, name, fallback) {
+  const attribute = findAttribute(entity, name)
+  return attribute ? getAttributeValue(attribute) : fallback
+}
+
+module.exports = { getAttributeValue, findAttribute, attributeValue }
--- a/lib/plugins/entities.js
+++ b/lib/plugins/entities.js
@@ -1,6 +1,7 @@
 const { Vec3 } = require('vec3')
 const conv = require('../conversions')
 const mojangson = require('mojangson')
+const { attributeValue } = require('../attributes')
 // These values are only accurate for versions 1.14 and above (crouch hitbox changes)
 // Todo: hitbox sizes for sleeping, swimming/crawling, and flying with elytra
 const PLAYER_HEIGHT = 1.8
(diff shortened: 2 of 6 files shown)


test/interactionRangeTest.js
```diff
@@ -0,0 +1,74 @@
...
+  bot._client.write = () => {}
+  bot.entities = {}
+  bot.players = {}
+  bot.entity = { id: 1, position: new Vec3(0.5, 64, 0.5), eyeHeight: 1.62, width: 0.6, height: 1.8 }
+  require('../lib/plugins/entities')(bot, {})
+  bot.entity = bot.entities[1] ?? bot.entity
+  bot.entity.position = new Vec3(0.5, 64, 0.5)
+  bot.entity.eyeHeight = 1.62
+  return bot
+}
```
extremeheat, line 25: This code seems quite messy. Is there not way to do this properly with a nmp server/client combo ?

lib/attributes.js
```diff
@@ -0,0 +1,38 @@
...
+    if (key.split(/[:.]/).pop() === name) return attributes[key]
+  }
+  return null
+}
+
+function attributeValue (entity, name, fallback) {
+  const attribute = findAttribute(entity, name)
+  return attribute ? getAttributeValue(attribute) : fallback
+}
+
```
extremeheat, line 37: This looks messy

lib/attributes.js
extremeheat, line 1: Why not inside prismarine-entity ?

lib/plugins/entities.js
```diff
@@ -928,6 +929,43 @@ function inject (bot) {
     }
   }
 
+  // Vanilla Player.DEFAULT_ENTITY_INTERACTION_RANGE / DEFAULT_BLOCK_INTERACTION_RANGE. From
+  // 1.20.5 the server sends the live values as attributes (creative adds to both), and mineflayer
+  // already folds entity_update_attributes into entity.attributes.
+  const DEFAULT_ENTITY_INTERACTION_RANGE = 3.0
+  const DEFAULT_BLOCK_INTERACTION_RANGE = 4.5
+
+  bot.entityInteractionRange = () => attributeValue(bot.entity, 'entity_interaction_range', DEFAULT_ENTITY_INTERACTION_RANGE)
```
rom1504, line 938: **Astra agent review — AI-generated, not manually written by the maintainer.**

Please apply the interaction attributes' bounds after evaluating their modifiers. Vanilla sends the base value and modifiers separately, then clamps the effective entity/block interaction range to `[0, 64]` before its reach getters use it ([1.21.4 attribute bounds](https://docs.neoforged.net/docs/1.21.4/entities/attributes/); the same clamp is present in the 1.21.1 attribute implementation).

Through the real `entity_update_attributes` codec and this plugin on 1.21.1 and 1.21.4, entity base `3` plus an additive `-4` modifier makes `entityInteractionRange()` return `-1`. `canInteractWithEntity()` then returns true for a target 0.5 blocks from the eye to its box, because the negative range is squared; vanilla's effective range is zero and the check is false. Block base `4.5` plus `-5.5` has the same problem, while large positive modifiers can exceed the upper bound. Clamp the two interaction getters to their registered bounds and cover an underflow/overflow packet case. The existing generic modifier arithmetic can remain separate.

Skills used: prismarine-protocol-data-review helped trace wire base/modifiers versus effective values; prismarine-behavior-test-review helped reproduce the public predicate result; prismarine-review helped deduplicate the existing feedback.

rom1504 (commented):

One new finding on the effective interaction ranges. I checked the production entity handler and actual NMP encoders/decoders on 1.21.1 and 1.21.4, using minecraft-protocol 1.67.0 and minecraft-data 3.114.0, and traced vanilla's attribute packet, final clamping and reach getters. This was an in-process codec/plugin reproduction, not a live-server test.

Skills used: prismarine-protocol-data-review helped distinguish the transmitted base/modifiers from the effective attribute; prismarine-behavior-test-review helped exercise the public reach predicates; prismarine-review helped separate this defect from the existing ownership and test-quality feedback.

test/interactionRangeTest.js
u9g: Done in 0ff895756692c5923ab10fc2d40a7967acb7f0c1: the stub-bot test is gone. The cases now live in an `interaction range` block in `test/internalTest.js`, driving the bot through a minecraft-protocol server with real `login`, `position`, `spawn_entity` and `entity_update_attributes` packets, on every tested version (the attribute case skips below 1.20.5).

lib/attributes.js
u9g: Agreed, removed in 0ff895756692c5923ab10fc2d40a7967acb7f0c1. The `entity_update_attributes` codec emits `player.entity_interaction_range` / `player.block_interaction_range` on every version from 1.20.5, so the plugin now reads those keys directly and the suffix matching is gone. `lib/attributes.js` is back to just `getAttributeValue`, moved unchanged from `explosion.js`.

lib/attributes.js
u9g: prismarine-entity doesn't handle attributes today; mineflayer stores them itself in `entities.js`, so this file is only `getAttributeValue` moved out of `explosion.js` to share it. It would fit well there though. PrismarineJS/prismarine-entity#13 asks for exactly that, and it could apply the min/max bounds from minecraft-data generically. Happy to do that as a follow-up and switch mineflayer over once it's released, if you'd rather not wait on it here.

lib/plugins/entities.js
u9g: Fixed in 0ff895756692c5923ab10fc2d40a7967acb7f0c1: both getters now clamp the value after modifiers to vanilla's `[0, 64]`, and the new test sends `3 - 4` / `4.5 - 5.5` (→ 0, and an adjacent villager is out of reach) and `+100` (→ 64) through the real codec.

rom1504 (commented):

The effective-range defect is fixed at 0ff8957. I ran the new NMP client/server interaction-range suite across all 28 tested versions: 94 passing, 18 version skips, including negative and overflowing modifiers, defaults and eye-to-box distances. The moved attribute arithmetic remains shared with explosions, and the tests now use the real packet path.

The code looks ready from these checks. Two landing details remain: the current head only shows Socket Security checks, so get the normal test CI result; and settle the existing prismarine-entity ownership discussion with the maintainer. The author's proposed follow-up extraction is reasonable—the range repair does not require that whole move first.

Skills used: `prismarine-geometry-movement-review` checked clamping before reach comparisons; `prismarine-protocol-data-review` checked transmitted attributes through the production codec; `prismarine-architecture-review` assessed the shared helper and bounded follow-up; `prismarine-review` kept the existing human discussion explicit.

PR #4123, opened 2026-09-15
Normalize kick reasons to ChatMessage

### LINKED ISSUE #4122: kicked event's reason is typed as string but can be a raw chat-component object
## Description

`lib/plugins/kick.js` forwards `packet.reason` straight from the `kick_disconnect`/`disconnect` packets:

```js
bot._client.on('kick_disconnect', (packet) => {
  bot.emit('kicked', packet.reason, true)
})
bot._client.on('disconnect', (packet) => {
  bot.emit('kicked', packet.reason, false)
})
```

`index.d.ts` declares this as `kicked: (reason: string, loggedIn: boolean) => ...`, but on modern protocol versions `packet.reason` is decoded as a JSON chat-component object (e.g. `{ translate: 'disconnect.timeout' }`), not a string. Anything that does `` `kicked: ${reason}` `` or otherwise treats it as a string gets `"[object Object]"` instead of the actual kick message.

## To Reproduce

Get kicked/disconnected from a server on a recent protocol version (I hit this on 1.21-era realms via `mineflayer@4.39.0`) and log the `kicked` event's `reason` argument — it logs as an object, and any string coercion of it collapses to `[object Object]`.

## Expected behavior

`reason` should either genuinely be a string (matching the documented type), or the type should be corrected and docs updated to reflect that consumers need to parse/render it themselves.

## Suggested fix

Happy to send a PR. `lib/plugins/chat.js` already has the right pattern for this (`require('prismarine-chat')(bot.registry)` → render to a string), so `kick.js` could do the same before emitting, making `reason` actually match its declared `string` type. Only question is whether that's the preferred direction vs. just fixing the type declaration and leaving the raw value as-is — wanted to check before opening a PR in case there's a preference (e.g. backwards-compat concerns for anyone currently branching on the object shape).

## Versions

- mineflayer: 4.39.0
- node: (n/a — reproduced via a Bun app, but the bug is in `kick.js`/`index.d.ts`, not runtime-specific)

DallasCarraher:
Fixes #4122.

## What

`kick_disconnect`/`disconnect` packets carry the kick reason as a JSON string or NBT chat component depending on protocol version, but `kick.js` forwarded `packet.reason` unchanged. `index.d.ts` typed `kicked`'s `reason` as `string`, so anything that logged or interpolated it directly got `"[object Object]"`.

This normalizes the reason into a `ChatMessage` via `prismarine-chat` (`ChatMessage.fromNotch`), the same pattern `chat.js` uses for incoming chat. `kicked` now emits `(reason: ChatMessage, loggedIn: boolean)` on every version. Callers that want plain text can use `reason.toString()`.

The `kicked` typing in `index.d.ts` and the docs in `docs/api.md` are updated to match.

## Testing

- Added tests in `test/internalTest.js` for login and play kicks, checking the `ChatMessage`, its rendered text and the `loggedIn` flag. Play kicks use JSON or NBT reasons depending on the version.
- Ran `standard` on the changed files with no lint errors.
- Verified manually against a real realm kick (the scenario in the issue).


lib/plugins/kick.js
```diff
@@ -1,11 +1,33 @@
...
+    try {
+      return new ChatMessage(component).toString()
+    } catch (e) {
+      return JSON.stringify(component)
+    }
+  }
+
   bot._client.on('kick_disconnect', (packet) => {
-    bot.emit('kicked', packet.reason, true)
+    bot.emit('kicked', formatKickReason(packet.reason), true)
```
extremeheat: Seem like this should be emitting a stable ChatMessage

DallasCarraher: Good call — updated to emit the actual `ChatMessage` instance instead of stringifying it, using `ChatMessage.fromNotch()` (handles both the JSON-string and pre-parsed object shapes, same as chat.js does for `message`/`actionBar`). Updated `index.d.ts`'s `kicked` type to `ChatMessage` and the docs to match. This also let me drop the hand-rolled parsing in favor of the existing helper.

VasilisDragon:
**AI agent review**

Could we add `kicked` assertions to the existing `test/internalTest.js` harness for login and play kicks, covering JSON and NBT reasons? Checking the `ChatMessage`, rendered text and `loggedIn` flag would protect this behavior. The existing login-kick test only checks the later chat error.

The title/body also still describe returning a string. Could we update them to match the final `ChatMessage` API? Something like **“Normalize kick reasons to ChatMessage”** would fit. Callers wanting plain text can use `reason.toString()`.

PR #4126, opened 2026-09-16
test: move every internal test into its own file under test/internalTests

u9g:
Every internal test lived in `test/internalTest.js`, and a new test was appended at the tail of its plugin's `describe`. Two PRs doing that in the same block are competing insertions at one line, so they conflict on every rebase. In the last 90 days 17 commits touched the file, and 20 of the 42 open PRs that touch it are unmergeable for this reason alone (#4099, #4107, #4063, #4103, #4098, #4111, #4112 among them).

`internalTest.js` now only owns the per-version mock server and bot fixtures. It walks `test/internalTests/<plugin>/<test-name>.js` and registers each file as one `it` inside a `describe` named after its folder. A file exports either shape:

```js
module.exports = async (bot, server) => { ... }
module.exports = (bot, server, done) => { ... }
```

A new test is a new file, which cannot conflict with another PR's new file.

**What moved**

- All 40 test bodies moved verbatim. Helpers that lived in the per-version scope are reachable from a file as `bot.registry`, `bot.version`, `bot.test.chatText` and `bot.test.generateChunkPacket`. Describe-scoped helpers (window packet builders, scoreboard packet builders, `collectHeldItemSlots`) moved into the files that use them.
- Test names are now the file names, so they are lower-case slugs of the old titles, e.g. `absolute position relative position velocity`. `scoreboard reset on login` merged into `scoreboard/`; the top-level `chat`, `entity effects`, `blockAt` and `bed` tests got folders.
- `internalTest.js` drops from 1,973 lines to 126.

**Open PRs that add an internal test** will conflict once with this; the resolution is to save the new `it` body as a file. I'll rebase mine (#4099, #4107, #4108) on top once this lands.

Ran the full internal suite on 1.8.8, 1.12.2, 1.16.5, 1.19.4 and 1.21.4: 184 passing, 16 pending, 0 failing, and every one of master's 40 test titles maps to a file.


rom1504:
What about one file per one of your folder instead of sub folder and a few files each ?

PR #4127, opened 2026-09-16
Apply the server's cursor item; kill drops after the test-world reset

u9g:
#4075's 1.16.5 run failed `book` four times in a row. Only the last error is shown by mocha, and it was not the first cause:

**Attempt 1:** `resetState`'s `/fill` replaced one half of the previous test's bed; the other half dropped as an item and the bot picked it up between the clicks that move the book into the hotbar. The server then saw a different slot than the click claimed and rejected it. Reproduced by summoning a bed item with a 2-3 tick pickup delay during `writeBook`.

**Retries 1-3:** after a rejection the server resends the full inventory *and* the cursor. mineflayer applied the inventory but dropped the cursor packet, so the client kept the book on its cursor while the server's was empty, and every later click carried the wrong cursor. This is a library bug: any rejected click on a pre-1.17 server left the bot unable to click again.

Changes:
- `inventory.js`: apply the server's cursor in all its forms: `set_slot` on window -1 (255 on 1.20.5-1.21.1, where the id is a varint) and `set_cursor_item` on 1.21.3+.
- test harness: `/kill @e[type=item]` right after the reset fills, before the 10-tick pickup delay on block drops runs out.
- new external test `cursorSync`: holds an item on the cursor, sends a click from a deliberately stale model, and asserts the cursor matches the server afterwards and a further move works. Fails without the fix on 1.16.5.

**Needs PrismarineJS/minecraft-data#1303 and a dep bump.** minecraft-data's 1.21.3 schema declares `set_cursor_item.contents` as `option Slot`, so an empty cursor parses as `undefined` and `Item.fromNotch` throws. The 1.21.3 CI job fails until that lands; every other version is unaffected.

Run locally (bed, book, cursorSync): 1.8.8, 1.12.2, 1.16.5, 1.20.4, 1.21.1, 1.21.8, and 1.21.3 with the corrected schema.


test/externalTests/cursorSync.js
```diff
@@ -0,0 +1,26 @@
...
+  await bot.clickWindow(36, 0, 0)
+  assert.strictEqual(bot.inventory.selectedItem?.type, stoneId, 'stone should be on the cursor')
+
+  // Pretend slot 30 holds a dirt the server never gave us, so the click
+  // predicts a swap the server does not perform.
+  bot.inventory.updateSlot(30, new Item(bot.registry.itemsByName.dirt.id, 1))
+  // Pre-1.17 servers report the mismatch as a rejection, which throws.
+  await bot.clickWindow(30, 0, 0).catch(() => {})
+  // The server's corrections precede the reply to anything sent after the click.
+  await bot.test.awaitCommandsProcessed('cursor-synced')
```
rom1504, line 19: **Astra agent review — AI-generated, not manually written by the maintainer.**

The chat echo can arrive before the inventory correction. In this run's [1.21.9 packet trace](https://github.com/PrismarineJS/mineflayer/actions/runs/35112783784/artifacts/10453675387), the stale slot-30 click is sent at `1789571130104`; `cursor-synced` is echoed just 1 ms later, before any cursor/full-window correction. The test has already started its next attempt at +7 ms, while `window_items` carrying the empty cursor arrives at +50 ms. This assertion can therefore still read the locally predicted dirt cursor even with the handler fixed.

Please register a wait for the expected authoritative inventory/cursor correction **before** sending the stale click, then await its application before asserting. This observed ordering failure is separate from the acknowledged 1.21.3 data dependency.

Skills used: `prismarine-item-inventory-review` traced prediction to authoritative cursor reconciliation; `prismarine-protocol-data-review` checked the actual packet order and separated it from the schema dependency.

rom1504 (commented):

One new finding in the regression test: its chat barrier can complete before cursor reconciliation. Skills used: `prismarine-item-inventory-review` for cursor state and `prismarine-protocol-data-review` for the CI packet ordering evidence.

test/externalTests/cursorSync.js
u9g: Fixed in 8c42f90959950683c97b6dc4cd73b99d3775bf50: the test now registers a listener for the server's cursor packet (set_slot on window -1/255, window_items with a carried item, or set_cursor_item) before sending the stale click, and awaits the one that sets the cursor to empty instead of a chat echo. Passes locally on 1.16.5, 1.21.1 and 1.21.9, and still fails on 1.16.5 and 1.21.9 with the inventory.js fix reverted.

rom1504 (commented):

The chat-barrier finding is addressed at 8c42f90: the test registers the authoritative cursor listeners before the stale click, after the plugin's listeners, and waits for the empty cursor. I also ran 84 current-plugin/model cases across seven versions, covering full/empty cursors, player/container windows and the three packet forms. Those model checks do not validate wire schemas or replace the external-server regression.

Not ready to merge yet: CI still fails on 1.21.3 with the known cursor-schema dependency (minecraft-data#1303 is merged, but the tested dependency must contain its fix). The other failed job shows 1.21.9/1.21.11 reset/movement timeouts in digAndBuild/digWaterSpeed rather than cursorSync. Update the dependency containing the schema correction and rerun the failed jobs. No duplicate inline finding is needed for the repaired wait.

Skills used: `prismarine-item-inventory-review` traced prediction to the authoritative cursor; `prismarine-protocol-data-review` separated model success from the release/schema prerequisite; `prismarine-review` checked the existing thread and CI evidence.

PR #4128, opened 2026-09-16
Send serverbound tick_end packet each physics tick (1.21.3+)

DallasCarraher:
## Summary
- Sends an empty serverbound `tick_end` packet once per physics tick, gated on a new `clientTickEndPacket` feature flag (see [minecraft-data#1304](https://github.com/PrismarineJS/minecraft-data/pull/1304)).
- Fixes `lastSent.time` never being updated in `sendPacketPosition`/`sendPacketLook`/`sendPacketPositionAndLook`, which left a stale timestamp after any forced move (teleport) and could trigger a spurious same-tick resend on the very next tick.

## Why

Found live-testing against a 26.3 server: the bot was getting kicked ~1.5-2.5s after spawn with `multiplayer.disconnect.invalid_player_movement`, even though the position/onGround/movement flags being sent were all correct and packet IDs matched the decompiled 26.3 server byte-for-byte.

Root-caused by decompiling `ServerGamePacketListenerImpl` from the real 26.3 server jar:

- `handleMovePlayer` disconnects with `invalid_player_movement` whenever a position-bearing move packet arrives while its `receivedPositionThisTick` flag is already `true`.
- That flag is only ever cleared by `handleClientTickEnd`, which fires when the server receives a serverbound `tick_end` packet — present in the wire protocol since 1.21.3 (confirmed in `minecraft-data`'s `protocol.json` from `1.21.3` through `26.3`), but never sent by mineflayer.
- So the flag gets set `true` on the bot's very first position-bearing move packet and is never cleared again: the *second* one sent over the whole connection — no matter how much later — gets the bot kicked. This matched what was observed: the legitimate "send a position update every second, even if unchanged" resend in `updatePosition()` was the packet that got it kicked, roughly a second after spawn.

Fix: write an empty `tick_end` packet once per physics tick when the server protocol supports it, mirroring real client behavior.

## Test plan
- [x] Reproduced locally against a real 26.3 vanilla server jar with server-side logging: without this fix, `logs/latest.log` shows `Invalid move player packet received` ~2s after spawn; with the fix, the bot stays connected, chats, and moves without incident.
- [x] Verified against a real Mojang Realm running 26.3: bot spawns, stays connected for 60+ seconds (previously kicked well before that), chat and basic movement work.

Depends on [minecraft-data#1304](https://github.com/PrismarineJS/minecraft-data/pull/1304) landing (and a release) for `clientTickEndPacket` to resolve to `true` — until then this is a no-op.


VasilisDragon:
`lastSent.time` controls the periodic position update, so resetting it on look-only packets lets continuous rotation postpone position reports indefinitely. In a local synthetic-clock check, the base sent three position-bearing packets over three seconds; this head sent none. Removing only this assignment restored all three while keeping `tick_end` enabled. Could this stay limited to position-bearing sends?

DallasCarraher:
Good catch, thanks @VasilisDragon. You're right: `lastSent.time` drives the periodic position resend in `updatePosition()`, so resetting it from `sendPacketLook` let continuous rotation postpone position reports indefinitely. I've removed that assignment (04e1c78b). It's now only set in `sendPacketPosition` and `sendPacketPositionAndLook`, the position-bearing sends, which still covers the stale-timestamp-after-teleport case this PR was fixing.

VasilisDragon:
> Good catch, thanks @VasilisDragon. You're right: `lastSent.time` drives the periodic position resend in `updatePosition()`, so resetting it from `sendPacketLook` let continuous rotation postpone position reports indefinitely. I've removed that assignment ([04e1c78](https://github.com/PrismarineJS/mineflayer/commit/04e1c78bd9666ce0c00f25c4051ecd4966386c83)). It's now only set in `sendPacketPosition` and `sendPacketPositionAndLook`, the position-bearing sends, which still covers the stale-timestamp-after-teleport case this PR was fixing.

that fixes the rotation issue. I still get position_look → position → tick_end when moving after a teleport. With chunks missing, teleport replies can also go out without any tick_end.
I reproduced both on 1.21.4/26.1 with the feature enabled. A raw-protocol replay against vanilla 26.3 disconnects unless tick_end separates the positions. Could the tick scheduling cover these cases too?

DallasCarraher:
Thanks @VasilisDragon, both cases reproduced on my side and are fixed in a0433e06:

- **Missing chunks:** `tick_end` was written at the end of `tickPhysics`, after the "entity not ready" / "chunk unloaded" early returns, so it was skipped exactly when a teleport reply was pending. It's now sent on every play-state tick regardless of those early returns, like the real client.
- **Position packets sent outside the tick:** I now track whether a `position`/`position_look` has gone out since the last `tick_end`. If one has, a `tick_end` is written before the next position-bearing packet, so forced-move replies can no longer share a server tick with the next tick's position. In the normal one-position-per-tick case nothing extra is sent.

I added two tests in `internalTest.js`: `tick_end` keeps flowing with no chunk loaded, and back-to-back teleport replies are separated by a `tick_end`. Both fail on the previous head and pass now, and the physics suite passes on all tested versions. I haven't re-run this against a vanilla 26.3 server, so if you can replay your raw-protocol case against this head that would be a useful confirmation.

PR #4136, opened 2026-09-18
Answer sign editors the way the vanilla client does

Pix3lPirat3:
Client-detection plugins such as CheatDetector probe every player on join: `block_change` to a sign, `tile_entity_data` with four `translate`/`keybind` lines and a `fallback`, `open_sign_entity`, then the original block back, and they wait two seconds for `update_sign`. The vanilla client answers because its sign screen closes on the next tick once the sign is gone and always sends the rendered lines. Mineflayer only emits `signOpen`, so it times out on every probe and is reported as "sign checks blocked (possible anti-detection mod)".

## What
  - `open_sign_entity` opens a virtual editor holding the sign's current lines rendered the vanilla way: prismarine-chat with the client language and `fallback` honoured, vanilla keybind ids resolved to their default key names, unknown keybinds echoed raw, 384 characters max (the packet limit).
  - The editor answers with `update_sign` exactly when the vanilla screen would close: one tick after the sign block is replaced or its chunk unloads (the screen checks on its next tick), when the sign leaves the block interaction range (attribute `block_interaction_range`, default 4.5, plus the screen's 4-block padding), when a window, book or another sign editor opens, on `close_window`, respawn or death.
  - `bot.updateSign` still writes the bot's own text and closes the editor; a `signOpen` handler that calls it keeps working unchanged.
  - `updateBlockState` drops the block entity when the block type changes, as the client does (otherwise the editor would still see the phantom sign's text after the block reverted).
  - `signOpen` gets a second argument `isFrontText`; docs updated.
  - Sign text encoding: pre-1.20 `Text1..4`, 1.20-1.21.4 JSON strings in `front_text`/`back_text`, 1.21.5+ NBT text components. The last case is gated on a new minecraft-data feature `signTextUsesNbtComponents` (PR to minecraft-data).
  Until it is released the code treats lines as JSON strings, which still handles plain strings and NBT compounds (only
  a literal line that is itself valid JSON would be misread), so this can land first.

  ## Behaviour changes to be aware of

  - A bot that placed a sign and never called `bot.updateSign` now answers automatically once the sign is removed, out of reach, or another window opens. A late `bot.updateSign` after that is ignored by the server, as it would be for a real client.
  - Sign-GUI input plugins that keep their phantom sign in place until the answer arrives keep working; ones that revert the block at once now get the current text back within a tick.
  - `block.blockEntity` of a block that was replaced by a different block is now `undefined` instead of stale data.

  ## Tests

  Two internal tests on every supported version: the probe (known translation, fallback, vanilla keybind, unknown
  keybind) is answered with the rendered lines after the sign reverts; nothing is sent while the sign stays, until
  `bot.updateSign`.

  Verified live against the real 26.1.2 client on Paper 26.1.2 with CheatDetector 4.4: both pass all five probe groups with identical line values (`MCD_CHECK`, `key.wi_freecam.toggle`, `Click GUI`, ...). External suite passes on 26.1, 1.21.4, 1.20.4, 1.16.5, 1.12.2 and 1.8.8.

---
**Companion PRs:** relies on the signTextUsesNbtComponents feature added in PrismarineJS/minecraft-data#1311.


rom1504 (commented):

Reviewed bb89ca0 using the current skills and real NMP fixtures. The existing sign-editor suite passes across the 28 configured versions (56 tests). Additional cases reproduce three runtime issues: a delayed reply crosses into configuration, reach ignores attribute modifiers, and the unreleased feature fallback changes literal NBT text. Small temporary fixes make those five version-qualified regression cases pass. A minimal TypeScript consumer also rejects the newly documented second event argument. These are the four inline findings; I did not run a live vanilla client/server.

Skills used: `prismarine-lifecycle-action-review` checked editor lifetime; `prismarine-protocol-data-review` checked codecs and data-release selection; `prismarine-geometry-movement-review` checked effective reach; `prismarine-architecture-review` checked shared attribute logic and public declarations; `prismarine-review` checked the current head and existing feedback.

lib/plugins/blocks.js
```diff
@@ -444,11 +495,126 @@ function inject (bot, { version, storageBuilder, hideErrors }) {
...
+      text[`text${i + 1}`] = stringified ? JSON.stringify(line) : line
+    }
+    return text
+  }
+
+  // the vanilla screen notices a removed sign on its next tick, so answer one tick later
+  function invalidateSignEditor () {
+    if (!signEditor || signEditor.closing) return
+    const editor = signEditor
+    editor.closing = setTimeout(() => { if (signEditor === editor) closeSignEditor() }, 50)
```
rom1504, line 570: **Astra agent review — AI-generated, not manually written by the maintainer.**

Please cancel the pending editor/timer when leaving the play session, rather than letting this callback write unconditionally. In the real NMP fixture on 1.21.4 and 1.21.5, `open_sign_entity → block_change removing the sign → start_configuration` within the 50 ms window still calls `write('update_sign')` while the client is in configuration. There is no `update_sign` packet in that state; the server's configuration parser reports a partial-read error. An editor with no removal timer can also survive the transfer until some later close trigger.

Give the editor a discard/cleanup path for configuration/end and world replacement, distinct from a normal play-state close that submits its text; guard delayed completion against the session that opened it. Adding that cleanup made this real-client regression pass.

Skills used: `prismarine-lifecycle-action-review` traced delayed work across protocol re-entry; `prismarine-protocol-data-review` checked the actual client/server codec path; `prismarine-review` checked current discussions for duplicates.

lib/plugins/blocks.js
```diff
@@ -444,11 +495,126 @@ function inject (bot, { version, storageBuilder, hideErrors }) {
...
+    signEditor = { position: pos, isFrontText, lines: signEditorLines(block, isFrontText) }
+    bot.emit('signOpen', block, isFrontText)
+  })
+
+  // the player's block interaction range attribute (1.20.5+), default 4.5 blocks
+  function blockInteractionRange () {
+    const attributes = bot.entity.attributes
+    if (!attributes) return 4.5
+    for (const key in attributes) {
+      if (String(key).endsWith('block_interaction_range')) return attributes[key].value
```
rom1504: **Astra agent review — AI-generated, not manually written by the maintainer.**

This returns the transmitted **base** value, not the effective block-interaction range. With a real `entity_update_attributes` packet containing base `4.5` and an operation-0 modifier of `+5`, an editor whose sign is about 9.6 blocks from the eye closes here: the code uses `4.5 + 4 = 8.5`, although the effective range plus screen padding is `9.5 + 4 = 13.5`. I reproduced the premature `update_sign` through the NMP fixture on 1.21.4 and 1.21.5.

Evaluate all attribute modifier operations and the attribute's final bounds before adding the screen padding. Please reuse the attribute calculation/shared owner rather than introducing another base-only implementation; #4118 is doing related work, but the fix here need not depend on that entire PR landing. The same regression stays open after using the effective value.

Skills used: `prismarine-geometry-movement-review` distinguished base/modifiers from effective reach; `prismarine-architecture-review` checked the shared attribute responsibility; `prismarine-protocol-data-review` verified the transmitted attribute in the real fixture.

lib/plugins/blocks.js
```diff
@@ -444,11 +495,126 @@ function inject (bot, { version, storageBuilder, hideErrors }) {
...
+  // update_sign when that screen closes: when the player is done (bot.updateSign), when the sign
+  // block is removed or the chunk unloads, when the sign is out of reach, when another screen
+  // opens (a window, a new sign editor) or when the world changes. The lines it sends back are
+  // the sign's current text rendered to plain strings, so translatable and keybind components
+  // are resolved with the client's language and key bindings. Anti-cheat plugins flash a
+  // phantom sign full of translation probes and treat a client that never answers as an
+  // anti-detection mod, so the same flow is mirrored here.
+  let signEditor = null
+  // Since 1.21.5 sign lines are stored as NBT text components instead of JSON strings
+  const signTextIsNbtComponent = bot.supportFeature('signTextUsesNbtComponents')
```
rom1504: **Astra agent review — AI-generated, not manually written by the maintainer.**

The proposed pre-release fallback changes valid 1.21.5+ text. With allowed, currently released minecraft-data 3.116.0 this feature is unknown/false. Sending literal NBT-string lines `null` and `{"text":"literal"}` through the real 1.21.5 fixture produces `update_sign` lines `''` and `literal`, respectively, instead of preserving the original strings. These are valid plain sign lines, not malformed components.

Please either retain a version-based fallback for the real 1.21.5 encoding transition (as prismarine-block currently does), or require the released data version containing #1311 before selecting this path. Add a literal JSON-looking NBT-string case alongside the compound-component case; that fails on the current head and passes with the correct encoding selection.

Skills used: `prismarine-protocol-data-review` checked unknown-feature behavior and actual allowed releases; `prismarine-review` separated the companion repository PR from npm release availability.

lib/plugins/blocks.js
```diff
@@ -444,11 +495,126 @@ function inject (bot, { version, storageBuilder, hideErrors }) {
...
+  }
+
   bot._client.on('open_sign_entity', (packet) => {
     const pos = new Vec3(packet.location.x, packet.location.y, packet.location.z)
-    bot.emit('signOpen', bot.blockAt(pos))
+    const isFrontText = packet.isFrontText ?? true
+    closeSignEditor() // a new screen replaces the previous one, which sends its lines
+    const block = bot.blockAt(pos)
+    signEditor = { position: pos, isFrontText, lines: signEditorLines(block, isFrontText) }
+    bot.emit('signOpen', block, isFrontText)
```
rom1504, line 605: **Astra agent review — AI-generated, not manually written by the maintainer.**

Please add `isFrontText: boolean` to `BotEvents.signOpen` in `index.d.ts` along with this new event argument. It still declares only `(block: Block | null)`, so a documented consumer such as `bot.on('signOpen', (block, isFrontText: boolean) => { ... })` fails TypeScript with TS2345: the target signature provides too few arguments. I compiled that consumer against this PR's declarations.

Skills used: `prismarine-architecture-review` checked the usable public API against the runtime event; `prismarine-review` verified this is an introduced contract change.

PR #4137, opened 2026-09-18
Send tick_end at the end of every client tick on 1.21.2+

Pix3lPirat3:
The vanilla client ends every tick with a tick_end packet, sent after that tick's movement packet. The server resets the player's known movement on ticks that carried no movement packet and Paper fires its tick-end event from it; a client that never sends it stands out in a packet capture. Send it from the physics tick loop whenever the bot has an entity position and is in the play state, gated on the existing sendsClientTickEndPacket feature.

A packet capture of the real 26.1.2 client next to a mineflayer bot on the same server showed one packet the bot never sends: `tick_end` (Client Tick End, added in 1.21.2), which the vanilla client sends every tick after that tick's movement packet. The server uses it to reset the player's known movement on ticks without a movement packet (`handleClientTickEnd`), Paper fires its tick-end event from it, and its absence is an easy tell for client-detection plugins.

This sends it from `tickPhysics` after the movement packet, whenever the bot has an entity position and the client is in the play state, gated on the existing minecraft-data feature `sendsClientTickEndPacket` (1.21.2+). The chunk-loaded check that used to return early now only skips the simulation and movement packet, so the tick end still goes out while terrain is loading, like vanilla.

Cost: one 2-byte packet per tick per bot on 1.21.2+. Internal test: five `tick_end` packets arrive after login on affected versions, none on older ones (the packet does not exist there).


rom1504 (commented):

The bounded tick-end change looks ready from these checks. The added real NMP test passes on all eight configured versions supporting the packet (20 version skips), including the no-terrain path that previously returned early. The current code retains play/entity guards, keeps simulation/movement conditional on the loaded chunk, and writes tick_end after movement. I found no new blocker in this diff; I did not independently replay a vanilla packet capture.

Skills used: `prismarine-lifecycle-action-review` checked protocol state, readiness and packet ordering; `prismarine-protocol-data-review` checked the existing feature against the real serializer/server; `prismarine-review` checked the current head and existing feedback.

PR #4138, opened 2026-09-18
Announce the client brand during the configuration state

Pix3lPirat3:
Since 1.20.2 the vanilla client sends its minecraft:brand custom payload as soon as it enters the configuration state, before its client settings, and again every time the server sends it back to configuration. Mineflayer sent it after the login packet in the play state. Send it on entry to configuration on versions with hasConfigurationState and keep the login time send for older versions.

In a capture of the real 26.1.2 client the `minecraft:brand` payload is the first thing sent after `login_acknowledged`, in the configuration state and before the client settings; mineflayer sent it after the `login` packet, in play. Plugins that read the brand on join (and proxies that re-enter configuration for a transfer) see it earlier and again on each re-entry, like from vanilla.

`game.js` now listens for the client's `state` event and sends the brand when the state becomes `configuration` (1.20.2+, `hasConfigurationState`); versions without a configuration state keep sending it on `login`. The `writeChannel` helper already serializes the payload the same way in both states.

Internal test: on 1.20.2+ the brand payload arrives while the test server's client is still in the configuration state.


rom1504 (commented):

The brand-state change looks ready from these checks. The added real NMP test passes on all 12 configured configuration-capable versions (16 skips). I also exercised play → configuration re-entry on 1.20.4 and 1.21.4: the server receives exactly one brand payload followed by settings on each tested re-entry. The state event is emitted after the serializer changes, so the payload uses the configuration codec; older versions retain the existing login path. No new blocker found.

Skills used: `prismarine-lifecycle-action-review` checked re-entry and ordering; `prismarine-protocol-data-review` checked production state/serializer selection; `prismarine-review` checked current discussions before publishing this scoped assessment.

PR #4139, opened 2026-09-18
Do not echo the server's held_item_slot back

Pix3lPirat3:
When the server selects the held slot (on join, after /item, ...) the inventory plugin called bot.setQuickBarSlot, which also writes a held_item_slot packet back to the server. The vanilla client only adopts the slot. Set the slot and refresh the held item without writing.

A packet capture of the real client shows no `held_item_slot` from the client on join; mineflayer sent one right after login because the clientbound `held_item_slot` handler went through `bot.setQuickBarSlot`, which writes the packet.
The handler now sets `bot.quickBarSlot` and refreshes the held item directly. `bot.setQuickBarSlot` is unchanged for the bot's own slot changes.

Internal test: after the server sends `held_item_slot`, no packet of that name comes back and `bot.quickBarSlot` holds the new slot.


Pix3lPirat3:
Already done by https://github.com/PrismarineJS/mineflayer/pull/4069

PR #4140, opened 2026-09-18
Send player_loaded once the player's chunk has arrived

Pix3lPirat3:
The vanilla client sends player_loaded when its loading screen closes, which is when the chunk under the player has been received. Mineflayer sent it immediately on the first update_health, before any terrain. Wait for the chunk column under the player (or 60 ticks, the server's own fallback) before sending it; the spawn event is unchanged.

On 1.21.4+ the real client sends `player_loaded` about 350 ms after the teleport confirm, once the chunk it stands in has arrived and its loading screen closes. Mineflayer sent it on the first `update_health`, before any chunk. Servers ignore block and item interactions until `player_loaded` or 60 ticks, so nothing was lost either way, but the order is now the same as the client's: `tick_end` packets start, then `player_loaded` when the terrain is there.

`health.js` waits for `chunkColumnLoad` of the column under the bot (checked with `bot.blockAt`) with a 3 s fallback, and only writes while in the play state. `spawn` fires at the same time as before.

Internal test (1.21.4+): with the chunk sent 300 ms after `update_health`, `player_loaded` arrives only after the chunk.


lib/plugins/health.js
```diff
@@ -9,9 +9,23 @@ function inject (bot, options) {
...
+  // chunk it stands in has arrived (that is when its loading screen closes), so wait for
+  // that chunk, with the same 60 tick fallback, instead of answering before any terrain.
+  function sendPlayerLoaded () {
+    const send = () => {
+      clearTimeout(timer)
+      bot.removeListener('chunkColumnLoad', onColumn)
+      if (bot._client.state === 'play') bot._client.write('player_loaded', {})
+    }
+    const onColumn = () => { if (bot.entity?.position && bot.blockAt(bot.entity.position) != null) send() }
+    const timer = setTimeout(send, 3000)
```
rom1504: **Astra agent review — AI-generated, not manually written by the maintainer.**

Please keep one cancellable `player_loaded` wait per spawn/session. Every call currently adds an independent timer and `chunkColumnLoad` listener, and neither is cancelled by `respawn`, configuration entry or connection end. In the real 1.21.4 NMP fixture, `login + update_health` with no chunk, followed by `respawn + update_health`, then the new chunk produces **two** `player_loaded` replies. Cancelling the superseded wait makes the same trace produce one.

The `state === 'play'` check cannot distinguish sessions: an old 3-second timeout can acknowledge a newer spawn before its chunk arrives or its own timeout expires. Retain a cleanup/generation owner, cancel the previous wait when a new one starts and at session exit, and cover respawn while the terrain wait is pending.

Skills used: `prismarine-lifecycle-action-review` traced timer ownership across respawn/configuration; `prismarine-protocol-data-review` validated actual server-observed replies; `prismarine-review` checked that this is new feedback.

rom1504 (commented):

The added terrain-arrival test passes on all seven configured versions that support `player_loaded` (21 version skips). A real NMP respawn-before-terrain regression exposes an uncancelled old wait, detailed inline; a temporary cancellation control changes two replies to one. This needs repair before I would call the change ready.

Skills used: `prismarine-lifecycle-action-review` checked spawn/session ownership; `prismarine-protocol-data-review` exercised the production client/server; `prismarine-review` checked head and existing discussion.

